PR #307, opened 2023-04-13
WIP: use QTimeZone for tz info

chriadam:
Problem: Qt for WebAssembly doesn't support QTimeZone...


chriadam:
Yes, closing.  This cannot be supported until QTBUG-68812 and QTBUG-91441 are resolved, which depend on Qt taking into use C++20 features, which is probably a couple of years away.

PR #458, opened 2023-09-04
Support initializing language from argument or query parameter

chriadam:


src/main.cpp
```diff
@@ -669,15 +685,20 @@ int main(int argc, char *argv[])
...
+	QString languageCode;
+	initBackend(&languageCode, &enableFpsCounter);
 
 	QQmlEngine engine;
 	engine.setProperty("colorScheme", Victron::VenusOS::Theme::Dark);
 
 	/* Force construction of translator */
 	int languageSingletonId = qmlTypeId("Victron.VenusOS", 2, 0, "Language");
 	Q_ASSERT(languageSingletonId);
-	(void)engine.singletonInstance<Victron::VenusOS::Language*>(languageSingletonId);
```
MikeTrahearn-Qinetic, line 680: This section seems out of date with main at the time of writing this comment, but it is noticed that the Victron::VenusOS::Language::create(&engine); does not specify QQmlEngine::setObjectOwnership(language, QQmlEngine::CppOwnership); and could cause multiple instances if the create() function was called multiple times. Suggest using Meyer's singleton pattern to prevent this.

PR #462, opened 2023-09-08
Remove translation files (*.ts, *.qm) from source control

DanielMcInnes:
Use 'qt_add_translations' instead of multiple 'add_custom_target' and 'add_custom_command' calls

Remove *.ts and *.qm from source control. We don't actually have any translations yet, so removing the .ts files doesn't cost us anything.

In future, when we actually do get translations, we should add the .ts files back to source control. Any code changes are then likely to change the .ts files, and these changes should be committed. I don't think the .qm files should ever be under source control, they are a build artifact. 

With these changes, newly added engineering strings are displayed in the UI without needed a separate build step.

This has only been tested on ubuntu so far, it needs to be tested on windows and mac.


DanielMcInnes:
When using 6.5.2, I don't see any of the intermittent issues that I saw with 6.2.4, it works fine

chriadam:
Still LGTM.  Thanks!

PR #473, opened 2023-09-27
Update the default maximum value for VeQuickItem

DanielMcInnes:
This was causing issues with SpinBoxes when the maximum value was less than the current value.

### DIFF
--- a/pages/settings/PageGeneratorAcLoad.qml
+++ b/pages/settings/PageGeneratorAcLoad.qml
@@ -64,7 +64,6 @@ Page {
 				suffix: "W"
 				stepSize: 5
 				from: stopValue.value + stepSize
-				to: 1602
 			}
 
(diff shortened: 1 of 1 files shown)


jhofstee:
I think this should be reasonably fine. The reason for the 100 to be there is just to return a number instead of undefined. The trouble with undefined is that typically min / max etc are used like 'property real max: something.max'. Since undefined cannot be assigned to real, all such bindings must check for undefined only to handle the moment the min/max is not available yet. Hence it returns a bogus number, but I don't think gui-v1 relies on the 100 anywhere.

DanielMcInnes:
@chriadam merge it?

chriadam:
It currently conflicts.  But, in general: so long as Jeroen is happy that it won't break anything, then it LGTM.

PR #495, opened 2023-11-03
MQTT: Detect and work-around stuck reconnection issue

chriadam:
Previously, we just stayed at the "Connected" state and were unable to progress to "Initializing" or "Initialized" state.

Note that we cannot use the missed-heartbeat state to detect connection loss, as it could be that we are properly connected to VRM but the device itself has lost internet connectivity. In that case, our UI is connected (to VRM) but the device is not.

We should show some indication (e.g. values not live, orange lamp) in that case...


chriadam:
Obsoleted by https://github.com/victronenergy/gui-v2/pull/585

PR #517, opened 2023-11-10
Move BriefPage animations to render thread

chriadam:
Instead of animating the endAngle of an Arc segment (which must be animated on the GUI thread, and involves a bunch of CPU-bound binding expression evaluations) we now animate the y property of a highlight rectangle (with opacity mask effect to clip to the remainder arc).  This y property can be animated with a YAnimator on the render thread instead (which just updates scene graph nodes and thus converts into pure GPU calls, basically).

There are two downsides:
1) we lose the rounded top of the progress arc
2) aliasing is worse, and CerboGX doesn't yet support MSAA

### DIFF
--- a/components/ArcGauge.qml
+++ b/components/ArcGauge.qml
@@ -34,15 +34,16 @@ Item {
 		layer.smooth: true
 		layer.textureSize: Qt.size(antialiased.width*2, antialiased.height*2)
 
-		ProgressArc {
+		CheapProgressArc {
 			id: arc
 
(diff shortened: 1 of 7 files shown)


mr-manuel:
I really like the new bubble animation. Is it intended, that it starts different then it proceeds after a few seconds?

guiv2_bubbles

chriadam:
Obsoleted by https://github.com/victronenergy/gui-v2/pull/602/files, closing.

PR #557, opened 2023-11-21
Use Scene Graph Animators for Overview battery charge animation

chriadam:
Previously, we animated the width of the rectangles, which was a NumberAnimation on the GUI thread.

Now, we animate X/Y/Opacity with Animators, on the render thread.


DanielMcInnes (approved):

chriadam:
Updated according to Serj's design.

components/widgets/BatteryWidget.qml
```diff
@@ -44,151 +43,109 @@ OverviewWidget {
...
+									to: 0
+									duration: Theme.animation.overviewPage.widget.battery.bubble.duration + 100*Math.random()*bubble.index
+									easing.type: Easing.InOutQuad
+									loops: Animation.Infinite
+									running: root._animationReady && delayedStartTimer.count >= bubble.index
+								}
+
+								XAnimator {
+									target: bubble
+									from: bubble.index % 6 === 0 ? chimney.width/2 - bubble.width
```
blammit, line 121: The Repeater model uses a themed value `Theme.animation.overviewPage.widget.battery.bubbles` rather than hardcoded values, is there a reason for the `from` and `to` bindings use hardcoded values instead?

blammit (approved):
Bubbles look great! Just have a minor comment.

components/widgets/BatteryWidget.qml
chriadam: Good catch, will fix

chriadam: Ah, actually, they aren't related to the bubbles value, but it's just meant to make sure the bubbles have slightly different tracks.  I will change to 3 instead of 6, and add a comment.

PR #563, opened 2023-11-23
Use XAnimator for PageStack transitions (and NavBar)

chriadam:


components/SplashView.qml
```diff
@@ -109,11 +109,15 @@ Rectangle {
 
 		running: Global.allPagesLoaded
 
-		NumberAnimation {
+		PropertyAction {
+			target: extraInfoColumn
+			property: "nextOpacity"
```
blammit, line 114: Does `nextOpacity` need to be reset to 1.0 at some point? E.g. if the SplashView is shown again if the MQTT connection is lost?

Same for other values, like loadingProgress opacity.

blammit (commented):
On mac desktop, sometimes (e.g. 1 in 5 times) when a page is pushed or popped, the animation doesn't run and the page just jumps to the end point instead. I can't repro this with wasm/chrome though. Do you see this at all?

components/SplashView.qml
chriadam: Good point, yes

components/PageStack.qml
```diff
@@ -10,37 +10,48 @@ import Victron.VenusOS
...
 			from: width
 			to: 0
 			duration: Theme.animation.page.slide.duration
 			easing.type: Easing.InOutQuad
 		}
 	}
 	pushExit: Transition {
-		NumberAnimation {
-			property: "x"
+		ScriptAction { script: pageStack.setNavbarX(-width, 2) }
```
DanielMcInnes, line 35: maybe: script: if(pageStack.depth === 2) { navbarX = -width } 
and for popEnter: 
    script: if(pageStack.depth === 1) { navbarX = 0 }

... and then you can get rid of the function 'setNavbarX(...)' and all the calls to it

DanielMcInnes (approved):
1 nit, lgtm

components/PageStack.qml
chriadam: will do

chriadam:
I can't repro the jumps on device... Maybe it is a Mac issue?

I had a think about it, and aside from visibility issues (which I don't think is the problem here) one possibility is that if the render thread is blocked syncing with the gui thread, then there will be missed render thread animation steps - so the animation will "jump" to the "appropriate" spot when the render thread is able to service animation ticks again.

But I don't know what would cause some an expensive sync, especially on a desktop platform...

components/SplashView.qml
chriadam: Fixed.  Can be tested with e.g. language change, as that triggers splash screen again.

components/PageStack.qml
chriadam: Fixed.

chriadam:
Rebased.  Merging, we can investigate the jumps further in future if we can repro on device or in webassembly.

PR #565, opened 2023-11-24
DC fixes

blammit:


DanielMcInnes (commented):
I get this error:
qrc:/qt/qml/Victron/VenusOS/components/widgets/OverviewWidget.qml:12:2: Unable to assign [undefined] to int
	property int type: VenusOS.OverviewWidget_Type_UnknownType

data/common/DcDevice.qml
```diff
@@ -0,0 +1,24 @@
...
+import QtQuick
+import Victron.VenusOS
+import Victron.Veutil
+
+Device {
+	id: dcDevice
+
+	readonly property real voltage: _voltage.value === undefined ? NaN : _voltage.value
+	readonly property real current: _current.value === undefined ? NaN : _current.value
+	readonly property real power: isNaN(voltage) || isNaN(current) ? NaN : voltage * current
```
DanielMcInnes, line 15: does gui-v1 calculate power like this for a dc device? If so, it's ok. Otherwise, I don't know if this is necessarily true.

data/mqtt/DcLoadsImpl.qml
```diff
@@ -0,0 +1,23 @@
...
+*/
+
+import QtQuick
+import Victron.VenusOS
+import Victron.Veutil
+
+QtObject {
+	id: root
+
+	property Instantiator objects: Instantiator {
```
DanielMcInnes, line 13: can the root object be an Instantiator?

DanielMcInnes (approved):
nits, lgtm

data/common/DcDevice.qml
blammit: Hm yes, there are other places where 'amps' is not available so it is calculated using power and voltage, but this is not the case here, and /Dc/0/Power is available. Thanks, fixed, and not this calculation only applies in mock mode.

data/mqtt/DcLoadsImpl.qml
blammit: Done

PR #571, opened 2023-11-25
Fix bug in CMake string comparison

DanielMcInnes:


DanielMcInnes:
When we source '/opt/venus/dunfell-arm-cortexa7hf-neon-vfpv4/environment-setup-cortexa7hf-neon-vfpv4-ve-linux-gnueabi' in order to build for the target, it exports the '-g' flag to CXXFLAGS, which implicitly gets appended to CMAKE_CXX_FLAGS,  which adds debug symbols to our cerbo builds, which balloons the size to ~220MB. I had edited this file locally, removing the debug flag, and missed a bug in CMakeLists.txt because of it.

blammit (commented):
Could you describe in the commit message what the bug is and how it's been fixed?
[DM: done}

DanielMcInnes:
String handling in cmake is awkward.

We were doing this:
string(FIND CMAKE_CXX_FLAGS " -g " FOUND_DEBUG_COMPILER_FLAG)

This doesn't work, it searches the string 'CMAKE_CXX_FLAGS' for ' -g ', and
not surprisingly, never finds it.

We would need to do this:
string(FIND "${CMAKE_CXX_FLAGS}" " -g " FOUND_DEBUG_COMPILER_FLAG)

We were checking for the existence of ' -g ' in the compiler flags prior to
stripping it, which isn't necessary. Just strip ' -g ' if we are doing a
release build.

Also, we were doing this:
    string(REPLACE " -g " " " CMAKE_CXX_FLAGS ${CMAKE_CXX_FLAGS})

which works fine if CMAKE_CXX_FLAGS is populated (as happens when we build on
the desktop for cerbo), but when we build for wasm, CMAKE_CXX_FLAGS is empty.
This causes a cmake error, it complains about running 'string(REPLACE ...)'
with 3 args instead of 4.

I.e. if CMAKE_CXX_FLAGS is empty,
    string(REPLACE " -g " " " CMAKE_CXX_FLAGS ${CMAKE_CXX_FLAGS})
resolves to this:
    string(REPLACE " -g " " " CMAKE_CXX_FLAGS)

We need to surround the possibly empty variable with quotes.

    string(REPLACE " -g " " " CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")       ### CMAKE_CXX_FLAGS is empty

resolves to:

    string(REPLACE " -g " " " CMAKE_CXX_FLAGS "")

CMakeLists.txt
```diff
@@ -16,19 +16,18 @@ set(CMAKE_AUTOMOC ON)
...
 
-# If we want a release build, and the '-g' debug compiler flag is set by the environment setup script, remove it.
+# If we want a release build, remove the '-g' debug compiler flag set by the environment setup script.
 # Otherwise, our executable size ballons from ~10MB to ~220MB.
-string(FIND CMAKE_CXX_FLAGS " -g " FOUND_DEBUG_COMPILER_FLAG)
-if((${CMAKE_BUILD_TYPE} STREQUAL "Release" OR ${CMAKE_BUILD_TYPE} STREQUAL "MinSizeRel") AND NOT(FOUND_DEBUG_COMPILER_FLAG EQUAL -1))
+if((${CMAKE_BUILD_TYPE} STREQUAL "Release" OR ${CMAKE_BUILD_TYPE} STREQUAL "MinSizeRel"))
     message("Removing '-g' debug flag set by /opt/venus/dunfell-arm-cortexa7hf-neon-vfpv4/environment-setup-cortexa7hf-neon-vfpv4-ve-linux-gnueabi")
-    string(REPLACE " -g " " " CMAKE_CXX_FLAGS ${CMAKE_CXX_FLAGS})
+    string(REPLACE " -g " " " CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
```
chriadam, line 30: Can you add a comment above the string(REPLACE ..) line which says:
```
# Match all instances of " -g " in the input "${CMAKE_CXX_FLAGS}" and replace it with the empty string " " and store the result into the output variable CMAKE_CXX_FLAGS
```
because the cmake macro is just insanity to parse without such a comment.

DanielMcInnes: Done

chriadam (approved):

PR #578, opened 2023-11-29
Allow developers to disable compiling qml to C++

DanielMcInnes:
To use, run: cmake -DNO_CACHEGEN=true path/to/CMakeLists.txt This reulsts in much faster builds.
The resulting executable will run slower on the target. Do not use this option for releases, this is just for development.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,13 @@
 cmake_minimum_required(VERSION 3.24) # earlier versions of cmake don't handle 'qt_add_qml_module' properly, you get errors at runtime like: "qrc:/main.qml:164 ApplicationContent is not a type\n"
 
+# CMAKE HINTS
+#
+# Different build types:
+#   cmake -DCMAKE_BUILD_TYPE=[Debug|Release|RelWithDebInfo|MinSizeRelease]
+#
+# Faster builds, for use in development. This stops qml code from being compiled into C++ at build time. Will run slower on the target.
+#   cmake -DNO_CACHEGEN=true
+
 cmake_policy(SET CMP0071 NEW) # process GENERATED source files in AUTOMOC and AUTOUIC. Added to silence a cmake warning.
 cmake_policy(SET CMP0079 NEW)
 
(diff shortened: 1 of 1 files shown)


chriadam (approved):

chriadam:
One interesting note: not only does ahead-of-time compilation / type cache generation mean that compile times are greatly increased, it also means that (since the generated typedata and bytecode is embedded into the executable as a resource) the size of the output binary increases.  According to my limited testing, MinSizeRel binary for CerboGX goes from 8 MB (with cached pre-generated typedata) to 4.5 MB (without cached pre-generated typedata).  MinSizeRel wasm blob goes from 30 MB (with cached pre-generated typedata) to 26 MB (without).

This isn't very interesting for the CerboGX build (where the increased performance afforded by the cache ALWAYS outweighs the increased binary size), it DOES raise an interesting question for the WebAssembly build.  That is: is it worth "suffering" from the lack of pre-generated typecache, in order to reduce the binary size (and thus download size for the client) by 4 MB?

(Note: the "suffering" cost is basically: some time spent compiling all of the QML-defined types at runtime; generally we do this during the splash screen - we have a "pre-load all types" component to trigger this - but there might be some cases where we compile the type lazily on first dynamic load (if that pre-load component missed some cases).)

@dirkjanfaber for discussion

dirkjanfaber:
First of all I probably can adjust the GitHub action script to run that `cmake -DNO_CACHEGEN=true path/to/CMakeLists.txt` on non-tagged builds. @DanielMcInnes : Does `cmake -DNO_CACHEGEN=false path/to/CMakeList.txt` also work as expected?

The question raised on whether or not it wise to suffer from the decreased filesize of the wasm file is interesting. My gut feeling would say that we should add the extra 4 MB's. But to be sure it needs testing. The side that is holding me against adding the 4 MB's, is that disk space on the Venus releases is currently very limited and in that sense 4MB is quite a lot. 

I suggest that I add the option to the  `.github/workflows/build-wasm.yml` file. And default it to `true` on non-tagged builds. Then I'll setup some performance tests with one of those wasm files. And have them also checked by others and report back the results.

chriadam:
Sounds great - thanks very much.

The performance improvement of pre-compiling might not even be noticeable (since that compilation should happen while we're waiting to receive all of the initial MQTT messages) - but yes, benchmarking is needed for sure.

chriadam (approved):

DanielMcInnes:
@dirkjanfaber , yes, `cmake -DNO_CACHEGEN=false path/to/CMakeList.txt` works as expected

PR #579, opened 2023-11-29
Fix crash in translation catalogue loading

chriadam:


DanielMcInnes (approved):
Does lupdate fail on windows if the templates aren't in the .ts files? That's weird, it translates / shows the engineering strings on linux if I delete all the .ts files before building.

chriadam:
On windows, during the first run (if no .ts files exist) lupdate simply creates the template files in the source dir.  On second run, it populates those with the updates.  I don't know why.

PR #585, opened 2023-11-30
MQTT: transition to Ready after "full_publish_complete" message

chriadam:
Previously, we used a timing heuristic to determine when we had received all of the initial-state messages from the broker. FlashMQ now sends a specific message to tell us once it has sent all of the initial-state messages, so depend on it instead. (Still keep the fallback timer, for VRM and non-FlashMQ brokers).

Also, add complete-connection teardown and rebuild, to handle disconnection cases which breaks QtMqtt (see QTBUG-118820).


blammit:
Yes, looks to fine to in afterf the DeviceInstance workaround is removed.

chriadam:
Done, and rebased

blammit (approved):

PR #589, opened 2023-12-05
Performance improvements

chriadam:


components/Utils.js
```diff
@@ -166,17 +166,6 @@ function degreesToRadians(degrees) {
 	return (degrees * Math.PI / 180)
 }
 
-// This considers whether the values are NaN. If both are NaN, the result is NaN.
```
chriadam, line 169: moved to units.cpp because it was hit quite often, and porting Utils.js to utils.cpp is ... a huge amount of work.

src/units.h
```diff
@@ -0,0 +1,91 @@
...
+
+#include "enums.h"
+
+namespace Victron {
+namespace Units {
+
+class Quantity
+{
+	Q_GADGET
+	Q_PROPERTY(QString number READ number WRITE setNumber)
```
blammit, line 20: Could just make `m_number` public and use `Q_PROPERTY(QString number MEMBER m_number)`? And for unit also.

CMakeLists.txt
```diff
@@ -545,14 +545,15 @@ qt_add_qml_module(Mock
 
 # Units_QML_MODULE
 set(Units_QML_MODULE_SOURCES
```
DanielMcInnes, line 547: Rename this to Units_SOURCES

src/units.h
```diff
@@ -0,0 +1,91 @@
...
+	Q_INVOKABLE QString scaledUnitString(Victron::VenusOS::Enums::Units_Type unit) const;
+
+	Q_INVOKABLE qreal convertVolumeForUnit(
+		qreal value_m3,
+		Victron::VenusOS::Enums::Units_Type toUnit) const;
+
+	Q_INVOKABLE qreal celsiusToFahrenheit(qreal celsius) const;
+	Q_INVOKABLE qreal fromKelvin(qreal value, Victron::VenusOS::Enums::Units_Type toUnit) const;
+	Q_INVOKABLE qreal toKelvin(qreal value, Victron::VenusOS::Enums::Units_Type fromUnit) const;
+	Q_INVOKABLE qreal convertFromCelsius(qreal celsius, Victron::VenusOS::Enums::Units_Type unit) const;
```
blammit, line 58: We really need to update gui-v2 to use the veutil temperature and unit conversions instead. Anyway, something for another day.

CMakeLists.txt
```diff
@@ -545,14 +545,15 @@ qt_add_qml_module(Mock
...
-    components/Units.js
+    src/units.h
+    src/units.cpp
 )
 list(APPEND QML_MODULE_SOURCES ${Units_QML_MODULE_SOURCES})
```
DanielMcInnes, line 551: list(APPEND SOURCES ${Units_SOURCES}). C++ code doesn't get processed by lupdate()

src/units.cpp
```diff
@@ -0,0 +1,278 @@
...
+	case Victron::VenusOS::Enums::Units_Hertz:
+		return QStringLiteral("kHz");
+	case Victron::VenusOS::Enums::Units_Volume_Liter:
+		// \u2113 = l, \u3398 = kl
+		return QStringLiteral("\u3398");
+	default:
+		return QString();
+	}
+}
+
```
blammit, line 114: stray newline?

src/units.cpp
```diff
@@ -0,0 +1,278 @@
...
+	return QStringLiteral("%1/%2%3").arg(r.number(), c.number(), c.unit());
+}
+
+// This considers whether the values are NaN. If both are NaN, the result is NaN.
+qreal Units::sumRealNumbers(qreal a, qreal b) const
+{
+	const bool aNaN = qIsNaN(a);
+	const bool bNaN = qIsNaN(b);
+
+	return (aNaN && bNaN)  ? qQNaN()
```
blammit, line 271: stray whitespace?

components/VerticalGauge.qml
```diff
@@ -4,29 +4,82 @@
...
+					anim.start()
+				}
+			}
+
+			YAnimator {
+				id: anim
+				target: fgRect
+				easing.type: Easing.InOutQuad
+				duration: Theme.animation.briefPage.sidePanel.sliderValueChange.duration
+				onRunningChanged: fgRect.resetYBinding()
```
DanielMcInnes, line 73: does this need to be:
`if (!running) fgRect.resetYBinding()`
?

src/units.cpp
```diff
@@ -0,0 +1,278 @@
...
+	case Victron::VenusOS::Enums::Units_Volume_CubicMeter:
+		// \u33A5 is not supported by the font, so use two characters \u006D\u00B3 instead.
+		return QStringLiteral("m³");
+	case Victron::VenusOS::Enums::Units_Volume_GallonUS: // fall through
+	case Victron::VenusOS::Enums::Units_Volume_GallonImperial:
+		return QStringLiteral("gal");
+	case Victron::VenusOS::Enums::Units_RevolutionsPerMinute:
+		return QStringLiteral("RPM");
+	default:
+		return QString();
```
DanielMcInnes, line 90: maybe a warning message for the default case? here and elsewhere

DanielMcInnes (approved):
nits

src/units.h
```diff
@@ -0,0 +1,91 @@
...
+		int precision,
+		const QString &baseUnit,
+		const QString &scaledUnit = QString()) const;
+
+	Q_INVOKABLE Victron::Units::Quantity getDisplayText(
+		Victron::VenusOS::Enums::Units_Type unit,
+		qreal value,
+		int precision = -1,
+		qreal unitMatchValue = 0) const;
+
```
blammit, line 72: The precision is not working for labels that have more than one decimal place now. E.g. in the Device List, this is the existing UI on the left, and then with this PR on the right:

image

CMakeLists.txt
chriadam: will that SOURCES be specific to this module, or would I be clobbering the main project's SOURCES variable?  my cmake-fu is weak

components/VerticalGauge.qml
chriadam: No.  In both the running changed and the root.animationEnabled changed cases, we need to check both conditions, otherwise we can fail to reset the binding appropriately when we should, or overwrite the value with a binding when we shouldn't.  So both check conditions go into the resetYBinding() function.

CMakeLists.txt
DanielMcInnes: ignore my comment, this is fine

src/units.cpp
```diff
@@ -0,0 +1,278 @@
...
+Victron::Units::Quantity Units::scaledQuantity(
+	qreal value,
+	qreal unitMatchValue,
+	int precision,
+	const QString &baseUnit,
+	const QString &scaledUnit) const
+{
+	Victron::Units::Quantity quantity;
+	qreal v = value;
+	const qreal scaleMatch = unitMatchValue != 0.0 ? unitMatchValue : value;
```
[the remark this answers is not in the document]
chriadam: I'd prefer to to just use `0.0` as "don't scale".  I would like to avoid NaN comparisons wherever possible, if there is no specific reason for them.

src/units.h
```diff
@@ -0,0 +1,91 @@
...
+		qreal unitMatchValue,
+		int precision,
+		const QString &baseUnit,
+		const QString &scaledUnit = QString()) const;
+
+	Q_INVOKABLE Victron::Units::Quantity getDisplayText(
+		Victron::VenusOS::Enums::Units_Type unit,
+		qreal value,
+		int precision = -1,
+		qreal unitMatchValue = 0) const;
```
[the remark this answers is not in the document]
chriadam: as above, I'd prefer to denote that zero means "don't scale".

src/units.h
chriadam: Good point, will do

src/units.cpp
```diff
@@ -0,0 +1,278 @@
...
+		return QStringLiteral("\u3398");
+	default:
+		return QString();
+	}
+}
+
+
+qreal Units::convertVolumeForUnit(qreal value_m3, Victron::VenusOS::Enums::Units_Type toUnit) const
+{
+	if (qIsNaN(value_m3)) {
```
DanielMcInnes, line 118: is 'qIsNan' the same as js `isNaN()`, or js `Number.isNaN()`? We want the former.

src/units.cpp
chriadam: will remove

src/units.cpp
chriadam: will remove

src/units.h
chriadam: Thanks, will fix.

src/units.cpp
chriadam: If the js ones produce different results, we should REMOVE ALL USES of NaN from our codebase.

chriadam: NaN is a very specific range of code points in floating point numbers.  If JS cannot produce floating point numbers which conform with IEEE floating points, then we should remove them all, and use boolean "numberIsInvalid" flags instead.

src/units.cpp
chriadam: will do (well, not for the scaling one, because I think it's intentional that there are not "scaled" units for some base units.

src/units.cpp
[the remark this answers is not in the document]
chriadam: Well, I guess there is value in being consistent, if we're not going to get rid of NaN everywhere else.  Will change.

src/units.h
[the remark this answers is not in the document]
chriadam: will fix.

src/units.cpp
chriadam: done

src/units.cpp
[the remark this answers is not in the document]
chriadam: done

src/units.h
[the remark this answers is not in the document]
chriadam: done

src/units.h
chriadam: done

src/units.cpp
chriadam: done

src/units.h
chriadam: done

src/units.cpp
chriadam: For now, I'm going to assume that it works as required.  Please file a bug if it doesn't.

PR #592, opened 2023-12-05
ModalDialog: extend correct type and fix fade in/out animations

blammit:
OpacityAnimator doesn't work on the background, so use NumberAnimation instead.


chriadam (approved):
Qt bugs annoy me.  LGTM.

PR #593, opened 2023-12-05
Use pre-calculated width for QuantityLabel geometry

chriadam:


chriadam:
The QuantityLabel commit is fairly horrible, there must be a better way to do it?

chriadam:
Would be better if we did this once per typography style (including font size), rather than once per QuantityLabel.

components/ElectricalQuantityLabel.qml
```diff
@@ -11,15 +11,12 @@ QuantityLabel {
 	id: root
 
 	property var dataObject
+	property bool unitAmps: Global.systemSettings.electricalQuantity.value === VenusOS.Units_Amp
```
DanielMcInnes, line 14: readonly?

components/ElectricalQuantityLabel.qml
```diff
@@ -11,15 +11,12 @@ QuantityLabel {
...
-			? NaN
-			: Global.systemSettings.electricalQuantity.value === VenusOS.Units_Amp
-				&& !isNaN(dataObject.current)
-			  ? dataObject.current
-			  : dataObject.power
-	unit: Global.systemSettings.electricalQuantity.value === VenusOS.Units_Amp
-			&& dataObject != null && !isNaN(dataObject.current)
-		  ? VenusOS.Units_Amp
-		  : VenusOS.Units_Watt
+	property bool _valueUnknown: dataObject == null
```
DanielMcInnes, line 16: any reason for '==' instead of '===' ?

components/ElectricalQuantityLabel.qml
```diff
@@ -11,15 +11,12 @@ QuantityLabel {
...
+	value: _valueUnknown ? NaN
```
blammit, line 17: Also need to use Watts if dataObject.current == NaN

components/QuantityLabel.qml
```diff
@@ -18,31 +18,54 @@ Item {
...
+	// Cache some values to reduce binding re-evaluations.
+	property int _unitLength: unitLabel.text.length
+	property int _valueLength: valueLabel.text.length
+	property bool _containsDot: (valueLabel.text.indexOf('.') >= 0) || (valueLabel.text.indexOf(',') >= 0)
+	property real _quantityLabelSpacing: 2*Theme.geometry.quantityLabel.spacing
+	property real _implicitValueWidth: Math.ceil(_valueLength*fontMetrics.numWidth + _valueLength*fontMetrics.advWidth - (_containsDot ? fontMetrics.dotDelta : 0))
+	property real _implicitUnitWidth: Math.ceil(_unitLength*fontMetrics.charWidth + _unitLength*fontMetrics.advWidth)
+	implicitWidth: _implicitValueWidth + _quantityLabelSpacing + _implicitUnitWidth
+
+	FontMetrics {
```
blammit, line 38: This seems thorough but I'm also wondering whether it's worth adding this complexity for every QuantityLabel. Maybe all we need to do is to set a fixed width for the cases where there are multiple QuantityLabels in a row, e.g. in ListQuantityGroup. Because, the main issue is that when other items are anchored to the left/right of a QuantityLabel, they shift when the quantity changes. But in many cases, a QuantityLabel is not displayed next to another item (e.g. in Brief and Overview pages, Tank gauges) and when they are, we end up needing to set a fixed maximum width anyway, to ensure items are evenly spaced (e.g. in QuantityTableSummary or the humidity gauges).

Also this change affects a lot of areas, could you check how it looks throughout the UI? E.g. 
- on the Brief and Overview pages, there's now a gap between the unit and the number, like "620 W", where there wasn't before
- the quantity numbers don't fit correctly below the temperature and humidity gauges
- in battery widget on Overview page, the units of the bottom numbers squeeze into the adjacent labels
- in solar charger drilldowns, the labels don't fit into the row

pages/BriefMonitorPanel.qml
```diff
@@ -159,7 +159,7 @@ Column {
 	Column {
 		width: parent.width
 		spacing: Theme.geometry.briefPage.sidePanel.generator.columnSpacing
-		visible: !generatorColumn.visible && !(isNaN(Global.acInputs.power) && isNaN(Global.dcInputs.power))
+		visible: root.visible && !generatorColumn.visible && !(isNaN(Global.acInputs.power) && isNaN(Global.dcInputs.power))
```
DanielMcInnes, line 162: I think 'visible' inherited anyway?
i.e.
```
		Rectangle {
			anchors.centerIn: parent
			width: parent.width / 2
			height: parent.height / 2
			color: "red"
			visible: false
			Rectangle {
				anchors.centerIn: parent
				width: parent.width / 2
				height: parent.height / 2
				color: "green"
				visible: true
			}
		}
```
the green rectangle is not visible

chriadam: Correct, visible is inherited.
But this binding will still re-evaluate, if one of the captured properties emits a change.
Thus this would previously re-evaluate if Global.acInputs.power or if Global.dcInputs.power emits a change signal.
Now, the only capture (when root.visible is false) is the root.visible property, so no re-evaluations while root.visible remains false.

components/QuantityLabel.qml
chriadam: Well, the main issue from my perspective is the performance cost of the item geometry being dependent upon the result of text layouts every time the value of the text property changes (even if the number of characters doesn't change).  Fixing the visual jumping around in the rows etc was secondary for me ;-)

chriadam: But, yes, we shouldn't construct metrics in every label, but instead do it in C++.  I will fix.

components/QuantityLabel.qml
```diff
@@ -18,31 +18,54 @@ Item {
...
 
 	// Restrict the height to the baseline to help align the baseline of labels in different
 	// QuantityLabel items with different font sizes. E.g. Environments tab may have multiple gauges
 	// with different font sizes, that need to align side-by-side at the font baseline.
-	implicitWidth: valueLabel.implicitWidth + Theme.geometry.quantityLabel.spacing + unitLabel.implicitWidth
 	implicitHeight: Math.ceil(valueLabel.baselineOffset) + Theme.geometry.quantityLabel.bottomMargin
 
+	// For width, we want quasi-fixed-width for the value and label.
+	// Cache some values to reduce binding re-evaluations.
+	property int _unitLength: unitLabel.text.length
```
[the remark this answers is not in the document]
chriadam: I was very surprised, too.  It could be that the operations are exponentially slower in Debug mode than they are in Release mode, so that they "show up" in the QML performance traces disproportionately to their actual impact.  I'm not sure.

But, short answer is: yes, adding these intermediate properties greatly reduced the self-time spent in QuantityLabel.qml, according to the profiler traces I collected.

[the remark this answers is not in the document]
chriadam: The down side, of course, is greater memory pressure due to more dynamic properties.  Swings and roundabouts...

components/QuantityLabel.qml
blammit: I see, my main concern was the visual jumping, as that was the bug I was looking at. The performance would be great to fix, as long as it doesn't make the code overly complex and difficult to maintain.

pages/BriefMonitorPanel.qml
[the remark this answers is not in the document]
chriadam: Good point, I'll try to unload the entire panel when it's not visible, instead of this commit.

components/ElectricalQuantityLabel.qml
chriadam: Can the value ever go to NaN if it was previously valid?
If so, this seems incredibly problematic (because we need to check the value every time it changes, which is terrible for performance).

What is the use-case?

components/ElectricalQuantityLabel.qml
chriadam: nope, fixed.

components/ElectricalQuantityLabel.qml
chriadam: fixed

components/ElectricalQuantityLabel.qml
chriadam: Well, added the check back, we can revisit again in future if it starts showing up in the traces again.

pages/BriefMonitorPanel.qml
chriadam: Ended up just loading the entire BriefMonitorPanel on demand, instead.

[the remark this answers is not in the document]
chriadam: Done.  We skip a frame on load, but it's probably better than the alternative.

chriadam:
Ready for review again, I think.

components/ElectricalQuantityLabel.qml
blammit: Yes I assume that if an input is disconnected, reading the power/current on that VeQuickItem will produce an `undefined` value.

blammit: Is there a benefit in having this `_valueUnknown` property? 

Also, seeing this warning:

`qrc:/qt/qml/Victron/VenusOS/components/ElectricalQuantityLabel.qml:18: TypeError: Cannot read property 'power' of undefined`

pages/BriefPage.qml
```diff
@@ -234,37 +234,44 @@ Page {
...
+		// and cause the object hierarchy to take multiple seconds to load.
+		asynchronous: false
 
 		// hidden by default.
-		property bool active: false
+		active: false
 		x: root.width
 		opacity: 0.0
-		visible: false
+		property bool showing // intermediate, used only to trigger loading.
```
blammit, line 257: I'm not understanding why this intermediate `showing` is needed, could you explain? I would've thought that toggling the Loader `active` would be great for disabling animations and calculations in the panel when it's hidden, instead of keeping `active=true` after the first time it's shown.

src/units.h
```diff
@@ -15,9 +15,10 @@
 namespace Victron {
 namespace Units {
 
-class Quantity
+class quantityStrings
```
blammit, line 18: The name might be inflexible if we change the member types. Also what was the reason for starting with a lowercase?

How about QuantityInfo instead?

components/QuantityLabel.qml
```diff
@@ -51,9 +59,9 @@ Item {
...
 			verticalCenter: parent.verticalCenter
 			left: valueLabel.right
-			leftMargin: Theme.geometry.quantityLabel.spacing
+			leftMargin: _quantityLabelSpacing
 		}
-		text: root.unit === VenusOS.Units_None ? "" : root._quantity.unit
+		text: root._quantity.unit
 		font: valueLabel.font
 		color: Theme.color.font.secondary
 	}
```
blammit, line 67: There's a larger spacing between the number and the unit now (as below). Can it be kept smaller as per the design?

image
image

blammit: If you click on the solar widget in Overview, the values are not very readable now:

image

Similarly when viewing solar history: 

image

and for EV charger drilldowns: ('Energy' column and also L2/L3 values)

image

src/units.h
DanielMcInnes, line 18: captialise

components/QuantityLabel.qml
blammit: Also alignments/eliding in temperature gauges: 

image

and alignments/overflow in humidity gauges:

image
 
(Both are reproducible with mock mode if you go to Levels page and press left/right to cycle through visual configurations)

components/QuantityLabel.qml
```diff
@@ -18,31 +18,39 @@ Item {
...
-		 : Units.getDisplayText(unit, value, precision)
+	readonly property quantityStrings _quantity: Units.getDisplayText(unit, value, precision)
 
 	// Restrict the height to the baseline to help align the baseline of labels in different
 	// QuantityLabel items with different font sizes. E.g. Environments tab may have multiple gauges
 	// with different font sizes, that need to align side-by-side at the font baseline.
-	implicitWidth: valueLabel.implicitWidth + Theme.geometry.quantityLabel.spacing + unitLabel.implicitWidth
 	implicitHeight: Math.ceil(valueLabel.baselineOffset) + Theme.geometry.quantityLabel.bottomMargin
 
+	// For width, we want quasi-fixed-width for the value and label.
```
blammit, line 28: Could you expand on what is meant by "quasi-fixed"? E.g. which part is actually fixed and which is not?

src/units.h
```diff
@@ -51,14 +52,14 @@ class Units : public QObject
...
-	Q_INVOKABLE Victron::Units::Quantity scaledQuantity(
+	Q_INVOKABLE Victron::Units::quantityStrings scaledQuantity(
 		qreal value,
 		qreal unitMatchValue,
 		int precision,
 		const QString &baseUnit,
 		const QString &scaledUnit = QString()) const;
 
-	Q_INVOKABLE Victron::Units::Quantity getDisplayText(
+	Q_INVOKABLE Victron::Units::quantityStrings getDisplayText(
```
DanielMcInnes, line 62: not related to your changes, but it was not clear to me what 'unitMatchValue' is used for. A comment here with an example of how this is used would be useful.

src/units.h
```diff
@@ -80,6 +81,6 @@ class Units : public QObject
 }
 }
 
-Q_DECLARE_METATYPE(Victron::Units::Quantity)
+Q_DECLARE_METATYPE(Victron::Units::quantityStrings)
```
DanielMcInnes, line 84: is Q_DECLARE_METATYPE needed after we use QML_ELEMENT? The doco is confusing. I tried commenting it out and it seems to work

src/theme.cpp
```diff
@@ -231,5 +233,46 @@ QColor Theme::statusColorValue(StatusLevel level, bool darkColor) const
...
+}
+
+qreal Theme::characterDotDeltaWidth(const QFont &font) const
+{
+	return fontInfo(font).dotDeltaWidth;
+}
+
+const FontInfo& Theme::fontInfo(const QFont &font) const
+{
+	for (const FontInfo &fi : qAsConst(m_fontInfo)) {
```
DanielMcInnes, line 258: would this be more performant if m_fontInfo was a QMap / std::map?

components/ElectricalQuantityLabel.qml
```diff
@@ -11,15 +11,11 @@ QuantityLabel {
...
-			  : dataObject.power
-	unit: Global.systemSettings.electricalQuantity.value === VenusOS.Units_Amp
-			&& dataObject != null && !isNaN(dataObject.current)
-		  ? VenusOS.Units_Amp
-		  : VenusOS.Units_Watt
+	readonly property bool unitAmps: !!dataObject && !isNaN(dataObject.current) && Global.systemSettings.electricalQuantity.value === VenusOS.Units_Amp
+	readonly property bool _valueUnknown: dataObject === null
+	value: _valueUnknown ? NaN
+		: unitAmps ? dataObject.current
+		: dataObject.power
```
DanielMcInnes, line 18: when running in mock mode, I get:
qrc:/qt/qml/Victron/VenusOS/components/ElectricalQuantityLabel.qml:18: TypeError: Cannot read property 'power' of undefined

components/QuantityLabel.qml
chriadam: it is "fixed-per-length" rather than "fixed".  That is, if you add more characters, you get more width, but the width is fixed for any given string length.

src/units.h
chriadam: quantityInfo might work.  It has to start with lower case because it's a Q_GADGET value type, the QML engine complains if it starts with uppercase

pages/BriefPage.qml
chriadam: because when you toggle the button off, we can't set active to false (or the entire panel will disappear).  By toggling "showing" we can "animate away" the panel before setting active to false.

chriadam: active doesn't stay true, though.  We set it to false in the script animation of the transition.

src/units.h
chriadam: nope.  Q_GADGET value type name has to start with lower case

components/QuantityLabel.qml
chriadam: > There's a larger spacing between the number and the unit now (as below). Can it be kept smaller as per the design?
> 
> image image

unfortunately, no.  That is the point of fixed width.  If they were all 4's (4 being the widest number glyph) then there would be "the same" space.  But if they are not 4's, then the gap between the numbers and the unit string will be larger.  There's no way around that.

chriadam: > Also alignments/eliding in temperature gauges:
> 
> image
> 
> and alignments/overflow in humidity gauges:
> 
> image
> 
> (Both are reproducible with mock mode if you go to Levels page and press left/right to cycle through visual configurations)

Huh!  Thanks, will fix.  What's the repro for this (the elision)?  It definitely shouldn't do that (since the label should be allocated the ... maximal possible width for the number of characters displayed).  Definitely a bug.

chriadam: For the temperature and percentages, it's interesting.  I guess we need to peek into the units, to determine if its either temp/percentage, and then instead of allocating enough space for "444" just allocate enough space for "100".

src/units.h
chriadam: Well, at least in the Q_GADGET example they declare the metatype, so I will leave it.

src/theme.cpp
chriadam: No.  Maps and hashes are super slow compared to vectors for small N.  See e.g.  https://isocpp.org/blog/2014/06/stroustrup-lists

components/ElectricalQuantityLabel.qml
chriadam: How is this possible?  Surely _valueUnknown should resolve to true in this case?  Ah, I guess it's because I changed the "dataObject == null" to "dataObject === null" so now that check doesn't catch the undefined case.  I will revert back to double-equals.

components/ElectricalQuantityLabel.qml
chriadam: the undefined thing is because of the triple-equals check, will fix.

components/QuantityLabel.qml
blammit: Sorry, I mean, could you expand the comment to explain that, since there are a lot bindings involved in understanding this bit?

pages/BriefPage.qml
blammit: Ah, I see. It might be nice to use two separate states here then, instead of using two `showing` and `showingAndLoaded` properties and managing the states that way.

src/units.h
blammit: Oh, I see. The lowercase complaint seems like a new thing for gadgets? Haven't seen that before

components/ElectricalQuantityLabel.qml
chriadam: Fixed

pages/BriefPage.qml
chriadam: I'd prefer not to add additional states if we don't have to.  I think the properties are pretty self-documenting: one goes to true when we need to show the panel, the other goes to true once the panel has been loaded and can thus be animated in.

src/units.h
chriadam: not new (e.g. "font", "color" etc types) but until recently users couldn't register their own (easily, at least)

chriadam: Renamed to quantityInfo

components/QuantityLabel.qml
chriadam: Fixed most of these.  I now align the value text with Text.AlignRight to get rid of the wide gap.  I added explicit test for "100" valued percentages/temperatures, and determine the width specially for those.  I solved the elision issue by discovering that font metrics occasionally underreports the width, and adding a fudge factor to allow for it.

components/QuantityLabel.qml
chriadam: Done

src/units.h
chriadam: Not in this PR. (I don't remember what it's for, I just raw ported the code.)

components/ElectricalQuantityLabel.qml
chriadam: Fixed (I hope, at least, couldn't repro this in mock mode)

components/QuantityLabel.qml
chriadam: Couldn't figure out how to view the "EV Charger" page you mentioned - hints appreciated

blammit: > Couldn't figure out how to view the "EV Charger" page you mentioned - hints appreciated

Looks much better with the latest changes -- this issue doesn't occur any more.

pages/BriefPage.qml
blammit: It seems a bit messy to use multiple properties to track the state, but ok.

components/QuantityLabel.qml
```diff
@@ -18,42 +18,54 @@ Item {
...
 
+	// For width, we want fixed-for-text-length-width for the value and units labels.
+	// Cache some values to reduce binding re-evaluations.
+	readonly property int _unitLength: unitLabel.text.length
+	readonly property int _valueLength: valueLabel.text.length
+	readonly property bool _containsDot: (valueLabel.text.indexOf('.') >= 0) || (valueLabel.text.indexOf(',') >= 0)
+	readonly property real _dotDeltaWidth: _containsDot ? Theme.characterDotDeltaWidth(valueLabel.font) : 0
+	readonly property bool _containsMinus: valueLabel.text.indexOf('-') >= 0
+	readonly property real _minusDeltaWidth: _containsMinus ? Theme.characterMinusDeltaWidth(valueLabel.font) : 0
+	readonly property real _oneHundredWidth: ((root.unit === VenusOS.Units_Percentage
```
blammit, line 36: Why is it that the _oneHundredWidth only applies for some units and not others?

blammit (commented):
QuantityLabel layouts are looking much better. I'm still seeing some odd alignments. E.g. on the Overview page  the numbers are no longer left-aligned to the icons -- I assume this is because the numbers are intentionally right-aligned, but it might look a bit untidy in the UI layout?

image

Also in the drilldown measurement summaries, like this:

image

And on the Brief page here the "100%" doesn't quite line up with 81% and 3%, which is a minor thing but just looks a bit odd.

image

components/QuantityLabel.qml
chriadam: for others, 100 isn't the maximum value, so e.g. it might be 104 or 444 (which are wider than 100).  But for temperature and for percentage, 100 is the maximum value (i.e. 99 will always be skinnier than 100, but no 104 is allowed etc).

chriadam: added a code comment about this

chriadam:
Fixed the alignments.  (Except the issue in Brief page multi circular gauges - that's not a quantity label, but a pair of normal Labels.)

blammit (approved):
As discussed, should get Serj to approve the UI changes. But looks good to go in for now and we can fix it up later to address any concerns.

chriadam:
Thanks, rebase-merging.

PR #600, opened 2023-12-08
Solar history dialog

blammit:


components/dialogs/SolarDailyHistoryDialog.qml
```diff
@@ -58,11 +58,11 @@ C.Dialog {
 					root._refresh()
 				}
 			}
-			OpacityAnimator { from: 0.0; to: 1.0; duration: Theme.animation.page.fade.duration }
+			NumberAnimation { property: "opacity"; from: 0.0; to: 1.0; duration: Theme.animation.page.fade.duration }
```
DanielMcInnes, line 61: Annoying that we can't use OpacityAnimator here. I see that it doesn't work, but I don't understand why. From the doco, it looks like it should. Do you know why?

DanielMcInnes (approved):

components/dialogs/SolarDailyHistoryDialog.qml
blammit: No idea -- I wonder if it's because there is some layering involved in showing a modal dialog, so it might be handled differently by the renderer.

PR #602, opened 2023-12-08
WIP: write customer shaders for circular gauges

chriadam:
Animating the Shape-based circular gauges is too expensive. So reimplement it as a single-pass shader.


chriadam:
WIP, don't merge.  I'll create implementations for the generic ProgressArcs too (they're trickier since the arc specified can be part of a notional circle whose diameter is far greater than the width or height of the gauge itself).

chriadam:
Still WIP.  Need to test on CerboGX.
Doesn't work on WebAssembly (but it definitely should, I think it's just a qsb issue).
Aside from that, basically finished.

chriadam:
Unfortunately, https://bugreports.qt.io/browse/QTBUG-120112 might mean that we have to wait until Qt 6.6.x for this to work on webassembly, at least.  This may not be a problem, since the performance problems don't show up on webassembly or desktop, but only on the CerboGX, so we can just load different types depending on the platform (shader vs non-shader gauges).

The bigger problem is: (I haven't tested yet, but it's very possible that) the shader performance is poor on CerboGX due to the complexity of the shader.  It may not be possible to draw everything using the shader, but instead draw only the progress bar with the shader.  But, let's see, once we have the shader tools added to the SDK.

DanielMcInnes:
@chriadam can you convert this PR to draft if it is not ready for review yet?

chriadam:
By changing the name of the fragment shader's input from "coord" to "qt_TexCoord0" (to match Qt's default vertex shader output for older GLSL targets) the shaders now work.

However, on the CerboGX they are incredibly slow, due to how complex the shaders are (e.g. the antialiasing, shine effect, etc).
I will look at simplifying them to see whether we can get them to acceptable performance.

src/main.cpp
```diff
@@ -343,6 +343,7 @@ int main(int argc, char *argv[])
 	engine.setProperty("screenSize", (round(screenDiagonalMm / 10 / 2.5) == 7)
 			? Victron::VenusOS::Theme::SevenInch
 			: Victron::VenusOS::Theme::FiveInch);
+engine.setProperty("screenSize", Victron::VenusOS::Theme::SevenInch);
```
chriadam, line 346: debugging, should remove.

shaders/circulargauge.frag
```diff
@@ -0,0 +1,107 @@
+#version 440
```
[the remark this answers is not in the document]
chriadam: Nope, not necessary.

shaders/circulargauge.frag
```diff
@@ -0,0 +1,106 @@
...
+#define CONSTANT_PI 3.141592653589793
+#define CONSTANT_TAU 6.283185307179586
+layout(location = 0) in vec2 coord;
```
chriadam, line 4: have to change this to qt_TexCoord0 to work around QTBUG-120112

PR #609, opened 2023-12-15
AC inputs rework

blammit:


data/AcInputs.qml
```diff
@@ -27,36 +24,66 @@ QtObject {
...
-	function removeInput(input) {
-		if (model.removeDevice(input.serviceUid)) {
-			if (input === connectedInput) {
-				connectedInput = null
-			}
-			if (input === generatorInput) {
-				generatorInput = null
+		uid: Global.system.serviceUid + "/Ac/ActiveIn/Source"
+		onValueChanged: {
+			sourceAsInt = value === undefined ? VenusOS.AcInputs_InputSource_NotAvailable : parseInt(value)
```
DanielMcInnes, line 33: does this need to be imperative?

data/AcInputs.qml
```diff
@@ -27,36 +24,66 @@ QtObject {
...
+	{
+		switch (source) {
+		case VenusOS.AcInputs_InputSource_Grid:
+			return "qrc:/images/grid.svg"
+		case VenusOS.AcInputs_InputSource_Generator:
+			return "qrc:/images/generator.svg"
+		case VenusOS.AcInputs_InputSource_Shore:
+			return "qrc:/images/shore.svg"
+		default:
+			break
```
DanielMcInnes, line 73: maybe get rid of the 'break' here, and just 'return ""'

data/common/AcInputPhaseModel.qml
```diff
@@ -0,0 +1,52 @@
...
+	readonly property VeQuickItem _numberOfPhases: VeQuickItem {
+		uid: Global.system.serviceUid + "/Ac/ActiveIn/NumberOfPhases"
+		onValueChanged: {
+			const count = value || 0
+			root._phaseObjects.model = null
+			root.clear()
+			for (let i = 0; i < count; ++i) {
+				root.append({ name: "L" + (i+1), current: NaN, power: NaN })
+			}
+			root._phaseObjects.model = count
```
DanielMcInnes, line 49: Can this be declarative instead?

DanielMcInnes (approved):
lgtm, nits / questions

data/common/AcInputPhaseModel.qml
chriadam: I don't think it can be - otherwise there might be a transiently invalid state.  It's why model is set to null, the objects are created, and then the model is set to the count.

In general, if you try to separate two operations into two separate bindings or a binding plus signal handler (binding which sets the phaseobjects model to value, and handler which injects appropriate objects into the root model) it can lead to order-of-operation errors (because the binding will be re-evaluated in some engine-specific order, so if the model binding re-evaluates before the  signal handler, then there is a transiently invalid state).

chriadam: But I may be wrong in this particular case, not sure

data/common/AcInputPhaseModel.qml
```diff
@@ -0,0 +1,52 @@
...
+			readonly property VeQuickItem _power: VeQuickItem {
+				uid: Global.system.serviceUid + "/Ac/ActiveIn/L" + (model.index + 1) + "/Power"
+				onValueChanged: root._updatePhaseValue(model.index, "power", value)
+			}
+		}
+	}
+
+	readonly property VeQuickItem _numberOfPhases: VeQuickItem {
+		uid: Global.system.serviceUid + "/Ac/ActiveIn/NumberOfPhases"
+		onValueChanged: {
```
chriadam, line 42: I hope this happens very infrequently?

data/common/AcInputServiceLoader.qml
```diff
@@ -196,9 +90,8 @@ Loader {
 	Component {
 		id: gridOrGensetComponent
 
-		Instantiator {
-			id: phaseObjects
-
+		QtObject {
+			readonly property bool hasPower: true
```
chriadam, line 94: should this be something different than just true?

data/common/ActiveAcInput.qml
```diff
@@ -0,0 +1,89 @@
...
+	}
+
+	// Phase measurements from com.victronenergy.system/Ac/ActiveIn/L<1|2|3>
+	readonly property AcInputPhaseModel _phases: AcInputPhaseModel {
+		id: _phases
+
+		property real totalPower: NaN
+		property real firstPhaseCurrent: NaN
+
+		onPhaseValueChanged: function(phaseIndex, propertyName, propertyValue) {
```
chriadam, line 57: is this run frequently?  If so, we should try to move it into C++

data/AcInputs.qml
blammit: Good point, no it doesn't. Fixed.

data/AcInputs.qml
blammit: Ah yes, fixed

data/common/AcInputPhaseModel.qml
blammit: Yes it's deliberately imperative so that the Instantiator is loaded with the correct number of objects, instead of having some intermediate state where instantiator.count != _numberOfPhases.value.

data/common/AcInputPhaseModel.qml
blammit: Seems to be infrequent in practice

data/common/AcInputServiceLoader.qml
blammit: Yes, looks weird. Removed.

data/common/ActiveAcInput.qml
blammit: It does, and we have the same issue with the AC loads phases. I think we can fix them together when the AC loads gauges are reworked.

PR #613, opened 2023-12-18
Write a code generator for the Theme object

chriadam:


tools/themeparser/CMakeLists.txt
```diff
@@ -0,0 +1,45 @@
+cmake_minimum_required(VERSION 3.16)
+
+project(themeparser VERSION 0.1 LANGUAGES CXX)
+
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+
+find_package(Qt6 6.5 REQUIRED COMPONENTS Quick ShaderTools)
```
chriadam, line 7: oops.  doesn't need either of these, just needs gui

CMakeLists.txt
```diff
@@ -66,6 +66,12 @@ if(QT_KNOWN_POLICY_QTP0001)
     qt_policy(SET QTP0001 NEW) # >Qt6.5 only. Enabling this policy ensures that your QML module is placed under a default import path, and its types can be found without manual calls to QQmlEngine::addImportPath.
 endif()
 
+# Not actually a subproject, but we need to build it somehow...
+# Eventually, we should implement the code generator as a python script
+# and invoke it directly via add_custom_command() + add_custom_target().
+# For now, we require the developer to run the code generator manually.
+add_subdirectory("tools/themeparser")
```
chriadam, line 73: should be commented out by default, as it's a developer tool and not needed for the "actual" build.

tools/themeparser/themeparser.cpp
```diff
@@ -0,0 +1,735 @@
...
+		Ok = 0,
+		Warning,
+		Critical
+	};
+	Q_ENUM(StatusLevel)
+
+	ThemeBase(QObject *parent = nullptr) : QObject(parent) {}
+
+	ScreenSize screenSize() const { return m_screenSize; }
+	ColorScheme colorScheme() const { return m_colorScheme; }
```
chriadam, line 82: I need to implement setters and emit the change signals, too.

DanielMcInnes:
@chriadam do you want this reviewed now, or after the cmake changes go in / rebasing /testing?

chriadam:
After.  I haven't tested anything so I'm sure there are a bunch of errors I need to fix, after I can test it properly.

chriadam:
Ready for review now.

blammit (commented):
As discussed:

- would be nicer if ThemeObjects didn't reproduce the existing code in theme.h
- current theme.h contents should probably be renamed to be in a `VenusApp` namespace or something instead (something to do separately)
- themeparser should be implemented as a Python script so that it can be run at compile time. Can you make a task to track this?

chriadam:
You're right, I will move the ThemeBase class declaration into Theme.h.

In future, we can look at renaming stuff (e.g. separate singletons for the theme-constants properties, versus the application state properties like color scheme, methods for various functionality, etc), but for now I will keep as-is.

I have created https://github.com/victronenergy/gui-v2/issues/625 to track the "rewrite as a python script" task.

chriadam:
Done.

blammit (approved):
As discussed, the json constants can also be renamed to use underscores instead of dots, so that the json names are the same as the names used in the source code.

Can done separately, LGTM.

Issue #615, opened 2023-12-18
Device List: PV Inverter issues


mr-manuel:
## Settings -> Device List -> PV Inverter

Phases without values are not hidden. I saw that `/NrOfPhases` is checked, but what, if the PV Inverter is only 1 phase, but connected to `L3`?

https://github.com/victronenergy/gui-v2/blob/8e57400d3fc631aa55f794cbd0cd62ae1ab29a1d/pages/settings/devicelist/ac-in/PageAcInModelDefault.qml#L57-L84

![grafik](https://github.com/victronenergy/gui-v2/assets/12592765/cc1c4974-4572-45a5-9ccb-71b26e845f46)

## Overview -> Solar Yield -> PV Inverter

The position is not shown correctly, since there is a typo in the code.

https://github.com/victronenergy/gui-v2/blob/8e57400d3fc631aa55f794cbd0cd62ae1ab29a1d/components/settings/PvInverterPositionRadioButtonGroup.qml#L14

![grafik](https://github.com/victronenergy/gui-v2/assets/12592765/05d6881f-2fc0-4a67-84e2-040d225cae9a)

mr-manuel:
https://github.com/victronenergy/gui-v2/pull/617#discussion_r1434713505

> @jepefe what do you think, should the entire phase data row be hidden in the list if the phase does not have a valid 'power' value? Same for 'energy' below. This would differ from the gui-v1 behavior.

Opened new PR for "PV Inverter hide phases without values" which has to be discussed #653

PR #617, opened 2023-12-18
Device List: PV Inverter issues

mr-manuel:
Fixes https://github.com/victronenergy/gui-v2/issues/615

### DIFF
--- a/components/settings/PvInverterPositionRadioButtonGroup.qml
+++ b/components/settings/PvInverterPositionRadioButtonGroup.qml
@@ -11,6 +11,6 @@ ListRadioButtonGroup {
 	optionModel: [
 		{ display: CommonWords.ac_input_1, value: VenusOS.PvInverter_Position_ACInput },
 		{ display: CommonWords.ac_input_2, value: VenusOS.PvInverter_Position_ACInput2 },
-		{ display: CommonWords.ac_output, value: VenusOS.PvInverter_Position_Output },
+		{ display: CommonWords.ac_output, value: VenusOS.PvInverter_Position_ACOutput },
 	]
 }


mr-manuel:
Ready for review :-)

pages/settings/devicelist/ac-in/PageAcInModelDefault.qml
```diff
@@ -79,6 +79,8 @@ ObjectModel {
 					id: phasePower
 					source: root.bindPrefix + "/Ac/L" + (model.index + 1) + "/Power"
 				}
+
+				visible: phasePower.valid
```
blammit, line 83: It looks like gui-v1 would still show this phase entry in the list if the phase is present (according to `NrOfPhases`) even if the data is not valid, so in that sense it would be better to keep this.

mr-manuel: Ok, then can we change it to if the setting is present?

I will also change https://github.com/victronenergy/venus/issues/1091

blammit: Which setting is that? Do you mean NrOfPhases? With the current code, phase 2 and 3 will not be shown if NrOfPhases=1.

mr-manuel: Theoretically and practically if `phasePower.value` is valid and not None, then the value should be shown, else not.

blammit: @jepefe what do you think, should the entire phase data row be hidden in the list if the phase does not have a valid 'power' value? Same for 'energy' below.

blammit: @mr-manuel if you want to put f92438ea834e65a516bc5b317e231add3896559a in a separate PR that could be merged separately already.

mr-manuel: Sure, I removed the commit and will open a new PR after you merge this one.

mr-manuel: @blammit could you merge now?

blammit: LGTM

blammit (approved):

pages/settings/devicelist/ac-in/PageAcInModelDefault.qml
jepefe: > @jepefe what do you think, should the entire phase data row be hidden in the list if the phase does not have a valid 'power' value? Same for 'energy' below. This would differ from the gui-v1 behavior.

This was discussed for gui-v1 not so long ago, and checking the code in commit https://github.com/victronenergy/gui/commit/b4c03f5f63315e973b95a83c9caba80cc4f73061 it was decided to assume that there are 3 phases when NumberOfPhases is not provided.

There are situations were the Number Of Phases is not provided and the power can be temporarily invalid so we can't just hide those lines.

Please, keep same behaviour than gui-v1.

PR #622, opened 2023-12-19
596 device list add sub pages for comvictronenergymeteoxxx

DanielMcInnes:


pages/settings/devicelist/PageMeteo.qml
```diff
@@ -0,0 +1,80 @@
...
+
+	DataPoint {
+		id: deviceInstance
+
+		source: bindPrefix + "/DeviceInstance"
+	}
+
+	GradientListView {
+		model: ObjectModel {
+			ListTextItem {
```
blammit, line 24: Should be a ListQuantityItem instead? To get the different font coloring for the unit etc.

Same for all the quantities below.

blammit (approved):

Issue #580, opened 2023-11-30
Generator auto-start toggle not working in control cards


blammit:
Reported issue with generator control cards:

- can stop the autostart but can't toggle it on from the device controls
- the "click" event is disabled or broken,  if you drag the knob, then it works

Issue #520, opened 2023-11-12
Generator start/stop alert not dismissing automatically


kwindrem:
If you start or stop the generator in the switch page, a message is displayed at the bottom of the screen. This message blocks the page selector at the bottom of the screen when going back to the other pages:

![Screenshot 2023-11-12 at 9 01 14 AM](https://github.com/victronenergy/gui-v2/assets/58538395/4709f645-1c48-48ff-9f0c-546905b793ea)

I would expect this message to disappear after a few seconds or disappear when switching pages.

Maybe the answer is to make it more obvious that the Start/Stop button is for manual run. The original did this, calling it MANUAL START / MANUAL STOP. These buttons ARE in the manual control area of the tile but maybe not obvious enough to avoid the confusion of having the generator not stop if auto run conditions are still active.

Also, in playing with this a little I've also come across situations where the controls in the Manual control section were not responding to clicks. Because of this I was never able to setup a timed manual run.

Issue #518, opened 2023-11-12
Generator tile in overview not updating with generator status changes


kwindrem:
The Generator tile is not updating the generator running/stopped state or the running time properly. Also at times, the tile remains on Grid/Shore even though the system AC input type is changed to generator.

The state displayed here should also include warm-up, cool-down and post-cool-down periods especially beause the AC input is ignored in these states. This leaves the user to wonder why if the generator is running it is not being used.

blammit:
> The Generator tile is not updating the generator running/stopped state or the running time properly. 

With https://github.com/victronenergy/gui-v2/pull/623, the Generator tile will only show a "stopped" status when stopped (and otherwise will just show the power input). When running, the detailed reason will be shown in the Generator "card" (shown when clicking the top-left buttons). The PR also adds the warm up / cooling down statuses.

>  Also at times, the tile remains on Grid/Shore even though the system AC input type is changed to generator.

This should be fixed by https://github.com/victronenergy/gui-v2/pull/609, but would be good to know if this is not the case.

PR #623, opened 2023-12-19
Various generator fixes and improvements

blammit:
This addresses various issues:

https://github.com/victronenergy/gui-v2/issues/580
https://github.com/victronenergy/gui-v2/issues/520
https://github.com/victronenergy/gui-v2/issues/518

Note that PageGenerator.qml doesn't work properly with MQTT at the moment. This will be fixed when the MQTT uid conversion is fixed.

### DIFF
--- a/components/dialogs/GeneratorDialog.qml
+++ b/components/dialogs/GeneratorDialog.qml
@@ -1,0 +1,136 @@
+/*
+** Copyright (C) 2023 Victron Energy B.V.
+** See LICENSE.txt for license information.
+*/
+
+import QtQuick
+import Victron.VenusOS
+import Victron.Utils
+
+
+// This dialog allows the generator to be started or stopped. It is does not close immediately
+// when the 'accept' button is clicked to start/stop the generator. Instead, it shows a color
+// animation over a short duration, and while the animation is running, the user can elect to cancel
+// the start/stop action.
+//
+// When the dialog is opened, the 'accept' button shows either 'Start' or 'Stop' to start/stop the
+// generator, and the 'reject' button shows 'Close'. When the Start/Stop button is clicked:
+//  - A color animation slides slowly over the button background
+//  - The 'reject' button text changes from 'Close' to 'Cancel'. If the button is clicked at
+//    this point, the color animation stops, the dialog does NOT close, and the reject button
+//    text changes again to 'Close'.
+//  - If the Cancel button is not clicked, when the color animation reaches the end of the
+//    button, the start/stop action is triggered. When the action completes, the dialog is closed.
+//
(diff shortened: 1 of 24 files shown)


components/dialogs/GeneratorDialog.qml
```diff
@@ -0,0 +1,103 @@
...
+*/
+
+import QtQuick
+import Victron.VenusOS
+import Victron.Utils
+
+ModalDialog {
+	id: root
+
+	property var generator
```
chriadam, line 13: can this one also be a Generator rather than var?

components/dialogs/GeneratorDialog.qml
```diff
@@ -0,0 +1,103 @@
...
+			},
+			State {
+				name: "tryingToAccept"
+				PropertyChanges {
+					target: root
+					rejectText: root.rejectTextCancel
+					canAccept: false
+					tryReject: function() {
+						acceptButtonBackground.animating = false
+						acceptButtonBackground.state = "default"
```
chriadam, line 57: It's strange to me that the tryReject() function has side-effects but then always returns false here.
It's even stranger that it's occurring from within a "tryingToAccept" state -> I don't really understand how someone can reject a dialog if they are currently trying to accept a dialog.

Could a comment be added here to explain what part of the flow this is?

components/dialogs/GeneratorDialog.qml
```diff
@@ -0,0 +1,103 @@
...
+				}
+			}
+		]
+	}
+
+	onAboutToShow: {
+		acceptButtonBackground.state = "default"
+	}
+
+	Connections {
```
chriadam, line 94: I really dislike connections.  I suspect that since the target can be set at runtime, it can't be easily inspected ahead of time...  Is there a different way we can do this?
e.g.: `property bool _acceptConditionMet: !!root.generator && root.open && root.generator.state === root.finalGeneratorState ; on_AcceptConditionMetChanged: if (_acceptConditionMet) root.accept()`

blammit: Sure, changed to bind to generator state instead

components/dialogs/GeneratorDialog.qml
blammit: Done

components/dialogs/GeneratorDialog.qml
blammit: Agreed, a comment would be good to explain what's going here. Added one at the top of the file to explain the state changes and interactions.

chriadam (approved):
LGTM, would be good if Daniel could check also.

PR #633, opened 2023-12-22
Remove use of /ServiceMapping, and remove UidHlper and SingleUidHelper

blammit:


blammit:
Building this with -DNO_CACHEGEN=true appears to make the app crash on desktop Mac and Linux when pushing some pages onto the stack (e.g. when clicking on AC or DC widgets from Overview or Device List). Doesn't occur if I remove the build flag. Pretty strange.

data/System.qml
```diff
@@ -46,9 +46,15 @@ QtObject {
...
 	property SystemDc dc: SystemDc {}
 
-	property QtObject veBus: QtObject {
-		property string serviceUid
-		property real power
+	readonly property QtObject veBus: QtObject {
+		readonly property string serviceUid: BackendConnection.type === BackendConnection.MqttSource
+				? (_deviceInstance.isValid ? "mqtt/vebus/" + _deviceInstance.value : "")
+				: (_serviceName.isValid ? BackendConnection.uidPrefix() + "/" + _serviceName.value : "")
+		readonly property real power: _power.value === undefined ? NaN : _power.value
```
chriadam, line 53: RHS could be simplified to: `_power.value || NaN` I think

data/mock/AcInputsImpl.qml
```diff
@@ -63,11 +63,17 @@ QtObject {
 		bindPrefix: Global.system.serviceUid + "/Ac/In/0"
 		Component.onCompleted: {
 			_deviceInstance.setValue(300)
-			_serviceName.setValue("com.victronenergy.vebus.tty01")
 			_serviceType.setValue("vebus")
 			_connected.setValue(1)
 		}
 
+		property Connections _serviceNameSetup: Connections {
```
chriadam, line 70: I really dislike Connections.
Can we just use:
```
property string vebusServiceUid: Global.system.veBus.serviceUid
onVebusServiceUidChanged: root.inputSysInfo._serviceName.setValue(vebusServiceUid)
```

chriadam: Well, it's just in mock, never mind :-)

pages/settings/debug/PagePowerDebug.qml
```diff
@@ -67,7 +67,20 @@ Page {
 	}
 
 	GradientListView {
-		model: ObjectModel {
+		// Hardcoded com.victronenergy.pvinverter.* uids do not work on MQTT, so the page has no
```
chriadam, line 70: I don't understand this comment - are you saying that there is a bug in (venus platform or something) which populates something with the wrong paths, and we can't map those to MQTT equivalents?

chriadam (approved):
A couple of nits but LGTM.

data/System.qml
blammit: Not in this instance, because if `_power.value === 0` (i.e. coerces to false) then the value should be `0` rather than `NaN`

data/mock/AcInputsImpl.qml
blammit: Updated to use the property. Wondering why you prefer to avoid Connections? Is it due to the extra QObject? Sometimes I'm also not sure whether to use property vs Connections.

pages/settings/debug/PagePowerDebug.qml
blammit: Not a bug in venus-platform etc, more that a path like `com.victronenergy.pvinverter.qwacs_di1` is hardcoded in D-Bus format, i.e. we can't automatically translate it to `mqtt/pvinverter/256` or such, as there's no way to figure out which pvinverter is referred to by the `qwacs_di1` part of the uid. But this is okay as the page is just for debugging.

PR #635, opened 2023-12-22
Remove DataPoint

blammit:
Now that the codebase always provides MQTT-specific uids, DataPoint is no longer required for converting D-Bus uids to MQTT uids, so VeQuickItem can always be used instead.


components/listitems/ListRangeSlider.qml
```diff
@@ -72,20 +62,20 @@ ListItem {
 				target: slider.second
 				function onPressedChanged() {
 					// Update data value when mouse is released, to avoid spamming data changes.
-					if (!slider.second.pressed && secondDataPoint.source.length > 0) {
+					if (!slider.second.pressed && secondDataItem.source.length > 0) {
```
chriadam, line 65: should this be `.uid.length` ?

components/listitems/ListTextField.qml
```diff
@@ -51,13 +46,13 @@ ListItem {
...
 		enabled: root.enabled
-		text: dataValid ? dataValue : ""
+		text: dataItem.isValid ? dataItem.value : ""
 
 		onAccepted: {
 			let newValue = text
 			_accepted = true
-			if (dataPoint.source) {
-				dataPoint.setValue(newValue)
+			if (dataItem.uid) {
```
chriadam, line 54: should this be `if (dataItem.uid.length)` ?

data/SystemSettings.qml
```diff
@@ -70,48 +71,48 @@ QtObject {
 
 	property QtObject electricalQuantity: QtObject {
 		// Values for /Settings/Gui/ElectricalPowerIndicator: 0 = watts, 1 = amps
-		readonly property var value: _electricalQuantityDataPoint.value === 1 ? VenusOS.Units_Amp : VenusOS.Units_Watt
+		readonly property var value: _electricalQuantityDataItem.value === 1 ? VenusOS.Units_Amp : VenusOS.Units_Watt
```
chriadam, line 74: unrelated to this PR, but can we narrow this type?  var seems suboptimal here

pages/settings/PageSettingsHub4.qml
```diff
@@ -167,10 +168,10 @@ Page {
 
 			//% "BatteryLife state"
 			text: qsTrId("settings_ess_batteryLife_state")
-			dataSource: Global.systemSettings.serviceUid + "/Settings/CGwacs/BatteryLife/State"
+			dataItem.uid: Global.systemSettings.serviceUid + "/Settings/CGwacs/BatteryLife/State"
 			visible: defaultVisible
 				&& essMode.value !== VenusOS.Ess_Hub4ModeState_Disabled
-				&& Global.ess.isBatteryLifeActive(dataValue)
+				&& Global.ess.isBatteryLifeActive(dataItem.value)
```
chriadam, line 174: nitpick: I'd prefer `batteryLifeState.dataItem.value` rather than just `dataItem.value` just in case a dataItem property is added somewhere in the parent object scope.

pages/settings/PageSettingsHub4Feedin.qml
```diff
@@ -71,9 +72,9 @@ Page {
 				text: qsTrId("settings_ess_feed_in_limiting_active")
 				visible: defaultVisible
 					&& hub4Mode !== VenusOS.Ess_Hub4ModeState_Disabled
-					&& dataValid
-				dataSource: BackendConnection.serviceUidForType("hub4") +"/PvPowerLimiterActive"
-				secondaryText: CommonWords.yesOrNo(dataValue)
+					&& dataItem.isValid
+				datauid: BackendConnection.serviceUidForType("hub4") +"/PvPowerLimiterActive"
+				secondaryText: CommonWords.yesOrNo(dataItem.value)
```
chriadam, line 77: nitpick: I'd prefer adding an id to this ListTextItem and then resolving `feedInLimitingActive.dataItem.value` or whatever

chriadam:
Mostly LGTM.  Feel free to ignore the nits.  Two things which do need fixing, I think.
Tested on device, works nicely.  Really great to get rid of DataPoint etc!  Nice work!

components/listitems/ListRangeSlider.qml
blammit: Oops, fixed

components/listitems/ListTextField.qml
blammit: Yes, thanks, fixed

data/SystemSettings.qml
blammit: Also fixed

pages/settings/PageSettingsHub4.qml
blammit: Done

pages/settings/PageSettingsHub4Feedin.qml
blammit: Done

chriadam (approved):

PR #638, opened 2024-01-02
634 allow non bundled gui v2 builds

DanielMcInnes:
The purpose of these changes is to allow cerbo users to edit qml sources directly on the cerbo, restart the app, and see the effect of their changes.
Previously, all qml resources were compiled into the venus-gui-v2 executable. Now, all qml resources are loaded from the filesystem.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,10 +7,28 @@ cmake_minimum_required(VERSION 3.24) # earlier versions of cmake don't handle 'q
 # CMAKE HINTS
 #
 # Different build types:
-#   cmake -DCMAKE_BUILD_TYPE=[Debug|Release|RelWithDebInfo|MinSizeRelease]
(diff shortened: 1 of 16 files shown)


chriadam:
Would it be useful to base this PR on top of some of the work you did for the qmltc builds first (i.e. especially I'm thinking of the changes you made to ensure that all types came from a single plugin, rather than multiple)?

DanielMcInnes:
> Would it be useful to base this PR on top of some of the work you did for the qmltc builds first (i.e. especially I'm thinking of the changes you made to ensure that all types came from a single plugin, rather than multiple)?

I found that putting all types in the same plugin had negligible effect. I tested this branch on a cerbo, and once it's started, there's no real noticeable difference in performance. Trace attached.
[trace.zip](https://github.com/victronenergy/gui-v2/files/13815915/trace.zip)

chriadam (approved):

PR #640, opened 2024-01-03
Patch qt-configure-module

dirkjanfaber:
Fix the reversed slashes in this file, which caused builds to fail.


DanielMcInnes (approved):
lgtm. I don't understand why this suddenly started causing problems - do you know? It was working fine before Christmas.

chriadam (approved):
Strange, but LGTM

PR #644, opened 2024-01-08
Use python script to do json theme conversion

blammit:


tools/themeparser.py
```diff
@@ -0,0 +1,263 @@
+import json
```
DanielMcInnes, line 1: does this file need a copyright notice of its own?

src/themeobjects.h
```diff
@@ -52,6 +29,192 @@ class ThemeSingleton : public Theme
 	QML_SINGLETON
 
 	// property declarations
+	Q_PROPERTY(int animation_acceptButtonBackground_expand_duration READ animation_acceptButtonBackground_expand_duration CONSTANT FINAL)
```
DanielMcInnes, line 32: can we remove this file from source control, seeing as it is auto-generated?

DanielMcInnes:
maybe run this by @dirkjanfaber , just so he's aware of the added dependency on python. It should be ok, seeing as the CI job went smoothly

DanielMcInnes (approved):
lgtm

tools/themeparser.py
chriadam: yes please

src/themeobjects.h
chriadam: We definitely should remove it from source control, I think.

But, do we need some cmake trick, to avoid it complaining that the target relies on a file which doesn't exist, or something?

chriadam:
Great work!

src/themeobjects.h
DanielMcInnes: no, I tried deleting it, and it built fine

blammit: Makes sense, removing this from source control and added src/themeobjects.h to .gitignore.

tools/themeparser.py
blammit: Done

chriadam (approved):
LGTM

dirkjanfaber:
> maybe run this by @dirkjanfaber , just so he's aware of the added dependency on python. It should be ok, seeing as the CI job went smoothly

Checked and lgtm. Python is available on the build server.

PR #645, opened 2024-01-08
Show BMS, charger and wakespeed errors

blammit:
Now that BmsError, ChargerError and WakespeedError are available in veutil, use them to show text descriptions of relevant errors.


src/main.cpp
```diff
@@ -252,6 +259,10 @@ int main(int argc, char *argv[])
 	initBackend(&enableFpsCounter);
 	QObject::connect(&engine, &QQmlEngine::quit, &app, &QGuiApplication::quit);
 
+	qmlRegisterSingletonInstance("Victron.Veutil", 1, 0, "BmsError", bmsError.get());
```
chriadam, line 262: I'd prefer not to do it this way, since then it won't be known by the QML compiler at compile time (since the types and value are only registered at runtime).

Can we add QML_ELEMENT and QML_SINGLETON to BmsError / ChargerError / WakespeedError types in veutil instead?

(Or, have our own shim types defined in gui-v2 which e.g. extend BmsError and inject those macros into the definition?)

chriadam:
As discussed offline, we should wait for Daniel's PRs to be merged first, so that this can make use of QML_ELEMENT + QML_SINGLETON more easily.

DanielMcInnes (approved):
lgtm, once Chris's comments re. QML_ELEMENT are resolved

PR #647, opened 2024-01-08
Improve conversion to preferred units and use veutil for conversions

blammit:
Rework the unit conversion handling and the API for the user-preferred unit access:

- Make temperature, volume and electrical quantity units accessible as Global.systemSettings 'temperatureUnit', 'volumeUnit' and 'electricalQuantity' integers instead of as VeQuickItem objects, so that they can be accessed as e.g. Global.systemSettings.volumeUnit instead of as Global.systemSettings.volumeUnit.value
- Add convenience conversion functions to Global.systemSettings e.g. convertFromCelsius() and convertFromCubicMeters() to convert to/from the user-preferred unit
- Use veutil unit_conversion.hpp for all unit conversions, instead of implementing the conversion in unit.cpp
- Ensure settings pages show temperature in user-preferred unit
- Fix TemperatureRelaySettings.qml to ensure temperature settings are saved in celsius

### DIFF
--- a/pages/settings/devicelist/dc-in/PageDcMeterModel.qml
+++ b/pages/settings/devicelist/dc-in/PageDcMeterModel.qml
@@ -44,11 +44,9 @@ ObjectModel {
 		}
 	}
 
-	ListQuantityItem {
+	ListTemperatureItem {
 		text: CommonWords.temperature
 		dataItem.uid: root.bindPrefix + "/Dc/0/Temperature"
-		value: Global.systemSettings.convertFromCelsius(dataItem.value)
-		unit: Global.systemSettings.temperatureUnit
 		visible: defaultVisible && dataItem.isValid
 	}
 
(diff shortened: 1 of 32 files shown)


pages/settings/devicelist/dc-in/PageDcMeterModel.qml
```diff
@@ -47,8 +47,8 @@ ObjectModel {
 	ListQuantityItem {
 		text: CommonWords.temperature
 		dataItem.uid: root.bindPrefix + "/Dc/0/Temperature"
-		value: dataItem.isValid ? Global.systemSettings.convertTemperature(dataItem.value) : NaN
-		unit: Global.systemSettings.temperatureUnit.value
+		value: Global.systemSettings.convertFromCelsius(dataItem.value)
```
DanielMcInnes, line 50: There are ~11 ListQuantityItems with:
```
    value: Global.systemSettings.convertFromCelsius(dataItem.value)
    unit: Global.systemSettings.temperatureUnit
```
Do you think it's worth adding a 'ListTemperatureItem' type that encapsulates this?

DanielMcInnes (approved):
lgtm

pages/settings/devicelist/dc-in/PageDcMeterModel.qml
blammit: Yes, that's much nicer. I saw you added a ListTemperatureItem earlier, I'll update the code to use that.

PR #648, opened 2024-01-09
Performance improvements and bug fixes

DanielMcInnes:
Reduce the number of qml modules to a minimum. Fix qml bugs found while investigating the qml type compiler.

I couldn't get rid of the Gauges module, as it imports Victron.VenusOS, and causes a circular dependency.
I couldn't include the changes I did for moving from QtQuick.Controls to Templates, as they caused build issues in some configurations. If NO_CACHEGEN was set to false(the default), then qmlcachegen would hang during the build. Killing the build did not kill qmlcachegen, I had to kill it via task manager.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -649,7 +649,7 @@ set(Dbus_QML_MODULE_SOURCES
 
 list(APPEND TRANSLATION_SOURCES ${Dbus_QML_MODULE_SOURCES})
 
-qt_add_qml_module(Dbus
+qt_add_qml_module(VictronDbus
     URI Victron.Dbus
     STATIC
     OUTPUT_DIRECTORY Victron/Dbus
@@ -696,7 +696,7 @@ set(Mock_QML_MODULE_SOURCES
 
 list(APPEND TRANSLATION_SOURCES ${Mock_QML_MODULE_SOURCES})
 
-qt_add_qml_module(Mock
+qt_add_qml_module(VictronMock
     URI Victron.Mock
     STATIC
     OUTPUT_DIRECTORY Victron/Mock
@@ -712,7 +712,7 @@ set(Gauges_QML_MODULE_SOURCES
 
 list(APPEND TRANSLATION_SOURCES ${Gauges_QML_MODULE_SOURCES})
 
-qt_add_qml_module(Gauges
+qt_add_qml_module(VictronGauges
     URI Victron.Gauges
     STATIC
     OUTPUT_DIRECTORY Victron/Gauges
@@ -752,7 +752,7 @@ set(Mqtt_QML_MODULE_SOURCES
 
 list(APPEND TRANSLATION_SOURCES ${Mqtt_QML_MODULE_SOURCES} )
 
-qt_add_qml_module(Mqtt
+qt_add_qml_module(VictronMqtt
     URI Victron.Mqtt
     STATIC
     OUTPUT_DIRECTORY Victron/Mqtt
@@ -965,10 +965,10 @@ target_link_libraries(${PROJECT_NAME} PRIVATE
     Qt6::Xml
     Qt6::Mqtt
     VenusQMLModuleplugin
-    Gaugesplugin
-    Dbusplugin
-    Mockplugin
-    Mqttplugin
+    VictronGaugesplugin
+    VictronDbusplugin
+    VictronMockplugin
+    VictronMqttplugin
 )
 
 if("${CMAKE_SYSTEM_NAME}" STREQUAL "Emscripten")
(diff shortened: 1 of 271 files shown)


chriadam:
We would definitely like veutil in a proper module, so that we can add QML_ELEMENT and QML_SINGLETON to the veutil types without requiring runtime registration.  I don't really understand why the veutil types would cause any issues with the build system - after all, as far as it's concerned, they're just "local" C++ files which declare QObject-derived types etc, right?

chriadam:
Can you check what the type of Global.dataServiceModel is?  If its type is "var" then that might be the problem.  We could narrow the type there.

DanielMcInnes:
> Can you check what the type of Global.dataServiceModel is? If its type is "var" then that might be the problem. We could narrow the type there.

I tried this, it changes the error message from:
`qrc:/qt/qml/Victron/Dbus/data/dbus/VeBusDevicesImpl.qml:30:4: Unable to assign QJSValue to VeQItemTableModel*)`
to
`qrc:/qt/qml/Victron/Dbus/data/dbus/VeBusDevicesImpl.qml:30: TypeError: Type error`

DanielMcInnes:
> We would definitely like veutil in a proper module, so that we can add QML_ELEMENT and QML_SINGLETON to the veutil types without requiring runtime registration. I don't really understand why the veutil types would cause any issues with the build system - after all, as far as it's concerned, they're just "local" C++ files which declare QObject-derived types etc, right?

Done

Global.qml
```diff
@@ -6,6 +6,7 @@
 pragma Singleton
 
 import QtQml
+import Victron.VenusOS
```
chriadam, line 9: if QTBUG-66976 is now fixed and we can import Victron.VenusOS here without it breaking the wasm build, then we should remove the comment below (or at least change it to say "this is the reason we did this, although the bug has since been fixed").

chriadam:
OOI what is special about the Victron.Gauges import such that it can't be folded into Victron.VenusOS namespace?

Global.qml
DanielMcInnes: QTBUG-66976 is now fixed, and the wasm build is ok. Removed comment.

CMakeLists.txt
```diff
@@ -610,7 +647,7 @@ set(Dbus_QML_MODULE_SOURCES
     data/dbus/VeBusDevicesImpl.qml
 )
 
-list(APPEND QML_MODULE_SOURCES ${Dbus_QML_MODULE_SOURCES})
+list(APPEND TRANSLATION_SOURCES ${Dbus_QML_MODULE_SOURCES})
 
 qt_add_qml_module(Dbus
```
chriadam, line 652: I think this module name should be something like VictronDbus so that the static library generated for wasm builds is libVictronDbus.a rather than libDbus.a, otherwise there is the potential for nasty conflicts.  Same for the Mqtt and Mock modules.

DanielMcInnes:
> OOI what is special about the Victron.Gauges import such that it can't be folded into Victron.VenusOS namespace?

It seems you can't have a .js library in module 'X' that imports module X. A .qml singleton works ok though, see Global.qml.

It builds ok, but at runtime, you get the following error:

```
root@einstein:~# ./venus-gui-v2 
Victron gui version: v0.2.3
Connecting to system bus...
Connecting to deprecated signal QDBusConnectionInterface::serviceOwnerChanged(QString,QString,QString)
BackendConnection state: Victron::VenusOS::BackendConnection::Ready
venus.gui: Unable to load translations for locale "C"
venus.gui: Falling back to English as locale catalogue failed to load.
venus.gui: Successfully loaded translations for locale "en_US"
qt.qml.typeresolution.cycle: Cyclic dependency detected between "qrc:/qt/qml/Victron/VenusOS/components/Gauges.js" and "qrc:/qt/qml/Victron/VenusOS/components/Gauges.js"
"qrc:/venus-gui-v2/Main.qml:71 Type DataManager unavailable\nqrc:/qt/qml/Victron/VenusOS/data/DataManager.qml:-1 Script qrc:/qt/qml/Victron/VenusOS/components/Gauges.js unavailable\n"
root@einstein:~# ```

If you remove the `import Victron.VenusOS`, it throws ReferenceErrors about undefined symbols.

chriadam:
Ok.  In future we should move the .js to .cpp anyway, and at that point we can fold it all into VenusOS import, I guess.  But not for this PR.

CMakeLists.txt
DanielMcInnes: Done

chriadam (approved):
LGTM, nice work.

PR #656, opened 2024-01-12
Device list pulsemeter

blammit:


pages/settings/PageSettingsDisplayUnits.qml
```diff
@@ -45,7 +45,11 @@ Page {
 				}
 			}
 
-			VolumeUnitRadioButtonGroup { }
+			VolumeUnitRadioButtonGroup {
+				//: Title for a list of units of volume (e.g. cubic meters, liters, gallons)
+				//% "Volume"
```
DanielMcInnes, line 50: CommonWords?

DanielMcInnes (approved):
lgtm

pages/settings/PageSettingsDisplayUnits.qml
blammit: The other 'Volume' is unrelated and specifies something about tank shapes, so I've left this as-is

PR #670, opened 2024-01-19
Fix crash in wasm build

DanielMcInnes:
When toast notifications are closed, they would destroy themselves, which causes problems for wasm builds.
Now, when we want to close a toast notification, start a timer and let the timer destroy the notificaiton.

### DIFF
--- a/pages/NotificationLayer.qml
+++ b/pages/NotificationLayer.qml
@@ -12,11 +12,21 @@ Item {
 	anchors.fill: parent
 
 	function showToastNotification(category, text, autoCloseInterval = 0) {
-		var toast = toaster.createObject(this, { "category": category, "text": text, autoCloseInterval: autoCloseInterval })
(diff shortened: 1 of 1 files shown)


chriadam (approved):
I don't understand why it's necessary, but LGTM.

pages/NotificationLayer.qml
```diff
@@ -26,13 +36,12 @@ Item {
 			bottomMargin: Theme.geometry_toastNotification_bottomMargin
 		}
 		width: parent.width
+		height: childrenRect.height
```
chriadam, line 39: can you just fill the parent, rather than using childrenRect.height, by any chance?

DanielMcInnes: no, otherwise the ListView covers the bottom bar, and steals the mouse events

PR #671, opened 2024-01-22
Quantity label numbers

blammit:


DanielMcInnes (approved):
lgtm. Just to test my understanding, this gives us a monospaced version of whatever font we plug into it?
Also, the results for `characterImageSize(...)` seem to be cached somewhere so that it doesn't get called more often than it needs to be. Do you know how that works?

chriadam (approved):
LGTM.  In future, we might need to clear the cache if the selected font family changes, I guess?

blammit:
> LGTM. In future, we might need to clear the cache if the selected font family changes, I guess?

Good point, will make a note to check that.

blammit:
> > LGTM. In future, we might need to clear the cache if the selected font family changes, I guess?
> 
> Good point, will make a note to check that.

Actually, the image provider can continue to use MuseoSans as the font for QuantityLabel digits and symbols. gui-v2 only needs to use a different font to render characters that can't be rendered in MuseoSans (e.g. for Arabic, Chinese and Thai at the moment).

PR #673, opened 2024-01-22
Add translations to gui-v2

dirkjanfaber:
Adding 2 scripts. One for uploading the used terms to poeditor.com and one for downloading the translations.
Added the translations for: Arabic, Chinese, Czech, Danish, Dutch, French, German, Italian, Polish, Romanian, Russian, Spanish, Swedish, Thai, Turkish and Ukrainian.  Note that not all terms have been fully translated for all of the languages yet, so expect more PR's to synchronise the translations in the future.
Other minor change in this PR is that Swedish is not the same as Northern Sami, so fixed the code for that. 

Also note that both Chinese and Thai use a different character set that are not included yet. See https://github.com/victronenergy/gui/commit/12872ba74a7a1caac8889f0ec9a32c9cd75b220e for inspiration on how that was added in gui-v1. 

So the Chinese and Thai characters should still be added to these patches, before merging to main.


blammit:
@dirkjanfaber I've added a patch to render Thai text properly, and removes the Chinese language setting as it can't be rendered in wasm, as that font is not bundled.

blammit:
@dirkjanfaber just thinking, for Chinese we could try to find the DroidSansFallback font if available, so that it would still load on the cerbo, even if it doesn't work in wasm. What do you think?

dirkjanfaber:
Well, the font is already on the GX device in `/usr/lib/fonts/DroidSansFallback.ttf`. So I can simply add a symbolic link to it in the directory where the `.wasm` file resides (currently `/var/www/venus/gui-beta`). So you could then load it from the root directory.

blammit:
> Well, the font is already on the GX device in `/usr/lib/fonts/DroidSansFallback.ttf`. So I can simply add a symbolic link to it in the directory where the `.wasm` file resides (currently `/var/www/venus/gui-beta`). So you could then load it from the root directory.

That would be great. Could you also add a symlink for "/usr/lib/fonts/DejaVuSans.ttf"? I discovered that the default MuseoSans font can't render Arabic, but DejaVuSans.ttf can do this just fine.

I've added a commit that looks for `/DroidSansFallback.ttf` in wasm if the language is QLocale::Chinese, and similarly `/DejaVuSans.ttf` for QLocale::Arabic.

dirkjanfaber:
Ok, for this to work a bit easier, I've added a `Makefile` to the zipped webassembly, which allows to do a `make install` when we add the files to the GX device. 

For the VRM portal I've copied the font over to there too and am automatically adding a symbolic link from the same root as on the GX device (so there too, the font will be available on the same directory as the `.wasm` file is) for new installs (this is effictive for 0.2.7 and above).

So the next step is to have this PR merged to the main branch. Then I'll adjust the https://github.com/victronenergy/meta-victronenergy-private/blob/master/recipes-ve/gui-v2-webassembly.bb file to call the `Makefile` during install, basically adding the symbolic link and making future improvements a lot easier.

blammit:
> For the VRM portal I've copied the font over to there too and am automatically adding a symbolic link from the same root as on the GX device (so there too, the font will be available on the same directory as the .wasm file is) for new installs (this is effictive for 0.2.7 and above).

Great, okay then we won't need the extra fonts in the bundle then. Could you / have you done this for `NotoSansThai.ttf` as well? If so, then I'll remove `NotoSansThai.ttf` from the gui-v2 bundle.

dirkjanfaber:
Sure. I've adjusted the `Makefile` to also symlink to `NotoSansThai.ttf` (and put that font also on the VRM portal).

If ok, then I think that @chriadam  should merge this into master before creating the next release.

blammit:
Updated to remove NotoSansThai.ttf since that is not supplied in wasm as well.

blammit:
@dirkjanfaber I noticed the `Makefile` EXTRAFONTS doesn't contain `DejaVuSans.ttf` which is required for rendering Arabic. Could you add that as well? And the symlink for it?

dirkjanfaber:
Sure. Added `DejaVuSans.ttf` to the Makefile and put it on the VRM portal.

blammit:
@chriadam or @DanielMcInnes could you also have a look at this before we merge?

scripts/download_from_poeditor.sh
```diff
@@ -0,0 +1,26 @@
...
+then
+  echo "Please set environment variable POEDITOR_TOKEN first"
+  exit 1
+fi
+
+sed -ne 's/^\s*addLanguage("\([^"]*\)", "\([^"]*\)".*/\2 \1/p' src/language.cpp |\
+grep -v "English" |\
+while read -r code language
+do
+  output_file="i18n/venus-gui-v2_${code}.ts"
```
chriadam, line 13: So, just checking: this parses the language.cpp and downloads the appropriate .ts file from the server, and puts it into `(source directory)/i18n/` ?

Does this script get run for both wasm builds and for device builds?

If so, maybe we can remove the .ts files from git, since they will magically appear prior to build anyway?

CMakeLists.txt
```diff
@@ -935,16 +934,44 @@ qt_add_resources(${PROJECT_NAME} "${PROJECT_NAME}_large_resources"
 # That way if new entries are added in code, the updates don't pollute the git diff.
```
chriadam, line 934: obviously, the script which grabs the .ts files from the poeditor server MUST run before the cmake build, otherwise these files may not exist.  (assuming that we remove all of the .ts files from git)

chriadam:
LGTM, except for one thing: I think we should not check into git the full .ts files, but instead just a minimal template/placeholder (similar to the current venus-gui-v2_fr.ts file)

CMakeLists.txt
chriadam: Or, actually, we should check into git just minimal template/placeholder files, similar to the current .fr one

scripts/download_from_poeditor.sh
chriadam: Or, not remove from git, but only check into git the minimal template .ts, similar to .fr one now.

.github/patches/Makefile
```diff
@@ -0,0 +1,32 @@
+ifeq ($(PREFIX),)
+	PREFIX := /var/www/venus/gui-beta
+endif
+
+ifeq ($(EXTRAFONTS),)
+	EXTRAFONTS=/usr/lib/fonts/DroidSansFallback.ttf /usr/lib/fonts/NotoSansThai.ttf /usr/lib/fonts/DejaVuSans.ttf
```
DanielMcInnes, line 6: If these fonts are required for gui-v2, they should be added to the git repo. Otherwise, we end up in a situation where we need a specific machine to produce builds

blammit: Do you mean, just add them to the repo, without bundling them with gui-v2? That sounds okay, but we'd also need some option to configure them into the build, to ensure they are not included in the app bundle in the typical case.

.github/workflows/build-wasm.yml
```diff
@@ -103,9 +103,10 @@ jobs:
         sed -i "s%^var \(createQtAppInstance\)%window.\1%" venus-gui-v2.js
```
DanielMcInnes, line 103: it looks like we are patching venus-gui.v2.js here. Is this something that should be done for our developer builds as well? If so, please create an issue and assign it to me.

chriadam: We could actually ship custom .js and .html files, and just copy them to the bin directory manually as part of our build via:
```
add_custom_command(
        TARGET <target> POST_BUILD
        COMMAND ${CMAKE_COMMAND} -E copy
                ${CMAKE_SOURCE_DIR}/<target>.html
                ${CMAKE_CURRENT_BINARY_DIR}/<target>.html)
```

scripts/download_from_poeditor.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+
```
DanielMcInnes, line 2: Maybe add a comment explaining when and why this script gets run. Does this only get run when we we somehow know that there are updated translations available on poeditor?

scripts/upload_new_terms.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+
+if [ "${POEDITOR_TOKEN}" == "" ]
+then
+  echo "Please set environment variable POEDITOR_TOKEN first"
+  exit 1
+fi
+
+dirs=$(find . -maxdepth 2 -type f -name '*.qml' -printf '%h\n' | sort  | uniq | sed -e 's/[\.\/]//g')
```
DanielMcInnes, line 9: Does this only search 2 directories deep?
We already run lupdate on all translatable sources at build time, and produce .../[build_dir]/i18n/venus-gui-v2.ts, which contains every translatable string in the project.

scripts/download_from_poeditor.sh
```diff
@@ -0,0 +1,26 @@
...
+if [ "${POEDITOR_TOKEN}" == "" ]
+then
+  echo "Please set environment variable POEDITOR_TOKEN first"
+  exit 1
+fi
+
+sed -ne 's/^\s*addLanguage("\([^"]*\)", "\([^"]*\)".*/\2 \1/p' src/language.cpp |\
```
DanielMcInnes, line 9: There is already a definitive list of downloadable .ts files in CMakeLists.txt, it would be nice to leverage that list, rather than building a list dynamically by scraping language.cpp. 
eg.
to invoke:
cmake -DDOWNLOAD_TRANSLATIONS=True ../

// CMakeLists.txt
if ("${DOWNLOAD_TRANSLATIONS}")
    message("Downloading translations from poeditor")
    ... 
endif()

.github/patches/Makefile
DanielMcInnes: yes, that's right, only add them for wasm

blammit: They won't be added for wasm either -- on wasm, language.cpp will look for the font symlinks in the root directory, so that the fonts don't need to be bundled with the app.

scripts/upload_new_terms.sh
```diff
@@ -0,0 +1,21 @@
...
+echo "### Finding the latest strings in the qml files in directories: ${dirs}"
+lupdate ${dirs} -extensions qml -ts i18n/venus-gui-v2.ts
+
+echo "### Uploading terms to POEdior"
```
DanielMcInnes, line 13: typo - 'POEditor'

.github/patches/Makefile
DanielMcInnes: Huh. I don't understand what this makefile is for

.github/patches/Makefile
```diff
@@ -0,0 +1,32 @@
+ifeq ($(PREFIX),)
```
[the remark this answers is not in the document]
dirkjanfaber: The function of the `Makefile` is to assist in deploying the webassembly files on the GX device (and can also be used for deploying those files on the VRM proxy). In order to install the files on the GX device (when building a new Venus OS release), we use this script: https://github.com/victronenergy/meta-victronenergy-private/blob/master/recipes-ve/gui-v2-webassembly.bb.
The `Makefile` here will replace all of the individual calls in the the `do_install()` part by a `make install`. This saves us a double administration where, if for instance we add a new font or language to gui-v2, the `gui-v2-webassembly.bb` does not need to be adjusted.

.github/patches/Makefile
dirkjanfaber: The fonts are already installed on a GX device in `/usr/lib/fonts`. To allow the webassemby project to access those, a symbolic link is created in a web exposed directory (the directory where the `.wasm` file also resides, `/var/www/venus/gui-beta` at the moment).

All the three fonts come from the `theme` package, which does need to be added as a `DEPENDS` to the `gui-v2-webassembly.bb` file. And probably also to the `gui-v2.bb` file. 

Also see [my comment](https://github.com/victronenergy/gui-v2/pull/673#discussion_r1475663227) above for more info on the Makefile.

scripts/download_from_poeditor.sh
dirkjanfaber: The way it is done now for gui-v1 is to periodically (manually) synchronize the translations from POEditor and adding committing the updated translations to the project. Which works, but sometimes the translations lag a bit on the latest work of the translators.

So far my idea was to keep this workflow initially and move to downloading the translations automatically if this is working, but we might as well add this to the automatic builds directly. 

I agree that we do need to be adding minimal `template.ts` files for each language. 

The token can be stored as a [GitHub secret](https://docs.github.com/en/actions/security-guides/using-secrets-in-github-actions), which I'll arrange.

My idea is to only download the latest translations for tagged builds/releases. If you need them in a develop tree, either call the script manually or copy the files over from a release build. Does that work for you, @chriadam ?

scripts/upload_new_terms.sh
dirkjanfaber: Yes, it is searching 2 levels deep. And your proposal of running it after the build has completed is definitely better.

.github/patches/Makefile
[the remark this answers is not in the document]
mpvader: Hey @DanielMcInnes , All three of you have access now.

gui-v2-wasm file is in here: https://github.com/victronenergy/meta-victronenergy-private/tree/master/recipes-ve.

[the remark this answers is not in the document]
DanielMcInnes: @dirkjanfaber , this makefile looks good to me

.github/patches/Makefile
DanielMcInnes: @dirkjanfaber , lgtm

src/language.cpp
```diff
@@ -10,30 +10,115 @@
...
+	const QString fontFileName = fontFileNameForLanguage(language);
+	if (!fontIds.contains(fontFileName)) {
+		fontIds.insert(fontFileName, QFontDatabase::addApplicationFont(fontFileName));
+	}
+
+	int fontId = fontIds.value(fontFileName, -1);
+	if (fontId < 0) {
+		qWarning() << "Fall back to default font, cannot load" << fontFileName << "for locale" << QLocale(language).name();
+		if (!fontIds.contains(defaultFontFileName())) {
+			fontIds.insert(fontFileName, QFontDatabase::addApplicationFont(defaultFontFileName()));
```
chriadam, line 83: For WebAssembly, we cannot load files from the filesystem, so instead we need to load the font files via a network request (the nginx should serve it, now that those files are symlinked to the /var/www/gui-beta/ directory, and then inject into the font database with QFontDatabase::addApplicationFontFromData() I think.

.github/workflows/build-wasm.yml
```diff
@@ -103,9 +103,10 @@ jobs:
...
         cd -
         mkdir -p artifacts/wasm
-        mv build-wasm/venus-gui-v2.{html,js,wasm} build-wasm/qtloader.js images/victronenergy.svg artifacts/wasm/
+        mv build-wasm/venus-gui-v2.{html,js,wasm} build-wasm/venus-gui-v2_*.qm build-wasm/qtloader.js images/victronenergy.svg artifacts/wasm/
```
chriadam, line 106: I don't think including the .qm files here makes sense.  Instead, we should generate these files (using lrelease) as part of a pre-build step, and then add the .qm files to a bundled resource file so that they get included in the wasm blob, I think.

scripts/download_from_poeditor.sh
chriadam: That's fine.  Yes I agree that developers can do things manually if required for testing.

.github/workflows/build-wasm.yml
chriadam: @DanielMcInnes are you able to help Dirk-Jan with this one?  Your cmake-fu is better than mine.

DanielMcInnes: will do

dirkjanfaber: Ideally I would have both of the builds (the native and the webassembly) work with the same translation files. But not sure if that is possible.

DanielMcInnes: I don't think you have to move the .qm files around, they are already bundled into the app. When I locally build a wasm app, I can run it, change languages to Danish, app restarts, have to hit f5 for some reason, then the app starts up in Danish. Please try not moving the .qm files around, and check whether the production build allows you to change languages. I expect it will work fine, if it doesn't, then something else weird is going on

chriadam:
This PR still includes "full" .ts files (rather than the minimal template ones); those will have to be changed to just minimal templates.

dirkjanfaber:
> This PR still includes "full" .ts files (rather than the minimal template ones); those will have to be changed to just minimal templates.

If we download the latest translations _after_ the branch has been tagged, the native build will run out of sync with that.
Talked about this with Daniel and Bea. In order to resolve that, we do want to have use the full `.ts` files. Thus ensuring that both the native as the webassembly build use (but don't share) the same translations.
This might get another review in the future.

.github/workflows/build-wasm.yml
dirkjanfaber: Tested and indeed, the `.qm` files do not seem to be used. I'll remove installing them from the Makefile.

dirkjanfaber:
I fully agree that developers are probably as well of with placeholder translations files while developing. A solution for that could be to move the translations into a git submodule. But that will need some testing and investigation. And I don't want that to hold up this PR either. I think the best way for now is to have the translations in first. And see how we can improve the workflow afterwards. We should make a new issue for that then if you agree @chriadam . 

Assuming that is ok, the remaining (blocking) issue is that the webassembly builds don't show the characters for Chinese and Thai right now (Note that Arabic does work). Fixing that should become a separate issue/PR and not block this any longer. But for this PR that means that for Chinese and Thai we should fall back to English first. Keep the selection to the set language, but show English. And on the wasm show this in the pulldown or option list:
Chinese (no browser support)
Thai (no browser support)

@blammit : Can you add those changes to this PR?

blammit:
> Assuming that is ok, the remaining (blocking) issue is that the webassembly builds don't show the characters for Chinese and Thai right now (Note that Arabic does work). Fixing that should become a separate issue/PR and not block this any longer. But for this PR that means that for Chinese and Thai we should fall back to English first. Keep the selection to the set language, but show English. And on the wasm show this in the pulldown or option list: Chinese (no browser support) Thai (no browser support)
> 
> @blammit : Can you add those changes to this PR?

@dirkjanfaber I've added this patch, could you try it out and verify please?

dirkjanfaber:
> > Assuming that is ok, the remaining (blocking) issue is that the webassembly builds don't show the characters for Chinese and Thai right now (Note that Arabic does work). Fixing that should become a separate issue/PR and not block this any longer. But for this PR that means that for Chinese and Thai we should fall back to English first. Keep the selection to the set language, but show English. And on the wasm show this in the pulldown or option list: Chinese (no browser support) Thai (no browser support)
> > @blammit : Can you add those changes to this PR?
> 
> @dirkjanfaber I've added this patch, could you try it out and verify please?

That looks good. Only remark is that Arabic did work before in the browser (not sure why). But this is good enough for me right now. We can worry about that once this has been merged.

blammit:
> That looks good. Only remark is that Arabic did work before in the browser (not sure why). But this is good enough for me right now. We can worry about that once this has been merged.
> 
> Update: Also noticed that when switching from English to Danish to Chinese, the resulting language is Danish. But I don't think that is problematic. It just remains as it was (so not the "fallback to English"). If you want that, switch via English to Chinese, resulting in a gui-v1 in Chinese and webassembly gui-v2 in English.

@dirkjanfaber sorry, I missed your point about Arabic working fine in the browser. I'm not sure what I did before where the Arabic text was not rendering correctly. I've updated the PR to restore Arabic support in wasm, and updated the commit message to avoid mentioning English as a fallback.

blammit:
@dirkjanfaber I hope you don't mind, but I took the liberty of rebasing this on top of main and force-pushing to this branch.

blammit:
@chriadam @DanielMcInnes would you be happy for this to go in, in the current state?

chriadam (approved):
Yes.

DanielMcInnes:
> @chriadam @DanielMcInnes would you be happy for this to go in, in the current state?

yes

DanielMcInnes (commented):
lgtm

PR #678, opened 2024-01-24
Remove translation ids for genset errors that should not be translated

blammit:
Many of the error texts should not be translated, as per the original code in gui-v1. Remove the translation ids for any text that was not translated in gui-v1, to avoid unnecessary translation work for debug-type values.


dirkjanfaber (approved):
This looks good to me. Thanks!

Issue #650, opened 2024-01-10
Issues with analog tank inputs (13,208 gallons + 3rd tank input issues)


mpvader:
For starters, @ReinvdZee will see to reproduce it.

https://community.victronenergy.com/questions/249509/venus-os-v32034-available-for-testing.html

![image](https://github.com/victronenergy/gui-v2/assets/5200296/a39417c7-4d88-4c35-80eb-c9e0fcb77a68)

ReinvdZee:
**Tested on Cerbo GX running V3.20\~37**
I did some testing and found the following issues with the tanks:

### 1. Switching from GUI v1 to Beta --> Tank capacity off by 1
I started off with two tanks configured in GUI v1:

1. Fuel tank of 100gal
2. Fresh water tank of 200gal
I connected some resistors to the tank input of the GX Tank 140 to end up with the following:
![Screenshot from 2024-01-17 10-20-21](https://github.com/victronenergy/gui-v2/assets/38864570/0cf95391-ffbf-4928-bbff-68ccd7925458)

After switching to the beta GUI, the tank capacity levels are decreased by 1:

![image](https://github.com/victronenergy/gui-v2/assets/38864570/7637062b-8346-42c9-82ee-f0ad765f53ed)

The setup section shows the correct value. However, the edit menu shows the value off by 1 again.
![image](https://github.com/victronenergy/gui-v2/assets/38864570/4803a2a0-2fc4-4f97-8117-3cd6de5f4ed7)

### 2. Incorrect capacity after changing
Next, I changed the capacity of the fresh water tank from 199 (or 200) to 205gal:
![image](https://github.com/victronenergy/gui-v2/assets/38864570/d4524e3d-2897-4bf2-b089-3acf0ab37b09)

The new capacity setting seems to be saved properly:
![image](https://github.com/victronenergy/gui-v2/assets/38864570/390958f1-cf3b-4d4f-bc4e-a4ac60e1ccd9)


However, switching to the "Levels" view yields a capacity of 54155 gal:
![image](https://github.com/victronenergy/gui-v2/assets/38864570/3a9013f4-ac7f-45f4-aafb-fc2995794d34)

**54155gal equals 205m3, issue in capacity conversion between units? dbus-spy also shows a capacity of 205m3**

[...]

PR #685, opened 2024-01-26
Fix various issues with tank setup and units handling

blammit:


chriadam (approved):
Code LGTM.  I did a bit of poke-testing and it all looks good.  Is there some specific issues which list the repro steps for bugs which this fixes?

data/common/TankModel.qml
```diff
@@ -0,0 +1,15 @@
+/*
+** Copyright (C) 2023 Victron Energy B.V.
```
DanielMcInnes, line 2: 2024

DanielMcInnes (approved):
lgtm

blammit:
> Code LGTM. I did a bit of poke-testing and it all looks good. Is there some specific issues which list the repro steps for bugs which this fixes?

Yes, see https://github.com/victronenergy/gui-v2/issues/650 (linked from this PR)

data/common/TankModel.qml
blammit: Thanks, fixed

PR #688, opened 2024-01-29
Don't crash in VeQItem when passed invalid UID from QML

chriadam:


pages/settings/PageHub4Debug.qml
```diff
@@ -9,7 +9,7 @@ import Victron.VenusOS
 Page {
 	id: root
 
-	readonly property string batteryService: Global.batteries.firstObject ? Global.batteries.firstObject.serviceUid : null
+	readonly property string batteryService: Global.batteries.model.deviceAt(0) ? Global.batteries.model.deviceAt(0).serviceUid : null
```
chriadam, line 12: not 100% sure if this is correct or not, I think it is, but would like someone to double check.

blammit: Should be `Global.batteries.model.firstObject` instead (that is the same thing as `model.deviceAt(0)`)

pages/MainView.qml
```diff
@@ -66,13 +66,28 @@ Item {
...
 				if (navBar.currentIndex === navBar.model.count - 1 && objectOrUrl !== Global.appUrl("/pages/ControlCardsPage.qml")) {
 					properties = properties || {}
 					properties.height = pageStack.height - navBar.height
 				}
+
+				if (typeof(obj) === "string") {
+					// pre-construct the object to make sure there are no errors
+					// to avoid messing up the page stack state.
+					var checkComponent = Qt.createComponent(objectOrUrl)
+					var checkObject = checkComponent.createObject(null, properties)
```
blammit, line 82: Would prefer to check whether the Component has `status == Component.Error` instead of the error string. Also, that can be checked before calling `createObject()`, to avoid creating the object if the component couldn't be created (which is the case for this bug).

Also, I've been using `let` or `const` instead of `var` now in the codebase.

pages/settings/PageHub4Debug.qml
DanielMcInnes: I can't test via PageHub4Debug, but I put the following code in Batteries.qml:

```
    property var firstObjectServiceUid: model.count === 0 ? null : model.deviceAt(0).serviceUid
    onFirstObjectServiceUidChanged: console.log("*************************************** firstObjectServiceUid:", firstObjectServiceUid)

```
... and received the following output:
```
qml: *************************************** firstObjectServiceUid: null qtloader.js:389:25
qml: Loading MQTT data backend... qtloader.js:389:25
qml: *************************************** firstObjectServiceUid: mqtt/battery/258 qtloader.js:389:25
qml: *************************************** firstObjectServiceUid: mqtt/battery/256

```
so it looks ok to me

pages/MainView.qml
DanielMcInnes, line 82: I put some debug in PageAcIn.qml:
```
    Component.onCompleted: console.log("********************************** PageAcIn: onCompleted")
    Component.onDestruction: console.log("********************************** PageAcIn: onDestruction")
```
If you go to Overview, and repeatedly open and close the Grid Meter drill-down, I can see this object never gets deleted.

chriadam: thanks, fixed

chriadam: thanks, fixed

pages/settings/PageHub4Debug.qml
chriadam: thanks

pages/MainView.qml
DanielMcInnes: This causes grief with radio buttons. If you select any option from a radio button, you get the following error message:
`qrc:/qt/qml/Victron/VenusOS/pages/MainView.qml:99: Error: Invalid attempt to destroy() an indestructible object`
Eg. go to inverter/charger drilldown, product page, alarm setup, and change any of the settings

DanielMcInnes: maybe the object should only be destroyed if `parent === null` ?

chriadam: you're right - that's definitely the case.  Will fix.

chriadam: done

PR #689, opened 2024-01-29
Use opaque colors

chriadam:
Alpha blending is expensive.  Reducing alpha blending by using opaque colors reduces GPU work, and thus `swap` time.


tools/themeparser.py
```diff
@@ -32,32 +32,43 @@
 
 class ThemeSingleton : public Theme
 {
-    Q_OBJECT
-    QML_NAMED_ELEMENT(Theme)
-    QML_SINGLETON
+	Q_OBJECT
```
DanielMcInnes, line 35: tabs look wonky

DanielMcInnes (approved):
lgtm

tools/themeparser.py
chriadam: The code which forms part of the output needs tabs, whereas the python code needs spaces.  Hence the discrepancy.  What is in the PR is correct, I believe.

components/QuantityLabel.qml
```diff
@@ -16,6 +16,7 @@ Item {
 	property alias unitColor: unitLabel.color
 	property int alignment: Qt.AlignHCenter
 	property int precision: Units.defaultUnitPrecision(unit)
+	property bool useOpaqueColor: false
```
chriadam, line 19: as discussed offline, this is not a good solution, instead we should provide an "opaqueColor" property which can be set directly by the client so they can provide their own, as the actual opaque colour to be used must depend on the underlying surface colour as well as the semitransparent font colour.

chriadam:
Putting this on hold for now - asked Serj instead if he could just provide opaque colours for all elements.  Hopefully he agrees.  If not, we can go down this PR path (and maybe provide a `Theme.blendColors(elementColor, surfaceColor)` mixer function, which calculates the blending for us, rather than manually providing opaque theme colour values).

chriadam:
Serj provided some opaque colours, but they didn't match the existing colour scheme (it might be intentional or might be a mistake) and a variety of colours were missing, unfortunately.  I've added a set of colours which match the previous design, and covers all of the colours we use in our code.  I have asked Serj whether these colours need any adjustments.

MikeTrahearn-Qinetic:
The ability to define opaque colors is dependent on how well defined the colors are in the Figma Design System.
Where there are a small number of common colors set on top of a fairly common set of background colors, it is possible to define sets of pre-multiplied alpha colours - one for each Dark and Light mode. Having gone through this exact scenario on a previous project, this is less than trivial, unless you build it in to the Design System from the beginning.

chriadam:
Closing.  Was modified and included in https://github.com/victronenergy/gui-v2/pull/1171 a long time ago.

PR #690, opened 2024-01-29
Ensure preloaded pages are not visible by default

chriadam:
We preload the main (NavBar-activated) pages to avoid stutters at runtime when navigating.  This commit ensures that these pages are all marked as visible=false to avoid rendering them offscreen.

### DIFF
--- a/pages/MainView.qml
+++ b/pages/MainView.qml
@@ -278,6 +278,7 @@ Item {
 		Loader {
 			y: root.height + 4 // avoid fractional scaling smearing a row of pixels into visible area
 			asynchronous: true
+			visible: false
 			source: url
 
 			onStatusChanged: {


chriadam:
There's a strange flicker the first time you switch pages, now.  Not sure why.  Will investigate..

But, without this PR, QSG_VISUALIZE=overdraw showed that the overview page was still rendered (offscreen) while the brief page was showing.  With this PR, we reduce rendering quite a lot, resulting in about a 6% CPU usage reduction on the brief page.

chriadam:
Fixed.

components/Page.qml
```diff
@@ -15,14 +15,19 @@ FocusScope {
 	property bool fullScreenWhenIdle
 	readonly property bool isCurrentPage: !!Global.pageManager && Global.pageManager.currentPage === root
 	property bool animationEnabled: isCurrentPage && BackendConnection.applicationVisible
+	property bool animatingInOrOut: false
```
DanielMcInnes, line 18: maybe just 'animating'?

components/Page.qml
```diff
@@ -15,14 +15,19 @@ FocusScope {
...
 
 	property int topLeftButton: C.StackView.view && C.StackView.view.depth > 1 ? VenusOS.StatusBar_LeftButton_Back : VenusOS.StatusBar_LeftButton_None
 	property int topRightButton: VenusOS.StatusBar_RightButton_None
 
 	property var tryPop // optional function: returns whether the page can be poppped
 
+	visible: isCurrentPage || animatingInOrOut
```
DanielMcInnes, line 25: why do we want this visible when it is not the current page, but is animating?

DanielMcInnes (approved):
lgtm, 1 question

components/Page.qml
chriadam: so that it's visible when animating into the screen from offscreen, or animating out of the screen from onscreen.

components/Page.qml
chriadam: true

chriadam:
I will check to see whether I can just set the preloader repeater visibility to false (it should be inherited by children).  Once we call pageStack.replace(obj) the pagestack should control visibility properly, so my bindings shouldn't be needed.  It's only on initial load (and prior to reparenting to the page stack) that we hit this issue, I think.

chriadam:
PR is much simpler now ;-)

blammit (approved):

PR #696, opened 2024-01-30
Add "Onscreen UI (GX Touch & Ekrano)" option to display settings

blammit:


mpvader:
Was it tested on a GX and does it work? Also from WASM ?

chriadam:
I have tested it on both CerboGX and Wasm.  It works in both, however in wasm it takes some time for QtMqtt to inform us that the connection to the MQTT server is lost, so it just sits in an idle state for a couple of minutes until QtMqtt deigns to reconnect.

pages/DialogLayer.qml
```diff
@@ -0,0 +1,41 @@
...
+				//% "Rebooting..."
+				? qsTrId("dialoglayer_rebooting")
+				//% "Device has been rebooted."
+				: qsTrId("dialoglayer_rebooted")
+
+			// On device, dialog cannot be dismissed; just wait until device is rebooted.
+			dialogDoneOptions: VenusOS.ModalDialog_DoneOptions_OkOnly
+			footer.enabled: BackendConnection.type !== BackendConnection.DBusSource
+			footer.opacity: footer.enabled ? 1 : 0
+			closePolicy: C.Popup.NoAutoClose
```
chriadam, line 38: should be Templates not Controls, I will update.

chriadam (approved):
LGTM.

PR #697, opened 2024-01-31
Read solar charger tracker name from /Pv/<index>/Name

blammit:
When showing tracker data in the solar history views, show the name of the tracker, instead of just defaulting to "<charger name> (#1)" for the first tracker, etc.

This only works for com.victronenergy.solarcharger.* services. This data is not available for com.victronenergy.inverter.* services, so when viewing history data for PV inverters from the Device List, the default naming scheme will continue to be used.

### DIFF
--- a/data/common/SolarHistory.qml
+++ b/data/common/SolarHistory.qml
@@ -23,6 +23,19 @@ QtObject {
 		return _historyObjects.dailyHistory(day, trackerIndex)
 	}
 
+	function trackerName(trackerIndex) {
+		const nameObject = _trackerNames.objectAt(trackerIndex)
+		const name = nameObject ? nameObject.value || "" : ""
+		return name ? name : Global.solarChargers.defaultTrackerName(trackerIndex, trackerCount, deviceName)
+	}
+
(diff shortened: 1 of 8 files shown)


data/common/SolarHistory.qml
```diff
@@ -12,6 +12,11 @@ QtObject {
 	property string bindPrefix
 	property string deviceName
 	property int trackerCount
+	property var trackerNameForIndex: function(trackerIndex) {
```
DanielMcInnes, line 15: qml syntax question - if you had:
`function trackerNameForIndex(trackerIndex) {...}`
... could you override the function in the same way?
I've seen this elsewhere, I'm just wondering why people do it like this

DanielMcInnes (approved):
lgtm

blammit:
Martin B mentioned that the /Pv/<index>/Name is actually available for inverters as well, so I've updated the PR to also fetch this when viewing inverter data from the Device List (i.e. using `PageInverter.qml`). 

This removes the function property previously discussed in this PR :)

chriadam (approved):

Issue #375, opened 2023-06-07
Update settings when venus-platform provides various features


blammit:
The gui-v1 settings uses C++ helpers such as "vePlatform" to set/get various system features. In gui-v2 these are not available, and the preferred method is to do this via venus-platform. 

These are the features (also marked with TODO in the code) which still need to be implemented, when they become available in venus-platform:

- Display settings: needs vePlatform.hasAutoBrightness, vePlatform.hasBacklight, vePlatform.brightness, ~something to set display color 'auto' mode (i.e. not fixed to dark/light)~  -> ~**implemented in per Venus OS v3.30~something; so its up for implementation in gui-v2**~ (Done in gui-v2)
- ~Display off timer handling: each gui needs to do that by itself (have a timer, reset it on keypress or touch event, and on elapse turn display off). Implementation in gui-v1 is done and released; Implementation in gui-v2 needs doing. -> Qinetic to prioritize this (filed #859)~ fixed
- General: Set root password -> **ON HOLD until MQTT security is fixed.**
- ~Services (CAN Bus): needs vePlatform.canBusStats()~
- ~Date/Time: set system date time~
- ~Relays: set relay state based on relay polarity and alarm status (in gui-v1 this is done via the 'Relay' singleton) ->  **implemented in per Venus OS v3.30~something; so its up for implementation in gui-v2**~ (no implementation required in gui-v2 since this is handled in venus-platform)
- ~VRM online portal: vePlatform.uniqueId~
- ~Remote console settings: set remote console password using vePlatform.setRemoteConsolePassword()~ -> Dropped by MQTT security changes.

mpvader:
Hey @blammit do you agree to remove `Debug: vePlatform.getMemInfo() and QuickView.imageCacheSize() `  from the list? We don't need that.

Thats in general btw: all those debug pages and debug items: please ignore them. If and when someone needs them at some time or date they'll come back.

blammit:
Great, sounds good to not do those if they're not needed. Will update the list.

ReinvdZee:
Hi @blammit,
I'm currently working on this issue, and I started with the display controls. As I am not all that familiar with gui-v2, I could use some feedback.

Would it be sufficient for the gui-v2 to have the following dbus settings to control the display/backlight:

1. `/Settings/Gui/Brightness` : The current brightness setting, in 0 to 100% (let venus-platform convert this setting in 0-100% to the supported brightness range, prevents an unnecessary dbus setting for the max brightness)
[...]

PR #699, opened 2024-01-31
Remove auto color mode

blammit:
This will not be supported at present; only dark/light modes will be allowed.

See discussion in https://github.com/victronenergy/gui-v2/issues/375#issuecomment-1899964269


DanielMcInnes (approved):
lgtm

PR #702, opened 2024-01-31
Delay reboot after changing /RunningVersion, and always reboot

blammit:
- Wait for /RunningVersion value change to take effect before rebooting
- Always reboot when changing value


chriadam (approved):
LGTM.

chriadam:
It's a pity that the value is reflected to MQTT before the sync() finishes to persist the write the disk.  Oh well.  Nothing we can do about it, other than delay the reboot, as done in this PR, I think.

chriadam:
Discussed with Matthijs further - the delay is necessary with the current behaviour of the local settings service (which delays the sync by 1 second after the value is accepted).  However, a delay is always going to be brittle, so better off either changing the local service so that (for this specific setting value) it will trigger the reboot (or reload of the UI process) itself, after its sync succeeds.

This task is now underway.  We need to remove the Timer and also the Reboot triggering stuff.

chriadam (changes requested):
As commented.

blammit:
Sounds good, updated to remove the reboot behavior.

chriadam:
I believe it will just restart the (appropriate) UI process, not reboot the system.  But @ReinvdZee would know for sure.
Also, we cannot merge this change until Rein's change to the local settings service is finished.
Until then, we need to revert all changes to this file.

blammit:
> I believe it will just restart the (appropriate) UI process, not reboot the system. But @ReinvdZee would know for sure. Also, we cannot merge this change until Rein's change to the local settings service is finished. Until then, we need to revert all changes to this file.

Okay, created https://github.com/victronenergy/gui-v2/pull/705 to remove the setting.

pages/settings/PageSettingsDisplay.qml
```diff
@@ -132,18 +132,14 @@ Page {
...
-					}
-				}
-
 				optionModel: [
 					{ display: "Standard version", value: 1 },
 					{ display: "Gui-v2 (beta) version", value: 2 },
 				]
+
+				onOptionClicked: function(index) {
+					Global.dialogLayer.showRebootDialog()
```
chriadam, line 141: this might need to change so that the text says "UI is restarting" rather than "Device is rebooting"

blammit: Done

chriadam (approved):
LGTM

PR #703, opened 2024-01-31
bugfix: Unable to change the MultiPlus Mode with VE.Bus BMS v2

### LINKED ISSUE #679: Unable to change the MultiPlus Mode with VE.Bus BMS v2
And expected behaviour is that that is indeed blocked with VE.Bus BMS (v1), but not with the much newer VE.Bus BMS v2 model.

See qml of the old gui and/or ask @vejpasop .

(sorry in case this reported issue is a double one, I've seen this complaint before but can't remember where / to whom I communicated it)

Reported here: https://community.victronenergy.com/questions/249509/venus-os-v32034-available-for-testing.html?page=2&pageSize=50&sort=oldest

image

DanielMcInnes:
Fixes #679

### DIFF
--- a/data/mock/VeBusDevicesImpl.qml
+++ b/data/mock/VeBusDevicesImpl.qml
@@ -69,7 +69,6 @@ QtObject {
 
 			property int state
 			property int mode: -1
-			property bool modeAdjustable
 			property int numberOfPhases: 3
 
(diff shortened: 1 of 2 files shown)


DanielMcInnes:
tested wasm build against local cerbo

blammit (approved):

PR #705, opened 2024-02-01
Remove "Running version" selection in settings

blammit:
This removal is temporary. The setting will be restored when some venus-platform changes are made to allow the system changes to take effect when the running version is changed.


blammit:
Don't need, https://github.com/victronenergy/gui-v2/pull/704 already exists to do this.

PR #708, opened 2024-02-02
Fix ArcGauge - don't shadow a FINAL property

chriadam:
Property shadowing semantics were tightened in 6.6.x, (or the Shape's status property was marked as FINAL).

In either case, we can no longer shadow this property (and we shouldn't shadow it, regardless).  Rename our property.


DanielMcInnes (approved):
lgtm

PR #715, opened 2024-02-06
Provide command-line parameter to skip the splash screen

jpetrell:
- Also fixes the nav bar not showing when skipping splash screen
- Skipping the splash screen reduces the launch time on desktop from 9 to 2 seconds


blammit (approved):
Nice, I'll definitely use this during development :)

PR #719, opened 2024-02-07
Enable autostart when the respective switch is checked

jpetrell:


jpetrell:
Added few small related fixes on top:
- Don't animate the switch state until user confirms disabling the autostart
- Manually adding and stopping the generator in the settings showed inverted toasts

blammit (approved):

PR #735, opened 2024-02-09
Fix nav bar getting stuck after visiting settings

jpetrell:


components/PageStack.qml
```diff
@@ -32,14 +35,14 @@ C.StackView {
 		}
 	}
 	popEnter: Transition {
-		ScriptAction { script: if (pageStack.depth === 1) pageStack.navbarX = 0 }
```
jpetrell, line 35: popEnter is not executed when switching from the settings to other navigation bar pages.

blammit (approved):

Issue #734, opened 2024-02-09
On iPhone, settings menu and more is broken per v0.2.9


mpvader:
Discussed here: https://victrondevelopment.slack.com/archives/C020QG3JVEW/p1707458487757799

Matthijs can reproduce.

Agreed first step is to test a build that was made with Qt 6.6 instead of Qt 6.5.something. Afaik Dirk-Jan already has that process working

dirkjanfaber:
@mpvader : Can you reproduce it on [this site](https://acceptancevrm.victronenergy.com/installation/237451/dashboard)?

That one is running the gui with a Qt 6.6.1 build. (using  [this](https://github.com/victronenergy/gui-v2/tree/faberd/qt-6.6.1-emscripten-3.1.37) branch and the files from [this action](https://github.com/victronenergy/gui-v2/actions/runs/7843682943)).

mpvader:
Hi @dirkjanfaber that solves it indeed.

So then plan remains the same: await Lucas from El Nino to finish and deploy his preps for Qt 6.6, and then switch the webassembly build to that; right?

dirkjanfaber:
I believe that it is only a matter of deploying for Lucas. I've just put this build on CDN and confirm that running that also works on acceptance VRM via the remote console 2 link. Meaning that both the Qt 6.6.1 as the Qt 6.5.2 builds work on all tested cases:

| Build  | local | CDN |
| --- | --- | --- |
| Qt 6.5.2 | ✅ | ✅  |
| Qt 6.6.1 | ✅ | ✅ |

With local I mean that VRM downloads the .wasm file from the local device. With CDN I mean that VRM downloads the .wasm file from the CDN.

So if Chris and Daniel agree (see https://github.com/victronenergy/gui-v2/issues/677), my branch can be merged and the code from acceptance VRM can be deployed on regular VRM. Both old and new builds work.

mpvader:
Nice

blammit:
PRs merged, closing.

PR #741, opened 2024-02-12
Qt update to 6.6.1, emscripten to 3.1.37

dirkjanfaber:
Update Qt version to 6.6.1 and emscripten to 3.1.37. Made the patching more generic, install `yq`, update and install python is now done after the cache check (should speed things up). Skip steps when using cached version.

Have packing + patching html and `js` files into a zip as separate step. The patchings are showing the Victron logo and to set the default argument for connecting to the mqtt broker. And to set the qtAppInstance as window (to make it work via VRM). Remove `qtloader-controller.js` (as it is no longer needed for 6.6.1).

Also see issue #734


chriadam (approved):

.github/workflows/build-wasm.yml
```diff
@@ -7,34 +7,37 @@ on:
 jobs:
   build-wasm-file:
     env:
-      EMSCRIPTEN: 3.1.25
-      QT_VERSION: 6.5.2
+      EMSCRIPTEN: 3.1.37
+      QT_VERSION: 6.6.1
```
DanielMcInnes, line 11: Should we change the required Qt version in CMakeLists.txt as well?

DanielMcInnes (approved):
lgtm, just 1 question

.github/workflows/build-wasm.yml
dirkjanfaber: AFAIK the non-webassembly build still uses 6.5.2 for building. Once that has changed to 6.6.1 I'd change the `CMakeLists.txt`.

DanielMcInnes: lgtm

Issue #574, opened 2023-11-28
Controls Card view enter transition animation should be "from top + fade"


chriadam:
Currently, the control cards view animates in from the left rather than the top.
It should slide down from the top (and fade in as it animates).
The view "below" it should NOT move, but just be faded out "below" it.
That is, the controls card view is kind of an "overlay" view, rather than a true "navigation page".
See discussion on Slack with Serj for the Figma prototype.

blammit:
Could also update the control cards open/close icon as part of this task. When it is "open", the icon should be highlighted. See current Figma design.

chriadam:
Simple task, needs to be done for 1.0.0.  Setting as medium_prio.

chriadam:
The icon task Bea mentioned is tracked separately as #584

jpetrell:
Draft PR https://github.com/victronenergy/gui-v2/pull/744

[access-controls-from-top.webm](https://github.com/victronenergy/gui-v2/assets/2203667/0ebb81d1-d9eb-4a6a-9dda-cbb78a65c03e)

Didn't update any icons yet. I will ask Serj for feedback.

PR #744, opened 2024-02-12
Access controls from top

jpetrell:
Fixes https://github.com/victronenergy/gui-v2/issues/574.

### DIFF
--- a/components/Page.qml
+++ b/components/Page.qml
@@ -24,5 +24,5 @@ FocusScope {
 	implicitWidth: C.StackView.view ? C.StackView.view.width : 0
 	implicitHeight: C.StackView.view ? C.StackView.view.height : 0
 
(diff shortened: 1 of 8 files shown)


jpetrell:
Showed the design to Serj. Updated the animation duration and easing curve to match the design. Qt doesn't have easing curve called ease-in, but chose Easing.InSine as that seemed close to the CSS ease-in.

blammit:
@jpetrell could you add the task number to the commit please? Something like "Issue: #xxx" or "Fixes #xxx" as the last line would be sufficient.

blammit (dismissed):

blammit (approved):
As discussed, noticed a minor issue where buttons in the control cards will change color when the cards begin to slide back upwards. Otherwise LGTM.

jpetrell:
Fixed the focus flicker.

PR #759, opened 2024-02-15
MQTT support for notifications & alarms

blammit:
Notifications can be tested in mock mode:

    - 'n' to add an info or warning notification
    - Shift+n to add an alarm notification

On device, if you have a generator running (e.g. with demo mode "Boat/Motorhom demo 1"), you can go to Device List > Generator > Setup > Enable alarm. Whenever the alarm is enabled, a new Generator notification will be sent. If you disable the alarm, the notification will become inactive.

When a new active notifications is received, the app should show the "Silence alarms" button. When this is clicked, all notifications will be acknowledged.

If the new notification is of an "alarm" type, the app should also jump to the Notifications page.


components/GradientListView.qml
```diff
@@ -11,9 +11,9 @@ ListView {
 	id: root
 
 	x: Theme.geometry_page_content_horizontalMargin
+	y: Theme.geometry_gradientList_topMargin    // don't set Flickable topMargin, causes bug in list positioning (QTBUG-87007)
```
jpetrell, line 14: Small margin value like this is probably not a problem, but in general having flickable margins outside the flickable can be dangerous as ListView only controls delegate visibility within its own bounds (sometimes causing issues like delegates to disappear abruptly). If topMargin is an issue bake the margin to the headerItem?

pages/MainView.qml
```diff
@@ -34,7 +34,7 @@ Item {
 	function _loadUi() {
 		console.warn("Data sources ready, loading pages")
 		preloader.model = navBar.model
-		navBar.currentIndex = 0
+		navBar.showIndex(0)
```
jpetrell, line 37: navBar.showPage("BriefPage.qml") feels bit more readable.

components/NavBar.qml
```diff
@@ -10,8 +10,31 @@ Rectangle {  // Use an opaque background so that page disappears behind nav bar
...
+
+	function showPage(pageFileName) {
+		for (let i = 0; i < navBarModel.count; ++i) {
+			const url = navBarModel.get(i).url
+			if (url.endsWith(pageFileName)) {
+				_currentIndex = i
+				_currentUrl = url
+				break
+			}
+		}
```
jpetrell, line 36: Print warning if the page is not found?

components/NavBar.qml
```diff
@@ -10,8 +10,31 @@ Rectangle {  // Use an opaque background so that page disappears behind nav bar
...
+
+	// External components should not write to these properties.
+	property int _currentIndex
+	property url _currentUrl
+
+	function showIndex(i) {
+		if (i >= 0 && i < navBarModel.count) {
+			_currentIndex = i
+			_currentUrl = navBarModel.get(i).url
+		}
```
jpetrell, line 25: Print warning if the index is out of bounds?

components/NotificationDelegate.qml
```diff
@@ -40,75 +35,88 @@ Rectangle {
...
+			id: descriptionLabel
+
+			width: parent.width
+			visible: root.notification.description.length > 0 || root.notification.value.length > 0
+			elide: Text.ElideRight
+			color: Theme.color_font_primary
+			font.pixelSize: Theme.font_size_body2
+			//: %1 = notification description (e.g. 'High temperature'), %2 = the value that triggered the notification (e.g. '25 C')
+			//% "%1 %2"
+			text: qsTrId("notification_description_and_value").arg(root.notification.description).arg(root.notification.value)
```
jpetrell, line 96: I understand this is in general short, but as this is common service to various features would sleep better if notification description labels wrapped instead of truncating with elide. In some senteces and languages the meaningful bits of the sentence can be at the end.

components/NotificationDelegate.qml
```diff
@@ -10,12 +10,7 @@ import Victron.VenusOS
 Rectangle {
 	id: root
 
-	property bool acknowledged
-	property bool alarmActive
-	property int category
-	property date date
-	property string source
-	property string description
+	required property Notification notification
```
jpetrell, line 13: Nice!

components/StatusBar.qml
```diff
@@ -81,40 +81,39 @@ Rectangle {
...
+		enabled: !!Global.notifications && Global.notifications.alert
 		opacity: enabled ? 1 : 0
 		Behavior on opacity { OpacityAnimator { duration: Theme.animation_toastNotification_fade_duration} }
 		border.width: Theme.geometry_button_border_width
 		border.color: Theme.color_critical
-		width: Theme.geometry_notificationsPage_snoozeButton_width
+		leftPadding: Theme.geometry_notificationsPage_snoozeButton_horizontalMargin
+		rightPadding: Theme.geometry_notificationsPage_snoozeButton_horizontalMargin
 		height: Theme.geometry_notificationsPage_snoozeButton_height
 		backgroundColor: Theme.color_darkCritical
```
jpetrell, line 96: The legibility on light style is not great, red contents drawn on top of semi-transparent red.

image

data/common/Notification.qml
```diff
@@ -0,0 +1,107 @@
...
+		uid: notification.serviceUid ? notification.serviceUid + "/DeviceName" : ""
+	}
+
+	readonly property VeQuickItem _description: VeQuickItem {
+		uid: notification.serviceUid ? notification.serviceUid + "/Description" : ""
+	}
+
+	readonly property VeQuickItem _value: VeQuickItem {
+		uid: notification.serviceUid ? notification.serviceUid + "/Value" : ""
+	}
```
jpetrell, line 44: Feels suboptimal do communicate notification state through 7 different pipes, though don't have better ideas.

jpetrell:
For some reason the icons look washed out compared to the design:
image

data/common/Notification.qml
```diff
@@ -0,0 +1,107 @@
...
+		}
+	}
+
+	acknowledged: !!_acknowledged.value
+	active: !!_active.value
+	type: _type.value === undefined ? -1 : parseInt(_type.value)
+	dateTime: _dateTime.value === undefined ? _invalidDate : new Date(_dateTime.value * 1000)
+	deviceName: _deviceName.value || ""
+	description: _description.value || ""
+	value: _value.value || ""
```
jpetrell, line 100: Seeing these VeQuickItems mostly used to bind to a property kind of makes me wish you could declare both VeQuickitem and the property on the same code line e.g. QML property value source, e.g.

`VeQuickItem on deviceName { uid: notification.serviceUid ? notification.serviceUid + "/DeviceName" : "" }
`

jpetrell: Though doesn't work if the input value needs to be somehow modified before setting it to the property.

themes/geometry/FiveInch.json
```diff
@@ -86,32 +86,17 @@
...
-    "geometry_notificationsPage_delegate_description_spacing_horizontal" : 24,
-    "geometry_notificationsPage_delegate_description_spacing_vertical" : 3,
-    "geometry_notificationsPage_delegate_topMargin" : 15,
-    "geometry_notificationsPage_delegate_rightMargin" : 24,
-    "geometry_notificationsPage_snoozeButton_topMargin" : 8,
-    "geometry_notificationsPage_snoozeButton_rightMargin" : 24,
-    "geometry_notificationsPage_snoozeButton_width" : 160,
     "geometry_notificationsPage_snoozeButton_height" : 32,
-    "geometry_notificationsPage_snoozeButton_radius" : 8,
-    "geometry_notificationsPage_snoozeButton_image_leftMargin" : 16,
```
jpetrell, line 113: Nice!!

ApplicationContent.qml
```diff
@@ -13,6 +13,14 @@ Item {
 
 	property var _inputComponent
 
+	readonly property bool _goToNotifications: !!pageManager.navBar && !!Global.notifications
+			&& Global.notifications.alarm
```
jpetrell, line 17: Wish you could write `Global.notifications?.alarm` as shorthand for `!!Global.notifications && Global.notifications.alarm`.

jpetrell: Using property state to signal change and call a function feels odd combination of declarative and imperative, but it works so why not. :smile:

jpetrell:
Overall LGTM, some small findings here and there. Awesome to get 1 kLOC away.

I was going to file a bug about notification layout, but you fixed most issues here already (could simplify the header item layout at some point, but that is separate issue and the current one works well enough).

components/GradientListView.qml
blammit: Good point, thanks -- I've moved the margin to the headerItem, and filed a separate Qt bug for the scroll issue (QTBUG-122422).

pages/MainView.qml
blammit: Done, and removed showIndex()

components/NavBar.qml
blammit: Removed the function

components/NotificationDelegate.qml
blammit: Yes, seems good to prepare for that case. Updated to do this as suggested.

components/NotificationDelegate.qml
```diff
@@ -40,75 +35,88 @@ Rectangle {
...
 		}
-		color: Theme.color_notificationsPage_text_color
-		text: _formatTimestamp(root.date)
-	}
-	MouseArea {
-		id: mouseArea
-		anchors.fill: parent
-		onClicked: model.acknowledged = true
+		color: Theme.color_listItem_secondaryText
+		text: root._formatTimestamp(root.notification.dateTime)
```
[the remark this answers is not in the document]
blammit: Moved to Utils.js.

data/common/Notification.qml
blammit: That would be neat :)

blammit:
> For some reason the icons look washed out compared to the design:
> 

The icons have a dimmed color when the notification is inactive. If you trigger a new notification (with "n" in mock mode) then the new active notification will have a brightly-coloured icon.

DanielMcInnes:
I notice some different behavior between running on the device and the wasm build. On the device, I can trigger an alarm by enabling alarms for the generator. I am taken to the Notifications page. If I tap the Notifications button on the NavBar, it silences the beeping and acknowledges the alarm. 

If I do trigger an alarm in the wasm build, I am taken to the Notifications page. If I tap the Notifications button on the NavBar, it doesn't silence the beeping or acknowledge the alarm. If I click on the delegate, it acknowledges the alarm (little red light goes off), but doesn't silence the beeping. The only way to stop the beeping is to click the new 'silence alarm' button.

components/NavBar.qml
```diff
@@ -10,8 +10,25 @@ Rectangle {  // Use an opaque background so that page disappears behind nav bar
...
+	function showPage(pageFileName) {
+		for (let i = 0; i < navBarModel.count; ++i) {
+			const url = navBarModel.get(i).url
+			if (url.endsWith(pageFileName)) {
+				_currentIndex = i
+				_currentUrl = url
+				break
+			}
+		}
+		console.warn("showPage(): cannot find page", pageFileName)
```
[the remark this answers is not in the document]
blammit: Oops, fixed, thanks

components/NavBar.qml
```diff
@@ -10,8 +10,25 @@ Rectangle {  // Use an opaque background so that page disappears behind nav bar
...
+	readonly property url currentUrl: _currentUrl
+
+	// External components should not write to these properties.
+	property int _currentIndex
+	property url _currentUrl
+
+	function showPage(pageFileName) {
+		for (let i = 0; i < navBarModel.count; ++i) {
+			const url = navBarModel.get(i).url
+			if (url.endsWith(pageFileName)) {
```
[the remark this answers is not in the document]
blammit: That's true, I've added a "/" to the condition to avoid a partial match, as the argument isn't the full file name.

components/NotificationDelegate.qml
```diff
@@ -10,105 +10,94 @@ import Victron.VenusOS
 Rectangle {
 	id: root
 
-	property bool acknowledged
-	property bool alarmActive
-	property int category
-	property date date
-	property string source
-	property string description
+	required property Notification notification
```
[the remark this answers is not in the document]
blammit: It's set automatically due to the `required property Notification notification` in this file. We haven't used the `required property` syntax much in this codebase but might be nice to do that more going forward, to make it explicit in terms of the properties that are passed from the model.

blammit:
> I notice some different behavior between running on the device and the wasm build. On the device, I can trigger an alarm by enabling alarms for the generator. I am taken to the Notifications page. If I tap the Notifications button on the NavBar, it silences the beeping and acknowledges the alarm.

Hm, interesting. Could you check that it's not because gui-v1 is running underneath and also getting the mouse event? (Since the checkmark button in the toolbar on the gui-v1 notifications page will silence the alarms.)

blammit:
> If I trigger an alarm in the wasm build, I am taken to the Notifications page. If I tap the Notifications button on the NavBar, it doesn't silence the beeping or acknowledge the alarm. If I click on the delegate, it acknowledges the alarm (little red light goes off), but doesn't silence the beeping. The only way to stop the beeping is to click the new 'silence alarm' button.

Indeed, this is because clicking a notification will set Acknowledge=1 just for that notification, but clicking the "Silence alarm" button will set AcknowledgeAll=1 and thus turn off the alert.

@jepefe If there is only one notification left to acknowledge, and the user clicks it to acknowledges it, should gui-v2 also set AcknowledgeAll=1 to silence the alert? Or should the user be forced to do both (click the notification *and* click "Silence alarm")?

jpetrell:
> The icons have a dimmed color when the notification is inactive. If you trigger a new notification (with "n" in mock mode) then the new active notification will have a brightly-coloured icon.

Ah true. :smile: What does active notification mean in practice? If I press "Silence alarm" button on the status bar the notification icon doesn't get dimmed. Still, feels odd to see the bright red dot (new notification indication) next to a washed out icon.

There are now not one, but three different visual states to indicate new notifications:
- full-brightness/washed out icon indicates active notification
- active/history grouping
- red dot unread notifications indicator

Ok well maybe not three, seems the red dot and active/history grouping are in fact the same state, the latter is just applied with a slight delay. Dunno, for me feels this could be simplified somehow. Also, I would try to show the icons in general without dimming, now most notification icons shown there are dimmed.

jpetrell:
Serj shared me the design. Works as specified. :+1: 

The logic feels a bit heavy, but I guess this is needed as many of the warnings and errors can highlight a critical issue and we want to make sure the user is informed every step on the way, so even e.g. warnings that are no longer active still need to be manually acknowledged. The design is kind of optimized for high priority messaging, hope the parties that send notifications don't misuse the capability, so we don't end up creating a spam folder with a lot of duplicate or unacknowledged items.

jepefe:
>  If there is only one notification left to acknowledge, and the user clicks it to acknowledges it, should gui-v2 also set AcknowledgeAll=1 to silence the alert? Or should the user be forced to do both (click the notification and click "Silence alarm")?

@blammit Setting AcknowledgeAll = 1 just sets Acknowledged = 1 to the rest of alarms. If there is only one alarm and is set to Acknowledged = 1 the buzzer should stop. No need to set AcknowledgeAll.

Is it not working this way? 
What I understand from Daniel is that it works as expected on local gui, but not in the wasm.

DanielMcInnes:
> > I notice some different behavior between running on the device and the wasm build. On the device, I can trigger an alarm by enabling alarms for the generator. I am taken to the Notifications page. If I tap the Notifications button on the NavBar, it silences the beeping and acknowledges the alarm.
> 
> @DanielMcInnes Hm, interesting. Could you check that it's not because gui-v1 is running underneath and also getting the mouse event? (Since the checkmark button in the toolbar on the gui-v1 notifications page will silence the alarms.)

Yes, this is what was happening. All good.

DanielMcInnes (approved):
lgtm

blammit:
> > If there is only one notification left to acknowledge, and the user clicks it to acknowledges it, should gui-v2 also set AcknowledgeAll=1 to silence the alert? Or should the user be forced to do both (click the notification and click "Silence alarm")?
> 
> @blammit Setting AcknowledgeAll = 1 just sets Acknowledged = 1 to the rest of alarms. If there is only one alarm and is set to Acknowledged = 1 the buzzer should stop. No need to set AcknowledgeAll.

@jepefe This isn't happening at the moment. I guess that could be fixed in venus-platform, then gui-v2 wouldn't need to set AcknowledgeAll as well just to silence the alarm.

PR #763, opened 2024-02-16
Add status LED setting

jpetrell:


jpetrell:
Finally verified with v3.30~11. Tried earlier with 20240216163909 v3.21, but the LED setting wasn't there yet.

blammit (approved):

PR #764, opened 2024-02-16
Show battery status along with battery voltage

### LINKED ISSUE #751: PageTemperatureSensor: Show battery status along with battery voltage
See this commit to gui-v1:
https://github.com/victronenergy/gui/commit/00836db21bb773fc0c928fae3f7392491b28c1bb

Related issue: https://github.com/victronenergy/venus/issues/889

jpetrell:
Fixes #751. Tested with dummy DBus service. Still needs to be verified by someone with access to Ruuvi tag device. 

Three possible variants (only one really shown at any one time):
image

### DIFF
--- a/pages/settings/devicelist/temperature/PageTemperatureSensor.qml
+++ b/pages/settings/devicelist/temperature/PageTemperatureSensor.qml
@@ -70,13 +70,48 @@ Page {
 				visible: defaultVisible && dataItem.isValid
 			}
 
-			ListQuantityItem {
+			ListItem {
+				id: sensorBattery
+
 				//% "Sensor battery"
 				text: qsTrId("temperature_sensor_battery")
-				dataItem.uid: bindPrefix + "/BatteryVoltage"
-				unit: VenusOS.Units_Volt
-				precision: 2
-				visible: defaultVisible && dataItem.isValid
+				visible: defaultVisible && batteryVoltage.isValid
+
+				content.children: [
+					QuantityLabel {
+						id: batteryVoltageLabel
+						anchors.verticalCenter: parent.verticalCenter
+						font.pixelSize: Theme.font_size_body2
+						value: batteryVoltage.value
+						unit: VenusOS.Units_Volt
+						precision: 2
+						VeQuickItem {
+							id: batteryVoltage
+							uid: bindPrefix + "/BatteryVoltage"
+						}
+					},
+					Label {
+						anchors.verticalCenter: parent.verticalCenter
+						text: {
+							if (lowBattery.isValid) {
+								const low = lowBattery.value === 1
+								//% "Low"
+								return low ? qsTrId("temperature_sensor_battery_status_low")
+										   : CommonWords.ok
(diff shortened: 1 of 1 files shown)


components/CommonWords.qml
```diff
@@ -289,7 +289,7 @@ QtObject {
 	//% "Offline"
 	readonly property string offline: qsTrId("common_words_offline");
 
-	//% "OK"
+	//% "Ok"
```
jpetrell, line 292: Requested by Serj. All other strings show on the respective labels avoid all caps.

blammit: I'm not sure about this. "OK" is generally seen as the correct form, and "Ok" just looks wrong. In informal use, "ok" would be fine, but "Ok" seems like a strange mismash of "OK" and "ok". It's true that this is not consistent in terms of capitalization, but this word is an exception.

pages/settings/devicelist/temperature/PageTemperatureSensor.qml
```diff
@@ -70,13 +70,51 @@ Page {
...
-				precision: 2
-				visible: defaultVisible && dataItem.isValid
+				visible: defaultVisible && batteryVoltage.isValid
+
+				content.children: [
+					QuantityLabel {
+						id: batteryVoltageLabel
+						anchors.verticalCenter: parent.verticalCenter
+						font.pixelSize: Theme.font_size_body2
+						value: batteryVoltage.value === undefined ? NaN : batteryVoltage.value
```
blammit, line 85: Just `value: batteryVoltage.value` is sufficient, as QuantityLabel internally checks if the value is NaN.

pages/settings/devicelist/temperature/PageTemperatureSensor.qml
```diff
@@ -70,13 +70,51 @@ Page {
...
+								//% "Low"
+								return low ? qsTrId("temperature_sensor_battery_status_low")
+										   : CommonWords.ok
+							} else {
+								return ""
+							}
+						}
+						color: lowBattery.value === 1 ? Theme.color_red : Theme.color_green
+						font.pixelSize: Theme.font_size_body2
+						wrapMode: Text.Wrap
```
blammit, line 108: The wrapping seems unnecessary?

components/CommonWords.qml
jpetrell: Discussed with Serj, agreed to switch back to "OK". :+1:  :smile:

pages/settings/devicelist/temperature/PageTemperatureSensor.qml
jpetrell: Fixed.

pages/settings/devicelist/temperature/PageTemperatureSensor.qml
jpetrell: True. Fixed.

blammit (approved):

PR #770, opened 2024-02-19
Fix values and spelling in SolarCharger_State

blammit:
The fault state is 2, not 1. This was pushing all the other states off by one.

Also fix the spelling for the bulk state.


blammit:
@ThiemoVanEngelen is it fine to leave out the "low power" state for solar chargers? I've left it out as the https://github.com/victronenergy/venus/wiki/dbus#solar-chargers documentation does not include it.

jpetrell:
Not sure how to verify, but the code LGTM.

PR #775, opened 2024-02-19
Fix invisible time-to-go

### LINKED ISSUE #768: Battery page: Time to go not visible
The time to go is not visible on the battery page, even if the item is present and valid.

https://github.com/victronenergy/gui-v2/blob/7873e22993e976916ca45489fe18023920e02694/pages/settings/devicelist/battery/PageBattery.qml#L182-L186

Probably because there is no `dataItem` in this case.

jpetrell:
Fixes #768.

### DIFF
--- a/pages/settings/devicelist/battery/PageBattery.qml
+++ b/pages/settings/devicelist/battery/PageBattery.qml
@@ -182,6 +182,7 @@ Page {
 			ListTextItem {
 				//% "Time-to-go"
 				text: qsTrId("battery_time_to_go")
+				dataItem.uid: root.battery.serviceUid + "/TimeToGo"
 				visible: defaultVisible && dataItem.seen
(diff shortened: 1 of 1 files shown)


jpetrell:
PR https://github.com/victronenergy/gui-v2/pull/775

jpetrell:
Checked that gui v1 and gui v2 showed the same values. gui v1 additionally had "show: item.seen" (VeQItem::getSeen()) visibility check, wonder if we should add the same for gui v2 also.

mr-manuel:
Yes, I would also add item.seem since not every battery provides this value.

blammit (commented):
Yes, we should keep the `seen` reference. The bug here is that `dataItem.uid` was not set to the `/TimeToGo` path.

blammit (approved):

PR #778, opened 2024-02-20
Polish device list items

jpetrell:
Adds a missing press effect and tunes list text group margins. These are minor issues , but big enough to irk me every time I visit the settings. :sweat_smile: 

Screenshot from 2024-02-20 17-00-30

### DIFF
--- a/components/listitems/ListTextGroup.qml
+++ b/components/listitems/ListTextGroup.qml
@@ -18,24 +18,19 @@ ListItem {
 		Repeater {
 			id: repeater
 
-			delegate: Label {
-				id: label
+			delegate: Row {
 				anchors.verticalCenter: !!parent ? parent.verticalCenter : undefined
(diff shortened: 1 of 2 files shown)


components/listitems/ListTextGroup.qml
```diff
@@ -18,28 +18,24 @@ ListItem {
 		Repeater {
 			id: repeater
 
-			delegate: Label {
-				id: label
+			delegate: Row {
+				property bool lastItem:  model.index === repeater.count - 1
```
blammit, line 22: Not sure if this is worth an extra property, but ymmv

blammit (approved):
It was irking me too, good to see it fixed

components/listitems/ListTextGroup.qml
jpetrell: Indeed. Fixed.

PR #782, opened 2024-02-20
Display a placeholder on empty energy meter settings

### LINKED ISSUE #767: Add explanation for empty energy meters menu
The Energy Meters menu, for historical reasons, show only Carlo Gavazzi meters connected through RS485. 
Make this clear by showing explanatory text.

See https://github.com/victronenergy/gui/commit/1443020d1b235dbe82c5c1454f0cb56f4056ff78

jpetrell:
Fixes #767.

### DIFF
--- a/pages/settings/PageSettingsCGwacsOverview.qml
+++ b/pages/settings/PageSettingsCGwacsOverview.qml
@@ -27,7 +27,18 @@ Page {
 	}
 
 	GradientListView {
+		id: energyMeterList
 		model: deviceIds.value ? deviceIds.value.split(',') : []
+
+		header: ListLabel {
+			visible: energyMeterList.count === 0
+			//% "No energy meters found\n\n"
+			//% "Note that this menu only shows Carlo Gavazzi meters connected over RS485. "
+			//% "For any other meter, including Carlo Gavazzi meters connected over ethernet, "
+			//% "see the Modbus TCP/UDP devices menu instead."
+			text: qsTrId("settings_cgwacs_no_energy_meters")
+		}
+
 		delegate: ListNavigationItem {
 			readonly property string devicePath: Global.systemSettings.serviceUid + "/Settings/Devices/cgwacs_" + modelData
 
--- a/components/GradientListView.qml
+++ b/components/GradientListView.qml
@@ -10,11 +10,11 @@ import Victron.VenusOS
 ListView {
 	id: root
 
-	x: Theme.geometry_page_content_horizontalMargin
-	width: parent.width - Theme.geometry_page_content_horizontalMargin
+	width: parent.width
 	height: parent.height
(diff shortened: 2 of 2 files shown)


jpetrell:
WIP Using ListLabel component as that was used by other pages when there was no content, though it doesn't work that well with the primary label + longer description here so still asking design for help with the visual structure.

jpetrell:
> so still asking design for help with the visual structure.

There is no empty state design yet, but this aligns with existing empty state labels so I guess ready for review.

pages/settings/PageSettingsCGwacsOverview.qml
```diff
@@ -27,7 +27,18 @@ Page {
 	}
 
 	GradientListView {
+		id: energyMeterList
 		model: deviceIds.value ? deviceIds.value.split(',') : []
+
+		ListLabel {
```
blammit, line 33: It looks like there's not enough space to the right of the label. Can you put this inside the ListView header? Otherwise I guess it doesn't get the extra left/right margins from the list view.

jpetrell: Oh true the horizontal margins are now asymmetrical, good catch. Fixed. I guess this explains why there are so many empty state labels defined on the header.  Though if there is no list background does the label still require double margins, both `geometry_page_content_horizontalMargin` and `geometry_listItem_content_horizontalMargin`? One problem with using the header is that it pushes the content down so you cannot do a nice cross-fade when the model gets populated (maybe at some point).  It is a bit odd that the left margin is done by moving GradientListView.x, but the right margin is done as margins within the content area one level lower on the item tree.

blammit: The main idea is that the ListLabel text should line up with the text of other list items, like this:

image

So if they don't line up due to the double margins, then we need to fix that somehow.

It's true that with the headerItem you can't do a nice cross-fade when the model is populated. Perhaps it can be done with two separate ObjectModels, for the empty and non-empty cases, and with a `populate` transition on the non-empty model to fade in the contents.

I don't remember if there was a good reason for changing the GradientListView `x` instead of `leftMargin`. If it works fine with `leftMargin`, let's do that instead :)

jpetrell: > The main idea is that the ListLabel text should line up with the text of other list items, like this:

Thanks, got it now. Makes sense. :+1: 

> I don't remember if there was a good reason for changing the GradientListView x instead of leftMargin

Changed it now. Browsed around the gui v2 pages, didn't see regression so replaced the x margin with leftMargin to avoid the asymmetry.

blammit (approved):
@jpetrell can you separate the GradientListView change into a separate commit? Otherwise, LGTM.

PR #783, opened 2024-02-20
Re-enable animation between battery and inverter/charger

### LINKED ISSUE #567: Overview Page: Battery connection line is not animated
The battery connection line on the overview page is not animated. I tried with v0.1.20 and the latest main build.

guiv2_bubbles

jpetrell:
Fixes #567.

Not sure if this is a correct fix, but after the change System.qml and SystemImpl.qml uid paths start to match and the animation starts flowing on both the Ekrano (demo mode) and mocked desktop,


blammit (approved):
Excellent, this is the correct fix :)

PR #787, opened 2024-02-21
Drop weather from the side panel for now

### LINKED ISSUE #717: Implement weather fallback
As discussed in the meeting today, @jepefe will talk to Serj about implementing a fallback design. And in v1.0.0 release we'll go for that.

Two reasons: making this work requires discussion + work outside of gui-v2; and we need a fallback anyway for systems that don't have internet.

jpetrell:
- Hide the unfinished weather widget by default
- Update the side panel component spacings to match the design
- To test in mock mode toggle weather on/off using keyboard shortcut "w"

Fixes #717.


blammit (commented):
Let's just remove the code altogether. If the weather UI is added later on it might look different anyway, or with updated icons or such. Also, "System/Weather/Enabled" is not a real value outside of the mock environment so best to avoid adding it in mock mode, to reduce confusion.

jpetrell:
> Let's just remove the code altogether.

I've implemented here both "weatherless" and "weatherful" layouts, but sure can drop the weather altogether also if there are no immediate plans to introduce it back the right after. Overall I agree on principle of keeping all the speculative code out, and the current weather implementation that was here seemed mostly a dummy skeleton.

blammit (approved):

PR #788, opened 2024-02-22
Add zh translation file to CMakeLists.txt and update translations

blammit:


blammit:
@dirkjanfaber or should we rename the ts file to `venus-gui-v2_zh_CN.ts` (or `zh-CN`?) as discussed?

dirkjanfaber:
> @dirkjanfaber or should we rename the ts file to `venus-gui-v2_zh_CN.ts` (or `zh-CN`?) as discussed?

Lets go for the minus, so `zh-CN`. That is also what POEditor uses.

blammit:
> > @dirkjanfaber or should we rename the ts file to `venus-gui-v2_zh_CN.ts` (or `zh-CN`?) as discussed?
> 
> Lets go for the minus, so `zh-CN`. That is also what POEditor uses.

Updated as suggested.

blammit:
@dirkjanfaber are you happy to approve this?

blammit:
> > > @dirkjanfaber or should we rename the ts file to `venus-gui-v2_zh_CN.ts` (or `zh-CN`?) as discussed?
> > 
> > 
> > Lets go for the minus, so `zh-CN`. That is also what POEditor uses.
> 
> Updated as suggested.

I changed it to `zh_CN` as the translations didn't load otherwise. Could you verify?

scripts/upload_new_terms.sh
```diff
@@ -9,9 +9,9 @@ then
   exit 1
 fi
 
-if [ ! -f i18n/venus-gui-v2.ts ]
+if [ ! -f build/i18n/venus-gui-v2.ts ]
```
DanielMcInnes, line 12: the comment above is a bit misleading. Could you please change:
`# is running `lupdate`, and thus producing i18n/venus-gui-v2.ts`
to:
`# is running `lupdate`, and thus producing .../path/to/build/dir/i18n/venus-gui-v2.ts`

(There is a file ./i18n/venus-gui-v2.ts checked into git. It doesn't get modified during the build, there is another one in the build directory that gets updated via lupdate)

CMakeLists.txt
```diff
@@ -1113,6 +1114,7 @@ qt_add_translations( ${PROJECT_NAME}
       "${CMAKE_CURRENT_BINARY_DIR}/i18n/${PROJECT_NAME}_th.ts"
       "${CMAKE_CURRENT_BINARY_DIR}/i18n/${PROJECT_NAME}_tr.ts"
       "${CMAKE_CURRENT_BINARY_DIR}/i18n/${PROJECT_NAME}_uk.ts"
+      "${CMAKE_CURRENT_BINARY_DIR}/i18n/${PROJECT_NAME}_zh_CN.ts"
```
DanielMcInnes, line 1117: I don't like the duplication of these lists, I've factored it out here: https://github.com/victronenergy/gui-v2/pull/801
I smoke tested it, seems to work (changing language to 'French' makes the 'Levels' button appear as 'Niveaux')

scripts/upload_new_terms.sh
```diff
@@ -9,9 +9,9 @@ then
...
 then
-  echo "No i18n/venus-gui-v2.ts found"
+  echo "No build/i18n/venus-gui-v2.ts found"
```
DanielMcInnes, line 14: I think both of these scripts (`download_from_poeditor.sh` and `upload_new_terms.sh`) should be done in cmake. If you agree, you can assign a ticket to me if you want.

dirkjanfaber: Yes, that is a good idea. Created issue #803 for this.

dirkjanfaber (approved):
lgtm. Tested a native linux build and got:
```
venus.gui: Successfully loaded translations for locale "zh_CN"
```

Also the screen shows Chinese characters.

PR #789, opened 2024-02-22
Fix Windows builds

chriadam:
Don't symlink venus-gui-v2.ts to venus-gui-v2_en.ts because that doesn't work properly on Windows.

Instead, copy venus-gui-v2.ts to builddir/i18n/venus-gui-v2_en.ts as well as to builddir/i18n/venus-gui-v2.ts.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1079,7 +1079,6 @@ file(COPY
       "i18n/${PROJECT_NAME}_cs.ts"
       "i18n/${PROJECT_NAME}_da.ts"
       "i18n/${PROJECT_NAME}_de.ts"
-      "i18n/${PROJECT_NAME}_en.ts"
       "i18n/${PROJECT_NAME}_es.ts"
       "i18n/${PROJECT_NAME}_fr.ts"
       "i18n/${PROJECT_NAME}_it.ts"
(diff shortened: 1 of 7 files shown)


blammit (approved):

blammit (approved):

dirkjanfaber (approved):
Lgtm.

chriadam:
> Getting following error now when calling cmake on desktop Linux.
> 
> ```
> CMake Error at CMakeLists.txt:1098 (file):
>   file COPY cannot find
>   "../gui-v2/i18n/venus-gui-v2.ts": No such
>   file or directory
> ```
> 
> After calling cmake venus-gui-v2.ts contents are gone
> 
> ```
> $ git status
> ...
> 	modified:   i18n/venus-gui-v2.ts
> 
> $ more i18n/venus-gui-v2.ts 
> <?xml version="1.0" encoding="utf-8"?>
> <!DOCTYPE TS>
> <TS/>
> ```

I cannot reproduce.

I don't have a desktop Linux dev environment set up for Victron at the moment (I use Windows for desktop+wasm builds, and a WSL2 ArchLinux instance for device builds, and all of those work for me).

Are you sure that you're doing a shadow build (i.e. building from within a "build" directory), so that the CMAKE_CURRENT_BINARY_DIR is not the same as the PROJECT_SOURCE_DIR ?  If you are doing an in-source build, I could see why the issue could occur, but we don't support in-source builds, I think.

DanielMcInnes (approved):
works for me on desktop linux

chriadam:
Thanks, rebase-merging now.

PR #790, opened 2024-02-22
Add swipe-based navigation for main pages

blammit:


blammit:
In light mode, the background changes color a little abruptly when dragging between the Brief and Overview pages. Could probably implement a nice transition that sets the color depending on the distance dragged between the two pages.

components/NavBar.qml
```diff
@@ -68,8 +68,7 @@ Rectangle {  // Use an opaque background so that page disappears behind nav bar
 				backgroundColor: "transparent"
 
 				onClicked: {
-					root.currentIndex = model.index
-					root.currentUrl = model.url
+					Global.mainView.showPage(model.index)
```
jpetrell, line 71: It is nice to use slide to hint swiping, but the cross-fade felt calmer than the movement. Also cross-fade behaves consistently, with the slide the travelled distance varies.

components/controls/SwipeView.qml
```diff
@@ -0,0 +1,48 @@
...
+*/
+
+import QtQuick
+import QtQuick.Templates as T
+
+T.SwipeView {
+	id: control
+
+	readonly property bool flicking: listView.flicking
+	readonly property bool dragging: listView.dragging
```
jpetrell, line 13: Use alias?

pages/LevelsPage.qml
```diff
@@ -117,4 +121,14 @@ Page {
 
 		visible: tabBar.currentIndex === 1
 	}
+
+	// Show gradients on the left/right edges to indicate the page bounds
+	ViewGradient {
+		x: -(width / 2) + (height / 2)
+		rotation: 90
```
jpetrell, line 128: visible: root.clip?

pages/MainView.qml
```diff
@@ -4,28 +4,43 @@
...
+			   : !!pageStack.currentItem ? pageStack.currentItem
+			   : swipeView.currentItem
+
+	// To reduce the animation load, disable page animations when the PageStack is transitioning
+	// between pages, or when flicking between the main pages. Note that animations are still
+	// allowed when dragging between the main pages, as it looks odd if animations stop abruptly
+	// when the user drags slowly between pages.
+	readonly property bool allowPageAnimations: BackendConnection.applicationVisible
+			&& !pageStack.busy && !swipeView.flicking
+			&& !Global.splashScreenVisible
```
jpetrell, line 25: Having two "page managers" i.e. PageManager and PageStack already felt odd, now MainView starts to feel like a page manager too as it maintains currentPage and allowPageAnimations. Would try to keep page "global public API", properties and methods in one component, now you push/pop with PageManager and currentPage state in MainView changes.

blammit: I wonder if we should get rid of PageManager. Most of the things it manages are in the MainView, and the rest could probably be moved to Global. E.g. I would consider moving pushPage() to become Global.pushPage() as a signal, and then MainView could listen for the signal to push a page. I'm not sure PageManager offers that much anymore, compared to when it was first added.

components/NavBar.qml
blammit: It's true it is calmer. I'll check with Serj.

blammit: I'm not sure about cross-fading if we're using the ListView internally, but we could jump straight to the position instead?

pages/LevelsPage.qml
blammit: Done

pages/MainView.qml
```diff
@@ -120,8 +90,13 @@ Item {
 			bottom: parent.bottom
 		}
 
+		Rectangle {
+			anchors.fill: parent
+			color: Theme.color_page_background
+		}
```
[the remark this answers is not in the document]
blammit: Must be leftover from a state where they were needed. They're not required now, removing.

components/controls/SwipeView.qml
blammit: Done

components/NavBar.qml
chriadam: The travelled distance does vary with the slide, but I think that's intentional, and matches the user's mental model of the UI (well, as it is now - a series of side-by-side pages, which they swipe between).  I think the current impl. is fine, personally.

components/controls/SwipeView.qml
chriadam: No, don't use an alias.

The contentItem should never be given an id (same with background item).
They can be overridden by the client at usage time, but if they have an id the QML engine has to construct the item in order to hold a reference in the component scope object.

If you use the qmltc it will complain vociferously about this sort of aliasing ("late binding") since it simply won't work if the client overrides the contentItem property.

chriadam: Oh, but note that you cannot bind to `listview.whatever` either.  When the values change in the listview, it has to promote them up to the root level.  That is, root needs `property bool flicking` and `property bool dragging`, and the listview needs onChanged handlers which set `root.flicking = flicking` and `root.dragging = dragging`

jpetrell: If qmltc just allows I would use alias as it allows reduce the number of moving parts. This is not really a public API that can deal with everything possible and impossible, but an internal UI component currently used in one place in a specific way. If you want SwipeView that is not tied to ListView then that is arguably a different swipe view altogether than this component.

chriadam: qmltc doesn't allow it.  But, now that I think about it - we can't use qmltc anyway, because we need the "load files from filesystem at runtime" support for cerbogx / device anyway.  So I guess we can just ignore the "late binding" problem, after all.  (I still think it's technically "wrong" but from a pragmatic perspective, I guess it doesn't matter, and it certainly is simpler to use an alias property, I agree.)

PR #791, opened 2024-02-22
Reload web browser page when firmware version change detected

chriadam:
If the device has its firmware updated, it may have a new build of the UI included.  To ensure that the version of the UI which is being used via the browser matches the version of the UI which is being used on device, we need to reload the page so that the up-to-date wasm binary is reloaded.

### DIFF
--- a/pages/DialogLayer.qml
+++ b/pages/DialogLayer.qml
@@ -9,13 +9,34 @@ import Victron.VenusOS
 Item {
 	id: root
 
+	anchors.fill: parent
+
 	function open(dialogComponent, properties) {
-		const dialog = dialogComponent.createObject(Global.dialogLayer, properties)
+		const dialog = dialogComponent.createObject(root, properties)
 		dialog.closed.connect(function() {
 			dialog.destroy()
 		})
 		dialog.open()
 	}
 
-	anchors.fill: parent
+	// For WebAssembly, if the firmware changed on device, this might
+	// mean that the webassembly blob served by its webserver has changed.
+	// We need to trigger a page reload to ensure we are running the right one.
+	property Component _firmwareVersionRestartDialog: Component {
+		ModalWarningDialog {
+			dialogDoneOptions: VenusOS.ModalDialog_DoneOptions_NoOptions
+			//% "Firmware version update detected"
+			title: qsTrId("firmware_installed_build_version_update_detected")
+			//% "Page will automatically reload in ten seconds to load the latest version."
+			description: qsTrId("firmware_installed_build_page_will_reload")
+			Timer {
+				running: true
+				interval: 10*1000
+				onTriggered: Qt.quit() // the aboutToQuit handler will trigger page reload.
+			}
+		}
+	}
+
+	property bool _needPageReload: Qt.platform.os == "wasm" && Global.firmwareInstalledBuildUpdated
+	on_NeedPageReloadChanged: if (_needPageReload) open(_firmwareVersionRestartDialog)
 }
--- a/Global.qml
+++ b/Global.qml
@@ -55,6 +55,9 @@ QtObject {
 	property bool allPagesLoaded
 	property bool changingLanguage
 
+	property string firmwareInstalledBuild // don't clear this on UI reload.  it needs to survive reconnection.
+	property bool firmwareInstalledBuildUpdated // as above.
+
 	signal aboutToFocusTextField(var textField, int toTextFieldY, var flickable)
 	signal keyPressed(var event)
 
--- a/components/FirmwareUpdate.qml
+++ b/components/FirmwareUpdate.qml
@@ -55,10 +55,16 @@ QtObject {
 				msg = qsTrId("settings_firmware_error_during_installation")
 				break
 			case FirmwareUpdater.Rebooting:
-				//% "Firmware installed, rebooting"
+				//% "Firmware installed, device rebooting"
 				msg = qsTrId("settings_firmware_installed_rebooting")
+				// Note: for WebAssembly, don't yet reload the page.
+				// If we did the reload now, we would still get the wrong blob.
+				// Instead, wait for the device to reboot (i.e. requires
+				// a full disconnect/reconnect cycle) then react to the
+				// updated build version we receive (reload the page).
 				break
 			}
+
 			if (msg) {
 				// TODO confirm whether we need to show "icon-firmwareupdate-active" instead of the normal notification icon
 				Global.showToastNotification(VenusOS.Notification_Info, msg, 10000)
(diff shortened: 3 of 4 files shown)


data/SystemSettings.qml
```diff
@@ -278,6 +278,74 @@ QtObject {
 		}
 	}
 
+	property VeQuickItem _firmwareInstalledBuild: VeQuickItem {
+		property string newValue
```
jpetrell, line 282: newValue seems unnecessary? it is only used within one signal handler?

data/SystemSettings.qml
```diff
@@ -278,6 +278,74 @@ QtObject {
...
+		uid: Global.venusPlatform.serviceUid + "/Firmware/Installed/Build"
+		onValueChanged: {
+			if (value != undefined) {
+				newValue = value
+				if (newValue.length > 0) {
```
jpetrell, line 287: If newValue gets dropped could combine the conditionals `if (value != undefined && value.length > 0) { ..`

data/SystemSettings.qml
```diff
@@ -278,6 +278,74 @@ QtObject {
...
+							+ " (" + Global.firmwareInstalledBuild + " != " + newValue + ")")
+						// If the value change was only detected after a reconnection (and rebuildUi)
+						// then the dialogLayer may not have been constructed yet.
+						// ApplicationContent is constructed asynchronously, so might require retries.
+						if (Global.dialogLayer) {
+							Global.dialogLayer.open(firmwareVersionRestartDialog)
+						} else {
+							timer.needRestart = true
+							timer.interval = 2000
+							timer.start()
```
jpetrell, line 302: Would consider creating another Timer instead to improve code readability.

data/SystemSettings.qml
```diff
@@ -278,6 +278,74 @@ QtObject {
...
+					Global.firmwareInstalledBuild = value
+				}
+			}
+		}
+
+		property Timer timer: Timer {
+			property bool needRestart
+			interval: 10 * 60 * 1000
+			repeat: true
+			running: true
```
jpetrell, line 314: running: Qt.platform.os == "wasm"?

data/SystemSettings.qml
```diff
@@ -278,6 +278,74 @@ QtObject {
...
+					// after a device restart, and MQTT comms can be unreliable.
+					// So, periodically force read the firmware installed build value.
+					root._firmwareInstalledBuild.getValue(true)
+				}
+			}
+		}
+
+		property Component firmwareVersionRestartDialog: Component {
+			ModalWarningDialog {
+				id: dlg
```
jpetrell, line 333: Unused id. I generally take visible id declaration as an indication that the element is accessed somewhere.

data/SystemSettings.qml
```diff
@@ -278,6 +278,74 @@ QtObject {
...
+				property bool languageChangeFailed
+				dialogDoneOptions: VenusOS.ModalDialog_DoneOptions_NoOptions
+				//% "Firmware version update detected"
+				title: qsTrId("firmware_installed_build_version_update_detected")
+				//% "Page will automatically reload in ten seconds to load the latest version"
+				description: qsTrId("firmware_installed_build_page_will_reload")
```
jpetrell, line 339: Is there a rule when to end the sentences in period? Anyhow most ModalWarningDialog I could find have titles that don't use periods, but descriptions that do.

data/SystemSettings.qml
chriadam: fixed

data/SystemSettings.qml
chriadam: done

data/SystemSettings.qml
chriadam: yes, done

data/SystemSettings.qml
chriadam: indeed, removed

ReinvdZee (approved):
LGTM

blammit (approved):
Can you add the task number to the commit?

blammit:
@chriadam can you remind me why the  `/Firmware/Installed/Build` value check is needed? If I update the firmware on the cerbo, the UI in wasm will trigger the "Firmware installed, rebooting" toast notification from https://github.com/victronenergy/gui-v2/blob/2a28d7b0ebc2b3893bf8a5f631ff9e4f7c3fbf62/components/FirmwareUpdate.qml#L57. Why can't we just trigger the reboot dialog from there on wasm, instead of having this separate `/Firmware/Installed/Build` check? Are there cases where that `FirmwareUpdate` state change is not triggered when the firmware is updated?

data/SystemSettings.qml
```diff
@@ -278,6 +278,70 @@ QtObject {
 		}
 	}
 
+	property VeQuickItem _firmwareInstalledBuild: VeQuickItem {
+		uid: Global.venusPlatform.serviceUid + "/Firmware/Installed/Build"
+		onValueChanged: {
+			if (value != undefined && value != null && value.length > 0) {
```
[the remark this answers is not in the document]
chriadam: indeed, did null and length>0

data/SystemSettings.qml
```diff
@@ -278,6 +278,70 @@ QtObject {
...
+				console.log("Firmware version received: " + value + ", previously seen value: " + Global.firmwareInstalledBuild)
```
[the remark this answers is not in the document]
chriadam: true, removed

data/SystemSettings.qml
```diff
@@ -278,6 +278,70 @@ QtObject {
...
+					} else {
+						reloadPageTimer.start()
+					}
+				}
+				Global.firmwareInstalledBuild = value
+			}
+		}
+
+		property Timer forceReadTimer: Timer {
+			property bool needRestart
```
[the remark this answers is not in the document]
chriadam: removed

data/SystemSettings.qml
```diff
@@ -278,6 +278,70 @@ QtObject {
 		}
 	}
 
+	property VeQuickItem _firmwareInstalledBuild: VeQuickItem {
+		uid: Global.venusPlatform.serviceUid + "/Firmware/Installed/Build"
+		onValueChanged: {
+			if (value != undefined && value != null && value.length > 0) {
+				console.log("Firmware version received: " + value + ", previously seen value: " + Global.firmwareInstalledBuild)
+				if (Qt.platform.os == "wasm"
+						&& Global.firmwareInstalledBuild.length > 0
```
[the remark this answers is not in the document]
chriadam: It is needed, because on first start it will initially be empty.  This is fine, and doesn't warrant a reboot, thus this forms part of the condition.

data/SystemSettings.qml
```diff
@@ -278,6 +278,70 @@ QtObject {
...
+						&& Global.firmwareInstalledBuild != value) {
+					console.warn("Firmware update detected, reloading page in 10 seconds"
+						+ " (" + Global.firmwareInstalledBuild + " != " + value + ")")
+					// If the value change was only detected after a reconnection (and rebuildUi)
+					// then the dialogLayer may not have been constructed yet.
+					// ApplicationContent is constructed asynchronously, so might require retries.
```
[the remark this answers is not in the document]
chriadam: Good points, done.

chriadam:
> Can you add the task number to the commit?

Done

chriadam:
> @chriadam can you remind me why the `/Firmware/Installed/Build` value check is needed? If I update the firmware on the cerbo, the UI in wasm will show the "Firmware installed, rebooting" toast notification due to this line in `FirmwareUpdate`:
> 
> https://github.com/victronenergy/gui-v2/blob/2a28d7b0ebc2b3893bf8a5f631ff9e4f7c3fbf62/components/FirmwareUpdate.qml#L57
> 
> Why can't we just trigger the reboot dialog from there on wasm, instead of having this separate `/Firmware/Installed/Build` check? Are there cases where that `FirmwareUpdate` state change is not triggered when the firmware is updated?

I've moved the impl to the FirmwareUpdate - that's definitely where it belongs.
We do still need the Firmware/Installed/Build check, though, for the case where we do a downgrade / reboot previously installed firmware version.
I've added a comment about why we don't want to trigger page reload immediately upon detecting that the device is rebooting.

pages/DialogLayer.qml
```diff
@@ -9,13 +9,34 @@ import Victron.VenusOS
 Item {
 	id: root
 
+	anchors.fill: parent
+
 	function open(dialogComponent, properties) {
-		const dialog = dialogComponent.createObject(Global.dialogLayer, properties)
+		const dialog = dialogComponent.createObject(root, properties)
```
chriadam, line 15: this change is because Global.dialogLayer won't be set until the Component.onCompleted handler is run, which might not happen until after the property change is noticed and reacted to.  I could have called Qt.callLater() in on_NeedPageReloadChanged, but I think it's more direct to simply specify the parent properly/directly here in this function (since Global.dialogLayer should always equal root anyway).

blammit (approved):

Issue #724, opened 2024-02-08
Align GUI text capitalization rules


jpetrell:
Some UI labels use all capitalized Title Case e.g. "Device List", "Date & Time", some Sentence case e.g. "System setup", "Remote support". The rule may differ between headings and body text, descriptive labels and interactive buttons, but the current strings don't seem to be any consistent rule.

Check the design intent, implement consistently across the current gui-v2.

chriadam:
We have been trying to standardise on sentence-case for all settings page texts, for example.  Maybe copilot could help us do this en-masse.

chriadam:
Merged to main, closing as fixed.

PR #793, opened 2024-02-22
Polish text

jpetrell:
Put few hours in polishing engineering English source strings:
- Don't show period on toast with only one sentence (most already didn't)
- Use period on modal warning dialog body text and descriptive labels written in proper sentences e.g. "Manually reboot the GX device after changing these settings." (most already did)
- Use left-aligned ListLabel when displaying descriptive labels on pages (most already did)
- Talk about "autostart" and not "auto start", both forms were in use, the former is recognized by the dictionaries.
- Renamed "Alarms and Errors" to "Alarms & Errors" to align with "Date & Time" and "Display & Language"
- Wrote "Yield today" on solar charger tables as "Yield Today" to align with the rest of column titles (also used in the design)

Overall small changes, 30 strings out of total 1.5k. Tried to align the naming with GX manual https://www.victronenergy.com/media/pg/CCGX/en/index-en.html

Contributes to #724.

image

### DIFF
--- a/components/CommonWords.qml
+++ b/components/CommonWords.qml
@@ -465,7 +465,7 @@ QtObject {
 	readonly property string yield_kwh: qsTrId("common_words_yield_kwh")
 
 	//: Solar charger yield for today, in kWh (kilowatt hours)
-	//% "Yield today"
+	//% "Yield Today"
 	readonly property string yield_today: qsTrId("common_words_yield_today")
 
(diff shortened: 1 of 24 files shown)


blammit (approved):
I think these small changes help a lot. A lot of the strings were copied directly from gui-v1 but seems fine to update them if the meaning if it makes the meaning more clear.

PR #799, opened 2024-02-22
Dim disabled time selector more

jpetrell:
Make it clear that the time selector is not interactive when "Timed run" switch is disabled.

image


blammit (approved):
Nice!

PR #802, opened 2024-02-26
Don't display '0% 0.00/0.000' for tanks without "capacity" and "remai…

DanielMcInnes:
…ning" values

### DIFF
--- a/components/TankGaugeGroup.qml
+++ b/components/TankGaugeGroup.qml
@@ -137,6 +137,7 @@ Rectangle {
 		width: Theme.geometry_levelsPage_gaugeDelegate_contentWidth
 		font.pixelSize: Theme.font_size_h1
 		unit: VenusOS.Units_Percentage
+		visible: !isNaN(root.level)
 		value: (isNaN(root.level) || root.level < 0) ? 0 : Math.round(root.level)
 	}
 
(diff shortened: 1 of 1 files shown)


components/TankGaugeGroup.qml
```diff
@@ -137,6 +137,7 @@ Rectangle {
 		width: Theme.geometry_levelsPage_gaugeDelegate_contentWidth
 		font.pixelSize: Theme.font_size_h1
 		unit: VenusOS.Units_Percentage
+		visible: !isNaN(root.level)
```
blammit, line 140: The level should always be present though, shouldn't it?

blammit (commented):
Part of the bug is also that the overall levels value is incorrect when there are multiple tanks without capacity/remaining -- were you able to reproduce this? (Even with mock values?)

components/TankGaugeGroup.qml
DanielMcInnes: I don't think so. If you look at the picture from the slack thread, the percentage text is 0%, while the gauge is showing some other non-zero value
image

DanielMcInnes:
> Part of the bug is also that the overall levels value is incorrect when there are multiple tanks without capacity/remaining -- were you able to reproduce this? (Even with mock values?)

No I wasn't. There is something screwy with the mock tanks - if you hard code the TankGauge value to 0.5:
`
// TankGaugeGroup.qml
value: 0.5 // (root.mergeTanks ? model.device.level : root.level) / 100`

... the gauges jump all over the place. If I run the same code in a wasm build, it behaves as you would expect, every gauge is at 50%.

Martin gave me the vrm portal id for the installation that reported the issue. I want to test against the real installation before I merge this. The site is currently offline.

DanielMcInnes:
I'll leave this PR back in 'draft' until I can test against the real installation (currently offline) and verify that it looks right

blammit:
This PR is obsoleted by https://github.com/victronenergy/gui-v2/pull/854, closing.

PR #805, opened 2024-02-27
Fix up brightness settings

blammit:


jpetrell:
> feels dangerous that you can basically blank the display by lowering the brightness to almost zero with the slider

Filed https://github.com/victronenergy/gui-v2/issues/809

blammit:
> Overall LGTM.
> 
> Yay press effects. \o/ One nitpick: the button press effect is quite subtle on light style, if I didn't know would not realize it is there.

Agreed -- for simplicity/consistency, this is using the same press effect as the back navigation button, but I think the press effects/coloring need an overall review throughout the app. Some buttons do not have any press effects, which is confusing.

> I would expect the slider to provide valueChanged signal, but I like the simplicity that if there is no use case that needs it currently don't provide it. I guess the QtQuick.Controls Slider.value still provides value change signaling if needed. Did you test all the existing sliders that they don't break? At least the slider on ESS debug page changed the values just fine, whether I dragged or tapped.

Yes, we only have two sliders at the moment -- in the ESS page, and the display settings. 


> Not caused by these changes, but feels dangerous that you can basically blank the display by lowering the brightness to almost zero with the slider. Also the minimum and maximum positions of the handles don't visually seem to go all the way to the end, which looks slightly off. And it is too bit too easy to accidentally touching the track when reaching for the minus and plus buttons. But again not caused by you changes so approved. 😄

Yes, would be nice to improve those visuals/interaction features. Maybe the hit box for various buttons is another thing that we need to review throughout the app.

I think the brightness min is supposed to be 1, but the app is getting 0 from the dbus/mqtt later, so need to investigate further there.

PR #813, opened 2024-02-28
Implement tank item redesign

### LINKED ISSUE #806: Update tanks to latest design
See Tanks visual improvement proposal from Serj, where the colored borders have been removed from the tank gauges, and the color coding is done behind the top tank label instead.

jpetrell:
While implementing this realized that the new tank style doesn't provide layout for the the tank group that you see when tapping a grouped tank item.

Fixes #806.


jpetrell:
Serj will finish the design next week, I will then finish the remaining bits.

jpetrell:
Ready for review.

blammit (approved):
I think it looks much better now :)

PR #814, opened 2024-02-28
Improve spacing between quantity table columns

### LINKED ISSUE #479: Adjust quantity table column widths to avoid eliding quantity values, where possible
Loading the PV inverter page on the local cerbo:

image

Would be nice if the energy in the QuantityTableSummary was not elided. Currently the first column has a fixed width, but would be nice to dynamically adjust this.

jpetrell:
Just small adjustment to the column sizes to make sure there are enough margins around the wider energy (kWh) column. Fixes #479.

image

### DIFF
--- a/components/QuantityTableMetrics.qml
+++ b/components/QuantityTableMetrics.qml
@@ -1,0 +1,41 @@
+/*
+** Copyright (C) 2024 Victron Energy B.V.
+** See LICENSE.txt for license information.
+*/
+
+import QtQuick
+import Victron.VenusOS
+
+FontMetrics {
+	property bool smallTextMode
+	property real availableWidth
+	property bool equalWidthColumns
+	property int count
+	property int spacing: Theme.geometry_quantityTable_horizontalSpacing
+
+	// If specified, allows for a custom column width for the 'Units_None' column.
+	// Eg. label columns with cells like "L1", "L2" can be thinner to allow wider columns elsewhere.
+	property int firstColumnWidth
+
+	font.family: Global.fontLoader.name
+	font.pixelSize: smallTextMode ? Theme.font_size_body2 : Theme.font_size_body3
+
+	function columnWidth(unit) {
+		if (!!firstColumnWidth) {
+			if (unit === VenusOS.Units_None) {
+				return firstColumnWidth
+			}
+			return (availableWidth - firstColumnWidth) / (count - 1)
+		}
+
+		if (equalWidthColumns) {
+			return availableWidth / count
+		}
+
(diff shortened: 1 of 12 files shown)


chriadam:
In 5-inch mode, this can cause the text to elide for solar chargers.  The text is "External control" and it renders properly without the patch applied.
venus-elide-text-issue


The width of the title label in QuantityTable.qml is set via `width: root.availableWidth - headerQuantityRow.width`  so I guess the patch causes the headerQuantityRow.width to increase in some cases?

jpetrell:
> The text is "External control" and it renders properly without the patch applied

That is actually unrelated, my changes actually reduced the total quantity row width a bit. But not enough, "External control" was still truncated. I fixed it now, also ended up refactoring the column width calculations to own component to avoid code duplication. Dropped the magical multipliers and width boost values, instead measuring now the maximum column size using FontMetrics (+ some spacing between the columns). Adjusted the seven inch variant to have more spacing so the layouts is still more evenly spread around the available space.

image

components/QuantityTableMetrics.qml
```diff
@@ -0,0 +1,40 @@
...
+			return (availableWidth - firstColumnWidth) / (count - 1)
+		}
+
+		if (equalWidthColumns) {
+			return availableWidth / count
+		}
+
+		// Give the unit symbol some extra space on the column.
+		const maxTextWidth = unit === VenusOS.Units_Energy_KiloWattHour ? advanceWidth("9999kWH")
+																		: advanceWidth("9999W")
```
blammit, line 37: Odd indentation? Maybe it's fine to just have it all on the same line

blammit (approved):
Looks great! The solution is a bit more complicated than I expected, but I think it's worthwhile to use the metrics to calculate things properly instead of just putting in random values and hoping it will work.

components/QuantityTableMetrics.qml
jpetrell: Fixed.

jpetrell:
> The solution is a bit more complicated than I expected

It may look daunting, but it fixes multiple issues and mostly it is just moving related properties and _quantityColumnWidth() function to a shared component to avoid code duplication between QuantityTable.qml and QuantityTableSummary.qml.

jpetrell:
> The solution is a bit more complicated than I expected

Also, the PR started with just small adjustments, but then Chris noticed other issues which could no longer be fixed by simply tweaking the multipliers. Also I was worried about the translations, any issue I fix with the magical values will be English-specific fix if I don't take the actual text dimensions into account.

PR #815, opened 2024-02-29
Fix startup glitch caused by QTBUG-115468

chriadam:
Don't dynamically add items to the SwipeView, since that is broken in Qt < 6.6.1.

Instead, declare them inline, and then load the whole thing asynchronously.

### DIFF
--- a/pages/MainView.qml
+++ b/pages/MainView.qml
@@ -14,45 +14,43 @@ Item {
 	property bool controlsActive
 	readonly property Page currentPage: controlsActive && controlCardsLoader.status === Loader.Ready ? controlCardsLoader.item
 			   : !!pageStack.currentItem ? pageStack.currentItem
-			   : swipeView.currentItem
+			   : !!swipeView ? swipeView.currentItem
+			   : null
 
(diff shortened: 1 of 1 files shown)


pages/MainView.qml
```diff
@@ -62,8 +60,24 @@ Item {
...
+		SwipeView {
+			anchors.fill: parent
+			onCurrentIndexChanged: navBar.setCurrentIndex(currentIndex)
+
+			// ensure the order matches NavBar.model
+			BriefPage {}
+			OverviewPage {}
+			LevelsPage {}
+			NotificationsPage {}
+			SettingsPage {}
```
jpetrell, line 80: Nice! It is a bit unfortunate there is no longer a direct mapping between the NavBar model and the page loading. Wonder if we should add // TODO about improving this once we migrate to newer Qt. :thinking: I guess no need, the risk of these going out-of-sync is small, and overall this is less code than before, which is nice.

jpetrell (approved):

pages/MainView.qml
chriadam: I agree

PR #816, opened 2024-02-29
Scan veutil (and other sources) for translation strings

chriadam:
veutil contains error messages which need to be translated. Ensure that we scan these sources for translation strings.


dirkjanfaber (approved):
Change looks good to me. Added a commit to get the new translations from poeditor. Feel free to skip that one when merging. 
In either case updated translations will need to be pushed and pulled periodically to and from poeditor.

blammit (approved):
Great, glad you picked this up.

PR #821, opened 2024-03-01
fix building with OpenEmbedded / Debian rules

jhofstee:
Since commit 59d894f73aa "Load qml files from the file system, not from compiled resources" the install target installs the files in a hardcoded directory build/install instead of the normal DESTDIR / bindir etc. As a consequence OpenEmbedded / Venus will fail to build the gui-v2. This patch restores installing files into the normal locations if CMAKE_INSTALL_PREFIX is set. If not set build/install will be used.

For completeness, this does not adjust / install the translation files. Translations do seem to work though..


DanielMcInnes:
These changes look good to me. I can't check out this branch to test it though, is this on a forked repo?
How is OpenEmbedded / Venus built, can you post a link to the repo please?
Also, please add reviewers to this code review, otherwise people won't know to review it.

jhofstee:
This is a branch in a fork, I will create a new pull request.

PR #826, opened 2024-03-04
make the wasm version work over https

jhofstee:
It is not tested as an action, since github warns for running actions on forks...


dirkjanfaber (approved):
Lgtm (even tested it via https://github.com/victronenergy/gui-v2/actions/runs/8153184483).

Issue #639, opened 2024-01-02
Electron/Ants animations on the Overview page cause high CPU usage


chriadam:
Currently, we construct an interpolator for each WidgetConnector path on the Overview page.
Then, for each WidgetConnector path, we animate a "progress" property.
Each electron on the path listens to the `progressChanged` signal and then asks the interpolator for the appropriate x/y/rotation values given the progress value (adjusted for the electron's index).

Most of this happens in QML and so the calculations are slow, and cause a variety of change signals / wasted side effects.
Since the animation is consistently performed when the Overview page is visible, this causes very high CPU usage.

We have a few different options to try:

--

1) move most of the math to a C++ class which pre-calculates the value at every progress interval on the interpolator spline, and populates a hash table with this information for each electron (i.e. `QHash<pathInterpolator*, QHash<int electronIndex, qreal progress> >`).  Each electron merely looks up the appropriate value from this hashtable at each animation step.  Advantages: most closely matches the current implementation, and may reduce the amount of work done in QML expressions significantly.  Disadvantages: we'd still be performing an extremely large number of x/y/rotation property updates in the GUI thread (and triggering all of the changes signals etc), so I don't expect huge perf improvement.

2) since each WidgetConnector path is quite simple, we should be able to split each one into 5 sections (horizontal section, first bend, vertical section, second bend, horizontal section) or maybe even just 3 sections (curve, straight, curve).  Each section defines x / y / rotation changes across their length.  Thus, we should be able to calculate just the start and end values for these three properties and use XAnimator/YAnimator/RotationAnimator to animate the values within each section.  Advantages: should greatly reduce CPU usage, as we'd be doing effectively (nbrSections * nbrProperties == 5 * 3) QQuickItem property writes per electron per animated path, rather than (nbrProperties * nbrAnimationTicks) QQuickItem property writes per electron per animated path.  Disadvantages: might be more difficult to calculate the start/end positions (and duration, i.e. based on pixel-velocity) than I predict.

3) we could just use a different method of animation altogether (e.g. could approximate the current design with a moving zebra-pattern texture opacity masked to the connector).  Advantages: performance would be much better.  Disadvantages: would look significantly worse.

[...]

PR #827, opened 2024-03-04
Reduce Overview page CPU load

jpetrell:
- Reduces the CPU consumption from 58% to 44%
- Optimize widget connector animation bindings
- Limit the widget connector animation to 15fps
- Replace the battery charging animation with a more light-weight transition (also got feedback from users that the the bubbling animation was too a bit too much #843)

[Screencast from 2024-03-13 02-00-36.webm](https://github.com/victronenergy/gui-v2/assets/2203667/5d308a3b-fd8f-477f-894f-486336d8005f)

Contributes to #639.


chriadam:
If we do it this way, it still keeps the property writes in the GUI thread (so signals will still be emitted, changes will still be synced to the render thread in updatePaintNode(), etc).  It has the advantage that it might be a good step in the direction of "replacing" all electrons with a single shader (the pre-calculated spline could be passed to the shader as a uniform), which could be even better than the option (2) listed in the task (because option (2) would keep all of the electron objects, and only reduce the amount of property writes which occur).  My one concern is that the shader might be expensive and the GPU might not be able to handle it.  But I may very well be wrong, and we'll only know by trying, I guess.

chriadam:
@jpetrell any chance you could recreate this PR from a branch of the gui-v2 repository rather than from your fork?  Just so that we get github actions etc building for it.

jpetrell:
> @jpetrell any chance you could recreate this PR from a branch of the gui-v2 repository rather than from your fork?

Sure, here https://github.com/victronenergy/gui-v2/pull/879.

PR #828, opened 2024-03-05
Fix broken idle mode

blammit:
After 3038b8cc5c74023a5f306bef28b0280069f7816f the application no longer went into idle mode, as currentPage moved into MainView.qml.


chriadam (approved):

jpetrell (approved):
LGTM. Works beautifully.

PR #830, opened 2024-03-05
Genset settings updates

blammit:


pages/settings/devicelist/ac-in/PageAcInModelGenset.qml
```diff
@@ -182,25 +182,29 @@ ObjectModel {
...
+							precision: 0
 						}
 
-						ListTextItem {
+						ListQuantityItem {
 							//% "Oil Pressure"
 							text: qsTrId("ac-in-genset_oil_pressure")
 							dataItem.uid: root.bindPrefix + "/Engine/OilPressure"
 							visible: defaultVisible && dataItem.isValid
+							unit: VenusOS.Units_Hectopascal
```
DanielMcInnes, line 199: Hectopascal? For oil pressure? That doesn't sound right

pages/settings/devicelist/ac-in/PageAcInModelGenset.qml
```diff
@@ -182,25 +182,29 @@ ObjectModel {
...
 						}
 
-						ListTextItem {
+						ListQuantityItem {
 							//% "Load"
 							text: qsTrId("ac-in-genset_load")
 							dataItem.uid: root.bindPrefix + "/Engine/Load"
 							visible: defaultVisible && dataItem.isValid
+							unit: VenusOS.Units_Percentage
+							precision: 0
```
jpetrell, line 191: Unnecessary as by default percentage already has precision 0?

pages/settings/devicelist/ac-in/PageAcInModelGenset.qml
```diff
@@ -182,25 +182,29 @@ ObjectModel {
...
 							visible: defaultVisible && dataItem.isValid
+							unit: VenusOS.Units_Hectopascal
 						}
 
 						ListTemperatureItem {
 							//% "Coolant temperature"
 							text: qsTrId("ac-in-genset_coolant_temperature")
 							visible: defaultVisible && dataItem.isValid
 							dataItem.uid: root.bindPrefix + "/Engine/CoolantTemperature"
+							precision: 0
```
jpetrell, line 207: Same as with percentages, the default precision for temperatures is already 0.

jpetrell: Ah while Units formats temperature with 0 precision ListTemperatureItem uses precision 1. There are 20 temperature lits items in settings, does this precision change affect only AC input settings for Genset models or should we visit all of them?

pages/settings/devicelist/ac-in/PageAcInModelGenset.qml
```diff
@@ -234,7 +234,7 @@ ObjectModel {
 							text: qsTrId("ac-in-genset_operating_time")
 							visible: defaultVisible && dataItem.isValid
 							dataItem.uid: root.bindPrefix + "/Engine/OperatingHours"
-							secondaryText: Utils.secondsToString(dataItem.value)
+							secondaryText: Utils.formatAsHHMM(dataItem.value, true)
```
jpetrell, line 237: Not caused by this PR, but it is not immediately obvious how does the second boolean parameter affect the formatting. Could either have an enumeration like formatAsHHMM(value, Units.ShowUnits) or a second method e.g. formatHHMMWithUnit(value). Also would be nice to have all formatting functions behind the Units API.

But both concerns are nice-to-haves so feel free to ignore. :smile:

DanielMcInnes (approved):
lgtm, one question about hectopascals.

pages/settings/devicelist/ac-in/PageAcInModelGenset.qml
blammit: Different settings have different precisions, but aside from using gui-v1 as a guide, it's not always clear when they can be generalized and applied across the board, and when they should be specific.

pages/settings/devicelist/ac-in/PageAcInModelGenset.qml
blammit: Oops, should be kPa. Fixed, thanks!

pages/settings/devicelist/ac-in/PageAcInModelGenset.qml
blammit: Thanks, removed

PR #834, opened 2024-03-05
fix building with OpenEmbedded / Debian rules

jhofstee:
Since commit 59d894f73aa "Load qml files from the file system, not from compiled resources" the install target installs the files in a hardcoded directory build/install instead of the normal DESTDIR / bindir etc. As a consequence OpenEmbedded / Venus will fail to build the gui-v2. This patch restores installing files into the normal locations if CMAKE_INSTALL_PREFIX is set. If not set build/install will be used.

For completeness, this does not adjust / install the translation files. Translations do seem to work though..


jhofstee:
Venus OS, see https://github.com/victronenergy/venus uses Open Embedded https://www.openembedded.org/wiki/Main_Page to create the images. https://github.com/victronenergy/meta-victronenergy/blob/master/meta-ve-software/recipes-ve/gui-v2.bb is the recipe for building gui-v2, which does include this patch to make it work.

chriadam (approved):
LGTM, thanks!

PR #844, opened 2024-03-08
Only show 'silence alarm' button when on Notifications page

blammit:


components/NavBar.qml
```diff
@@ -11,9 +11,11 @@ Rectangle {  // Use an opaque background so that page disappears behind nav bar
 
 	property alias model: buttonRepeater.model
 	readonly property int currentIndex: _currentIndex
+	readonly property string currentPage: _currentPage
```
jpetrell, line 14: Would expect a property called currentPage to reference the page object and not a string.

jpetrell:
This works, but now that the silence alarm is Notifications page specific, I kind of expect it to be also declared there, and parented to the status area. Dunno. Could also argue that it is still part of StatusBar layout. Up to you. Maybe at least consider moving SilenceAlarmButton to own component to improve code readability.

E.g. NotificationsPage.qml:

```
SwipeViewPage {
  id: root
  Binding {
     target: Global.statusArea
     property: "middleButton"
     value: silenceAlarmButton
  } 

  SilenceButton {
     id: silenceAlarmButton
     enabled: !!Global.notifications && Global.notifications.alert && root.isCurrentPage
  }
} 
```

components/NavBar.qml
blammit: Yes, maybe currentPageName instead? Have removed this part now.

blammit:
> This works, but now that the silence alarm is Notifications page specific, I kind of expect it to be also declared there, and parented to the status area. Dunno. Could also argue that it is still part of StatusBar layout. Up to you. Maybe at least consider moving SilenceAlarmButton to own component to improve code readability.

Agreed, that would be much cleaner. Updated to move the button to NotificationsPage.

The StatusBar left/right button organization needs some improvements. Maybe the left/right areas could be generic row containers that can then be filled by individual pages with their own buttons as necessary, instead of having enums that define the button types and their behaviours. Currently the API is quite restrictive, which makes it difficult to add things like a silence button. Anyway, something to do in a separate PR.

jpetrell (approved):

PR #854, opened 2024-03-13
Fix tank gauge display when /Remaining and /Capacity values are missing for a tank

blammit:


data/mock/config/LevelsPageConfig.qml
```diff
@@ -142,6 +142,20 @@ QtObject {
 				{ type: VenusOS.Tank_Type_RawWater, level: 95, capacity: .6 },
 			]
 		},
+		{
+			name: "Merge 2 tanks with average level of 50%, no capacity/remaining",
+			tanks: [
+				// Should get an average level of 50%
+				{ type: VenusOS.Tank_Type_Fuel, level: 25 },
+				{ type: VenusOS.Tank_Type_Fuel, level: 75 },
```
chriadam, line 150: there's also some code to calculate a level based on only remaining and capacity, wondering if it's worth adding an example for that case?  i.e. unkjnown "/Level" value.

components/ExpandedTanksView.qml
```diff
@@ -122,6 +123,7 @@ Rectangle {
 					fontSizeMode: Text.HorizontalFit
 					font.pixelSize: Theme.font_size_caption
 					color: Theme.color_font_secondary
+					visible: !isNaN(model.device.capacity) && !isNaN(model.device.remaining)
 					text: Units.getCapacityDisplayText(Global.systemSettings.volumeUnit,
 							model.device.capacity,
```
DanielMcInnes, line 128: ExpandedTanksView and TankGaugeGroup each have a percentageText and a valueText that are almost identical. Is it worth factoring this out? Also, TankGaugeGroup's valueText.text checks for isNaN, while ExpandedTanksView's valueText.text does not - does it need to?

DanielMcInnes (approved):
lgtm, nits

components/ExpandedTanksView.qml
jpetrell: The duplication goes away with https://github.com/victronenergy/gui-v2/pull/813 PR.

components/ExpandedTanksView.qml
```diff
@@ -107,7 +107,8 @@ Rectangle {
 					}
 					font.pixelSize: Theme.font_size_h1
 					unit: VenusOS.Units_Percentage
-					value: (isNaN(model.device.level) || model.device.level < 0) ? 0 : Math.round(model.device.level)
+					value: model.device.level
+					precision: 0
```
jpetrell, line 111: Percentages should already follow 0 precision.

components/TankGaugeGroup.qml
```diff
@@ -138,7 +138,8 @@ Rectangle {
 		width: Theme.geometry_levelsPage_gaugeDelegate_contentWidth
 		font.pixelSize: Theme.font_size_h1
 		unit: VenusOS.Units_Percentage
-		value: (isNaN(root.level) || root.level < 0) ? 0 : Math.round(root.level)
+		value: root.level
+		precision: 0
```
jpetrell, line 142: Same

data/common/Tank.qml
```diff
@@ -67,13 +67,19 @@ Device {
...
+			if (capacityValue > 0) {
 				levelValue = remainingValue / capacityValue * 100
 			}
 		}
+
+		// If there is no /Remaining, calculate from other values.
+		if (isNaN(remainingValue) && !isNaN(levelValue) && !isNaN(capacityValue)) {
+			remainingValue = capacityValue * (levelValue / 100)
+		}
+
```
jpetrell, line 82: Nice!

jpetrell:
I would have maybe expected the layouts between tanks to vertically align, though that will create empty space below the tanks which lack the data.

image

jpetrell:
Asked Serj, he suggested to show --/-- ℓ instead?

blammit:
> Asked Serj, he suggested to show --/-- ℓ instead?

Done, thanks for checking :)

components/TankGaugeGroup.qml
blammit: Fixed

components/ExpandedTanksView.qml
blammit: Fixed

data/mock/config/LevelsPageConfig.qml
blammit: Done

jpetrell (approved):

PR #870, opened 2024-03-18
Implement display off button and timer

### LINKED ISSUE #583: Screen Sleep function / feature
See https://victrondevelopment.slack.com/archives/C020QG3JVEW/p1701166638589059

Add a new button to the top-right of the brief page (actually the status bar) which puts the screen to sleep (not just black, but turn off backlight).  This allows people to go to sleep if their unit is in a caravan etc.

Needs support from venus-platform for hardware hooks (display on/off) if that doesn't exist already.

Once a screen is asleep, any touch input on the screen should wake it up again.

### LINKED ISSUE #859: Implement display off timer
Follow "Display off time" from Display settings, have an inactivity timer that gets reset on each key or touch event. Follow gui-v2 implementation.

jpetrell:
Fixes #859. Fixes #583.


src/screenblanker.h
```diff
@@ -0,0 +1,75 @@
...
+	Q_PROPERTY(bool allowed READ allowed WRITE setAllowed NOTIFY enabledChanged)
+	Q_PROPERTY(int displayOffTime READ displayOffTime WRITE setDisplayOffTime NOTIFY displayOffTimeChanged)
+
+public:
+	explicit ScreenBlanker(QQuickItem *parent = nullptr);
+
+	bool supported() const;
+
+	bool blanked() const;
+    void setBlanked(bool blanked);
```
chriadam, line 35: nitpick: wsp mismatch

src/screenblanker.h
```diff
@@ -0,0 +1,75 @@
...
+
+	bool allowed() const;
+	void setAllowed(bool allowed);
+
+	Q_INVOKABLE void setDisplayOn();
+	Q_INVOKABLE void setDisplayOff();
+
+	int displayOffTime() const;
+	void setDisplayOffTime(int time);
+signals:
```
chriadam, line 45: style nitpick: prefer Q_SIGNALS just for potential future proofing

tests/screenblanker/tst_screenblanker.qml
```diff
@@ -0,0 +1,87 @@
...
+	name: "ScreenBlanker"
+
+	ScreenBlanker {
+		id: blanker
+		displayOffTime: 100
+	}
+
+	SignalSpy {
+        id: spy
+        target: blanker
```
chriadam, line 19: nitpick: wsp in this block is a bit wrong

src/screenblanker.cpp
```diff
@@ -0,0 +1,212 @@
...
+		}
+		emit displayOffTimeChanged();
+	}
+}
+
+void ScreenBlanker::handleWindow(QQuickWindow *window)
+{
+	m_window = window;
+	if (m_window) {
+		m_window->installEventFilter(this);
```
chriadam, line 110: if an eventual child of the window (e.g. some button) receives and accepts some touch event, will the filter ever receive that event?  Or do we have to set e.g. setFiltersChildMouseEvents(true) on the window also?

jpetrell: Good question, had to check my assumptions. :smile:  So yes, all input events received by the items go through their parent window. QQuickWindow has no setFiltersChildMouseEvents() functionality, it is QQuickItem -specific thing.

src/screenblanker.h
jpetrell: Fixed.

src/screenblanker.h
jpetrell: Fixed.

tests/screenblanker/tst_screenblanker.qml
jpetrell: Fixed.

chriadam (approved):
LGTM but would like Bea to review also if possible

src/screenblanker.h
```diff
@@ -0,0 +1,75 @@
...
+namespace VenusOS {
+
+class ScreenBlanker : public QQuickItem
+{
+	Q_OBJECT
+	QML_ELEMENT
+
+	Q_PROPERTY(bool blanked READ blanked NOTIFY blankedChanged)
+	Q_PROPERTY(bool supported READ supported CONSTANT)
+	Q_PROPERTY(bool allowed READ allowed WRITE setAllowed NOTIFY enabledChanged)
```
blammit, line 26: Should the NOTIFY be allowedChanged?

I'm not sure about the `allowed` name but can't think of a better one if `enabled` can't be used. And `active` is ambiguous if it might indicate whether blanking is active.

src/screenblanker.h
```diff
@@ -0,0 +1,75 @@
...
+	bool writeToFile(QString filename, int value) const;
+	QString getFeature(QString const &name) const;
+
+private:
+	bool m_blanked = false;
+	bool m_allowed = false;
+
+	QQuickWindow *m_window;
+	QTimer m_blankingTimer;
+	QString mBlankDevice;
```
blammit, line 69: m_blankDevice to follow the other naming?

src/screenblanker.h
```diff
@@ -0,0 +1,75 @@
...
+	Q_PROPERTY(bool allowed READ allowed WRITE setAllowed NOTIFY enabledChanged)
+	Q_PROPERTY(int displayOffTime READ displayOffTime WRITE setDisplayOffTime NOTIFY displayOffTimeChanged)
+
+public:
+	explicit ScreenBlanker(QQuickItem *parent = nullptr);
+
+	bool supported() const;
+
+	bool blanked() const;
+	void setBlanked(bool blanked);
```
blammit, line 35: This function doesn't exist -- looks like `blankScreen()` is meant to be `setBlanked`?

src/screenblanker.cpp
```diff
@@ -0,0 +1,216 @@
...
+#include <QDir>
+#include <QFile>
+#include <QQmlInfo>
+#include <QRegularExpression>
+
+#include "screenblanker.h"
+
+using namespace Victron::VenusOS;
+
+ScreenBlanker::ScreenBlanker(QQuickItem *parent)
```
blammit, line 15: Does this need to extend QQuickItem? Could it just be a QObject and have a `window` property assigned to it to do the blanking?

pages/MainView.qml
```diff
@@ -34,6 +34,16 @@ Item {
 		}
 	}
 
+	property ScreenBlanker blanker: ScreenBlanker {
+		parent: root
+		allowed: !Global.splashScreenVisible && !(!!Global.notifications && Global.notifications.alarm)
+		displayOffTime: displayOffItem.isValid ? 1000*displayOffItem.value : 0.0
+		property VeQuickItem displayOffItem: VeQuickItem {
+			uid: !!Global.systemSettings ? Global.systemSettings.serviceUid + "/Settings/Gui/DisplayOff"
+										 : ""
```
blammit, line 43: Odd indentation, could just put this on previous line

pages/MainView.qml
```diff
@@ -34,6 +34,16 @@ Item {
 		}
 	}
 
+	property ScreenBlanker blanker: ScreenBlanker {
```
blammit, line 37: Not sure if `blanker` needs to be in `MainView`, wondering if it should be in `ApplicationContent` instead, and accessible with `Global.screenBlanker`.

tests/screenblanker/CMakeLists.txt
```diff
@@ -0,0 +1,44 @@
+#
+# Copyright (C) 2024 Victron Energy B.V.
+# See LICENSE.txt for license information.
+#
+
```
blammit, line 5: Is there a straightforward way to build the tests automatically as part of the project, e.g. by building with some flag? I'm concerned the tests will become stale if we don't build/run them regularly.

blammit (commented):
Looks great, some minor comments.

src/screenblanker.h
jpetrell: AFAICS enabled can be used (I remember even MouseArea overriding Item.enabled in C++ side). Renamed.

src/screenblanker.h
jpetrell: Fixed.

src/screenblanker.h
jpetrell: Yeah it is not used, good catch. Fixed.

pages/MainView.qml
jpetrell: On Qt Creator it shows ? and : horizontally aligned. But yeah GitHub doesn't appreciate tabs.

pages/MainView.qml
jpetrell: Good idea. Moved.

tests/screenblanker/CMakeLists.txt
jpetrell: Added dependency now to the root cmake project file. Tests are always built with rest of the project. It adds dozen seconds to the build times, but that subsequent builds shouldn't be too much affected. The tests are not installed unless -DVENUS_INSTALL_TESTS=ON flag is defined.

pages/MainView.qml
jpetrell: But maybe there is not much value seeing the empty string condition here. Fixed.

src/screenblanker.cpp
jpetrell: Done. I guess it is good to have external dependencies explicit. Also this was not really a visual item, just needed the handle to the window.

ApplicationContent.qml
```diff
@@ -38,6 +38,18 @@ Item {
 		Component.onCompleted: Global.firmwareUpdate = firmwareUpdate
 	}
 
+	onWindowChanged: function (window) { screenBlanker.window = window }
```
blammit, line 41: Any particular reason for not using `window: root.window` in `ScreenBlanker {}` instead?

src/screenblanker.cpp
```diff
@@ -0,0 +1,216 @@
...
+
+void ScreenBlanker::setDisplayOff()
+{
+	if (m_enabled) {
+		m_blankingTimer.stop();
+		blankScreen(true);
+	}
+}
+
+void ScreenBlanker::blankScreen(bool blanked)
```
blammit, line 155: Should this be called `setBlanked`? It sets the `m_blanked` value, and emits `blankChanged`.

blammit (approved):
Minor comments, otherwise LGTM.

ApplicationContent.qml
jpetrell: QQuickItem does not expose a property or a getter for QQuickItem::window() that would be accessible from QML.

image

From https://doc.qt.io/qt-6/qquickitem.html

src/screenblanker.cpp
jpetrell: Good point. Done.

PR #872, opened 2024-03-19
Fix to show inverter/charger details when clicking on overview box

blammit:


jpetrell (approved):

jpetrell:
Yikes how did that end up there. :sweat_smile:  Thanks!

PR #876, opened 2024-03-19
Remove the "Alarm relay on" relay setting

### LINKED ISSUE #855: Relay: Remove manual toggle when functionality is set to Alarm relay
A recent change in gui v1 removed the manual toggle when the relay 1 is set to "Alarm relay". It has to be removed in the gui-v2 as well.
This only applies to relay 1 as relay 2 function can't be set to "Alarm relay"

Please note that the relay referred as "1" in the gui is actually enumerated as relay "0" in DBus.

Check these screenshots to notice which item has been removed. The highlighted item in the first screenshot is the one to be hidden when alarm relay is selected.

Screenshot 2024-03-13 at 09 43 36
Screenshot 2024-03-13 at 10 38 04

blammit:
This corresponds to change 260305cd92cb405374fc093a770e62c769773d8a in gui-v1. The toggle is no longer required as venus-platform now manages the alarm state dependingo nthe relay configuration.

Fixes #855

### DIFF
--- a/pages/settings/PageSettingsRelay.qml
+++ b/pages/settings/PageSettingsRelay.qml
@@ -52,23 +52,6 @@ Page {
 			}
 
 			ListSwitch {
-				id: relaySwitch
-
-				//% "Alarm relay on"
-				text: qsTrId("settings_relay_alarm_relay_on")
-				updateOnClick: false
-				checked: root._relay0Object && _relay0Object.state === VenusOS.Relays_State_Active
-
-				visible: relayFunction.currentValue === VenusOS.Relay_Function_Alarm
-				onClicked: {
-					// TODO in gui-v1 the relay state change considers relay polarity and alarm status.
-					// In gui-v2 we will connect to venus-platform or some backend to do this.
-					const newState = checked ? VenusOS.Relays_State_Inactive : VenusOS.Relays_State_Active
-					root._relay0Object.setState(newState)
-				}
-			}
-
-			ListSwitch {
 				id: manualSwitch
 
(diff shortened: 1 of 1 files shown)


pages/settings/PageSettingsRelay.qml
```diff
@@ -51,23 +51,6 @@ Page {
 				]
 			}
 
-			ListSwitch {
```
chriadam, line 54: Maybe I am misunderstanding the task, but shouldn't this switch be hidden ONLY if the Relay1 is set to be the Alarm relay, but otherwise should be visible?  So isn't it just changing the "visible" binding?

Or does this switch only apply to the relay in the case where it is set to be the Alarm relay anyway (in which case, I agree, we can just remove it altogether)?

blammit: Yes, the latter. The removed code has this binding:`visible: relayFunction.currentValue === VenusOS.Relay_Function_Alarm`. So it was only visible when the relay function is 'alarm' anyway, which means the switch can just be removed.

This is a port of 260305cd92cb405374fc093a770e62c769773d8a from gui-v1:
https://github.com/victronenergy/gui/commit/260305cd92cb405374fc093a770e62c769773d8a

chriadam: Ok, LGTM

chriadam (approved):

jpetrell (approved):

jpetrell:
Tested, works as advertised. :+1:

PR #877, opened 2024-03-19
Use visual indications on Brief page for negative AC input values

blammit:


jpetrell (approved):

blammit:
Yes, it was nice to have that alignment. Also when we implement the 3-phase design, will have to check this doesn't push the quantity label into the centre gauges.

PR #884, opened 2024-03-19
Hide accepted generator dialogs quickly

### LINKED ISSUE #837: Generator Manual start/stop dialog stuck on the screen after accepting
Go to the Control cards page using the top-left icon button on the status bar. Tap "Manual Start" or "Manual Stop" button on the Generator card. Accept the confirmation dialog.

Actual result: The dialog performs the action, but remains on the display for half-dozen seconds before fading out.

Expected results:  The dialog performs the action and fades out right away.

Doesn't reproduce on desktop with mock backend.

https://github.com/victronenergy/gui-v2/assets/2203667/d270c526-49fa-4ff4-a02a-e7c2d7be91ab

jpetrell:
- The stopped generator goes quickly from the running to the cool-down state, but takes dozen seconds to go to the stopping and finally to the stopped states
- Similarly the started generator goes quickly from the stopped to the warmup state, but 20+ seconds to reach the running state
- Instead of waiting for the end state, hide the dialog already during cool-down and warmup
- Allow quickly toggling the generator on/off like with the switch used in gui v1 and gui v2 settings

Fixes #837.

### DIFF
--- a/components/dialogs/GeneratorStopDialog.qml
+++ b/components/dialogs/GeneratorStopDialog.qml
@@ -11,7 +11,17 @@ GeneratorDialog {
 
 	//% "Stop Now"
 	acceptText: qsTrId("controlcard_generator_stopdialog_stop_now")
-	finalGeneratorState: VenusOS.Generators_State_Stopped
+
+	onGeneratorStateChanged: {
+		if (root.open) {
+			if (generatorState == VenusOS.Generators_State_Stopped
+					|| generatorState == VenusOS.Generators_State_Stopping
+					|| generatorState == VenusOS.Generators_State_CoolDown) {
+				root.accept()
+			}
+		}
+	}
+
 	runGeneratorAction: function() {
 		root.generator.stop()
 	}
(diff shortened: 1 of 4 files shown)


components/dialogs/GeneratorStopDialog.qml
```diff
@@ -11,7 +11,17 @@ GeneratorDialog {
 
 	//% "Stop Now"
 	acceptText: qsTrId("controlcard_generator_stopdialog_stop_now")
-	finalGeneratorState: VenusOS.Generators_State_Stopped
+
+	onGeneratorStateChanged: {
+		if (root.open) {
+			if (generatorState == VenusOS.Generators_State_Stopped ||
```
blammit, line 17: nitpick: put `||` on the start of the next line instead

blammit (commented):
Right, I guess the bug is not reproducible on systems with only started/stopped states. I still have to test this on device, will check this tomorrow.

Also could you fix the line wrapping in the commit message?

components/dialogs/GeneratorStopDialog.qml
jpetrell: Fixed.

jpetrell:
> Also could you fix the line wrapping in the commit message?

Fixed.

blammit (approved):
LGTM, though tested with system that only used started/stopped states.

Issue #728, opened 2024-02-08
Opening Wi-Fi access point settings randomly crashes on gui-v2


jpetrell:
Go to Settings -> Wi-Fi -> Open access point. Now go back and open it again. After 3-12 tries the gui-v2 crashes and the terminal shows error

```
free(): invalid pointer
Aborted
```

Tested on Ekrano GX, firware v.3.20~46, application version v0.2.9.

DanielMcInnes:
@jpetrell , do you mean: Settings -> WiFi -> toggle the 'Create access point' switch?

jpetrell:
> do you mean: Settings -> WiFi -> toggle the 'Create access point' switch?

No sorry. I meant repeatedly opening the Wi-Fi networks page. Updated the description now to be more accurate.

jpetrell:
Ran venus-gui-v2 with gdbserver. The stack trace didn't give much, other than the gui process throws SIGABRT due to critical error (assert, unhandled throw).

```
Thread 1 "venus-gui-v2" received signal SIGABRT, Aborted.
0x441417e8 in raise () from target:/lib/libc.so.6
(gdb) bt
#0  0x441417e8 in raise () from target:/lib/libc.so.6
#1  0x4412bc84 in abort () from target:/lib/libc.so.6
#2  0x441796bc in ?? () from target:/lib/libc.so.6
```

The only asserts on Victron side are inside VeQItem, but they are not called. Though "free(): invalid pointer" seems to suggest something somewhere is trying to delete an object that has already been freed from the memory. Will keep looking.

Bumbed into another page stack issue #832 while trying to reproduce this one.

jhofstee:
For what it is worth, this (likely) happened to me as well:

```
[  979.930326] potentially unexpected fatal signal 6.
[  979.935201] CPU: 3 PID: 1439 Comm: venus-gui-v2 Tainted: G           O      5.10.109-venus-15 #1
[  979.944005] Hardware name: Allwinner sun8i Family
[  979.948735] PC is at 0x454217e8
[  979.951925] LR is at 0xbec6e698
[  979.955069] pc : [<454217e8>]    lr : [<bec6e698>]    psr: 00070010
[  979.961346] sp : bec6e698  ip : 00000000  fp : bec6e920
[  979.966600] r10: 45515ef0  r9 : 4552e2bc  r8 : 00000001
[...]

PR #885, opened 2024-03-19
Avoid crash when QML tries to delete global CmAgent

jpetrell:
PageSettingsTcpIp.qml fetches CmAgent from singleton:

```
property CmAgent _agent
Component.onCompleted: ... _agent = Connman.registerAgent(_agentPath);
```

CmAgent is created in C++ side, but has not been assigned a parent, making QML think it has the ownership of the object (["ownership IS transferred to QML if and only if the QObject has no parent"](https://wiki.qt.io/Shared_Pointers_and_QML_Ownership)). When the page is closed and QML tries to delete CmAgent object free() throws critical error and the process is closed with SIGABRT.

Fixes https://github.com/victronenergy/gui-v2/issues/728.


jpetrell:
> ERROR: Write access to repository not granted.

Also needs to be fixed to gui v1, but I don't have access rights to make a PR.

chriadam (approved):
Sure, LGTM.  Eventually, I assume we will be removing all connman code from the repo anyway, and all network settings will be specified via dbus/mqtt topics, and venus-platform will be the only thing which does connman stuff?

PR #886, opened 2024-03-19
Avoid relayouting list items on view visibility change

### LINKED ISSUE #851: Navigating from the Main settings page stutters
Navigate from the Settings page either using a swipe or by tapping the other nav bar buttons.

Expected result: All the main view pages are constructed in advance so  the transition is smooth.

Actual result: There is ~200ms momentary freeze, a hiccup when leaving the settings page, because 13 list items are constructed unnecessarily.

Measured on Cerbo GX.

jpetrell:
Fixes #851.

### DIFF
--- a/components/listitems/ListItem.qml
+++ b/components/listitems/ListItem.qml
@@ -23,22 +23,22 @@ Item {
 	readonly property bool userHasWriteAccess: Global.systemSettings.canAccess(writeAccessLevel)
 	readonly property bool userHasReadAccess: Global.systemSettings.canAccess(showAccessLevel)
 
-	readonly property bool defaultVisible: userHasReadAccess
+	readonly property bool defaultAllowed: userHasReadAccess
 	readonly property alias primaryLabel: primaryLabel
-	readonly property int defaultImplicitHeight: visible
-		? Math.max(Math.max(primaryLabel.implicitHeight + Theme.geometry_listItem_content_verticalMargin*2, content.height)
-					+ (bottomContent.height > 0
-							? bottomContent.height + bottomContentMargin
-							: 0),
-				   Theme.geometry_listItem_height)
-		: 0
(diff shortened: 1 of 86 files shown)


blammit (commented):
Looks great! But could you add a brief explanation in the commit of why this change is needed and what it provides?

components/listitems/ListItem.qml
```diff
@@ -23,22 +23,22 @@ Item {
...
+	readonly property int defaultImplicitHeight: {
+		const bottomHeight = bottomContent.height > 0 ? bottomContent.height + bottomContentMargin : 0
+		const labelHeight = primaryLabel.implicitHeight + Theme.geometry_listItem_content_verticalMargin*2
+		return Math.max(Theme.geometry_listItem_height,
+						Math.max(content.height, labelHeight) + bottomHeight)
+	}
 
 	readonly property int availableWidth: width - primaryLabel.anchors.leftMargin - content.anchors.rightMargin - content.spacing
 	property int maximumContentWidth: availableWidth * 0.7
+	property bool allowed: defaultAllowed
```
blammit, line 37: gui-v1 uses `show` for something equivalent, instead of `allowed`. I wonder if using `show` would make porting a little easier / more clear. What do you think?

chriadam: the problem with `show` is that it's a verb, so it feels like it should be an active method rather than a passive property.  But that's a nitpick, I don't think it should stop the change if you think it's much better than `allowed`.

blammit: It's true, I don't really like that about `show`. I suppose `canShow` would make more sense, but doesn't look nice. Well, let's keep it as `allowed`.

blammit (approved):

jpetrell:
> Looks great! But could you add a brief explanation in the commit of why this change is needed and what it provides?

Done.

PR #888, opened 2024-03-20
Avoid crash when returning from battery 'Settings' page

### LINKED ISSUE #887: Leaving Battery settings sub-page crashes
Seeing on desktop with mock backend, not sure if an issue on the real devices yet.

Go to Overview -> Battery -> Settings. Go back to Battery main page.

Actual result: The gui v2 process segmentation faults. The issue seems to be caused by ObjectModelMonitor, at least the crash goes away after commenting them out.

```
#1  0x00007ffff6b673d2 in QQmlInstantiatorPrivate::clear (this=0x3f76a00)
    at /home/qt/work/qt/qtdeclarative/src/qmlmodels/qqmlinstantiator.cpp:44
#2  0x00007ffff6b674d5 in QQmlInstantiator::~QQmlInstantiator (
    this=this@entry=0x3f76920, __in_chrg=<optimized out>)
    at /home/qt/work/qt/qtdeclarative/src/qmlmodels/qqmlinstantiator.cpp:211
#3  0x00007ffff6b5ea23 in QQmlPrivate::QQmlElement<QQmlInstantiator>::~QQmlElement (this=0x3f76920, __in_chrg=<optimized out>)
    at /home/qt/work/qt/qtdeclarative/src/qml/qml/qqmlprivate.h:99
#4  QQmlPrivate::QQmlElement<QQmlInstantiator>::~QQmlElement (
    this=0x3f76920, __in_chrg=<optimized out>)
    at /home/qt/work/qt/qtdeclarative/src/qml/qml/qqmlprivate.h:99
#5  0x00007ffff5bb31d2 in QObjectPrivate::deleteChildren (
    this=this@entry=0x3f6d130)
    at /home/qt/work/qt/qtbase/src/corelib/kernel/qobject.cpp:2206
#6  0x00007ffff5bb7b83 in QObject::~QObject (this=this@entry=0x3f6d050, 
    __in_chrg=<optimized out>)

```

blammit:
Declare the ObjectModelMonitor before declaring the model that it references. Otherwise, QQmlInstantiator crashes.

Fixes #887.

### DIFF
--- a/pages/settings/devicelist/battery/PageBatterySettings.qml
+++ b/pages/settings/devicelist/battery/PageBatterySettings.qml
@@ -11,6 +11,16 @@ Page {
 
 	property string bindPrefix
 
+	// Declare ObjectModelMonitor before the model that it is monitoring. See QTBUG-123496
+	ObjectModelMonitor {
+		id: alarmSettingsMonitor
+		model: batterySettingsAlarmModel
+	}
+	ObjectModelMonitor {
+		id: relaySettingsMonitor
+		model: batterySettingsRelayModel
+	}
+
 	BatterySettingsAlarmModel {
 		id: batterySettingsAlarmModel
 		bindPrefix: root.bindPrefix
@@ -39,11 +49,6 @@ Page {
 					Global.pageManager.pushPage(emptySettingsComponent,
 							{ "title": text, "model": batterySettingsAlarmModel })
 				}
-
-				ObjectModelMonitor {
-					id: alarmSettingsMonitor
-					model: batterySettingsAlarmModel
-				}
 			}
 
 			ListNavigationItem {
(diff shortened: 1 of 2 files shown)


blammit:
This seems like a fragile fix, but it is useful to keep the Instantiator. If we use a Repeater then the items are visible when created.

chriadam:
In the bug report you say that putting the monitor first crashes, and that it needs to be declared after the instantiator, but in this PR you seem to be putting the monitor first / above the object models?

blammit:
> In the bug report you say that putting the monitor first crashes, and that it needs to be declared after the instantiator, but in this PR you seem to be putting the monitor first / above the object models?

In the bug report, putting the ObjectModel first causes the crash. In this PR the ObjectModelMonitor is the Instantiator, not the ObjectModel.

chriadam:
Ah, I see!  Thanks.  LGTM.

chriadam (approved):

jpetrell (approved):
Fixes the issue (not exactly sure how, but yeah Qt bug).

PR #889, opened 2024-03-20
Hide display brightness settings on wasm

blammit:
The brightness settings do not affect the browser display, only the device display, so it is misleading to show these settings in the wasm build.

Contributes to #375

### DIFF
--- a/pages/settings/PageSettingsDisplay.qml
+++ b/pages/settings/PageSettingsDisplay.qml
@@ -19,6 +19,7 @@ Page {
 				//% "Adaptive brightness"
 				text: qsTrId("settings_adaptive_brightness")
 				dataItem.uid: Global.systemSettings.serviceUid + "/Settings/Gui/AutoBrightness"
+				allowed: Qt.platform.os != "wasm"
 			}
 
 			ListSlider {
@@ -26,7 +27,7 @@ Page {
 				text: qsTrId("settings_brightness")
 				dataItem.uid: Global.systemSettings.serviceUid + "/Settings/Gui/Brightness"
 				writeAccessLevel: VenusOS.User_AccessType_User
-				allowed: defaultAllowed && !autoBrightness.checked
+				allowed: defaultAllowed && !autoBrightness.checked && Qt.platform.os != "wasm"
 			}
 
 			ListRadioButtonGroup {


chriadam (approved):

jpetrell (approved):
Feels odd to still have dataItem communication going underneath in the invisible list item, but I guess the overhead is negligible.

Wonder if display off time should be hidden also. For some of these device-specific settings it may be nice to be able to configure them also remotely (though quite low prio use case overall), and if we choose to still display those would be good to make it clear when the setting only affects the local device HW, and not the browser environment.

pages/settings/PageSettingsDisplay.qml
```diff
@@ -19,14 +19,15 @@ Page {
 				//% "Adaptive brightness"
 				text: qsTrId("settings_adaptive_brightness")
 				dataItem.uid: Global.systemSettings.serviceUid + "/Settings/Gui/AutoBrightness"
+				visible: Qt.platform.os != "wasm"
```
jpetrell, line 22: -> allowed

pages/settings/PageSettingsDisplay.qml
```diff
@@ -19,14 +19,15 @@ Page {
...
 			}
 
 			ListSlider {
 				//% "Brightness"
 				text: qsTrId("settings_brightness")
 				dataItem.uid: Global.systemSettings.serviceUid + "/Settings/Gui/Brightness"
 				writeAccessLevel: VenusOS.User_AccessType_User
-				visible: defaultVisible && !autoBrightness.checked
+				visible: defaultVisible && !autoBrightness.checked && Qt.platform.os != "wasm"
```
jpetrell, line 30: -> allowed: defaultAllowed

blammit:
> Feels odd to still have dataItem communication going underneath in the invisible list item, but I guess the overhead is negligible.
> 
> Wonder if display off time should be hidden also. For some of these device-specific settings it may be nice to be able to configure them also remotely (though quite low prio use case overall), and if we choose to still display those would be good to make it clear when the setting only affects the local device HW, and not the browser environment.

Yeah, I wonder if they should still be displayed, but on wasm have some caption text like "This only applies on the device" or such. Anyway, I'll do this for now and see if we want to change it later.

PR #891, opened 2024-03-20
Avoid calculating tank level on Brief Page gauges if possible

### LINKED ISSUE #680: Brief tank levels for Black and Waste are inverted
The same data is shown correctly on the Tanks view.

Reported here: https://community.victronenergy.com/questions/249509/venus-os-v32034-available-for-testing.html?page=2&pageSize=50&sort=oldest

image

chriadam:
Affects WasteWater and BlackWater tanks.

Fixes #680

### DIFF
--- a/components/GaugeModel.qml
+++ b/components/GaugeModel.qml
@@ -84,11 +84,10 @@ ListModel {
 			readonly property string tankName: _tankProperties.name
 			readonly property string tankIcon: isBattery ? Global.batteries.system.icon : _tankProperties.icon
 			readonly property var tankModel: isBattery ? null : Global.tanks.tankModel(tankType)
-			readonly property real tankLevel: isBattery
-					? Math.round(Global.batteries.system.stateOfCharge || 0)
-					: (tankModel.count === 0 || tankModel.totalCapacity === 0
-					   ? 0
-					   : (tankModel.totalRemaining / tankModel.totalCapacity) * 100)
+			readonly property real tankLevel: isBattery ? Math.round(Global.batteries.system.stateOfCharge || 0)
+					: !isNaN(tankModel.averageLevel) ? tankModel.averageLevel
+					: (tankModel.count === 0 || tankModel.totalCapacity === 0) ? 0
+					: ((Math.min(tankModel.totalRemaining / tankModel.totalCapacity, 1.0) * 100))
 
 			readonly property var _tankProperties: Gauges.tankProperties(tankType)
 


components/GaugeModel.qml
```diff
@@ -81,14 +81,17 @@ ListModel {
...
 			readonly property string tankName: _tankProperties.name
 			readonly property string tankIcon: isBattery ? Global.batteries.system.icon : _tankProperties.icon
 			readonly property var tankModel: isBattery ? null : Global.tanks.tankModel(tankType)
 			readonly property real tankLevel: isBattery
 					? Math.round(Global.batteries.system.stateOfCharge || 0)
-					: (tankModel.count === 0 || tankModel.totalCapacity === 0
+					: ((tankModel.count === 0 || tankModel.totalCapacity === 0)
 					   ? 0
-					   : (tankModel.totalRemaining / tankModel.totalCapacity) * 100)
+					   : Math.min(tankModel.totalRemaining / tankModel.totalCapacity, 1.0) * 100)
```
chriadam, line 92: added Math.min() because I was seeing something strange in this particular case: the `remaining` value was larger than the `capacity` value somehow, resulting in a spurious level value.  Specifically, it was `remaining = 0.15142, capacity = 0.1514164714`.

jpetrell: We are effectively hiding the root issue, but I guess we cannot control all the tank configurations in the world and in the actual case the difference is close to a rounding error.

components/GaugeModel.qml
```diff
@@ -53,7 +53,7 @@ ListModel {
 	function updateGauge(gauge) {
 		const gaugeIndex = findGauge(gauge)
 		if (gaugeIndex >= 0) {
-			set(gaugeIndex, { name: gauge.tankName, icon: gauge.tankIcon, value: gauge.tankLevel })
+			set(gaugeIndex, { name: gauge.tankName, icon: gauge.tankIcon, value: gauge.tankValue })
```
jpetrell, line 56: Nice that tank value is now set to value (vs. previous tank level becoming a value). :blush:

components/GaugeModel.qml
```diff
@@ -81,14 +81,17 @@ ListModel {
...
 			readonly property var tankModel: isBattery ? null : Global.tanks.tankModel(tankType)
 			readonly property real tankLevel: isBattery
 					? Math.round(Global.batteries.system.stateOfCharge || 0)
-					: (tankModel.count === 0 || tankModel.totalCapacity === 0
+					: ((tankModel.count === 0 || tankModel.totalCapacity === 0)
 					   ? 0
-					   : (tankModel.totalRemaining / tankModel.totalCapacity) * 100)
+					   : Math.min(tankModel.totalRemaining / tankModel.totalCapacity, 1.0) * 100)
+			readonly property real tankValue: tankValueType == VenusOS.Gauges_ValueType_RisingPercentage
+					? (100 - tankLevel) : tankLevel
```
jpetrell, line 94: I would consider writing the logic with one JavaScript function using traditional if..else style, multiple conditionals (battery, valid values, rising percentage case, capacity going over remaining value) written as tertiary operations and two properties  (looks like tankLevel is not used for anything else than for determining tankValue) becomes a bit difficult read, though I guess it is a matter of taste.

jpetrell (approved):
Some comments. Overall LGTM.

jpetrell:
Thinking this a bit more it is a bit unfortunate that all the clients need to handle such hardware-specific variations, to me seems the middleware is reporting the value of **totalRemaining** incorrectly?

blammit (approved):
Nice! Could you add more detail to the commit message?

blammit:
> Thinking this a bit more it is a bit unfortunate that all the clients need to handle such hardware-specific variations, to me seems the middleware is reporting the value of **totalRemaining** incorrectly?

That's true. @ReinvdZee is this something you could assist with? This bug in gui-v2 was triggered because a tank's `/Remaining` value was  greater than its `/Capacity`. It would be nice if this could be handled at a lower level so that the UI layer doesn't have to check for this.

chriadam:
Well, to be clear, the bug itself was just that we didn't invert the level when the tank type was RisingPercentage (as opposed to the usual FallingPercentage).  But yes, the fact that the level could exceed 100% without my clamp is worrying.

blammit:
Now I'm confused. With the PR, the levels are inverted in an unexpected manner, showing 25% on the Levels page and 75% on the Brief page. So I thought the bug was related to the capacity/remaining not matching up. Would you mind putting those in separate commits so it's clear they are not related?

chriadam:
After discussion with Bea, it became apparent that instead of inverting the value (which we should NOT do, instead we should just invert the COLOUR), we need to fix the algorithm so that it doesn't assume that level = remaining/capacity.  Actually, level = capacity - remaining.  I will fix here, and check if we made similar mistake elsewhere.

components/GaugeModel.qml
```diff
@@ -84,11 +84,10 @@ ListModel {
...
 			readonly property string tankIcon: isBattery ? Global.batteries.system.icon : _tankProperties.icon
 			readonly property var tankModel: isBattery ? null : Global.tanks.tankModel(tankType)
-			readonly property real tankLevel: isBattery
-					? Math.round(Global.batteries.system.stateOfCharge || 0)
-					: (tankModel.count === 0 || tankModel.totalCapacity === 0
-					   ? 0
-					   : (tankModel.totalRemaining / tankModel.totalCapacity) * 100)
+			readonly property real tankLevel: isBattery ? Math.round(Global.batteries.system.stateOfCharge || 0)
+					: (tankModel.count === 0 || tankModel.totalCapacity === 0) ? 0
+					: !isNaN(tankModel.averageLevel) ? tankModel.averageLevel
```
blammit, line 89: I think the level check should come first, before the capacity -- that way if a level is available, but not the capacity, then we can still show a valid level.

chriadam: Good point, done

blammit (approved):
LGTM, thanks! Verified the gauges now look correct on the Brief page of the remote installation.

jpetrell (approved):

PR #894, opened 2024-03-21
Increase status bar button dimensions

### LINKED ISSUE #808: The back button feels bit small on Ekrano GX
Returning back from the sub-pages is one of the most common navigation actions. Currently the back button feels a tad too small on Ekrano GX. You are able to hit the button, but targeting such a small item causes unnecessary cognitive stress to the user. In comparison the navigation bar icon buttons and settings page list items have comfortable touch areas.

image

jpetrell:
- Increase status bar height to make the corner buttons easier to tap
- Reduce and unify vertical page margins
- Introduce a new back button icon with slightly longer tails

Fixes #808.


blammit (approved):

blammit:
I'm getting "Warning: theme constants need to be updated" from OverviewPage, not sure if it came from this PR or not.

chriadam:
I agree, it looks like the widget_compact_xl_height is now wrong (too large, by a few pixels) now that the status bar height has increased a bit.

blammit:
LGTM

PR #895, opened 2024-03-22
Ensure AC input phase data is displayed

blammit:


jpetrell (approved):
Works nicely. Weird that parenting did not work.

PR #896, opened 2024-03-25
Show grid and generator simultaneously

blammit:


components/widgets/WidgetHeader.qml
```diff
@@ -28,13 +28,13 @@ Item {
 
 	Label {
 		id: titleLabel
-
-		anchors {
-			left: icon.right
-			leftMargin: Theme.geometry_widgetHeader_spacing
-			right: parent.right
-		}
+		width: parent.width - icon.width - secondaryLabel.implicitWidth - Theme.geometry_widgetHeader_spacing
```
jpetrell, line 31: There are potentially two spacings to deduct with icon and two labels in a row?

pages/BriefMonitorPanel.qml
```diff
@@ -23,6 +23,7 @@ Column {
 		WidgetHeader {
 			id: solarHeader
 
+			width: parent.width
```
jpetrell, line 26: Are there use cases of using implicit width? If not move to the component?

pages/OverviewPage.qml
```diff
@@ -76,6 +79,27 @@ SwipeViewPage {
...
 				break
 			case 3:
+				if (preferLargeWidgetCount === 3) {
+					// If all three prefer L size, then use M for all three.
+					widget.size = VenusOS.OverviewWidget_Size_M
+				} else if (preferLargeWidgetCount >= 2) {
+					// If only two prefer L size, then use L for the bottom one, M for the other that
+					// prefers L size, and S for the last.
+					if (i === _leftWidgets.length - 1) {
+						widget.size = VenusOS.OverviewWidget_Size_L
```
jpetrell, line 89: What if the bottom one doesn't prefer the large side?

data/AcInputs.qml
```diff
@@ -112,5 +126,13 @@ QtObject {
 		return match ? match.name : "--"
 	}
 
+	function _updateActiveInfo(info) {
+		if (info.isActiveInput) {
+			root.activeInputInfo = info
+		} else if (root.activeInputInfo == info) {
+			root.activeInputInfo = null
+		}
```
jpetrell, line 134: Would maybe move the two signal handlers and the update function to a readonly binding:

```
readonly property AcInputSystemInfo activeInputInfo: {
    if (input1info.isActiveInput) return input1info
    else if (input2info.isActiveInput) return input2info
    else return null
}
```

data/common/AcInputSystemInfo.qml
```diff
@@ -6,28 +6,37 @@
...
+		uid: root.bindPrefix ? root.bindPrefix + "/ServiceName" : ""
 	}
 
 	readonly property VeQuickItem _serviceType: VeQuickItem {
-		uid: root.bindPrefix + "/ServiceType"
+		uid: root.bindPrefix ? root.bindPrefix + "/ServiceType" : ""
+	}
+
+	readonly property VeQuickItem _source: VeQuickItem {
+		uid: root.bindPrefix ? root.bindPrefix + "/Source" : ""
```
jpetrell, line 40: These "root.bindPrefix ? .." checks are unfortunate, but I guess no way around them.

data/common/ActiveAcInput.qml
```diff
@@ -9,40 +9,20 @@ import Victron.VenusOS
 Device {
 	id: root
 
-	readonly property bool connected: !!_activeInputInfo && _activeInputInfo.connected
-	readonly property string serviceType: !!_activeInputInfo ? _activeInputInfo.serviceType : ""
-	readonly property string serviceName: !!_activeInputInfo ? _activeInputInfo.serviceName : ""
+	property AcInputSystemInfo inputInfo
 
-	property int source: VenusOS.AcInputs_InputSource_NotAvailable
+	readonly property bool connected: !!inputInfo && inputInfo.connected
```
jpetrell, line 14: is "!!" prefix really needed?

data/mock/AcInputsImpl.qml
```diff
@@ -11,128 +11,161 @@ QtObject {
...
+			// Cycle between positive -> negative -> zero energy.
+			// Positive energy value = imported energy, flowing towards inverter/charger.
+			// Negative energy value = exported energy, flowing towards grid.
+			const negativeEnergyFlow = testEnergyCounter < 0
+			const zeroEnergyFlow = testEnergyCounter === 0
+
+			const phases = Global.acInputs.activeInput._phases._phaseObjects
+			let totalPower = NaN
+			let i
+			for (i = 0; i < phases.count; ++i) {
```
jpetrell, line 140: Why not for (let i = 0; ..

jpetrell (approved):
LGTM Added some small comments here and there. In testing could not find issues, works beautifully. :blush:

jpetrell:
L1, L2 and L3 labels are shown as "L1:", "L2:" and "L3:" in the design, and without ":" here?

jpetrell:
Why is the generator be stopped on control cards page but disconnected on the Overview page?

blammit:
> L1, L2 and L3 labels are shown as "L1:", "L2:" and "L3:" in the design, and without ":" here?

Could you check with Serj about whether the colons are necessary? They weren't there previously and we don't use this elsewhere in the UI. Generally the color distinction is enough.

blammit:
> Why is the generator be stopped on control cards page but disconnected on the Overview page?

Looks like it should be 'Stopped' for generators instead. Thanks, will update.

jpetrell:
> Could you check with Serj about whether the colons are necessary? They weren't there previously and we don't use this elsewhere in the UI. Generally the color distinction is enough.

Asked Jesus, please add colons ":" to the L1-3 phase labels.

> This layout doesn't look optimal to me, though it is pretty close to the design.

Works as specified. But maybe check that "(A)" label has correct margin on the right side.

pages/BriefMonitorPanel.qml
blammit: Done

components/widgets/WidgetHeader.qml
blammit: Oh yes, thanks

data/AcInputs.qml
blammit: Ah yes, the function is not needed. Thanks, fixed

data/common/ActiveAcInput.qml
blammit: Not here, removed

data/mock/AcInputsImpl.qml
blammit: Done

blammit:
>  Asked Jesus, please add colons ":" to the L1-3 phase labels.

Done

> Works as specified. But maybe check that "(A)" label has correct margin on the right side.

It is the same as the left margin, but might look a bit off due to the padding within the icon image.

PR #898, opened 2024-03-26
Align quantity group columns horizontally

jpetrell:
- Calms down the quantity&text group layouts by forcing labels to even-sized columns
- Calms down the stutter that happens when individual values get updated
- Don't apply everywhere, e.g. the status text columns are a lot wider than the quantity labels so following the same layout columns don't make sense

I thought this would be quick thing to do between bigger tasks (with quantity table alignments still fresh on my memory), but this became a bit complex due to different ways the quantity and text groups are used. Also, this may end up being a stopgap solution as the upcoming settings redesign likely introduces various design changes to these settings list items.

Requires more testing and some performance profiling before merging. The quantity repeater is constantly recreating all the quantity labels when the value update, which also causes a lot of unnecessary relayout calculations.

image


pages/settings/devicelist/battery/BatteryDetails.qml
```diff
@@ -53,14 +51,6 @@ QtObject {
...
 			uid: root.bindPrefix + "/System/NrOfModulesBlockingDischarge"
 		},
-		VeQuickItem {
-			id: nrOfModulesOnline
-			uid: root.bindPrefix + "/System/NrOfModulesOnline"
-		},
-		VeQuickItem {
-			id: nrOfModulesOffline
-			uid: root.bindPrefix + "/System/NrOfModulesOffline"
-		},
```
jpetrell, line 63: Unused and duplicated (see modulesOnline and modulesOffline VeQuickItems above).

jpetrell:
Discussed with Bea and Serg that while the current layout is suboptimal, maybe not worth it to integrate the improvement to 1.0 now, but visit this during the settings redesign work.

blammit:
This is no longer needed as the feature has been implemented in `main`.

PR #903, opened 2024-03-27
Improve support for editing QML source giles on the cerbo

DanielMcInnes:
I think this now works for all QML source files, except for Main.qml - this is loaded from .cpp, and needs to use the resource system. We could move most stuff out of Main.qml and into a new file if we think it's necessary.

### DIFF
--- a/ApplicationContent.qml
+++ b/ApplicationContent.qml
@@ -88,18 +88,15 @@ Item {
 	// Qt for WebAssembly due to QTBUG-104109).
 	// Note the VKB layer is the top-most layer, to allow the idleModeMouseArea beneath to call
 	// testCloseOnClick() when clicking outside of the focused text field, to auto-close the VKB.
-	Component.onCompleted: {
-		if (Qt.platform.os !== "wasm") {
-			_inputComponent = Qt.createComponent(Global.appUrl("/components/InputPanel.qml"), Component.Asynchronous)
-			_inputComponent.statusChanged.connect(function() {
-				if (_inputComponent.status === Component.Ready) {
-					Global.inputPanel = _inputComponent.createObject(root, { mainViewItem: mainView })
-				} else if (_inputComponent.status === Component.Error) {
-					console.warn("Cannot load InputPanel:", _inputComponent.errorString())
-				}
-			})
-		} else {
-			console.log("Not creating InputPanel, this is a wasm build")
+
+	Loader {
+		id: loader
+
+		asynchronous: true
+		active: Qt.platform.os !== "wasm"
+		sourceComponent: InputPanel {
+			mainViewItem: mainView
 		}
+		onLoaded: Global.inputPanel = item
 	}
 }
--- a/components/NavBar.qml
+++ b/components/NavBar.qml
@@ -108,7 +108,7 @@ Rectangle {  // Use an opaque background so that page disappears behind nav bar 
 					height: width
 					radius: Theme.geometry_notificationsPage_delegate_marker_radius
 					color: Theme.color_critical
-					visible: model.url === Global.appUrl("/pages/NotificationsPage.qml")
+					visible: model.url.endsWith("NotificationsPage.qml")
 							 && !!Global.notifications
 							 && Global.notifications.alert
 				}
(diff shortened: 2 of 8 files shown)


ApplicationContent.qml
```diff
@@ -88,16 +88,27 @@ Item {
...
 		if (Qt.platform.os !== "wasm") {
-			_inputComponent = Qt.createComponent(Global.appUrl("/components/InputPanel.qml"), Component.Asynchronous)
-			_inputComponent.statusChanged.connect(function() {
-				if (_inputComponent.status === Component.Ready) {
-					Global.inputPanel = _inputComponent.createObject(root, { mainViewItem: mainView })
-				} else if (_inputComponent.status === Component.Error) {
-					console.warn("Cannot load InputPanel:", _inputComponent.errorString())
-				}
-			})
+			loader.sourceComponent = inputPanelComponent
```
jpetrell, line 111: Could just write

```
Loader {
    active: Qt.platform.os !== "wasm"
    asynchronous: true
    sourceComponent: InputPanel { .. }
}
```

components/PageStack.qml
```diff
@@ -69,7 +69,7 @@ C.StackView {
 				return
 			}
 
-			let objectOrUrl = typeof(obj) === "string" ? Global.appUrl(obj) : obj
+			let objectOrUrl = typeof(obj) === "string" ? ".." + obj : obj
```
jpetrell, line 72: If you make this relative did you test that the loading from qrc still works in non-file system build?

I would remove Global.appUrl(), it is now only used in one place in NavBar.qml.

jpetrell:
Tested that I could modify MainView, PageStack, OverviewPage, SettingsPage, NavBar, InputPanel, PageSettingsWifi, PageSettingsGpsList. :partying_face:

components/PageStack.qml
DanielMcInnes: Yes I did, it works ok. Removed Global.appUrl().

components/NavBar.qml
```diff
@@ -108,7 +108,7 @@ Rectangle {  // Use an opaque background so that page disappears behind nav bar
 					height: width
 					radius: Theme.geometry_notificationsPage_delegate_marker_radius
 					color: Theme.color_critical
-					visible: model.url === Global.appUrl("/pages/NotificationsPage.qml")
+					visible: model.url === "qrc:/qt/qml/Victron/VenusOS/pages/NotificationsPage.qml"
```
chriadam, line 111: I'm confused about this.  What if it is loaded from the filesystem rather than qrc?

pages/MainView.qml
```diff
@@ -88,7 +88,9 @@ Item {
 
 		z: 1
 		opacity: 0.0
-		source: Global.appUrl("/pages/ControlCardsPage.qml")
+		sourceComponent: Component {
```
chriadam, line 91: I think QML will automatically add Component { } sugar, so you can just do sourceComponent: ContolCardsPage {} for brevity, I believe.

chriadam:
Typo in commit message (giles -> files)

chriadam:
A couple of comments, but mostly LGTM

components/NavBar.qml
DanielMcInnes: Whoops, thanks Chris

pages/MainView.qml
DanielMcInnes: Done

ApplicationContent.qml
```diff
@@ -88,18 +88,22 @@ Item {
...
+
+	Component {
+		id: inputPanelComponent
+
+		InputPanel {
+			id: inputPanel
+
+			mainViewItem: mainView
+			Component.onCompleted: Global.inputPanel = inputPanel
 		}
```
jpetrell, line 107: Could just write 

```
Loader {
    onLoaded: Global.inputPanel = item
    active: Qt.platform.os !== "wasm"
    asynchronous: true
    sourceComponent: InputPanel {
        mainViewItem: mainView
    }
}
```

DanielMcInnes: Done

jpetrell (approved):

PR #904, opened 2024-03-27
Don't draw the out-of-screen keyboard

jpetrell:
Reduces the overall CPU usage caused by drawing. On Cerbo minimal animation CPU consumption goes from 40% to 24%.

Contributes to #639.


blammit (approved):
Nice! Is it even better if we put it in a Loader that is `active=false` when not in use? Or maybe that would cause a noticeable delay when showing the keyboard? Anyway, LGTM.

PR #906, opened 2024-03-27
Don't draw off-screen swipe view

### LINKED ISSUE #905: Top-level parent pages stay visible when opening the sub-pages
The CPU usage caused by the drawing is proportional to how much gets drawn on every frame. Noticed that opening sub-pages from the Overview and Settings main pages doesn't properly hide the top-level parent Overview or Settings main pages. In the Overview page case it means that the 34% CPU usage caused by the animations keep playing in the background. In the settings case it means that two instead of one page is always drawn on every frame, increasing the CPU consumption e.g. during scrolling from 40% to 50%.

The issue is limited to the top-level pages like the Overview page and the Settings main page. The sub-pages are hidden and shown correctly.

jpetrell:
Reduces the total CPU consumption caused by drawing (e.g. during interaction and animations) when navigating inside the sub-pages as only one page instead of two get drawn at one time.

Also helped stop the Overview page animations when you have opened sub-pages by tapping the Overview boxes, reducing the CPU consumption from 34% to 1%.

Fixes #905.


blammit (approved):
Nice. Is this also triggered when the control cards page is shown?

jpetrell:
>  Is this also triggered when the control cards page is shown?

Good catch! Fixed.

PR #907, opened 2024-03-27
Define default scroll wheel speed

### LINKED ISSUE #901: GUI-v2 scroll speed is frustratingly slow on the browser
Open GUI-v2 Beta from https://vrm.victronenergy.com/. Open main settings page. Scroll the page using the mouse scroll wheel or from the desktop touch pad.

Expected result: Each scroll wheel step and flick on the touch pad scrolls the page by multiple list items, ideally with equivalent velocity and distance to other apps running in the surrounding OS.

Actual result: The scrolling with the mouse wheel or touch pad is way too slow.

The scroll velocity feels good if you flick the cursor in similar style as you would perform a flick gesture on touch screen (but on desktop you are likely to scroll with a mouse wheel or touch pad).

jpetrell:
Fixes #901.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -23,6 +23,7 @@
 #include <QQmlEngine>
 #include <QQuickWindow>
 #include <QCommandLineParser>
+#include <QStyleHints>
 
 #include <QtDebug>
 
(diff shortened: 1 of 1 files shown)


blammit (approved):
Nice, this was annoying me too.

PR #908, opened 2024-03-28
Support displaying absolute quantities on the circular gauges

### LINKED ISSUE #551: Allow switching Brief page circular gauges from percentages to units
Matthijs has asked if we could add a setting to allow switching the Brief page circular gauge values from Percentages to being "raw" units (e.g. Battery 87% to Battery 126 AH).

jpetrell:
Show the remaining battery level always in percentages, but allow showing the remaining tank levels in absolute volumes.

Fixes #551.

### DIFF
--- a/components/GaugeModel.qml
+++ b/components/GaugeModel.qml
@@ -37,8 +37,11 @@ ListModel {
 				return
 			}
 		}
+
+		const value = Global.systemSettings.briefView.unit.value === VenusOS.BriefView_Unit_Percentage || gauge.isBattery ? gauge.tankLevel : gauge.tankRemaining
+
 		insert(_insertionIndex(gauge.tankType, sortedGaugeTypes),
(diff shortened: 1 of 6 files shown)


components/GaugeModel.qml
```diff
@@ -53,7 +56,9 @@ ListModel {
 	function updateGauge(gauge) {
 		const gaugeIndex = findGauge(gauge)
 		if (gaugeIndex >= 0) {
-			set(gaugeIndex, { name: gauge.tankName, icon: gauge.tankIcon, value: gauge.tankLevel })
+
```
DanielMcInnes, line 59: extra whitespace?

DanielMcInnes (approved):
lgtm

components/GaugeModel.qml
jpetrell: Fixed.

Issue #892, opened 2024-03-20
The animation stutters when opening the Levels page with a swipe


jpetrell:
Go to Notifications page. Swipe to the right to access the Levels page.

Expected result: The Levels page slides in smoothly.

Actual result: The Levels page opening noticeably stutters.

chriadam:
This was resolved via page preloading in the page model.  Closing as fixed.

Issue #852, opened 2024-03-12
Navigating to the Brief page stutters


jpetrell:
Navigate to and from settings page either using a swipe or by tapping the nav bar buttons.

Expected result: All the main view pages are constructed in advance so the transition is smooth.

Actual result: There is ~100-200ms momentary freeze, a hiccup when the some of gauges are reconstructed during the reopening of the Brief page.

There may be good reason for doing this, but ideally any heavy operations would be done outside the interaction. Maybe the gauges could be kept constructed, just the updates would be paused when the page is inactive.

Measured on Cerbo GX.

PR #913, opened 2024-03-31
911 hide the levels page from the main view if there are no tanks and environmental sensors

DanielMcInnes:


DanielMcInnes:
There is a minor issue with these changes. If the Levels page becomes invisible (i.e. all tanks & sensors disappear) while you are on a page to the right of it, the screen goes blank for about 10 seconds. I don't think this will be an issue in the field, people either have tanks / sensors or they don't. I don't think the requirement is to handle adding and removing tanks at runtime. 

Having said that, if you understand why this happens, please let me know, I spent too long trying to figure it out.

pages/MainView.qml
```diff
@@ -65,21 +65,42 @@ Item {
...
+					}
+				}
+
+				Loader {
+					id: loader
+
+					source: url
+					// Once https://bugreports.qt.io/browse/QTBUG-115468 is fixed, the following expression for 'visible'
+					// can be replaced with: visible: C.SwipeView.view.pageInView(x, width, Theme.geometry_page_content_horizontalMargin)
+					visible: SwipeView.isCurrentItem || SwipeView.isNextItem || SwipeView.isPreviousItem
```
blammit, line 96: If you're on the BriefPage then click to the SettingsPage, then as you flick past, the intermediate pages disappear, because at that point, the Settings page is the current and the Notifications page is the previous item, so only those two pages are visible.

data/mock/config/MockDataSimulator.qml
```diff
@@ -11,6 +11,8 @@ QtObject {
 
 	property bool timersActive: !Global.splashScreenVisible
 	property int deviceCount
+	property bool tanksEnabled: true
+	property bool environmentalInputsEnabled: true
```
blammit, line 15: This is not modified anywhere?

components/NavBar.qml
```diff
@@ -44,46 +44,12 @@ Rectangle {  // Use an opaque background so that page disappears behind nav bar
 		x: Theme.geometry_page_content_horizontalMargin
 		width: parent.width - 2*Theme.geometry_page_content_horizontalMargin
 		height: parent.height
-		spacing: Theme.geometry_navigationBar_spacing
+		spacing: model.count === 5 ? Theme.geometry_navigationBar_5_icons_spacing : Theme.geometry_navigationBar_4_icons_spacing
```
blammit, line 47: The spacing could be calculated dynamically using the available space to avoid adding another theme constant.

components/SwipePageModel.qml
```diff
@@ -0,0 +1,51 @@
...
+	ListElement {
+		//% "Settings"
+		text: qsTrId("nav_settings")
+		icon: "qrc:/images/settings.svg"
+		url: "qrc:/qt/qml/Victron/VenusOS/pages/SettingsPage.qml"
+	}
+
+	onShowLevelsPageChanged: {
+		if (showLevelsPage) {
+			console.log("adding levels page")
```
blammit, line 37: Stray debug

pages/MainView.qml
```diff
@@ -65,21 +65,42 @@ Item {
...
-			// ensure the order matches NavBar.model
-			BriefPage {}
-			OverviewPage {}
-			LevelsPage {}
-			NotificationsPage {}
-			SettingsPage {}
+			Repeater {
+				id: repeater
+
+				model: Global.pageModel
```
blammit, line 80: Since `MainView` has a reference to `navBar`, maybe could just have an alias to the model in `NavBar` to make it accessible here, then we don't need another Global property that is only used here.

blammit (commented):
When the app is loaded on desktop, the app initially appears on the Settings page and then quickly scrolls to the Brief page. Can't repro this in the browser, but would be good to set the initial SwipeView index to ensure this doesn't happen, since it seems like a timing-dependent issue.

pages/MainView.qml
DanielMcInnes: Done. Did you know that there is no property in a ListView to tell you that it is scrolling due to changing the current index? The `moving` property only covers swiping and flicking

data/mock/config/MockDataSimulator.qml
DanielMcInnes: In mock mode, you can press 'V' to toggle tanksEnabled, and make the Levels page disappear / reappear

components/NavBar.qml
DanielMcInnes: Done

components/SwipePageModel.qml
DanielMcInnes: Removed

pages/MainView.qml
DanielMcInnes: Done

pages/MainView.qml
blammit: Yeah, that's basically why pageInView is needed :(

data/mock/config/MockDataSimulator.qml
```diff
@@ -230,9 +232,15 @@ QtObject {
 						   ? "Gallons (US)"
 						   : "Gallons (Imperial)")
 			event.accepted = true
+			break
+		case Qt.Key_V:
+			tanksEnabled = !tanksEnabled
+			if (!tanksEnabled) {
```
blammit, line 238: Empty braces?

DanielMcInnes:
> When the app is loaded on desktop, the app initially appears on the Settings page and then quickly scrolls to the Brief page. Can't repro this in the browser, but would be good to set the initial SwipeView index to ensure this doesn't happen, since it seems like a timing-dependent issue.

Done

pages/MainView.qml
```diff
@@ -65,21 +65,53 @@ Item {
...
+				model: navBar.model
+
+				onPagesLoadedChanged: {
+					if (pagesLoaded === count && !Global.allPagesLoaded) {
+						_swipeView.setCurrentIndex(0)
+						navBar.setCurrentIndex(0)
+					}
+				}
+
+				Loader {
```
blammit, line 101: Could you remind me, what was the reason for needing a Loader?

pages/MainView.qml
```diff
@@ -65,21 +65,53 @@ Item {
...
+					id: loader
+
+					source: url
+					// Once https://bugreports.qt.io/browse/QTBUG-115468 is fixed, the following expression for 'visible'
+					// can be replaced with: visible: C.SwipeView.view.pageInView(x, width, Theme.geometry_page_content_horizontalMargin)
+					visible: Global.allPagesLoaded && (_swipeView.moving || SwipeView.isCurrentItem)
```
blammit, line 107: Does this affect performance when flicking on the device? It seems like this would cause e.g. the Brief view to become visible even if you are only flicking between the Notifications and Settings pages.

data/mock/config/MockDataSimulator.qml
DanielMcInnes: Removed

pages/MainView.qml
DanielMcInnes: We don't really need a Loader. The motivation was to have a single model driving both the NavBar and the SwipeView.

pages/MainView.qml
DanielMcInnes: You are right, this does cause the Brief view to become visible when flicking between Notifications and Settings. With the frame rate turned on, this does reduce fps slightly when dragging slowly between pages by ~10fps, but I can't tell  the difference when I am using it.

jpetrell (commented):
Overall LGTM, some small comments here and there. :blush:

pages/MainView.qml
blammit: It would be nice to avoid the Loader if it's not needed, if the items are always loaded. Seeing as SwipeView is a `Container` with `contentModel`, `count` and `currentItem`: if each `SwipePage` stored the page icon and url, then NavBar could use the SwipeView `contentModel` as the `model`, and access the icon and url `model.icon` or such. And it could change the `currentItem` of the SwipeView as well.

pages/MainView.qml
jpetrell: In my profiling earlier I saw noticeable 100ms+ stutter that happened when Brief page became visible https://github.com/victronenergy/gui-v2/issues/852

The system would be more robust if pages that are not on screen would not become visible.

pages/MainView.qml
DanielMcInnes: Do you mean make SwipePageModel an ObjectModel? Done. This is slightly complicated by problems with using an ObjectModel in multiple views, see the comment in NavBar.qml line 52.

data/mock/config/MockDataSimulator.qml
```diff
@@ -11,6 +11,8 @@ QtObject {
 
 	property bool timersActive: !Global.splashScreenVisible
 	property int deviceCount
+	property bool tanksEnabled: true
+	property bool environmentalInputsEnabled: tanksEnabled
```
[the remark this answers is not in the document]
DanielMcInnes: updated as suggested

pages/MainView.qml
```diff
@@ -5,6 +5,7 @@
 
 import QtQuick
 import Victron.VenusOS
+import QtQuick.Controls as C
```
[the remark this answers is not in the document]
DanielMcInnes: removed

pages/MainView.qml
```diff
@@ -65,21 +65,53 @@ Item {
...
-			SettingsPage {}
+			Connections {
+				enabled: !Global.allPagesLoaded
+				target: _swipeView
+				function onMovingChanged() {
+					if (!parent.moving && repeater.pagesLoaded === repeater.count) {
+						Global.allPagesLoaded = true
+					}
+				}
+			}
```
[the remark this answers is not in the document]
DanielMcInnes: updated as suggested

pages/MainView.qml
```diff
@@ -199,7 +231,11 @@ Item {
 		color: root.backgroundColor
 		opacity: 0
 
-		onCurrentIndexChanged: if (swipeView) swipeView.currentIndex = currentIndex
+		onCurrentIndexChanged: {
+			if (swipeView) {
+				swipeView.currentIndex = currentIndex
+			}
+		}
```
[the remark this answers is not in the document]
DanielMcInnes: done

pages/MainView.qml
DanielMcInnes: @blammit , does it really matter if we don't display every intermediate page when flicking from the BriefPage to the SettingsPage? Currently, the 'Levels' page is not displayed in this scenario. There are enough pages displayed flicking past to give the sensation of flicking left -> right. There is currently no reliable way to determine which page is currently in view, due to https://bugreports.qt.io/browse/QTBUG-115468 . The only way I can display every intermediate page in this scenario is to make all pages visible while flicking, which hits the frame rate ~10fps. This is fixed in 6.5.3, I think we should revisit it then.

components/SwipePageModel.qml
```diff
@@ -0,0 +1,47 @@
...
+	}
+
+	onShowLevelsPageChanged: {
+		if (showLevelsPage) {
+			insert(2, {
+								 //% "Levels"
+								 text: qsTrId("nav_levels"),
+								 icon: "qrc:/images/levels.svg",
+								 url: "qrc:/qt/qml/Victron/VenusOS/pages/LevelsPage.qml"
+							 })
```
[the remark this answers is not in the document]
DanielMcInnes: I think this is ok now

pages/MainView.qml
```diff
@@ -65,21 +65,53 @@ Item {
...
+				id: repeater
+
+				property int pagesLoaded: 0
+
+				model: navBar.model
+
+				onPagesLoadedChanged: {
+					if (pagesLoaded === count && !Global.allPagesLoaded) {
+						_swipeView.setCurrentIndex(0)
+						navBar.setCurrentIndex(0)
```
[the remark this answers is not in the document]
DanielMcInnes: removed

pages/MainView.qml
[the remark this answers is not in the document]
DanielMcInnes: I changed the 'visible' logic to:
`visible: SwipeView.isCurrentItem || (view.moving && (SwipeView.isPreviousItem || SwipeView.isNextItem))`
So, while stationary, we only display the current page. While swiping, we display 1 page unnecessarily, but I don't think there is any way around that until https://bugreports.qt.io/browse/QTBUG-115468 is fixed (fixed in qt6.5.3)

data/mock/config/MockDataSimulator.qml
```diff
@@ -230,9 +232,15 @@ QtObject {
 						   ? "Gallons (US)"
 						   : "Gallons (Imperial)")
 			event.accepted = true
+			break
+		case Qt.Key_V:
+			tanksEnabled = !tanksEnabled
+			if (!tanksEnabled) {
+
+			}
```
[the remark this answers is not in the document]
DanielMcInnes: removed

pages/MainView.qml
blammit: Hmm. Yeah, you're right, let's drop the change to display every intermediate page during a swipe, then. I think it looks jumpy -- you see a bit of flicking but not the pages you expect -- but that's probably okay, and like you said we can't get around this problem if we want to be able to show/hide the Levels page dynamically.

DanielMcInnes:
> Overall LGTM, some small comments here and there. 😊 The main issue that needs to be fixed is the page visibility regression. On desktop the actual behavior seems to work nicely, though I am a bit concerned about the timings on real device and different sensors and tanks, how fast do the models update before GUI is ready or can the GUI go through different states (and cause unnecessary work or unexpected behavior) while the models update. I can re-test also on my side after the page visibility issue has been solved.

I think the 'page visibility' issue is as good as it can be at the moment. 1 page visible when stationary, up to 3 pages visible while flicking.

jpetrell:
> There is currently no reliable way to determine which page is currently in view, due to https://bugreports.qt.io/browse/QTBUG-115468

What is wrong with the old workaround? It worked before.

`visible: C.SwipeView.view.pageInView(x, width, Theme.geometry_page_content_horizontalMargin)
`

> I think the 'page visibility' issue is as good as it can be at the moment. 1 page visible when stationary, up to 3 pages visible while flicking.

This is regression from the main branch, drawing 3 instead of 2 pages adds roughly 50% more items that need drawing during interaction. It is less important how many pages are visible when the system is stationary and nothing gets drawn on the screen, many of the performance issues arise specifically during the interaction. The system is already struggling to achieve smooth 60fps during main view swiping on Cerbo GX see https://github.com/victronenergy/gui-v2/issues/892. 

> Tested on cerbo-gx, the performance seems to be unchanged compared to main.

On my profiling both the CPU load and frame rates are affected by how much gets drawn on the screen. E.g. hiding one extra item, the virtual keyboard from the screen reduced the CPU usage from 40% to 24%. Here the effect will likely be lower, but still notable.

DanielMcInnes:
Converted this review to 'draft' pending discussions on slack

data/mock/config/MockDataSimulator.qml
```diff
@@ -231,8 +232,10 @@ QtObject {
 						   : "Gallons (Imperial)")
 			event.accepted = true
 			break
+		case Qt.Key_V:
+			levelsEnabled = !levelsEnabled
+			break
```
jpetrell, line 237: When tapping "V" getting warning `qml: setCurrentIndex(): invalid index -1 nav bar count is: 1`

jpetrell:
Forgot to document here yesterday that pushed a fix commit on top, which restates pageInView(). There was a calculation error that occured then the page got removed due to originX becoming non-zero. Also fixed some jumping issues when removing/adding the Levels page to the swipe view.

There is still some issues before this can be merged, e.g. seeing Settings page sometimes flickering on top of the Brief page on launch.

DanielMcInnes:
Link to slack discussion for reference: https://victrondevelopment.slack.com/archives/C020QG3JVEW/p1713220802811429

pages/MainView.qml
DanielMcInnes: Found a magic combination that allows us to display every intermediate page during a swipe

pages/MainView.qml
DanielMcInnes: Fixed - we now only display pages that are on the screen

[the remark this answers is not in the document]
DanielMcInnes: Fixed - we now only display pages that are on the screen

DanielMcInnes:
> Forgot to document here yesterday that pushed a fix commit on top, which restates pageInView(). There was a calculation error that occured then the page got removed due to originX becoming non-zero. Also fixed some jumping issues when removing/adding the Levels page to the swipe view.
> 
> There is still some issues before this can be merged, e.g. seeing Settings page sometimes flickering on top of the Brief page on launch.

@jpetrell , I pushed a fix for the flickering Settings page issue. I've tested on desktop, wasm and cerbo gx, would you mind testing on ekrano?

jpetrell (approved):
> would you mind testing on ekrano?

Overall LGTM.

It is not perfect, but the remaining issues are quite minor:
- On Ekrano the Levels, Notifications and Settings pages momentarily gain visibility during launch, but the user doesn't see anything suspicious as they are shown behind the splash screen (unless you disable splash with --skip-splash command-line option), so relatively minor performance issue (some extra drawing happening, slowing down the launch slightly)
- If the Levels page gets hidden while its open the Notifications page momentarily becomes visible, and then the Overview page slides in. The slide transition is unnecessary, ideally it would just directly show the Overview page with nothing extra flickering on the screen, but again overall quite minor issue as happens very rarely
- If the Levels page gets added while the Notifications page is open, the Levels page momentarily becomes visible, and then Notifications page slides back in. The slide transition is again unnecessary here, but a minor issue as happens very rarely
- Other combinations work like expected, the current page stays intact and only the Levels icon button is added/hidden on the navigation bar

Do we want to merge this to 1.0.0 or wait? It is a nice feature, but carries some regression risk due to the extent of the changes.

components/controls/SwipeView.qml
```diff
@@ -32,6 +34,8 @@ T.SwipeView {
 		model: control.contentModel
 		interactive: control.interactive
 		currentIndex: control.currentIndex
+		onCurrentIndexChanged: scrollingTimer.restart()	// 'listView.moving' stays false when we are moving to a different page due to clicking on the nav bar.
+														// The scrolling timer is needed to tell us when the listView is in motion due to a nav bar click.
```
blammit, line 38: Odd indentation here?

blammit (approved):

components/controls/SwipeView.qml
DanielMcInnes: It only looks weird on a browser, it looks right in QtCreator.

blammit: Definitely looks worse in a browser, but to avoid the extra weirdness from tabs, could just put the two lines of comments above the onCurrentIndexChanged instead

DanielMcInnes: I've already merged it sorry, can I tidy it up next time I commit something?

blammit: Eh, was a minor thing, don't worry about it now :) We can rearrange it next time the code is touched.

Issue #914, opened 2024-04-02
Pause the Overview page animations on high CPU usage


jpetrell:
Animations cause 35% CPU load on Cerbo GX. When the system is under heavy load pause the animations.

jpetrell:
Serj didn't want any special graphics, pausing the animation is enough. The pausing has been implemented. Closing.

PR #915, opened 2024-04-02
Pause Overview page animations during high CPU load

jpetrell:
To simulate high CPU load call `while : ; do : ; done` inside the Ekrano GX terminal.

Calculates the total CPU usage caused by all the processes running on the system so different to what you get from top that displays CPU usage per process. On my laptop venus-gui-v2 uses 8% CPU (top) while the system total CPU (from /proc/stat) usage hovers around 4%. On Cerbo venus-gui-v2 uses 35% CPU while the system total CPU usage hovers around 40-50%. On Ekrano venus-gui-v2 uses 7% while the system total CPU usage overs around 12%.

Hands-on video demonstrating the pausing of animations:
https://github.com/victronenergy/gui-v2/assets/2203667/fc40dffa-6cbe-4f80-9d2d-d00b06f5d561

Also introduces an option to display CPU usage on the debug page.

Contributes to #914.

### DIFF
--- a/components/CpuMonitor.qml
+++ b/components/CpuMonitor.qml
@@ -1,0 +1,24 @@
+/*
+** Copyright (C) 2024 Victron Energy B.V.
+** See LICENSE.txt for license information.
+*/
+
+import QtQuick
+import Victron.VenusOS
+
+Rectangle {
+	height: label.height
+	width: label.width + Theme.geometry_button_spacing
+
+	QuantityLabel {
+		id: label
+
+		value: cpuInfo.usage
+		unit: VenusOS.Units_Percentage
+		anchors.centerIn: parent
+
+		CpuInfo {
+			id: cpuInfo
+		}
+	}
+}
--- a/components/widgets/WidgetConnector.qml
+++ b/components/widgets/WidgetConnector.qml
@@ -74,6 +74,11 @@ Item {
 			// Animate at a constant rate of pixels/sec, based on the diagonal length of the shape
 			electronAnim.duration = electronTravelDistance / Theme.geometry_overviewPage_connector_electron_velocity * 1000
 		}
+
+		// Force drawing manually if the animations are paused
+		if (overviewPageRootAnimation.paused) {
+			pathUpdater.update()
+		}
 	}
 
 	visible: defaultVisible
@@ -90,13 +95,13 @@ Item {
 		PropertyChanges { target: connectorPath; y: connectorPath.expandedY; yDistance: connectorPath.expandedYDistance }
 		PropertyChanges { target: connectorPath; startAnchorY: connectorPath.startAnchorExpandedY }
 		PropertyChanges { target: connectorPath; endAnchorY: connectorPath.endAnchorExpandedY }
+		PropertyChanges { target: root; _transitionUpdating: 1.0 }
 	}
 
 	transitions: Transition {
 		enabled: root.animateGeometry
-
 		NumberAnimation {
-			properties: "y,yDistance,startAnchorY,endAnchorY"
+			properties: "y, yDistance, startAnchorY, endAnchorY, _transitionUpdating"
 			duration: Theme.animation_page_idleResize_duration
 			easing.type: Easing.InOutQuad
 		}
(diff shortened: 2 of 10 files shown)


components/widgets/WidgetConnector.qml
```diff
@@ -253,6 +258,10 @@ Item {
 		}
 	}
 
+	// Force drawing during the expanding transition even if the animations are paused
+	property real _transitionUpdating
+	on_TransitionUpdatingChanged: if (overviewPageRootAnimation.paused) pathUpdater.update()
```
DanielMcInnes, line 263: Historically, we have always done this:

on_TransitionUpdatingChanged: {
    if (overviewPageRootAnimation.paused) {
        pathUpdater.update()
    }
}

However, I prefer the way you do it, it saves 4 lines of cruft. Fine with me

components/CpuMonitor.qml
```diff
@@ -0,0 +1,19 @@
+import QtQuick 2.0
```
DanielMcInnes, line 1: Remove the '2.0' please.

components/CpuMonitor.qml
```diff
@@ -0,0 +1,19 @@
...
+import Victron.VenusOS
+
+Rectangle {
```
DanielMcInnes, line 4: Could just be an Item?

pages/MainView.qml
```diff
@@ -354,4 +354,15 @@ Item {
...
 	}
+
+	Loader {
+		active: Global.displayCpuUsage
+		anchors {
+			bottom: parent.bottom
+			right: parent.right
+		}
+		sourceComponent: CpuMonitor {
+			color: root.backgroundColor
```
DanielMcInnes, line 365: If you make CpuMonitor an Item instead of a Rectangle, you won't need to set 'color'

src/cpuinfo.cpp
```diff
@@ -0,0 +1,69 @@
...
+#include "cpuinfo.h"
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <unistd.h>
+
+using namespace Victron::VenusOS;
+
+int get_cpu_usage(unsigned long long &previousBusy, unsigned long long &previousIdle) {
+  // Read CPU stats
```
DanielMcInnes, line 15: the formatting is weird for this file

src/cpuinfo.cpp
```diff
@@ -0,0 +1,69 @@
...
+  FILE *file = fopen("/proc/stat", "r");
+  if (file == NULL) {
```
DanielMcInnes, line 17: NULL -> nullptr

src/cpuinfo.cpp
```diff
@@ -0,0 +1,69 @@
...
+	perror("Could not open /proc/stat");
```
DanielMcInnes, line 18: perror -> qCWarning ?

pages/settings/debug/PageDebug.qml
```diff
@@ -11,28 +11,43 @@ Page {
...
+				checked: FrameRateModel.enabled
+				onClicked: FrameRateModel.enabled = !FrameRateModel.enabled
+			}
+
+			SwitchItem {
+				//% "Display CPU usage"
+				text: qsTrId("settings_page_debug_display_cpu_usage")
+				checked: Global.displayCpuUsage
+				onClicked: Global.displayCpuUsage = !Global.displayCpuUsage
+				allowed: Qt.platform.os === "linux"
```
DanielMcInnes, line 48: Do we want this for wasm as well?

components/CpuMonitor.qml
jpetrell: Done.

components/widgets/WidgetConnector.qml
jpetrell: I was commenting yesterday on your PR about considering one line more readable (just forgot to submit the review comments...). :smile: 

I can do either way, but yeah the prefer concise one.

pages/MainView.qml
jpetrell: It is actually intentional to maintain legibility when there is stuff drawn below the figures.

pages/settings/debug/PageDebug.qml
jpetrell: Not sure, but doesn't hurt either. Added. :+1:

src/cpuinfo.cpp
jpetrell: Fixed.

src/cpuinfo.cpp
jpetrell: Fixed.

src/cpuinfo.cpp
jpetrell: Fixed.

DanielMcInnes (approved):
lgtm

PR #920, opened 2024-04-03
Align dialog scaling with the application content

### LINKED ISSUE #918: Dialogs are opened in a different scale than the rest of the UI (wasm)
Open GUI-v2 Beta from https://vrm.victronenergy.com/. Open a modal dialog, e.g. generator or inverter dialogs from the Control cards page.

On the browser, the modal dialogs are shown proportionally larger than the rest of the UI below:

image
image

jpetrell:
GUI v2 has a dialog layer, which is a child of the transform item that gets scaled down on wasm builds, but Qt Quick Control dialogs get reparented outside the dialog layer before being shown. The new parent (Overlay.overlay) itself is parented to the invisible root item of the scene (Window.contentItem). To include the dialog in the scaling move the viewport transformations to the root item instead. Adapt DialogShadow to counter the scaling as we want the dimming to be applied to the whole Qt window.

Fixes #918.

### DIFF
--- a/Main.qml
+++ b/Main.qml
@@ -80,20 +80,18 @@ Window {
 		}
 	}
 
-	Item {
-		anchors.horizontalCenter: parent.horizontalCenter
-		width: Theme.geometry_screen_width + wasmPadding
-		height: Theme.geometry_screen_height
-
+	contentItem {
 		// on wasm just show the GUI at the top of the screen,
 		// otherwise browser chrome can cause problems on mobile devices...
-		y: (Qt.platform.os != "wasm" || scale == 1.0) ? (parent.height-height)/2 : 0
+		y: (Qt.platform.os != "wasm" || scale == 1.0) ? (root.height - contentItem.height)/2 : 0
 		transformOrigin: Qt.platform.os != "wasm" ? Item.Center : Item.Top
+		x: (root.width - contentItem.width)/2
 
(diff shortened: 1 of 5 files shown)


Main.qml
```diff
@@ -151,51 +149,46 @@ Window {
 
 		// Ideally each item would use focus handling to get its own key events, but in wasm the
 		// pagestack's pages do not reliably receive key events even when focused.
-		focus: true
```
blammit, line 154: Should the comment above this be removed as well?

blammit (approved):
Nice! LGTM

blammit:
> GUI v2 has a dialog layer, which is a child of the transform item that gets scaled down on wasm builds, but Qt Quick Control dialogs get reparented outside the dialog layer before being shown. The new parent (Overlay.overlay) itself is parented to the invisible root item of the scene (Window.contentItem). To include the dialog in the scaling move the viewport transformations to the root item instead. Adapt DialogShadow to counter the scaling as we want the dimming to be applied to the whole Qt window.

> Fixes #918.

@jpetrell could you put this explanation in the commit message?

Main.qml
jpetrell: The comment is about the Keys.onPressed below. Focus is not needed since we can be certain that root object has focus in the item focus chain, but couldn't earlier with our own extra Item {} parent.

Issue #900, opened 2024-03-26
GUI-v2 opens scaled down in the Victron VRM website


jpetrell:
Open GUI-v2 Beta from https://vrm.victronenergy.com/. 

Expected result: GUI-v2 is shown maximised to the available space to optimize legibility and ease interaction.

Actual result: GUI-v2 is shown inside a popup, but failing to expand to the available popup dimensions.

Seen on the desktop browser, reproes both on Chrome and Firefox.

![image](https://github.com/victronenergy/gui-v2/assets/2203667/6c199f55-46a8-47b8-be5a-a507c572c522)

Furthermore, modal dialogs are shown proportionally larger than the rest of the UI.

![image](https://github.com/victronenergy/gui-v2/assets/2203667/0508a39a-5921-429e-b993-e3d058a62876)

jpetrell:
> Furthermore, modal dialogs are shown proportionally larger than the rest of the UI.

Filed as a separate issue https://github.com/victronenergy/gui-v2/issues/918

jpetrell:
Found a way to reproduce the issue locally. Seems the 20-30% empty area below the viewport is not caused by the VRM website integration, but by GUI v2. GUI v2 initially scales nicely to full viewport on my desktop browsers (Chrome, Firefox), but similar empty area appears if I make the browser window smaller. Seems there is some height limit (though didn't find any @media height rules there, so probably Qt internal mechanism) after which the GUI v2 starts to scale differently. I will dig in deeper.

The popup dimensions on the VRM website should be fixed to follow the GUI v2 aspect ratio, which currently seems to be 5:3 (800:480 viewport, 5" variant) on the wasm build. There is also 7" 1024:600 variant, which was ideally used on the bigger desktop and tablet screens. Could also at some point consider making the GUI v2 dynamically scalable to any mobile, tablet and laptop screen dimension. Most existing views would already adapt nicely, just some crowded ones which don't allow scrolling like the Overview page would need some work.

jpetrell:
GUI v2 side scaling improvements have been released as part for v0.2.15 few days ago. Still need to be verified on https://vrm.victronenergy.com , which now was still running an older version. Also, the aspect ratio of the popup needs to be aligned with 1024:600 (128/75) aspect ratio of the wasm build.

jpetrell:
> Still need to be verified on https://vrm.victronenergy.com/ , which now was still running an older version. 

[...]

PR #923, opened 2024-04-04
Always maximise GUI v2 to the viewport dimensions

jpetrell:
See the individual commits. Done on top of https://github.com/victronenergy/gui-v2/pull/920

Resolution doesn't tell you much about DPI so better to just maximize the touch GUI to the available viewport. Usually would center the GUI to the available area, but on browsers you kind of expect the website to appear on the top so positioning to the top there.

Video demonstration, first running on Linux, then on browser:
[Screencast from 2024-04-04 16-58-49.webm](https://github.com/victronenergy/gui-v2/assets/2203667/a305fbe7-9f0b-43a8-9205-00da37086e82)

Contributes to #900.

### DIFF
--- a/Main.qml
+++ b/Main.qml
@@ -81,71 +81,15 @@ Window {
 	}
 
 	contentItem {
-		// on wasm just show the GUI at the top of the screen,
-		// otherwise browser chrome can cause problems on mobile devices...
-		y: (Qt.platform.os != "wasm" || scale == 1.0) ? (root.height - contentItem.height)/2 : 0
-		transformOrigin: Qt.platform.os != "wasm" ? Item.Center : Item.Top
-		x: (root.width - contentItem.width)/2
+		// on wasm just show the GUI at the top of the screen
+		transformOrigin: Qt.platform.os !== "wasm" ? Item.Center : Item.Top
 
 		// In WebAssembly builds, if we are displaying on a low-dpi mobile
 		// device, it may not have enough pixels to display the UI natively.
 		// To fix, we need to downscale everything by the appropriate factor,
 		// and take into account browser chrome stealing real-estate also.
 		onScaleChanged: Global.scalingRatio = contentItem.scale
-		scale: {
-			// no scaling required if not on wasm
-			if (Qt.platform.os != "wasm") {
-				return 1.0
-			}
-			// no scaling required if we can pad and zoom instead
-			if ((Screen.height >= Theme.geometry_screen_height)
-					&& (Screen.width >= Theme.geometry_screen_width)) {
-				return 1.0
-			}
-			var hscale = Screen.width / Theme.geometry_screen_width
-			var vscale = Screen.height / Theme.geometry_screen_height
-			// in landscape mode, give even more room, as the browser chrome
-			// will take up significant vertical space.
-			var chromeFactor = (Screen.height > Screen.width) ? 1.0 : 0.75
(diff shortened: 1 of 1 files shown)


Main.qml
```diff
@@ -80,122 +80,72 @@ Window {
...
-			var expectedWidth = Theme.geometry_screen_width * verticalRatio
-			var chromeFactor = 1.2 // browser doesn't give whole screen to content area
-			var delta = (Screen.width - expectedWidth) * chromeFactor
-			if (delta < 0) {
-				return 0
-			}
-			return Math.ceil(delta)
-		}
+		onScaleChanged: Global.scalingRatio = contentItem.scale
+		scale: Math.min(root.width/Theme.geometry_screen_width, root.height/Theme.geometry_screen_height)
```
jpetrell, line 92: Scale is 1.0 on desktop window and GX screen.

chriadam:
I haven't had a chance to review this properly yet.  Just double checking: did you test this on Android and iOS devices, in both portrait and landscape orientations, to make sure that no interior scrolling is required when the navigation chrome is visible?

jpetrell:
>  Just double checking: did you test this on Android and iOS devices, in both portrait and landscape orientations

Yes I've tested this with Android (Sony smartphone) and iOS (iPad), on portrait and landscape. Also on various simulated iPhone, Galaxy, Pixel devices using Chrome and Firefox Developer tools. But the layouting logic happens inside the Qt window so different size configurations can also be tested on the desktop environment.

iPad screenshot with hacked red background to highlight the content area:
image

jpetrell:
Rebased. No changes to the actual commit.

chriadam (approved):
Seems to work nicely, and the old hidden navigation bar issue doesn't repro on the devices I tested with, so LGTM.  Nice work.

mr-manuel:
On a 2560x1440px screen with 100% DPI the GUI now looks very pixelish. Shouldn't all elements be vectors? If yes, then the display get's only zoomed instead of scaled?

grafik

jpetrell:
> On a 2560x1440px screen with 100% DPI the GUI now looks very pixelish. Shouldn't all elements be vectors?
> If yes, then the display get's only zoomed instead of scaled?

Good catch! I didn't notice the issue on the devices I tested, but now that you point the issue out they do look a bit pixelated on some screens. The quantity label digits are currently drawn as bitmaps and don't take the viewport scaling into account like the normal font rendering. The digits will become crisp on the high res displays if https://github.com/victronenergy/gui-v2/pull/941 gets approved (also downscaled fonts will look better). As a stopgap solution we can limit the maximum scaling value to 1.0 like before until we get the quantity label rendering improved (from legibility and touch-usability point of view would still prefer to maximize the touch GUI to the available viewport as the resolution is not a good measure of DPI nowadays).

mr-manuel:
But the gauges should be already vectors or are they also rendered as bitmap?

jpetrell:
> But the gauges should be already vectors or are they also rendered as bitmap?

Looks like another issue, the gauges are drawn on FBO bitmap to achieve anti-aliasing effect on GX device, needs adjusting/disabled for high resolution displays.

Reverted upscaling for now in https://github.com/victronenergy/gui-v2/pull/943. Filed https://github.com/victronenergy/gui-v2/issues/944 about re-enabling upscaling after the rendering issues have been fixed.

PR #927, opened 2024-04-05
Use slightly different press effect for list items

### LINKED ISSUE #925: Press effect causes the list item background to become invisible
Press list item in settings.

Expected result: List item looks pressed down.

Actual result: Press effect 10% darkens the background just enough for the background to fade out.

Serj has defined an own press effect color for the list items, update.

image

jpetrell:
The default one caused list item background to fade out.

Fixes #925.

### DIFF
--- a/components/RadioButtonControlValue.qml
+++ b/components/RadioButtonControlValue.qml
@@ -25,7 +25,7 @@ ControlValue {
 		onClicked: root.clicked()
 	}
 
-	PressArea {
+	ListPressArea {
 		id: mouseArea
 
 		anchors.fill: parent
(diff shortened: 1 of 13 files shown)


components/RadioButtonControlValue.qml
```diff
@@ -25,7 +25,7 @@ ControlValue {
 		onClicked: root.clicked()
 	}
 
-	PressArea {
```
blammit, line 28: RadioButtonControlValue is also used in the control cards as well, so it will affect the press effects there. Is that OK?

blammit (approved):
One question, LGTM otherwise

components/RadioButtonControlValue.qml
jpetrell: Yes this works there also.

PR #929, opened 2024-04-08
Split AC/DC input side gauge into separate AC and DC input gauges 

blammit:
This is the first step in adding fixed min/max values for AC and DC gauges; I'm working on adding the fixed min/max values in a separate PR.

### DIFF
--- a/pages/BriefPage.qml
+++ b/pages/BriefPage.qml
@@ -143,7 +143,7 @@ SwipeViewPage {
 
 			name: properties.name
 			icon.source: battery.icon
-			value: (!visible || isNaN(battery.stateOfCharge)) ? 0 : Math.round(battery.stateOfCharge)
+			value: visible && !isNaN(battery.stateOfCharge) ? battery.stateOfCharge : 0
 			voltage: battery.voltage
 			current: battery.current
 			status: Gauges.getValueStatus(value, properties.valueType)
@@ -181,7 +181,7 @@ SwipeViewPage {
 				x: root._gaugeArcMargin
 				opacity: root._gaugeArcOpacity
 				animationEnabled: root.animationEnabled
-				value: !visible ? 0 : acInputRange.valueAsRatio * 100
+				value: visible ? acInputRange.valueAsRatio * 100 : 0
 
 				// Gauge color changes only apply when there is a maximum value.
 				valueType: isNaN(acInputRange.maximumValue)
@@ -234,7 +234,7 @@ SwipeViewPage {
 				x: root._gaugeArcMargin
 				opacity: root._gaugeArcOpacity
 				animationEnabled: root.animationEnabled
-				value: !visible ? 0 : dcInputRange.valueAsRatio * 100
+				value: visible ? dcInputRange.valueAsRatio * 100 : 0
 
 				ArcGaugeQuantityRow {
 					id: dcInGaugeQuantity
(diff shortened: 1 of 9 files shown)


pages/BriefPage.qml
```diff
@@ -91,9 +91,7 @@ SwipeViewPage {
 			leftMargin: Theme.geometry_briefPage_edgeGauge_horizontalMargin
 			right: mainGauge.left
 		}
-		// Show gauge even if there are no active AC inputs, so that the gauge visibility doesn't
-		// jump on/off when inputs are connected/disconnected
-		active: visible && (!!Global.acInputs.activeInput || Global.dcInputs.model.count > 0)
+		active: !!Global.acInputs.activeInput || Global.dcInputs.model.count > 0
```
jpetrell: In the commit message could refer to https://github.com/victronenergy/gui-v2/issues/852.

pages/BriefPage.qml
```diff
@@ -37,10 +102,8 @@ SwipeViewPage {
 	Loader {
 		id: mainGauge
 
-		anchors {
-			top: parent.top
-			topMargin: Theme.geometry_mainGauge_topMargin
-		}
+		// vertically center to the unexpanded height of the page
+		y: (Theme.geometry_screen_height - Theme.geometry_statusBar_height - Theme.geometry_navigationBar_height - height) / 2
```
jpetrell, line 115: Ideally the was some default content area that you could anchor the content to to avoid having to be aware of status bars and navigation bar dimensions..

pages/BriefPage.qml
```diff
@@ -85,30 +148,33 @@ SwipeViewPage {
...
-			startAngle: solarYieldGauge.active ? 270 : (270 - Theme.geometry_briefPage_largeEdgeGauge_maxAngle / 2)
+			startAngle: gaugeParams.start || 0
+			endAngle: gaugeParams.end || 0
+			height: gaugeParams.height || 0
+			alignment: gaugeParams.alignment || 0
+
+			x: root._gaugeArcMargin
+			opacity: root._gaugeArcOpacity
 			animationEnabled: root.animationEnabled
+			value: !visible ? 0 : inputsRange.valueAsRatio * 100
```
jpetrell: Nitpick: Personally I find `visible ? inputsRange.valueAsRatio * 100 : 0` i.e. no negatives more readable.

jpetrell (commented):
Overall LGTM.

The Brief page gauges look really cool, but so closer inspection some feel a bit gimmicky. They work well with batteries and tanks, anything with a well-defined and known value range. For inputs and loads feels that you should really understand the installation/configuration and the expectations. E.g. solar range could somehow be based on the the average maximum and minimum within the given month and day time. Now you may see solar yield almost maxed out just because you happened to boot up the device 5 minutes ago, or almost at 0 just because the usual value dipped few percentage points. To understand the range you would need to indicate to the user clearly how the range is determined, i.e. this gauge means the consumption is near the maximum for the running week, but minimum for today.

As somebody who still doesn't understand the domain that well feels odd to see 300W or 1069W consumption and the gauge indication near 0. Similarly felt odd to see tiny blue bar on the range in the ESS demo mode grid meter for negative values -2300W and -2400W. With the negative values would expect the values to be inverted (but with different color to indicate these are not positive values), i.e. the largest negative value. Whether the calculated dynamic minimum or maximum makes sense depends a lot if they present normal range variation or are seen as unwanted outliers that things should not be measured against.

pages/BriefPage.qml
blammit: Yes, hmm. The problem is that here we want to ignore the way the page resizes in "expanded" mode, whereas other pages want to adjust/resize together with that mode. So the content area would have to stay fixed in the Brief page, but resize in the others. Will keep it mind and see if we can make this nicer later on.

pages/BriefPage.qml
blammit: Done

pages/BriefPage.qml
blammit: Agreed, updated

blammit:
Yeah these are good points about the gauges. They are not that useful if you don't understand the min/max range or the colours. Certainly with the dynamic min/max, you would have to know what the device normally produces to know whether the colours or the value indicates a real problem or not.

For the negative values, I wonder if we could show the darker/lighter variations of the same colour in the positive range. Or show some marker at the zero point, like the temperature gauges.

PR #930, opened 2024-04-09
Migrate network settings to venus-platform

jpetrell:
The functionality is limited (no way to sign in to hidden networks, no WPA certs, identity, domain, etc.), but should now match the existing functionality. 

Fixes #618.

PR #940, opened 2024-04-12
Fix issues with setting date/time

chriadam:


chriadam:
Not yet ready for review, I still need to implement a simple date picker.

chriadam:
Now ready for review.  I based the DateSelector off the TimeSelector, basically, but had to massage stuff so it fit on 5-inch form factor.

src/cpuinfo.cpp
```diff
@@ -24,14 +25,14 @@ int get_cpu_usage(unsigned long long &previousBusy, unsigned long long &previous
...
 
-
 	// Calculate CPU usage
 	unsigned long long busy = busyTotal - previousBusy;
 	unsigned long long idle = idleTotal - previousIdle;
 	previousBusy = busyTotal;
 	previousIdle = idleTotal;
 	return qRound((busy * 100.0) / (busy + idle));
 }
+#endif
```
jpetrell, line 35: Sorry for the regression!

jpetrell:
Why should the user care about UTC time or the browser local time? These are GX device settings, the time that matters is the GX device time. We shouldn't complicate the UX of the common cases (change the time from GX device, change the time within the same time zone) due to relatively uncommon edge case. From using other devices the user expects to see one time and date, specified in the selected time zone.

Can we have a separate setting for time and date? The combined dialog becomes quite unwieldy.

image

If the time zone conversions are difficult can we just provide a setter API over MQTT to set the hour, minute, day, month and year and let the device side do the time setting? Everything is represented in device time, and user just modifies bunch of integers. If that is not possible would just make the date and time setting read-only in the problematic case (accessed from the browser + time zones don't match).

image

DanielMcInnes:
I tried this on a cerbo, and it segfaulted the first time I changed the time. Haven't been able to reproduce it. Also, the buttons on the date / time pickerare too sensitive, often when I try to press a button once, it increments multiple times

src/cpuinfo.cpp
chriadam: All good, simple fix :-)

chriadam:
> Why should the user care about UTC time or the browser local time? These are GX device settings, the time that matters is the GX device time. We shouldn't complicate the UX of the common cases (change the time from GX device, change the time within the same time zone) due to relatively uncommon edge case. From using other devices the user expects to see one time and date, specified in the selected time zone.

I agree - but unfortunately we cannot (in WASM builds) convert any date-time to the CerboGX device's timezone, as Qt For WebAssembly doesn't support arbitrary timezones.  The only supported timezones are UTC and the Browser's local timezone.

> Can we have a separate setting for time and date? The combined dialog becomes quite unwieldy.

Hmm, ok, I can prototype that.

> image
> 
> If the time zone conversions are difficult can we just provide a setter API over MQTT to set the hour, minute, day, month and year and let the device side do the time setting? Everything is represented in device time, and user just modifies bunch of integers. If that is not possible would just make the date and time setting read-only in the problematic case (accessed from the browser + time zones don't match).

The timezone conversions are difficult (in the WebAssembly case only).  It's a possibility, I suppose, to expose year/month/day/hour/minute values on MQTT, and just allow the user to set those directly.  @ReinvdZee do you have thoughts about this?  The alternative is, yes, we could make the setting read-only for WebAssembly case (and in fact just hide the "local time" list entry altogether, since we can only display the browser-local time and not the device local time; so all that would remain is the utc time and the device's timezone).

> image

chriadam:
> I tried this on a cerbo, and it segfaulted the first time I changed the time. IIRC, I changed the hours from 22 -> 23. FW v3.40~2. Difficult to reproduce, but I managed to do it by pressing 'Set' immediately after changing the time. Also, the buttons on the date / time picker are too sensitive, often when I try to press a button once, it increments multiple times.
> 
> Update: I haven't been able to reproduce with the latest FW version v3.40~6

Interesting, I will try to repro the crash.  But it might go away altogether if I split the dialogs into date / time separate pickers.

Regarding spinbox button sensitivity - that's not something I will look into in this PR - if it's broken in this PR then it's broken for every use of SpinBox in our UI, and needs to be fixed separately, I think.

chriadam:
Ok, ready for re-review.  Two separate buttons (one to set date, one to set time).  Simplified a bunch of code also.  And fixed a bug in the timezone handling.

pages/settings/PageTzInfo.qml
```diff
@@ -30,51 +30,82 @@ Page {
...
 			}
 		}
 	}
 
+	// Ensure time is up-to-date while this page is open.
+	Timer {
+		interval: 10000
+		repeat: true
+		triggeredOnStart: true
+		running: BackendConnection.applicationVisible
```
jpetrell, line 78: Could also use page.animationEnabled

pages/settings/PageTzInfo.qml
```diff
@@ -30,51 +30,82 @@ Page {
...
 
-				onClicked: {
-					root._openTimeSelector()
+				buttonTexts: [ClockTime.currentDate, ClockTime.currentTime]
+				onButtonClicked: (index) => {
+					if (index == 0) {
+						root._openDateSelector()
+					} else {
+						root._openTimeSelector()
+					}
```
jpetrell, line 108: Personally would prefer (and find more readable) to declare the buttons explicitly i.e. something like

```
ListItem {
    content: Row {
        Button {
            text: ClockTime.currentDate
            onClicked: root._openDateSelector()
        }
        Button {
            text: ClockTime.currentTime
            onClicked: root._openTimeSelector()
        }
    }
}
```

src/clocktime.h
```diff
@@ -23,44 +23,84 @@ class ClockTime : public QObject
...
+	// that the dateTime will be in the systemTimeZone.
+	Q_PROPERTY(QDateTime dateTime READ dateTime NOTIFY dateTimeChanged)
+
+	Q_PROPERTY(int year READ year NOTIFY yearChanged)
+	Q_PROPERTY(int month READ month NOTIFY monthChanged)
+	Q_PROPERTY(int day READ day NOTIFY dayChanged)
+	Q_PROPERTY(int hour READ hour NOTIFY hourChanged)
+	Q_PROPERTY(int minute READ minute NOTIFY minuteChanged)
+	Q_PROPERTY(int second READ second NOTIFY secondChanged)
+	Q_PROPERTY(int msec READ msec NOTIFY msecChanged)
```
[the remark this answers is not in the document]
chriadam: True, should not be necessary any more, will remove.

pages/settings/PageTzInfo.qml
chriadam: Sure, can do.

pages/settings/PageTzInfo.qml
```diff
@@ -67,20 +64,28 @@ Page {
...
+					? qsTrId("settings_tz_date_time_browser_local")
+					//% "Date/Time local"
+					: qsTrId("settings_tz_date_time_local")
+				button.text: "" + ClockTime.currentDateTime.year
+					+ "-" + Utils.pad(ClockTime.currentDateTime.month, 2)
+					+ "-" + Utils.pad(ClockTime.currentDateTime.day, 2)
+					+ " " + Utils.pad(ClockTime.currentDateTime.hour, 2)
+					+ ":" + Utils.pad(ClockTime.currentDateTime.minute, 2)
 				writeAccessLevel: VenusOS.User_AccessType_User
 				enabled: Global.systemSettings.time.isValid
```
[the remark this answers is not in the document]
chriadam: Mock data exists, it does: `ClockTime.clockTime = (new Date().getTime() / 1000)` which should set the appropriate setting, I believe.

jpetrell:
Serj made following design:

image

pages/settings/PageTzInfo.qml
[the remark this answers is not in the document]
DanielMcInnes: If I run on the desktop, the 'set Data / Time' button is disabled
Screenshot from 2024-04-17 07-31-54

DanielMcInnes (approved):

chriadam:
> Serj made following design (#930 has SectionHeader component for "UTC" and "Local"):
> 
> image

I really don't think we should be accepting design changes for 1.0.0 at this point.

pages/settings/PageTzInfo.qml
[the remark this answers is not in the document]
chriadam: You're right!  I was testing on desktop with `--mqtt <cerbo_ip>` not `--mock` before, but yes the mock system settings impl never set the mock MQTT value.  Fixed now.

src/clocktime.h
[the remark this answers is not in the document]
chriadam: More testing shows that I was wrong - they are still needed.

Setting the timezone setting on the device sets the dbus/mqtt setting, but venus-platform does not change the actual system timezone setting (on the office cerbo, the timezone reported by `date` always returns `UTC`.  As such, the JS engine in QML always does QDateTime->Date conversion which involves a timezone shift.

I need to expose the individual parts.

chriadam:
Added back the int accessors to avoid having to return a QDateTime to QML (i.e. JS Date conversion with associated timezone shift), and also fixed a check in the setDateTime mutator.

pages/settings/PageTzInfo.qml
```diff
@@ -67,11 +67,16 @@ Page {
 			}
```
blammit, line 90: Noticed this second commit says "UTC to ?TZ" -> a stray question mark?

components/dialogs/DateSelectorDialog.qml
```diff
@@ -9,85 +9,20 @@ import Victron.VenusOS
 ModalDialog {
 	id: root
 
-	property var date: ClockTime.currentDateTime
```
blammit, line 12: The previous DateSelectorDialog design was from Serj -- is there a particular reason for changing this? (Personally I prefer the current one with larger text and buttons, which is more consistent with other dialogs where we use spinboxes.) Do we really need to do both the date and time setting in the same dialog?

Also this commit message says DateSelectorDialog is not used, but it is currently used by ListDateSelector, which is used by PageGeneratorTestRun. Opening "Settings > Generator start/stop > Settings > Conditions > Periodic Run > Run interval start date" now gives this error:

`qrc:/qt/qml/Victron/VenusOS/components/listitems/ListDateSelector.qml: Setting initial properties failed: DateSelectorDialog does not have a property called date`

src/clocktime.cpp
```diff
@@ -162,6 +162,20 @@ bool ClockTime::isDateValid(int year, int month, int day) const
...
+	static const QCalendar calendar;
+	if (calendar.isDateValid(year, month, 31)) {
+		return 31;
+	} else if (calendar.isDateValid(year, month, 30)) {
+		return 30;
+	} else if (calendar.isDateValid(year, month, 29)) {
+		return 29;
+	} else {
+		return 28;
+	}
```
blammit, line 176: Could this just use `calendar.daysInMonth()`?

components/DateSelector.qml
```diff
@@ -0,0 +1,81 @@
...
+		SpinBox {
+			id: yearSpinbox
+			anchors.verticalCenter: parent.verticalCenter
+
+			smaller: true
+			from: 1970
+			to: 2100
+			label.text: value
+			//% "YYYY"
+			//~ A label identifying the "year" portion of a date string
```
blammit, line 33: I've been using `//:` for translation hints in this codebase.

components/dialogs/DateSelectorDialog.qml
chriadam: Date and time settings aren't in the same dialog, though.
Oh, I thought the previous one was only a placeholder, not a proper design - the reason I believed so is that its spinboxes are vertically oriented unlike the horizontal time ones.  It was jarring to go from horizontal time to vertical date.

src/clocktime.cpp
chriadam: yes, somehow I missed that.  will update.

components/DateSelector.qml
chriadam: my mistake, will fix

pages/settings/PageTzInfo.qml
chriadam: No, it's meant to suggest that the target timezone can be anything.

components/dialogs/DateSelectorDialog.qml
chriadam: Basically reverted my changes to the DateSelectorDialog (at least, as much as I could.  The old DateSelectorDialog API won't work properly for our purposes since its API includes a JS Date, which necessitates a potential tz shift.  So, I fixed that, then updated that case you identified where it was used.  Visually, it should look the same as the old one.)

pages/settings/PageTzInfo.qml
jpetrell: Generally would try to avoid UI timers when the page is not visible, but in practice a timer that goes of once in 10 seconds has very little impact since AFAICS device never suspends.

pages/settings/PageTzInfo.qml
```diff
@@ -30,51 +30,85 @@ Page {
...
+				// so we can't display the device-local date/time,
+				// as we don't know what it is.  Just hide the setting.
+				allowed: Qt.platform.os != "wasm"
+
 				//% "Date/Time local"
 				text: qsTrId("settings_tz_date_time_local")
-				button.text: ClockTime.currentTimeText
 				writeAccessLevel: VenusOS.User_AccessType_User
-				enabled: Global.systemSettings.time.isValid
+				enabled: allowed && Global.systemSettings.time.isValid
```
jpetrell, line 100: allowed=false means invisible, no need to disable

pages/settings/PageTzInfo.qml
```diff
@@ -30,51 +30,85 @@ Page {
...
-				secondaryText: ClockTime.currentTimeUtcText
+				secondaryText: ClockTime.currentDateTimeUtc
 			}
 
-			ListButton {
+			ListItem {
+				// Qt for WebAssembly doesn't support timezones,
+				// so we can't display the device-local date/time,
+				// as we don't know what it is.  Just hide the setting.
+				allowed: Qt.platform.os != "wasm"
```
jpetrell, line 95: should this have "&& defaultAllowed" (aka userHasReadAccess)?

pages/settings/PageTzInfo.qml
jpetrell: Nice! :pray:

jpetrell (approved):
Looks good. Didn't find issues in testing. :ok_hand:

components/dialogs/DateSelectorDialog.qml
blammit: Great, works well!

blammit (approved):

pages/settings/PageTzInfo.qml
chriadam: yes, good catch, will fix this.

chriadam: done.

chriadam:
Thanks, merging.

PR #941, opened 2024-04-12
Replace the digit image provider with a monospaced digit font

### LINKED ISSUE #890: QuantityLabel numbers have odd sizing
Reported issue from https://community.victronenergy.com/questions/265862/venus-os-v33010-available-for-testing.html: after some period of time, the gauge numbers on the Brief page are incorrectly sized. Issue goes away on reboot.

The numbers are images provided by DigitImageProvider, so the provider is returning images of the wrong size for some numbers. 

Once an image is created, it will be cached by the engine, so if the image is wrong it will be reused until the cache is cleared; in this regard it makes sense that the issue cleared on reboot.

jpetrell:
image

Fixes #890.


blammit (approved):
Awesome! Numbers look much cleaner now. 

LGTM but would like @chriadam to approve as well.

chriadam:
Has this been approved by legal / does our license for Museo Sans allow distributing a modified font?  I cannot approve this unless the answer is Yes.

chriadam:
The last time we discussed this with Matthijs and Serj, Serj indicated that we didn't have the license to modify the font - see https://victrondevelopment.slack.com/archives/C04RYU0UMDE/p1701683775252659?thread_ts=1701519502.485569&cid=C04RYU0UMDE

But, if that is no longer the case, then this solution of course seems perfect.

jpetrell:
The license is being clarified, changing the state to draft.

chriadam (approved):

PR #942, opened 2024-04-15
Used fixed min/max values for Brief side gauges

blammit:
The main changes here update the Brief page to:

- Show a 3-phase gauge for AC input (top left) with fixed min/max values
- Show a 3-phase gauge for AC load (top right) with fixed max value
- Show fixed max value for DC input gauge (bottom left)
- Show fixed max value for DC load gauge (bottom right)

Note:

- Requires Venus OS v3.40~5 to fetch the fixed min/max values
- There is no UI setting yet to control whether the UI shows dynamic  vs fixed values.


pages/BriefPage.qml
```diff
@@ -241,6 +241,7 @@ SwipeViewPage {
 				opacity: root._gaugeArcOpacity
 				animationEnabled: root.animationEnabled
 				value: visible ? dcInputRange.valueAsRatio * 100 : 0
+				valueType: VenusOS.Gauges_ValueType_NeutralPercentage
```
chriadam, line 259: the commit message says "use NeutralPercentage as the gauge type as it does not need warning/critical gauge colours" but could you expand the comment to explain why it doesn't need warning/critical gauge colours?  I mean, why is a DC input gauge less "dangerous" to the system than an AC input, for example?

chriadam: Or is it only loads for which we display the warning colours?

pages/BriefPage.qml
jpetrell, line 1: Typo in the commit message "paramters"

pages/BriefPage.qml
```diff
@@ -27,11 +27,17 @@ SwipeViewPage {
...
+	}
+
+	function _leftGaugeCount() {
+		return (acInputGauge.active ? 1 : 0) + (dcInputGauge.active ? 1 : 0) + (solarYieldGauge.active ? 1 : 0)
+	}
 
-	readonly property int _rightGaugeCount: (acLoadGauge.active ? 1 : 0) + (dcLoadGauge.active ? 1 : 0)
-	readonly property real _rightGaugeHeight: Theme.geometry_briefPage_largeEdgeGauge_height / _rightGaugeCount
+	function _rightGaugeCount() {
+		return (acLoadGauge.active ? 1 : 0) + (dcLoadGauge.active ? 1 : 0)
```
jpetrell, line 39: In JavaScript could probably write acLoadGauge.active + dcLoadGauge.active as true + true = 2, true + false = 1 and false + false = 0, but not sure if that is more confusing or clear... :smile:

pages/BriefPage.qml
```diff
@@ -245,7 +257,7 @@ SwipeViewPage {
 
 				ArcGaugeQuantityRow {
 					id: dcInGaugeQuantity
-					alignment: root._leftGaugeCount === 2
+					alignment: root._leftGaugeCount() === 2
```
jpetrell, line 263: Personally I would prefer to see binding to a property than a function here.

pages/BriefPage.qml
```diff
@@ -248,7 +248,10 @@ SwipeViewPage {
 							// top, or is the first (top) gauge, so label aligns to the bottom.
 							? Qt.AlignLeft | (acInputGauge.active ? Qt.AlignTop : Qt.AlignBottom)
 							: Qt.AlignLeft| Qt.AlignVCenter
-					icon.source: root._inputsIconSource
+					icon.source: Global.dcInputs.model.count === 1
+							? VenusOS.dcInputIcon(Global.dcInputs.model.firstObject.source)
+							: "qrc:/images/icon_input_24.svg"
+
```
jpetrell, line 271: > Now that there are separate gauges for AC and DC input, there is no
> need to use a combined icon.

Is something still using the combined icon?

components/DynamicValueRange.qml
```diff
@@ -0,0 +1,48 @@
...
+		}
+
+		if (!isNaN(_max) && value >= _max) {
+			_valueAsRatio = 1
+			return
+		}
+		const ratio = (value - _min) / (_max - _min)
+		_valueAsRatio = ratio === Infinity ? 0 : ratio
+	}
+}
```
jpetrell, line 48: Ideally there were unit tests for such algorithmic functions.

DanielMcInnes (approved):
lgtm

pages/BriefPage.qml
blammit: Oh, it should be done for the AC input gauge as well. From the design it appears only the AC and DC loads should have warning/critical levels, but I'll check in slack.

pages/BriefPage.qml
blammit: Yes, the side panel -- that will need to be fixed later.

components/DynamicValueRange.qml
blammit: Yes, agreed... maybe this should be in C++ as well.

blammit:
> With the 3 gauge design there are now gauges without quantity labels. With dynamic range I am afraid it becomes difficult to reason with all that information, kind of wish I could tap the gauge group to see more info with concrete phase values shown somehow, somewhere (not sure how, below how you see they are shown on the Overview page).

AFAIK this is intentional, and you can go to the Overview page to get more concrete data. But maybe it'd be nice to have the individual phase data here too; wonder where it can be squeezed in without crowding things even more.

pages/BriefPage.qml
blammit: Yeah, this is a bit icky. Maybe it could be provided via the gauge parameters, since using the property binding was causing other issues where the sideGaugeParamters() function and the leftGaugeCount property had different values.

blammit:
Extracted the data-related changes to https://github.com/victronenergy/gui-v2/pull/948

jpetrell (approved):
Overall LGTM. Approving already, but would still like to get the text overlap issue fixed. While only happens on certain installations those users will regularly see overlapping text, which is not nice.

> If you have generator connected and connect solar the generator gauges will draw incorrectly for 1-2 seconds before correcting:

Not caused by this PR I guess and relatively minor, filed https://github.com/victronenergy/gui-v2/issues/954.

> Typo in one the commit messages "paramters"

Nitpick: Still there, see "paramters" on commit d751e46b611ea46788b6de302df81a24fec7b3ae Avoid jittery gauge effects when Brief gauge layout changes.

pages/BriefPage.qml
blammit: Clarified with Jesús -- the AC/DC loads need warning/critical colours (to warn of overload), but the AC/DC inputs and PV gauges do not (there's nothing wrong with high or low input).

pages/BriefPage.qml
blammit: Fixed

pages/BriefPage.qml
blammit: That's true -- I'd slightly prefer the more explicit version, though it does make it more wordy.

pages/BriefPage.qml
blammit: Fixed

blammit:
> There seems to be a notable risk of text overlap with the 3 gauges.

Will fix that separately, working on it now.

> In mock mode when changing to 15. and 16. the right gauges take multiple seconds to appear. Can that happen with the real-world data models? The delay is notable on desktop alone, wonder if gets worse on GX device.

This was because the mock data didn't update the AC/DC load data immediately. Fixed now.

Also, I've updated this to always show the AC load gauge, to be consistent with the Overview page, which does the same thing.

> Again not sure if can happen with real-world data, but the mock modes 8.-15. show "-- W" quantity for shore/grid, but the gauge next to the value label shows non-zero, animated values.

This was another issue in the mock data, where the power could be invalid even when the current was set. Fixed now.

PR #947, opened 2024-04-15
Overview connector fixes

blammit:


pages/OverviewPage.qml
```diff
@@ -146,11 +146,6 @@ SwipeViewPage {
 		for (let i = 0; i < widgets.length; ++i) {
 			const connectors = widgets[i].connectors
 			for (let j = 0; j < connectors.length; ++j) {
-				// For all left widgets except for solar (which has two connectors), straighten the
-				// connector if there is only one widget (as it will expand to full size).
-				if (widgets === _leftWidgets) {
-					connectors[j].straight = connectors.length === 1 && _leftWidgets.length === 1
```
DanielMcInnes, line 152: `WidgetConnector.straight` appears to be unused now

data/AcInputs.qml
```diff
@@ -44,13 +44,27 @@ QtObject {
...
 		}
 	}
 
+	function sourceValid(source) {
+		return source !== VenusOS.AcInputs_InputSource_NotAvailable && source !== VenusOS.AcInputs_InputSource_Inverting
+	}
+
+	function findValidSource() {
+		if (activeInput) {
+			return activeInput.source
```
DanielMcInnes, line 59: the return value of this function is only used when it is AcInputs_InputSource_NotAvailable. Might be clearer if the function was `bool hasValidSource()`?

DanielMcInnes (approved):
lgtm, nits

pages/OverviewPage.qml
blammit: Thanks, removed

data/AcInputs.qml
blammit: It was to accomodate changes in https://github.com/victronenergy/gui-v2/pull/948 that need `hasValidSource()`, but yeah probably best to keep this change separate. Have updated this to be `hasValidSource()`.

PR #950, opened 2024-04-16
Add overview bar gauges

blammit:


DanielMcInnes:
This bit looks funny

Screenshot from 2024-04-17 11-20-17

DanielMcInnes (approved):
Apart from the overlap on the Brief page, lgtm

components/widgets/InverterChargerWidget.qml
```diff
@@ -32,6 +32,7 @@ OverviewWidget {
 	type: VenusOS.OverviewWidget_Type_VeBusDevice
 	enabled: !!Global.inverterChargers.first
 	quantityLabel.visible: false
+	rightPadding: Theme.geometry_overviewPage_widget_sideGauge_margins
```
jpetrell, line 35: Can we count on the three phase bar gauge is always present and needed?

components/ThreePhaseDisplay.qml
```diff
@@ -19,10 +20,25 @@ Flow {
...
 		id: phaseRepeater
 
 		delegate: Item {
+			id: phaseDelegate
+
+			readonly property int valueStatus: root.phaseModelProperty
+					? Gauges.getValueStatus(valueRange.valueAsRatio * 100, root.valueType)
+					: Theme.Ok
+			readonly property color textColor: (valueStatus === Theme.Critical || valueStatus === Theme.Warning)
+					? Theme.statusColorValue(valueStatus)
```
jpetrell, line 39: Status color value is a bit generic name, but I guess the use case is also generic / can be applied widely to different values. Maybe Theme.valueStatusColor i.e. color of value status vs. color value of status?

jpetrell (approved):
LGTM. Really nice, cleanly structured code, couldn't really find issues. :blush:

blammit:
> This bit looks funny

It does -- I will fix it in https://github.com/victronenergy/gui-v2/pull/942 (Brief updates)

components/ThreePhaseDisplay.qml
blammit: Theme.valueStatusColor() sounds better, yes. Would be really nice to do an API review at some point. E.g. over the course of writing various ListItem components, I didn't know whether ListSwitch or ListSwitchItem was better anymore, and then all these other new list-based components popped up as well.

components/widgets/InverterChargerWidget.qml
blammit: The AC loads box is always present, so from that point of view, I think it's fine to always show this gauge.

PR #953, opened 2024-04-17
Usability improvements to Wi-Fi settings

### LINKED ISSUE #934: Implement few improvements to the Wi-Fi settings page
While rewriting GUI v2 network settings in #618 agreed with Serj to implement few straight-forward improvements to the Wi-Fi settings:
- Combine Wi-Fi access point and Wi-Fi networks settings pages to save one page navigation step when accessing the networks
- You cannot edit IP configuration in automatic mode, but the fields are readonly. Display the read-only IP address, netmask, gateway and DNS fields the same way as usual read-only list items, now they looked editable when they actually were not
- Display unavailable network values as "--" instead of displaying placeholder text "000.000.000.000", "--" is less confusing and aligns with the rest of the UI

jpetrell:
- Combine Wi-Fi access point and Wi-Fi networks settings pages
- Make it clear IP address, netmask, gateway and DNS are read-only in automatic mode
- Display unavailable network values as "--" instead of displaying placeholder text "000.000.000.000"
    
image

Fixes #934.


blammit:
@jpetrell is this meant to target https://github.com/victronenergy/gui-v2/pull/930 since it includes the network settings migration? Or are the venus-platform changes in the latest VenusOS now?

jpetrell:
> @jpetrell is this meant to target https://github.com/victronenergy/gui-v2/pull/930 since it includes the network settings migration?

It was part of the migration PR earlier, but it was discussed that no design changes should go to 1.0.0 so split it now to a separate PR. The migration PR should go in first, and once the main is open for design changes this could be integrated. 

> are the venus-platform changes in the latest VenusOS now?

AFAIK Not yet.

jpetrell:
Fixed merge conflict.

blammit (approved):
The settings page is way nicer now :) I think this should go in to 1.0.0.

I have mixed feelings about substituting the TextField with a Label when the TextField is not enabled, because then you lose the hint to the user that the TextField may be editable under different circumstances. It seems like it would be better to improve the design of the disabled TextField instead. For example, maybe the border could be a different colour, in addition to using a different text colour.

PR #959, opened 2024-04-17
Open notifications when receiving alarm inside a sub-page

### LINKED ISSUE #957: GUI v2 shows no visual indication of the alarms outside the top-level main views
If you have audible alarms setting disabled in General settings there is nothing indicating there is an alarm on the GUI v2 unless you happen to visit the Notifications page. Some features like the display off button on the top-right corner will be disabled, which may look like a bug if you don't realize what is going on. In comparison GUI v1 status bar displays a blinking alarm icon.

jpetrell:
Fixes #957.

### DIFF
--- a/ApplicationContent.qml
+++ b/ApplicationContent.qml
@@ -17,6 +17,8 @@ Item {
 			&& Global.notifications.alarm
 	on_GoToNotificationsChanged: {
 		if (_goToNotifications) {
+			pageManager.popAllPages()
+			mainView.controlsActive = false
 			pageManager.navBar.setCurrentPage("NotificationsPage.qml")
 		}
 	}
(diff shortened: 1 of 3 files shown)


blammit (commented):
- Works nicely from the setting sub-pages.

- In mock mode, if I go to the Brief page and press Shift+N to generate an alarm, it goes to the Notification page as expected, but I get:

`qrc:/qt/qml/Victron/VenusOS/components/PageStack.qml:146: TypeError: Cannot call method 'destroy' of null`


- Would be nice if you could add a change to do the same thing if an alarm is triggered while the control cards are open, while you're at it?

jpetrell:
Found another potential issue, filed a separate #979.

blammit (approved):

Issue #936, opened 2024-04-11
Cannot select an item on the radio button list if you tap the rounded indicator


jpetrell:
Open any radio button list item (easiest to test is Settings -> Debug -> Radio button examples). Select by tapping the circular radio button indicator.

Actual result: If you tap the radio button list item background, the item gets selected. If you tap the rounded radio button list indicator, nothing happens.

[Screencast from 2024-04-11 16-05-07.webm](https://github.com/victronenergy/gui-v2/assets/2203667/18777e7a-d426-4db2-b712-bf8286d2a5e9)

Seems the radio button indicator doesn't like enabled binding disabling the touch in the middle of the interaction:

```
enabled: !checked && (Array.isArray(root.optionModel)
    ? !modelData.readOnly
    : !model.readOnly)
```
From components/listitems/ListRadioButtonGroup.qml

PR #962, opened 2024-04-17
Fix list switch toggling directly on the switch item

### LINKED ISSUE #955: Toggling settings switch requires two taps when done on top of the switch
Go to Display & Language settings. Turn adaptive brightness on or off. Go to General settings.

Expected result: The setting follows the switch state.

Actual result: Need to tap twice to make the change (see the video below).

Potentially affects more settings (there are 73 ListSwitches in GUI v2) since there is nothing particular special about the code:
```
ListSwitch {
  ..
}
```

Can also reproduce on Audible alarm setting, tapping the switch once doesn't change the actual setting.

Tested on Ekrano GX.

https://github.com/victronenergy/gui-v2/assets/2203667/e3f79566-da23-42bf-9021-897287669265

jpetrell:
Fixes #955.

### DIFF
--- a/components/listitems/ListSwitch.qml
+++ b/components/listitems/ListSwitch.qml
@@ -48,6 +48,7 @@ ListItem {
 		Switch {
 			id: switchItem
 			checked: invertSourceValue ? dataItem.value === valueFalse : dataItem.value === valueTrue
+			checkable: false
 			onClicked: root._setChecked(!checked)
 		}
 	]


components/listitems/ListSwitch.qml
```diff
@@ -48,7 +48,7 @@ ListItem {
 		Switch {
 			id: switchItem
 			checked: invertSourceValue ? dataItem.value === valueFalse : dataItem.value === valueTrue
-			onClicked: root._setChecked(!checked)
```
chriadam, line 51: The original code seems correct.  Do we need to specify `checkable: false` in the base type, to avoid its check state switching automatically on click, perhaps?

jpetrell: Well strictly speaking the original code was broken, but yeah you need to know that by default Qt Quick Control switches have already updated checked state before clicked emitted, which may not be clear. But agree `checkable: false` is better as then the clicked signal handlers between the PressArea and Switch align, and it is nice if the state only changes after the actual value updates. Updated.

chriadam: According to docs, https://doc.qt.io/qt-6/qml-qtquick-controls-abstractbutton.html#checkable-prop checkable is false by default.  I know that in Qt 6.2 and earlier, setting checked also affected checkable, but it should not be the case in 6.5 or 6.6.  But, given that the code was originally written for Qt 6.2.x you're right, it was broken originally ;-)

Can the `checkable: false` be set explicitly on the base type rather than here in ListSwitch?  Next question is: why does this now fix it, given that `checkable: false` should be the default in Qt 6.5.x / Qt 6.6.x?

chriadam: (Also, I assume the same fix will be needed for RadioButton and all of the other button-like controls?)

jpetrell:
>  Next question is: why does this now fix it, given that checkable: false should be the default in Qt 6.5.x / Qt 6.6.x?

Looking at the code the Switch and RadioButton have always had checkable true, and the base class AbstractButton has always had checkable false since [Checkable was merged to AbstractButton](https://github.com/qt/qtdeclarative/commit/97cd1782744f45f1c9c249bd5549dd7f1ea53b3c) in 2015 . But agree the documentation is confusing. :see_no_evil:

components/listitems/ListSwitch.qml
jpetrell: >  Next question is: why does this now fix it, given that checkable: false should be the default in Qt 6.5.x / Qt 6.6.x?

Looking at the code the Switch and RadioButton have always had checkable true, and the base class AbstractButton has always had checkable false since [Checkable was merged to AbstractButton](https://github.com/qt/qtdeclarative/commit/97cd1782744f45f1c9c249bd5549dd7f1ea53b3c) in 2015 . But agree the documentation is confusing. :see_no_evil:

chriadam: Ah!  I see!  I didn't realise Switch and RadioButton overrode the default!

jpetrell: > Can the checkable: false be set explicitly on the base type rather than here in ListSwitch?

`checkable: true` is good default for the regular Switch? `checkable: false` is good if the toggling may fail and you want explicit control. in comparison ListSwitch is not really an UI component, but a widget that combines the layout and model (VeQuickItem). But I don't have strong opinion on the matter, both approaches work, can also disable checkable for all switches if you want?

> (Also, I assume the same fix will be needed for RadioButton and all of the other button-like controls?)

I tested radio buttons yesterday, didn't see issues and I think the code there looks ok. There was a issue #936 with radio buttons I fixed earlier that looked very similar on the surface, but was related to enabled property and not checkability.

chriadam: My view is that checkable should almost always be false.  In general, the checked state of a switch or radio button should be bound to some canonical system state (exposed through the applicatoin data model), and upon click the appropriate system state mutator should be called.  If it succeeds, the system state will change, and the checked binding will cause the switch to display the appropriate new state.

The alternative (that the ui element toggles its switch, and then you attempt to change the system state based upon the new UI state) seems backwards to me (e.g. what if the system state mutation fails, or is rejected?).

But, perhaps I am wrong or overlooking something important.

Last Friday I found a similar issue in radio buttons too, see https://github.com/victronenergy/gui-v2/pull/969/commits/cc733013c170f29f03478c25686f5fe547d689fe

jpetrell: > But, perhaps I am wrong or overlooking something important.

No, I agree in general it is better to bind the UI status to actual model state (unless e.g. the model updates really slowly or doesn't provide a convenient property to bind to). In practice most settings work reliably and don't include extra confirmation steps so probably could implement use cases in either way, I've just been wary of making behavioral changes as then I need to go through and verify 86 radio button groups, 73 list switches, 3 switch control values and 2 radio button control values against regression.

I will now change all the radio button and switch components to default to `checkable=false`. It is nice to make all the checkable components behave consistently. :blush: 

> Last Friday I found a similar issue in radio buttons too, see https://github.com/victronenergy/gui-v2/commit/cc733013c170f29f03478c25686f5fe547d689fe

The code you removed tried to keep the old behavior of only emitting clicked signal when the radio button is not checked. The press effect below still behaves like that so needs to be also adapted.

jpetrell: Here is an alternative PR with the more intrusive behavioral changes https://github.com/victronenergy/gui-v2/pull/980

I tried to test the 164 use cases as well as I could, but it is hard to be sure no regression happens.

chriadam: Great work, thanks for that.

I suggest we merge this PR 962 now for 1.0.0, and get 980 in after the release - do you agree?

chriadam (approved):

components/listitems/ListSwitch.qml
jpetrell: > I suggest we merge this PR 962 now for 1.0.0, and get 980 in after the release - do you agree?

Sounds good. I'll merge this one, and keep testing the #980 a bit more, it is good to verify GUI v2 works reliably when wasm and device build usage are simultaneously used.

PR #964, opened 2024-04-18
Fix memory leak

### LINKED ISSUE #949: GUI v2 seems to be leaking memory
Initially venus-gui-v2 takes ~80MB of unique memory, but can grow to consume most of the memory (e.g. 953m or 95%) before forced reboot.

DanielMcInnes:


chriadam (dismissed):
LGTM, although I doubt the engine calls instance() more than once to instantiate the singleton.  See e.g. https://doc.qt.io/qt-6/qml-singleton.html#defining-singletons-in-c example where the instance() method returns a new instance.

chriadam:
Yikes!  Good catch, thanks.

src/units.cpp
```diff
@@ -40,7 +40,8 @@ namespace Units {
 
 QObject* Units::instance(QQmlEngine *, QJSEngine *)
 {
-	return new Units;
+	static QObject* units = new Units();
```
chriadam, line 43: it looks line we don't actually need the engine ptr, so the Units type should be default constructible.  See https://doc.qt.io/qt-6/qml-singleton.html#defining-singletons-in-c - I suggest we get rid of the instance method, if possible.

chriadam: Ah, it's for QuantityInfo which calls instance() manually.

chriadam: style nitpick: prefer just `static QObject *units = new Units;  return units;`

chriadam (approved):
One style nitpick but LGTM.

src/units.cpp
DanielMcInnes: Done. OOI, Why do we prefer:
`static QObject *units = new Units; `
over:
`static QObject* units = new Units();`
? Google won't tell me

PR #965, opened 2024-04-18
Various brief gauge fixes

blammit:


blammit:
Oh, I forgot to adjust the central gauge text size so that it doesn't overlap with the AC gauge label so easily. I'll do another PR for that.

data/SystemSettings.qml
```diff
@@ -104,6 +104,10 @@ QtObject {
 		}
 	}
 
+	function essFeedbackToGridEnabled() {
+		return _hubSetting.value === 4 && _overvoltageFeedIn.value === 1 && _preventFeedback.value === 0
```
DanielMcInnes, line 108: Can we use an enum instead of '4' for `_hubSetting.value === 4` ?

pages/BriefPage.qml
```diff
@@ -351,7 +370,7 @@ SwipeViewPage {
 
 				x: -root._gaugeArcMargin
 				opacity: root._gaugeArcOpacity
-				animationEnabled: root.animationEnabled
+				animationEnabled: false//root.animationEnabled && !pauseRightGaugeAnimations.running
```
DanielMcInnes, line 373: intentional?

DanielMcInnes (approved):
lgtm, nits

pages/BriefPage.qml
blammit: Oops! Fixed, thank you

data/SystemSettings.qml
blammit: That would be nicer, but I don't know what the magic number means :-(

pages/BriefPage.qml
```diff
@@ -408,6 +427,16 @@ SwipeViewPage {
...
 	}
 
+	Timer {
+		id: pauseLeftGaugeAnimations
+		interval: Theme.animation_progressArc_duration
+	}
+
+	Timer {
+		id: pauseRightGaugeAnimations
+		interval: Theme.animation_progressArc_duration
```
jpetrell, line 437: Ideally this would be bound to the actual animation state instead of separate timers, but I guess timers are bound to the same heartbeat so LGTM.

PR #966, opened 2024-04-18
Don't require QtVirtualKeyboard to be installed

chriadam:
Some platforms (desktop and WebAssembly) don't require the QtQuick.VirtualKeyboard module to be installed, as they either have hardware keyboards (desktop) or native platform VKB support (iOS/Android).

It is only on CerboGX/EkranoGX that we require QtVirtualKeyboard.

As such, don't declare a component whose type requires the existence of the QtVirtualKeyboard module.  Instead, load it at runtime via a source url, and only on Linux platform.

This does mean that in order to edit the InputPanel.qml file, the user also needs to edit the ApplicationContent.qml file to load it from disk rather than from qrc.

### DIFF
--- a/ApplicationContent.qml
+++ b/ApplicationContent.qml
@@ -84,19 +84,22 @@ Item {
 		Component.onCompleted: Global.notificationLayer = notificationLayer
 	}
 
+	// We only want the VKB on CerboGX/EkranoGX devices.
+	// WebAssembly should use the native platform VKB.
+	// Desktop platforms should use hardware keyboard.
 	// Create the InputPanel dynamically in case QtQuick.VirtualKeyboard is not available (e.g. on
 	// Qt for WebAssembly due to QTBUG-104109).
 	// Note the VKB layer is the top-most layer, to allow the idleModeMouseArea beneath to call
 	// testCloseOnClick() when clicking outside of the focused text field, to auto-close the VKB.
-
 	Loader {
 		id: loader
 
 		asynchronous: true
-		active: Qt.platform.os !== "wasm"
-		sourceComponent: InputPanel {
-			mainViewItem: mainView
+		active: Qt.platform.os == "linux" && !Global.isDesktop
+		source: "qrc:/qt/qml/Victron/VenusOS/components/InputPanel.qml"
+		onLoaded: {
+			item.mainViewItem = mainView
+			Global.inputPanel = item
 		}
-		onLoaded: Global.inputPanel = item
 	}
 }
--- a/components/InputPanel.qml
+++ b/components/InputPanel.qml
@@ -7,6 +7,11 @@ import QtQuick
 import QtQuick.VirtualKeyboard as QtVirtualKeyboard
 import Victron.VenusOS
 
+// *** This file can be edited directly on the cerbo filesystem,
+// *** but you will also need to edit ApplicationContent.qml
+// *** so that the loader's source property is:
+// *** "file:///opt/victronenergy/gui-v2/Victron/VenusOS/components/InputPanel.qml"
+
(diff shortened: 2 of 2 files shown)


chriadam:
Oops, unrelated changes in here, accidentally git added.  Will remove.

chriadam:
Fixed.

ApplicationContent.qml
```diff
@@ -84,19 +84,22 @@ Item {
...
 		id: loader
 
 		asynchronous: true
-		active: Qt.platform.os !== "wasm"
-		sourceComponent: InputPanel {
-			mainViewItem: mainView
+		active: Qt.platform.os == "linux"
+		source: "qrc:/qt/qml/Victron/VenusOS/components/InputPanel.qml"
+		onLoaded: {
+			item.mainViewItem = mainView
```
DanielMcInnes, line 101: it seems odd to set 'mainViewItem' like this. Should this be done in InputPanel?

ApplicationContent.qml
```diff
@@ -84,19 +84,22 @@ Item {
...
-
 	Loader {
 		id: loader
 
 		asynchronous: true
-		active: Qt.platform.os !== "wasm"
-		sourceComponent: InputPanel {
-			mainViewItem: mainView
+		active: Qt.platform.os == "linux"
+		source: "qrc:/qt/qml/Victron/VenusOS/components/InputPanel.qml"
```
DanielMcInnes, line 99: I think this will defeat the 'load components from the file system' functionality. Prefer: sourceComponent: InputPanel {}

chriadam: Basically, the current code means that if QtVirtualKeyboard plugin isn't installed, the application crashes on startup.

Note, for example, that the github action which produces the official WebAssembly build of gui-v2 does not install QtVirtualKeyboard plugin (as QtVKB isn't supposed to be used on WebAssembly, but instead the platform native VKB should be used instead).

DanielMcInnes: Does my suggested code crash for you? I had hoped it wouldn't go looking for InputPanel unless Qt.platform.os == "linux".

chriadam: Yes, it crashes.  Your suggested code is the current code in main.  The compiler still needs to compile the type assigned to the sourceComponent property, regardless of the (runtime) value of the expression assigned to the active property.

DanielMcInnes: Maybe a comment at the top of InputPanel.qml mentioning that it can't be edited on the filesystem?

DanielMcInnes (approved):

ApplicationContent.qml
chriadam: Good point, will do

chriadam: Done.

DanielMcInnes: > Please read the commit message.

Ah, I understand this now, I was looking at the commit message in github, and only saw the first line

jpetrell (approved):

ApplicationContent.qml
```diff
@@ -84,19 +84,22 @@ Item {
...
 	// testCloseOnClick() when clicking outside of the focused text field, to auto-close the VKB.
-
 	Loader {
 		id: loader
 
 		asynchronous: true
-		active: Qt.platform.os !== "wasm"
-		sourceComponent: InputPanel {
-			mainViewItem: mainView
+		active: Qt.platform.os == "linux"
```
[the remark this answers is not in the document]
chriadam: Good point, will do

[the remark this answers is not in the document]
chriadam: done, thanks

PR #969, opened 2024-04-18
Add warning dialog when enabling gui-v2 via VRM

chriadam:


chriadam:
Don't merge, may not be necessary for 1.0.0.

I'm not 100% about the second commit (it seems to solve my issue, but I couldn't really understand the purpose of the change in #936 so I might have broken something here).

chriadam:
(Is it expected that v1 of gui will always be the thing which provides the VRM Remote Console service?)

PR #970, opened 2024-04-18
Brief: prevent AC input gauge label from overlapping with center text

blammit:
Adjust the font sizes and margins, and elide the center gauge labels if necessary, to prevent the text overlap. This is particularly a problem when the center gauge is configured to shows the name, percentage and icon.

### DIFF
--- a/data/mock/config/MockDataSimulator.qml
+++ b/data/mock/config/MockDataSimulator.qml
@@ -198,6 +198,22 @@ QtObject {
 			event.accepted = true
 			break
 		case Qt.Key_U:
+			// Change the unit display of the Brief view center gauges
+			if (event.modifiers & Qt.ControlModifier) {
+				const v = root.mockValue(Global.systemSettings.serviceUid + "/Settings/Gui/BriefView/Unit")
+				let newBriefUnit = ""
+				if (v === VenusOS.BriefView_Unit_None) {
+					newBriefUnit = VenusOS.BriefView_Unit_Absolute
+				} else if (v === VenusOS.BriefView_Unit_Absolute) {
+					newBriefUnit = VenusOS.BriefView_Unit_Percentage
+				} else {
+					newBriefUnit = VenusOS.BriefView_Unit_None
+				}
+				root.setMockValue(Global.systemSettings.serviceUid + "/Settings/Gui/BriefView/Unit", newBriefUnit)
+				return
+			}
+
+			// Change the system unit
 			Global.systemSettings.setElectricalQuantity(
 					Global.systemSettings.electricalQuantity === VenusOS.Units_Watt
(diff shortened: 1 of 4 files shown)


blammit:
Text elides like this if necessary for 5":

image

And for 7":

image

themes/geometry/SevenInch.json
```diff
@@ -169,7 +171,7 @@
     "geometry_briefPage_edgeGauge_spacing": 44,
     "geometry_briefPage_edgeGauge_spacingAngle": 4,
     "geometry_briefPage_edgeGauge_quantityLabel_spacing": 2,
-    "geometry_briefPage_edgeGauge_quantityLabel_horizontalMargin": 54,
+    "geometry_briefPage_edgeGauge_quantityLabel_horizontalMargin": 60,
```
jpetrell, line 174: Its this accidentally here? It is not touched on the five inch layouts? Also conflicts with same line change in #965.

data/mock/config/MockDataSimulator.qml
```diff
@@ -198,6 +198,22 @@ QtObject {
...
+				if (v === VenusOS.BriefView_Unit_None) {
+					newBriefUnit = VenusOS.BriefView_Unit_Absolute
+				} else if (v === VenusOS.BriefView_Unit_Absolute) {
+					newBriefUnit = VenusOS.BriefView_Unit_Percentage
+				} else {
+					newBriefUnit = VenusOS.BriefView_Unit_None
+				}
+				root.setMockValue(Global.systemSettings.serviceUid + "/Settings/Gui/BriefView/Unit", newBriefUnit)
+				return
+			}
```
jpetrell, line 214: Nice! :muscle:

jpetrell:
Serj commented that ideally we would only move the arrow above in the rare problematic case.

When there are two gauges on the left side there should not be risk of overlapping text?

image

blammit:
> Serj commented that ideally we would only move the arrow above in the rare problematic case.
> When there are two gauges on the left side there should not be risk of overlapping text?

In the design, it appears that the spacing between the side gauge and its icon/label are the same regardless of the number of whether it is a single or 3-phase gauge, and regardless of the number of gauges on that side. I'm not sure of the limits we should test for, but if this is true, then if we move the arrow to between the gauge icon and label, then there would still be a risk of overlap in this last screenshot, if it was something like "-15.35kW" or larger, for example, and the gauge was "Hydraulic oil" or similarly longer text.

themes/geometry/SevenInch.json
blammit: Yes, it's true -- this was deliberate, as I found the extra pixels I added in the other patch just creates too much space in the 7" version.

PR #974, opened 2024-04-19
Drop confirmation toast

jpetrell:
Fixes warning "Could not set initial property category"..

Notification type Notification_Confirmation got dropped by recent notification page design changes and I don't see anything using it either. Alternatively could re-instate the enumeration.

Contributes to #975


blammit (approved):
Nice to clean this up, thanks!

PR #977, opened 2024-04-22
Add green bars for AC inputs on Brief and Overview page

blammit:


components/SideGauge.qml
```diff
@@ -5,17 +5,22 @@
...
+	property int valueType: VenusOS.Gauges_ValueType_NeutralPercentage
+	readonly property int valueStatus: Gauges.getValueStatus(value, valueType)
 
 	width: parent.width
 	height: parent.height
 	radius: Theme.geometry_briefPage_edgeGauge_radius
 	useLargeArc: false
 	strokeWidth: Theme.geometry_arc_strokeWidth
+	progressColor: Theme.statusColorValue(valueStatus)
+	remainderColor: Theme.statusColorValue(valueStatus, true)
```
jpetrell, line 23: Nitpick: just reading this code line it is hard to know what the second parameter `true` does.

jpetrell (approved):
LGTM

components/SideGauge.qml
blammit: Agreed, we should revisit that API. Will consider that separately.

PR #980, opened 2024-04-22
Improve switches and radio buttons

### LINKED ISSUE #955: Toggling settings switch requires two taps when done on top of the switch

jpetrell:
- Fix list switch toggling directly on the switch item
- Don't update checked status until the data item updates accordingly
- Avoid nuking the checked bindings (so if somebody updates the value remotely they still update)
- Set update updateOnClick=false if you don't want automatic behavior
- Earlier ListSwitch updated the checked status twice, once by Switch checkable, secondly by the select() function (latter nuked any app side binding)
- Tapping the current item now dismisses the radio button page, but still doesn't emit optionClicked()
- Expose checkable on the component API, many vanilla Switches and RadioButton assumed onCheckedChanged

Also, changes the default behavior of switches and radio buttons to not automatically toggle the checked state.

Fixes #955.

### DIFF
--- a/components/listitems/ListRadioButtonGroup.qml
+++ b/components/listitems/ListRadioButtonGroup.qml
@@ -81,14 +81,12 @@ ListNavigationItem {
 					id: radioButton
 
 					function select(password) {
-						if (root.updateOnClick) {
-							if (dataItem.uid.length > 0) {
+						if (!checked) {
+							if (root.updateOnClick && dataItem.uid.length > 0) {
 								dataItem.setValue(Array.isArray(root.optionModel) ? modelData.value : model.value)
-							} else {
-								root.currentIndex = model.index
 							}
+							root.optionClicked(model.index, password)
 						}
-						root.optionClicked(model.index, password)
 
 						if (root.popDestination !== undefined) {
 							popTimer.restart()
--- a/components/listitems/ListRadioButtonGroup.qml
+++ b/components/listitems/ListRadioButtonGroup.qml
@@ -36,7 +36,7 @@ ListNavigationItem {
 			? optionModel[currentIndex].value
 			: undefined
 
-	property bool updateOnClick: true
+	property bool updateDataOnClick: true
 	property var popDestination: null   // if undefined, will not automatically pop page when value is selected
 
 	property int defaultIndex: -1
@@ -82,7 +82,7 @@ ListNavigationItem {
 
 					function select(password) {
 						if (!checked) {
-							if (root.updateOnClick && dataItem.uid.length > 0) {
+							if (root.updateDataOnClick && dataItem.uid.length > 0) {
 								dataItem.setValue(Array.isArray(root.optionModel) ? modelData.value : model.value)
 							}
 							root.optionClicked(model.index, password)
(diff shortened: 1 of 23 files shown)


components/listitems/ListRadioButtonGroup.qml
```diff
@@ -80,14 +80,15 @@ ListNavigationItem {
 					id: radioButton
 
 					function select() {
-						if (root.updateOnClick) {
-							if (dataItem.uid.length > 0) {
-								dataItem.setValue(Array.isArray(root.optionModel) ? modelData.value : model.value)
-							} else {
-								root.currentIndex = model.index
```
blammit, line 87: Will removing this affect any radio button groups that don't set `dataItem.uid` and also do not change `currentIndex` in `onOptionClicked`? In these cases, I would expect that `currentIndex` doesn't update anymore. I can't find any instances where we do this, but just wondering if you found any.

blammit: Never mind, I see you found at least one of them, in PageVeBusAdvanced :)

blammit: With this change, I think it means that if a ListRadioButtonGroup does not have a `dataItem.uid`, it no longer needs to set `updateOnClick=false`. Is that right? E.g. in DvccCommonSettings, the `updateOnClick=false` has no effect now, since it only applies if `dataItem.uid` is set.

Also, could do `if (root.updateOnClick && dataItem.uid.length > 0)` instead of having two conditions here.

blammit (commented):
Feels strange to have a Switch that's not checkable by default, but makes sense given the way we use it.

Is `updateOnClick` necessary anymore? Wondering if it basically means the same thing as `checkable`, at least for the ListSwitch case.

blammit:
^ @jpetrell ping, in case you haven't seen these comments

blammit:
@jpetrell seems like some of this would be nice to have for 1.0.0? E.g. being able to click on switch toggles.

blammit:
I'll take over this one and look into integrating the changes. My understanding is that the main improvement/fix is to set checkable=false for buttons, switches and radio buttons.

components/listitems/ListRadioButtonGroup.qml
blammit: Added a patch to remove unnecessary `updateOnClick` bindings.

blammit:
I've rebased Joona's original patch and added some other commits. Ready for review.

DanielMcInnes (approved):
lgtm

Issue #532, opened 2023-11-14
Improve calculations of min/max values for AC input/loads gauges


blammit:
AC gauges should have meaningful minimum and maximum values where possible. This applies to:
- AC gauge and AC loads gauge on Brief page (i.e. top-left and top-right gauges)
- AC gauge and AC loads gauge in Brief side panel (i.e. the two bottom gauges). Note these are combined with DC values if they are available.
- AC gauge in Overview page (i.e. the right-edge vertical gauge for Grid, Shore, etc)
- AC loads gauge in Overview page (i.e. the right-edge vertical gauge for Inverter/Charger box, not yet implemented)

Some questions/complications:
- What to do when no maximum is available (currently in this case, the UI shows the max value seen during runtime)
- Should the min value seen during runtime be used as the minimum gauge value? (This is what gui-v2 uses at present)
- How to indicate progress/maximum value on multi-phase systems, since current design shows a single line for all phases combined
- How to show these values for arc gauges (Brief page) and straight-line gauges (Brief side panel, Overview boxes)? Should they be different?

Some fixes that can be done as a starting point:

- Use CurrentLimit for max value of AC gauge even if showing watts. Also for Overview page (side gauge for Shore, Grid etc)
- Use NominalInverterPower for max value of Loads bar (if only showing AC loads, otherwise use runtime max)
- Use NominalInverterPower for max value of vertical gauge bar for Inverter/Charger overview

## Subtasks ##

There are a number of sub-tasks to be done here, but discussed that it'll be easier to keep it all in this issue instead of making new ones. Here is a list of subtasks:

- General: (DONE)

  - ~Use fixed min/max values from backend~ https://github.com/victronenergy/gui-v2/pull/948
  - ~Add settings page to change `/Automax` and other configuration values~ https://github.com/victronenergy/gui-v2/pull/981 

- Brief page side gauges: (DONE)

  - ~Split AC+DC input arc gauge into separate AC + DC gauges~   https://github.com/victronenergy/gui-v2/pull/929
  - ~Use 3-phase arc gauge for AC input with fixed min/max~ https://github.com/victronenergy/gui-v2/pull/942
  - ~Use 3-phase arc gauge for AC load with fixed max~ https://github.com/victronenergy/gui-v2/pull/942
  -  ~Use fixed max for solar gauge~ https://github.com/victronenergy/gui-v2/pull/942

- Overview page: (DONE)

[...]

PR #981, opened 2024-04-23
Min/max gauge settings

blammit:


blammit:
If you change the min/max values in the settings, you should see the min/max range change in the gauges on the Brief and Overview page.

jpetrell (commented):
How do you get the min/max settings on the device? Quantity labels show values, but the gauges are empty.

jpetrell:
Some UX musings (maybe for the next version?):
- If I was a boat, RV or home owner with Victron installation I would probably want to define sane ranges for the gauges, but not sure I would have skills to actually do it with the current settings. I have math and engineering background, but I am no electrical engineer
- For one there is very little handholding, the terminology and layouts are quite different between the Brief, Overview and Min/Max gauge settings
- System setup settings page is able to map AC input 1 to Generator and AC input 2 to Shore power, could we show those titles somewhere?
- One reason why the settings appears complex is because the AC, current, min, max, 0, and 1 labels are repeated so many times
- Ideally the settings would somehow adapt/highlight the current configuration (e.g. if you don't have solar the "Solar - PV: Max power" could be dimmed)

components/listitems/ListNumberField.qml
```diff
@@ -0,0 +1,14 @@
...
+*/
+
+import QtQuick
+import Victron.VenusOS
+
+ListTextField {
+	id: root
+
+	textField.validator: RegularExpressionValidator { regularExpression: /\-?[0-9\.]+/ }
+	textField.inputMethodHints: Qt.ImhDigitsOnly
```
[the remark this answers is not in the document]
blammit: Thanks for filing the task, that looks great.

>  because they block the whole string input if the string has even one invalid character

I agree if it's a more complex validator, but this seems okay if it's just to validate numbers? 

We could show a toast notification here if it's invalid -- would be nicer than showing an extra label, which forces a change in the layout.

data/DcInputs.qml
```diff
@@ -138,7 +138,7 @@ QtObject {
 	}
 
 	readonly property VeQuickItem _maximumPower: VeQuickItem {
-		uid: Global.system.serviceUid + "/Dc/Input/Power/Max"
+		uid: Global.systemSettings.serviceUid + "/Settings/Gui/Gauges/Dc/Input/Power/Max"
```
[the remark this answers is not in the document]
blammit: Discussed with Rein that systemcalc will use the other one as the main "system" reference, and that gui-v2 can just use this /Gui setting value for reading/writing the user-loaded value.

pages/settings/PageSettingsDisplay.qml
```diff
@@ -70,6 +70,14 @@ Page {
 				}
 			}
 
+			ListNavigationItem {
+				//% "Minimum and maximum gauge ranges"
+				text: qsTrId("settings_display_minmax")
+				onClicked: {
+					Global.pageManager.pushPage("/pages/settings/PageSettingsDisplayMinMax.qml", {"title": text})
```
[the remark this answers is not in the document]
blammit: Yeah, it's true that would be nice, but it does cause a lot of string duplication, and then we would have to update the title in multiple places if it changed.

blammit:
> System setup settings page is able to map AC input 1 to Generator and AC input 2 to Shore power, could we show those titles somewhere?

We could show this, but the min/max applies regardless of whether they are connected to generator, shore, grid etc. Would it confuse things if we showed these titles, and the device input source changed?

> One reason why the settings appears complex is because the AC, current, min, max, 0, and 1 labels are repeated so many times

I agree -- would like to have a solution that simplified this without stretching the UI to multiple pages, and without having multiple text fields in one line. What do you think?

> Ideally the settings would somehow adapt/highlight the current configuration (e.g. if you don't have solar the "Solar - PV: Max power" could be dimmed)

I wonder how best to do this -- e.g. in settings pages the Relays options are not displayed if they are not connected, so we don't have a precedence for dimming as an indication of something not being connected.

blammit:
> How do you get the min/max settings on the device? Quantity labels show values, but the gauges are empty.

Have you updated to the latest firmware version? If so, can you check if the settings are available under com.victronenergy.settings/Settings/Gui/Gauges ?

pages/settings/PageSettingsDisplayMinMax.qml
```diff
@@ -0,0 +1,95 @@
...
+					delegate: Column {
+						required property int index
+						width: parent ? parent.width : 0
+
+						ListNumberField {
+							//: %1 = AC input number (0 or 1)
+							//% "AC in %1: Min current"
+							text: qsTrId("settings_minmax_ac_in_min_current").arg(index)
+							dataItem.uid: Global.systemSettings.serviceUid + "/Settings/Gui/Gauges/Ac/In/%1/Current/Min".arg(index)
+						}
```
jpetrell, line 49: I guess most realize the current is in amperes and power in watts, but ideally would good to see the quantity you are adjusting to speed up reasoning and clarify the context. Seems that before the text fields in GUI v2 have been used for device names, Bluetooth PIN codes, Wi-Fi passphrases, IP addresses, etc. , but not that much for adjusting electrical quantities. Most quantities have been adjusted using spin boxes, and often with fixed number of steps.

There is no comfirmation step so you lose the previous value when you start typing. Ideally there was a way to reset any manual adjustment to the automatic value that derives from the measurement history.

jpetrell:
> Have you updated to the latest firmware version?

I am running on v3.40~2. Tried the latest available 20240422090512 v3.31 earlier, but reverted back.

> can you check if the settings are available under com.victronenergy.settings/Settings/Gui/Gauges

I have some /Gui/* settings there, but no /Gui/Gauges/*.

blammit:
> > Have you updated to the latest firmware version?
> 
> I am running on v3.40~2. Tried the latest available 20240422090512 v3.31 earlier, but reverted back.

Could you try with v3.40~6?

> Very good point. I wonder how often do those change in practice? If you go through System setup and pick the right ones during the installation maybe not that often? Could also write "AC 1" + "Currently connected to Generator" if just "AC 1" + "Generator" is seen risky.
> If you change AC input connection to a new source shouldn't the system reset the min/max configs?

I'm not sure about this one. @ReinvdZee does the min/max configuration get reset if the source changes?

pages/settings/PageSettingsDisplayMinMax.qml
blammit: Yes, it might be nice to have "W" or "A" as a suffix in the text field? But I'm not sure about adding this here, as elsewhere I think we just use the unit type in the description (e.g. "Maximum voltage"). Maybe this could be considered as part of the broader Settings redesign instead.

jpetrell: > Maybe this could be considered as part of the broader Settings redesign instead.

Yeah definitely not for 1.0.

jpetrell:
> Could you try with v3.40~6?

Hmm... why doesn't the device offer me the latest versions any more. I must have changed some setting accidentally.

ReinvdZee:
>I'm not sure about this one. @ReinvdZee does the min/max configuration get reset if the source changes?

No, we only keep gauge limits for all AC scenarios (AC1, AC2, none) separately, but the gauge limit for a scenario is not reset when the connected source changes.

pages/settings/PageSettingsDisplayMinMax.qml
blammit: I'm in favour of adding these. E.g. in this case it would clarify whether power was in W vs kW if it's usually a big number.

jpetrell (approved):
LGTM. Reset could have confirmation to shield it from accidental presses, but discussed that it can be done later.

jpetrell:
>  Reset could have confirmation to shield it from accidental presses, 

Filed #990. Also noticed that #991 keyboard opens on top of the solar max power field.

kwindrem:
Regarding resetting max/min when AC input type changes:

I thought the AC input max was to be based on the AC input current limit. If so, the gauge should track that change and not require any special handling or even menu items to change the max value for the gauge at all.

If separate gauge max values for AC input are used,  feel it is wrong to reset them when switching the type. If the value is reset, then the gauge will show maximum values when in fact load might be small until the system is stressed to create realistic max values.

The one AC input to Multi inverter/chargers used on RVs, boats, etc are often switched between shore and generator power.Currently, this requires the user to switch the AC input type and the AC input current limit manually. 

GuiMods includes a mechanism to make these changes automatically for the user when a transfer switch change occurs. (The GX device is signaled via a digital input.) I will expand the external transfer switch logic to handle changing gauge max values if that ends up being needed, but again it seems like the AC input current limit is the right value to use for the AC input gauge's maximum.

mr-manuel:
Hello,

the AC output settings are a bit hard to understand. Wouldn't it better, if they are summed automatically? Like there should be only one AC output Maximum power field of the MultiPlus/Quattro. If AC In 1 and/or 2 is connected, then this values are added. Maybe selectable by the user, if this values should sum up or checked, if power assist or ESS is enabled.

Do I understand right, that the AC output Maximum power should be used for the AC Loads? In my case it's unfortunately not working. Probably because I have AC coupled PV Inverter (connected to AC Out) and another AC output Maximum power field would be required to add this value. I tried to play around with the values, but the max value of the AC Loads did not change. Maybe I'm missing something.

Another thing that is a bit tricky, is that some ranges are for current and some for power. Wouldn't it be easier to make all the same?

I tested with Venus OS `3.40~10` and GUIv2 `v0.2.17`.

Bildschirmfoto am 2024-04-27 um 21 09 08
Bildschirmfoto am 2024-04-27 um 21 09 55
Bildschirmfoto am 2024-04-27 um 21 09 58
Bildschirmfoto am 2024-04-27 um 21 01 37

kwindrem:
I'd agree that having a max for AC output is not correct. That value depends on inverter state. The value to set is what the inverter can supply on its own. Then the AC output gauge can use the combination of AC input max and inverter max depending on mode: AC in max + inverter max if not inverting (and assist active), interter max if inverting (no AC input used).

Plus, AC input current max can be the AC input current limit.

The inverter max power could also be pulled from a table based on the inverter model rather than having the system designer enter it.

blammit:
@mr-manuel @kwindrem thanks for the feedback; this work is still preliminary and will be improved after some initial use and testing. I'll comment in https://github.com/victronenergy/gui-v2/issues/532 to refer to this PR so this isn't lost.

mpvader:
For AC input min/max, Jesus and I made a separate and new issue: https://github.com/victronenergy/venus-private/issues/371. Pls don't review & drop comments in there yet and lets first have Jesus and Izak complete it.

PR #986, opened 2024-04-24
Protect gui-v2 from text overflows

### LINKED ISSUE #720: Protect gui-v2 from text overflows
In general UI labels should not get truncated, but limiting the text bounding box is better than the resulting overlapping UI elements on the screen and you don't always fully control the text length (translations, dynamic text coming from the middleware or user, unintended screen and font size changes, etc.).

Go through the UI labels
- Always define the available width of text
- In case the text doesn't fit either wrap to multiple lines or truncate with "..."
- Use long strings to test that nothing overlaps in the GUI

jpetrell:
Fixes #720.

### DIFF
--- a/components/controls/RadioButton.qml
+++ b/components/controls/RadioButton.qml
@@ -10,8 +10,6 @@ import Victron.VenusOS
 CT.RadioButton {
 	id: root
 
-	property alias label: label
-
 	implicitWidth: Math.max(
 		implicitBackgroundWidth + leftInset + rightInset,
 		implicitContentWidth + leftPadding + rightPadding,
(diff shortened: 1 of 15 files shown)


components/controls/RadioButton.qml
```diff
@@ -10,8 +10,6 @@ import Victron.VenusOS
 CT.RadioButton {
 	id: root
 
-	property alias label: label
```
jpetrell, line 13: Unused

components/controls/Switch.qml
```diff
@@ -67,11 +67,13 @@ CT.Switch {
...
-		text: root.text
-		color: Theme.color_font_primary
-		verticalAlignment: Text.AlignVCenter
+	contentItem: Item {
+		Label {
+			text: root.text
+			color: Theme.color_font_primary
+			width: parent.width - Theme.geometry_switch_groove_width
+			elide: Text.ElideRight
+		}
```
jpetrell, line 76: Cannot define width for the autosized contentItem so had to wrap the label inside an Item {}.

components/controls/RadioButton.qml
```diff
@@ -44,13 +42,4 @@ CT.RadioButton {
...
 	}
-
-	contentItem: Label {
-		id: label
-
-		font.pixelSize: Theme.font_size_body2
-		text: root.text
-		verticalAlignment: Text.AlignVCenter
-		color: root.enabled || root.checked ? Theme.color_font_primary : Theme.color_font_disabled
-	}
```
jpetrell, line 55: Unused

blammit (approved):

Issue #995, opened 2024-04-26
Some label overlap and truncation issues in different languages


jpetrell:
Browsed through the main gui-v2 views with all the languages. Most found issues need to be fixed in the translations (reported to Dirk-Jan), but here are some that require code-level fixes.

Gauge text truncation is bit too strong on many languages: ![image](https://github.com/victronenergy/gui-v2/assets/2203667/214510ba-1c00-4403-850b-62ad8514451e)

Ukrainian (Inverter / Charger box label goes outside the box):
![image](https://github.com/victronenergy/gui-v2/assets/2203667/6e74fce9-ab3d-4c97-9b3c-5c67a2097140)

German (hint label overlaps with the icon):
![image](https://github.com/victronenergy/gui-v2/assets/2203667/04873072-84f7-4913-8386-afd2718a0257)

Spanish and French (placeholder text goes over the border or lacks nice margins):
![image](https://github.com/victronenergy/gui-v2/assets/2203667/3bd1d069-0bf4-4922-a760-2119d58091dc)
![image](https://github.com/victronenergy/gui-v2/assets/2203667/ec9bffd3-8406-48f3-b3af-907ea1c8f04d)

jpetrell:
All reported issues fixed.

PR #999, opened 2024-04-29
Some text wrapping and truncation improvements

jpetrell:
- Overview widget label overflow
- ESS widget info label overflow
- Too aggressive truncation of the Brief page gauge labels

image

Contributes to #995.

### DIFF
--- a/components/CircularMultiGauge.qml
+++ b/components/CircularMultiGauge.qml
@@ -17,6 +17,7 @@ Item {
 	property bool animationEnabled
 	property real labelMargin
 	property alias labelOpacity: textCol.opacity
+	property int leftGaugeCount
 
 	// Step change in the size of the bounding boxes of successive gauges
 	readonly property real _stepSize: 2 * (strokeWidth + Theme.geometry_circularMultiGauge_spacing)
@@ -104,7 +105,13 @@ Item {
 					font.pixelSize: valueLabel.visible ? Theme.font_size_body1 : Theme.font_size_body2
 					color: Theme.color_font_primary
 					text: model.name
+
+					// With three gauges on the left there is a risk that the last labels on
+					// on the multi-gauge overlap with the labels on the top-left gauge.
+					//
+					// Increase the space for the two top-most labels or if there are less left gauges.
 					width: textCol.width - valueLabel.width - iconImage.width
+						+ (model.index < 2 || gauges.leftGaugeCount < 3 ? Theme.geometry_circularMultiGauge_label_extraWidth : 0)
 					elide: Text.ElideRight
 				}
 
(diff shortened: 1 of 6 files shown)


components/CircularMultiGauge.qml
```diff
@@ -105,6 +106,7 @@ Item {
 					color: Theme.color_font_primary
 					text: model.name
 					width: textCol.width - valueLabel.width - iconImage.width
+						+ (model.index < 2 || gauges.leftGaugeCount < 3 ? Theme.geometry_circularMultiGauge_label_extraWidth : 0)
```
jpetrell, line 109: Initially tried to calculate the available space, but tracking quantity label sizes and al the margins became a bit too complicated so added additional theme parameter instead.

```
root.width - Theme.geometry_mainGauge_size/2 - Theme.geometry_briefPage_edgeGauge_horizontalMargin - ...
```

blammit: It seems a bit complicated for the centre gauge to depend on the left gauge count, depending on the model index here; I think these dependencies will become hard to manage. Can we do it some other way? Maybe have some "Max width of labels" that can be passed in by the Brief page, and is updated when the left gauge count changes?

jpetrell: >  Can we do it some other way? Maybe have some "Max width of labels" that can be passed in by the Brief page, and is updated when the left gauge count changes?

Good idea. :+1:

jpetrell: > Can we do it some other way? Maybe have some "Max width of labels" that can be passed in by the Brief page

That kind of approach would fix the aggressive truncation of the labels in the common case (1 or 2 left gauges) when there is ample space, but not the issue of top gauge labels truncating just because the last gauge label potentially conflicts (but more often that not, doesn't).

> I think these dependencies will become hard to manage

Brief page layout is complicated in its nature with lot of moving parts and different layout variants, I am not sure we can avoid that if we want to do a good job with the layouts. It is one of the hero views that you see right after the boot.

Might be good to have a text truncation debug mode,  where you could easily highlight the bounding box of all the labels `Rectangle { anchors.fill: parent; color: "red"; opacity: 0.1 }` and automatically append extra text at the end to force truncation `Component.onCompleted: text = text + "ABCDEFG..."`. But adding such changes manually to the shared Label.qml component is also pretty trivial and wouldn't want to make such a common visual component more bloated with testing code.

blammit: Hmm. Okay, let's go with this for now. Can you add a comment to this part to explain why this extra width is needed?

blammit (approved):
Minor comments, LGTM otherwise

PR #1001, opened 2024-04-29
Confirm that user really wants to reset min/max values

### LINKED ISSUE #990: Add confirmation step to min/max gauge range reset button
Go to Settings > Display & Languages > Minimum and maximum gauge ranges. Accidentally tap "Reset all values to zero" button.

Actual result: You lose up to 10 min/max values with no way to set them back

Expected result: Before committing ask users if they are really sure.

jpetrell:
Fixes #990.

### DIFF
--- a/pages/settings/PageSettingsDisplayMinMax.qml
+++ b/pages/settings/PageSettingsDisplayMinMax.qml
@@ -31,16 +31,30 @@ Page {
 				//% "Reset all range values to zero"
 				text: qsTrId("settings_minmax_reset")
 				secondaryText: CommonWords.reset
-				onClicked: {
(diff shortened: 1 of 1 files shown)


blammit (commented):
With Qt 6.6.1 on MacOS I get:

```qrc:/qt/qml/Victron/VenusOS/pages/settings/PageSettingsDisplayMinMax.qml:12:5: QML GradientListView: Delegate must be of Item type```

If I move the component to within the ListButton then it works fine.

jpetrell:
> GradientListView: Delegate must be of Item type
> If I move the component to within the ListButton then it works fine.

Good catch! Thanks for the hint. Adapted and tested the error went away.

blammit (approved):

PR #1002, opened 2024-04-29
Make sure the editor stays visible when the keyboard is opened

### LINKED ISSUE #991: Keyboard opens on top of the text fields located lower on the page
Go to Settings > Display & Languages > Minimum and maximum gauge ranges. Scroll to the end. Tap "Maximum power" field under the "Solar" section. Keyboard opens.

Expected result: You can see the text field you are editing.

Actual result: The keyboard comes on top of the field. You can still edit, but won't see the result until you close the keyboard.

Happens with any text field that is located near the end of the scrollable page area.

jpetrell:
- Tracking cursor visibility is more robust to dynamic flickable content
  area changes than trying to predict in advance where the text field will land
- Easier to extend text editing use cases later to multi-line text areas,
  copy&paste, allowing scrolling during more complex text editing use cases
- Also adjusts the viewport if the field is too close to the top edge of the viewport
- Maintain nice margin around the focused text field

Fixes #991.

### DIFF
--- a/pages/settings/PageSettingsDisplayMinMax.qml
+++ b/pages/settings/PageSettingsDisplayMinMax.qml
@@ -10,6 +10,7 @@ Page {
 	id: root
 
 	GradientListView {
+
 		model: ObjectModel {
 			ListSwitch {
 				//: Whether to adjust the min/max values in the range dynamically, based on the lowest and highest values observed on the system.
--- a/Global.qml
+++ b/Global.qml
@@ -61,7 +61,7 @@ QtObject {
 	property bool isDesktop
 	property real scalingRatio: 1.0
 
-	signal aboutToFocusTextField(var textField, int toTextFieldY, var flickable)
+	signal aboutToFocusTextField(var textField, var flickable)
 	signal keyPressed(var event)
 
 	function showToastNotification(category, text, autoCloseInterval = 0) {
--- a/components/InputPanel.qml
+++ b/components/InputPanel.qml
@@ -47,18 +47,17 @@ QtVirtualKeyboard.InputPanel {
 
 	states: State {
 		name: "visible"
-		when: Qt.inputMethod.visible
+		when: Qt.inputMethod.visible && !!root.focusedFlickable
 
 		PropertyChanges {
 			target: root.focusedFlickable
-			contentY: root.toContentY
-			height: root.toHeight
+			height: Theme.geometry_screen_height - root.height - Theme.geometry_statusBar_height
 		}
 	}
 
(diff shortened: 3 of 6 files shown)


pages/settings/PageSettingsDisplayMinMax.qml
```diff
@@ -80,11 +82,13 @@ Page {
 							id: acInputMinCurrent
 							text: CommonWords.minimum_current
 							dataItem.uid: Global.systemSettings.serviceUid + "/Settings/Gui/Gauges/Ac/In/%1/Current/Min".arg(index)
+							flickable: listView
```
blammit, line 85: Can this find the parent listview dynamically instead of assigning it explicitly?

blammit (commented):
Could you explain in the commit message how this fixes the problem and why the previous implementation didn't work?

pages/settings/PageSettingsDisplayMinMax.qml
jpetrell: Initially I tried Component.onCompleted: if (!flickable) { /* loop through parent chain */ }, but at that point the repeater hadn't yet parented their child items so gave up. But I'll try again.

jpetrell:
> Could you explain in the commit message how this fixes the problem and why the previous implementation didn't work?

That is a good question. I can try to see if there is a less intrusive way to fix the visibility issues. The existing one-shot calculation felt brittle to me, though it could be enough for the limited text editing use cases currently supported.

pages/settings/PageSettingsDisplayMinMax.qml
jpetrell: > Initially I tried Component.onCompleted: if (!flickable) { /* loop through parent chain */ }, but at that point the repeater hadn't yet parented their child items

Works when using onWindowChanged instead (parent guaranteed to exist).

jpetrell:
> Could you explain in the commit message how this fixes the problem and why the previous implementation didn't work?

Here is an alternative fix https://github.com/victronenergy/gui-v2/pull/1011

blammit:
> > Could you explain in the commit message how this fixes the problem and why the previous implementation didn't work?
> 
> Here is an alternative fix #1011

I'm happy to go with this one if it is a general improvement. One thing is that with this PR, the fix for the scrolling seems confined to ListTextField. Which is the only place where we used text fields, but not sure if this is an improvement in this case.

jpetrell:
Alternative fix https://github.com/victronenergy/gui-v2/pull/1011 was merged, closing this one.

PR #1009, opened 2024-04-30
Show small 3-phase view for AC Loads widget when size <= S

blammit:
The AC input widget shows per-phase data even when its size <= S. The AC loads widget should do this too.

This adds AcInputWidget to do the common AC phase display configuration for AC input and AC load widgets.

### DIFF
--- a/components/widgets/AcInputWidget.qml
+++ b/components/widgets/AcInputWidget.qml
@@ -6,32 +6,28 @@
 import QtQuick
 import Victron.VenusOS
 
-OverviewWidget {
+AcWidget {
 	id: root
 
 	// If this is the currently active AC input, then fetch the measurements and phase data from
 	// Global.acInputs.activeInput. Otherwise, just show icon and title with 'Disconnected' status.
 	property AcInputSystemInfo inputInfo
 	property ActiveAcInput input: inputInfo && inputInfo.isActiveInput ? Global.acInputs.activeInput : null
-	property ListModel phaseModel: connected ? input.phases : null
 	readonly property bool connected: input && input.connected
-	readonly property bool isConnectedSinglePhase: connected && phaseModel && phaseModel.count === 1
 
 	type: VenusOS.OverviewWidget_Type_AcGenericInput
 	title: !!inputInfo ? Global.acInputs.sourceToText(inputInfo.source) : ""
-	secondaryTitle: root.size <= VenusOS.OverviewWidget_Size_S && extraContentLoader.sourceComponent === phaseDisplayComponent
-			? "(%1)".arg(Units.defaultUnitString(Global.systemSettings.electricalQuantity))
-			: ""
 	icon.source: !!inputInfo ? Global.acInputs.sourceIcon(inputInfo.source) : ""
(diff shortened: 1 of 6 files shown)


components/widgets/AcInputWidget.qml
```diff
@@ -6,33 +6,32 @@
 import QtQuick
 import Victron.VenusOS
 
-OverviewWidget {
+AcWidget {
```
jpetrell, line 9: Nice!

components/widgets/AcWidget.qml
```diff
@@ -0,0 +1,41 @@
...
+			left: parent.left
+			leftMargin: Theme.geometry_overviewPage_widget_content_horizontalMargin
+			right: parent.right
+			rightMargin: Theme.geometry_overviewPage_widget_content_horizontalMargin + root.rightPadding
+			bottom: parent.bottom
+			bottomMargin: root.size <= VenusOS.OverviewWidget_Size_M
+				  ? root.verticalMargin / 3
+				  : root.verticalMargin
+		}
+		sourceComponent: root.phaseModel && root.phaseModel.count > 1 ? phaseDisplayComponent : null
```
jpetrell, line 39: I find `active:  root.phaseModel && root.phaseModel.count > 1` more readable.

components/widgets/AcWidget.qml
```diff
@@ -0,0 +1,41 @@
...
+	secondaryTitle: size <= VenusOS.OverviewWidget_Size_S && extraContentLoader.sourceComponent === phaseDisplayComponent
+			? "(%1)".arg(Units.defaultUnitString(Global.systemSettings.electricalQuantity))
+			: ""
+	quantityLabel.visible: quantityLabel.dataObject
+		&& (size >= VenusOS.OverviewWidget_Size_M || (phaseModel && phaseModel.count === 1))
+	quantityLabel.font.pixelSize: size <= VenusOS.OverviewWidget_Size_S
+		|| (size === VenusOS.OverviewWidget_Size_M && extraContentLoader.sourceComponent === phaseDisplayComponent)
+			  ? Theme.font_overviewPage_widget_quantityLabel_minimumSize
+			  : Theme.font_overviewPage_widget_quantityLabel_maximumSize
+	preferLargeSize: !!phaseModel && phaseModel.count > 1
```
jpetrell, line 24: Sometimes State + PropertyChanges with small and large variants and some state extending the other similar state leads to more readable and maintainable presentation. But could also unnecessarily blow up the amount of code here, dunno. Tricky to find the right balance. :smile:

jpetrell (approved):
Small nitpicks, overall LGTM.

jpetrell:
Feels the small 3-phase row view could be improved, now it looks bit sparse and disconnected, particularly on 7". But not caused by these changes here.

image

components/widgets/AcWidget.qml
blammit: Agreed, this was doing my head in too. Updated to use states and PropertyChanges instead.

blammit:
> Feels the small 3-phase row view could be improved, now it looks bit sparse and disconnected, particularly on 7". But not caused by these changes here.

Yes, wonder if we should use larger text for the 3-phase display in this case.

PR #1011, opened 2024-04-30
Fix keyboard opening on top of the text field

### LINKED ISSUE #991: Keyboard opens on top of the text fields located lower on the page

jpetrell:
Fixes #991.

### DIFF
--- a/components/InputPanel.qml
+++ b/components/InputPanel.qml
@@ -83,8 +83,7 @@ QtVirtualKeyboard.InputPanel {
 				if (flickable.contentY + delta + flickable.height > flickable.contentHeight) {
 					// Item is too close to bottom of flickable, so it will still be hidden after
 					// scrolling upwards. Reduce the flickable height so that item can be seen.
-					root.toHeight = flickable.height - inputPanelY
-						+ ((!!Global.pageManager && !!Global.pageManager.navBar) ? Global.pageManager.navBar.height : 0)
+					root.toHeight = flickable.height - root.height
 				} else {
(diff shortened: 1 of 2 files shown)


components/InputPanel.qml
```diff
@@ -83,8 +83,7 @@ QtVirtualKeyboard.InputPanel {
 				if (flickable.contentY + delta + flickable.height > flickable.contentHeight) {
 					// Item is too close to bottom of flickable, so it will still be hidden after
 					// scrolling upwards. Reduce the flickable height so that item can be seen.
-					root.toHeight = flickable.height - inputPanelY
-						+ ((!!Global.pageManager && !!Global.pageManager.navBar) ? Global.pageManager.navBar.height : 0)
+					root.toHeight = flickable.height - root.height
```
jpetrell, line 86: Should now work nicely for all the current text fields.

I would sleep better if the viewport resizing was handled higher in the stack, but that is a bigger change. This is still brittle in sense that it doesn't really know how does the keyboard position relate to the flickable, what kind of fixed top or bottom menus, margins, etc. have been added around the viewport.

blammit (approved):
Nice!

PR #1013, opened 2024-04-30
Avoid label overlap on the Overview page battery widget

### LINKED ISSUE #1012: Battery widget quantity labels may overlap
[Reported](https://community.victronenergy.com/questions/270759/venus-os-v3401-available-for-testing.html) by customer 

image

jpetrell:
Reduces font size if there is a risk of overlap:
image

Long value labels are possible, big batteries produce power in thousands of watts range and if the battery is discharging the amps may be negative with additional "-" character.

Not too proud of this solution, but couldn't come up with more elegant solution either. One long value label is not a problem, but two large adjacent ones start overlapping. Alternatively could always make the font smaller, but that would make the common layout variant look sparse.

Fixes #1012.

### DIFF
--- a/components/QuantityLabel.qml
+++ b/components/QuantityLabel.qml
@@ -15,6 +15,8 @@ Item {
 	property alias valueColor: valueLabel.color
 	property alias unitColor: unitLabel.color
 	property alias unitVisible: unitLabel.visible
+	readonly property alias valueText: valueLabel.text
+	readonly property alias unitText: unitLabel.text
 	property int alignment: Qt.AlignHCenter
 	property alias precision: quantityInfo.precision
 	property alias formatHints: quantityInfo.formatHints
--- a/components/widgets/BatteryWidget.qml
+++ b/components/widgets/BatteryWidget.qml
@@ -23,6 +23,26 @@ OverviewWidget {
 	readonly property int _normalizedStateOfCharge: Math.round(batteryData.stateOfCharge || 0)
 	readonly property bool _animationReady: animationEnabled && !isNaN(batteryData.stateOfCharge)
 
+	// Calculate whether current and power quantities fit on the footer together, if not use smaller font.
+	// Discharging battery has negative amperes and its not unusual for the watts to be in the 1k+ range.
+	// No need to do the same check for the shorter voltage label on the left.
+	readonly property bool _useSmallFont: root.width/2 - 2*batteryPowerDisplay.anchors.rightMargin
+										  < quantityLabelWidth(batteryCurrentDisplay.valueText, batteryCurrentDisplay.unitText)/2
+										  + quantityLabelWidth(batteryPowerDisplay.valueText, batteryPowerDisplay.unitText)
+
+	function quantityLabelWidth(valueText, unitText){
+		const valueTextRect = quantityLabelFont.tightBoundingRect(valueText)
+		return quantityLabelFont.font, (valueTextRect.x + valueTextRect.width
+										+ Theme.geometry_quantityLabel_spacing
+										+ quantityLabelFont.advanceWidth(unitText))
+	}
+
+	FontMetrics {
+		id: quantityLabelFont
+		font.pixelSize: Theme.font_size_body2
+		font.family: "Museo Sans 500 Mono digits"
+	}
+
 	title: CommonWords.battery
 	icon.source: batteryData.icon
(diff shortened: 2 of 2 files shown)


components/widgets/BatteryWidget.qml
```diff
@@ -22,6 +22,7 @@ OverviewWidget {
 
 	readonly property int _normalizedStateOfCharge: Math.round(batteryData.stateOfCharge || 0)
 	readonly property bool _animationReady: animationEnabled && !isNaN(batteryData.stateOfCharge)
+	readonly property bool _useSmallFont: Theme.screenSize === Theme.FiveInch && batteryCurrentDisplay.valueText.length >= 4 && (batteryVoltageDisplay.valueText.length >= 5 || batteryPowerDisplay.valueText.length >= 4)
```
blammit, line 25: The solution does seem brittle. What do you think about:

- using `fontSizeMode` to size the labels
- using font/text metrics to calculate the width at which the text will overflow
- or even having two sets of labels with different sizes, and choosing the smaller one if the large one overflows

jpetrell: Great ideas, ended up going with the font metrics route. Font size mode would otherwise be nice and automatic, but don't want one footer label appearing in different size to others.

components/QuantityLabel.qml
```diff
@@ -15,6 +15,8 @@ Item {
 	property alias valueColor: valueLabel.color
 	property alias unitColor: unitLabel.color
 	property alias unitVisible: unitLabel.visible
+	property alias valueText: valueLabel.text
+	property alias unitText: unitLabel.text
```
blammit, line 19: Should we make these two properties read-only?

blammit (approved):
Minor comment, LGTM

components/widgets/BatteryWidget.qml
```diff
@@ -23,6 +23,24 @@ OverviewWidget {
...
 
+	// Calculate whether current and power quantities fit on the footer together, if not use smaller font.
+	// Discharging battery has negative amperes and its not unusual for the watts to be in the 1k+ range.
+	// No need to do the same check for the shorter voltage label on the left.
+	readonly property bool _useSmallFont: root.width/2 - 2*batteryPowerDisplay.anchors.rightMargin
+										  < quantityLabelWidth(batteryCurrentDisplay.valueText, batteryCurrentDisplay.unitText)/2
+										  + quantityLabelWidth(batteryPowerDisplay.valueText, batteryPowerDisplay.unitText)
+
+	function quantityLabelWidth(valueText, unitText){
+		return quantityLabelFont.advanceWidth(valueText) + Theme.geometry_quantityLabel_spacing
```
chriadam, line 34: maybe "return quantityLabelFont.font, (quantityLabelFont.tightBoundingRect(valueText).x + quantityLabelFont.tightBoundingRect(valueText).width + Theme.geometry_quantityLabel_spacing + quantityLabelFont.advanceWidth(unitText))" since the advanceWidth isn't necessarily correct depending on the font and the text involved, and to ensure that the font metrics font property is captured for re-evaluation.

jpetrell: Thanks! Applied you improvements.

I have never seen the issue of the font not being loaded, but reading comments on QTBUG-124588 I understand I have been lucky.

I am used to using advancedWidth, and for these truncation cases it is probably good enough (only measured 0.75-1.625 pixel differences, and tight bounding rect x+width was actually smaller than the advance width). Also, in this particular use case the monospaced digit glyphs fit comfortably within their bounding boxes so even if the bounding boxes would touch at edges there would be no risk of overlap (in addition the calculations here include additional margin).

Still mindblown reading about QTBUG-124588. I understand font rules are complicated, but IMHO it is the job of the framework to provide simple bounding rect API that 3rd party developers can trust contains all the drawn content.

components/QuantityLabel.qml
jpetrell: Yes! Done.

components/widgets/BatteryWidget.qml
chriadam: I agree.  My understanding is that the font metrics calculations were completely broken prior to Qt 6.3.x so advanceWidth provided the "best"/"closest" outcome prior to Qt 6.3.  Post 6.3 I think tBR.x+tBR.width is the way to go, since the normal boundingRect calculation is still broken in Qt 6.6.

PR #1023, opened 2024-04-30
Prevent Levels page from disappearing when changing mock configs

blammit:
Add new test objects before removing the old ones. Otherwise, the Levels page disappears as soon as the models are cleared, and the user is moved to the Overview page.

### DIFF
--- a/data/mock/EnvironmentInputsImpl.qml
+++ b/data/mock/EnvironmentInputsImpl.qml
@@ -81,15 +81,25 @@ Item {
 		target: Global.mockDataSimulator || null
 
 		function onSetEnvironmentInputsRequested(config) {
-			Global.environmentInputs.model.clear()
-			while (_createdObjects.length > 0) {
-				_createdObjects.pop().destroy()
+			// Do not remove all objects in the model before adding the new ones, as Levels page
+			// will disappear.
+			const hasCreatedObjects = _createdObjects.length > 0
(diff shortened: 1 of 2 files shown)


data/mock/EnvironmentInputsImpl.qml
```diff
@@ -81,16 +81,24 @@ Item {
...
 
 		function onSetEnvironmentInputsRequested(config) {
-			Global.environmentInputs.model.clear()
-			while (_createdObjects.length > 0) {
-				_createdObjects.pop().destroy()
+			// Add new inputs before removing the old ones, so that the Levels page doesn't disappear
+			let i
+			let objectsToRemove = []
+			for (i = 0; i < _createdObjects.length; ++i) {
+				objectsToRemove.push(_createdObjects[i])
```
jpetrell, line 88: Form `let object; for (object in _createdObjects) { objectsToRemove.push(object) }` and `for (object in objectsToRemove) { ..; object.destroy() `} would be slightly more concise and readable.

Could also just only leave one item there `while(_createdObjects.length > 1) { _createdObjects.pop().destroy() }` to avoid book-keeping a full array and at the end pop the item `object = _createdObjects.shift(); ..; object.destroy()`.

blammit: Nice, I kept the last item in the list as suggested, thanks.

PR #1037, opened 2024-05-08
Rework Brief side panel to use fixed minimum/maximum gauge ranges

blammit:


components/BarGauge.qml
```diff
@@ -6,16 +6,20 @@
 import QtQuick
 import QtQuick.Effects as Effects
 import Victron.VenusOS
+import Victron.Gauges
 
 Rectangle {
 	id: bgRect
 
-	property alias backgroundColor: bgRect.color
-	property alias foregroundColor: fgRect.color
```
jpetrell: Personally I prefer alias as that guarantees the two properties are the same, but this works too, understand that it is nice to define all the colors in top-level.

mr-manuel:
I tested this and on my side the AC Loads gauge is not moving anymore. The Grid gauge looks nice now. The confusing thing is that the gauges are not moving as fast as the values, but this is another story. I will update a new issue for that.

Top this build, bottom current main build.

venus-os-graph

components/ValueRange.qml
```diff
@@ -13,8 +14,6 @@ QtObject {
 		if (isNaN(value) || isNaN(minimumValue) || isNaN(maximumValue)) {
 			return 0
 		}
-		const range = maximumValue - minimumValue
-		const normalizedValue = Math.max(minimumValue, Math.min(maximumValue, value))
-		return range === 0 ? 0 : (normalizedValue - minimumValue) / range
+		return Utils.scaleNumber(value, minimumValue, maximumValue, 0, 1)
```
jpetrell, line 17: Not the most readable statement with 5 parameters. Could consider adding comments inline `.., maximumValue, 0 /* toMin */, /* toMax */ 1)`.

data/DcInputs.qml
```diff
@@ -12,6 +12,7 @@ QtObject {
 	property real power: NaN
 	property real current: NaN
 	readonly property real maximumPower: _maximumPower.value === undefined ? NaN : _maximumPower.value
+	readonly property string multiInputIcon: "qrc:/images/icon_dc_24.svg"
```
jpetrell, line 15: Having icons in the data/model side feels slightly wrong.

data/mock/config/MockDataSimulator.qml
```diff
@@ -181,17 +181,25 @@ QtObject {
...
+				if (event.modifiers & Qt.ControlModifier) {
+					phases.setProperty(i, "current", 0)
+					phases.setProperty(i, "power", 0)
+				} else if (event.modifiers & Qt.ShiftModifier) {
+					phases.setProperty(i, "current", phaseCurrent + 5)
+					phases.setProperty(i, "power", phasePower + 100)
+				} else {
+					phases.setProperty(i, "current", phaseCurrent - 5)
+					phases.setProperty(i, "power", phasePower - 100)
+				}
```
jpetrell, line 198: Nice! I also really should add more such mock test controls.

jpetrell (commented):
Overall LGTM, some observations, not sure if they are exact issues and/or if they need to be fixed now.

- Calling the side panel both side panel and brief monitor panel feels confusing

- Side panel with only 1 or 2 items looks a bit sparse, wonder what are the common configurations

Tested with Boat/Motorhome demo 2:

- Alternator and AC Loads change all the time, but show leveled lines on , but I guess that is because of miniscule changes in the total scale

- AC Loads graph resets and starts from the beginning every time you visit the Overview page and come back. Alternator graph doesn't

Tested with Boat/Motorhome demo 1:

- Generator range resets every time you visit the Overview page and come back

components/BarGauge.qml
blammit: This was partly to avoid the warning about aliasing to a property in the root item.

components/BriefMonitorWidget.qml
```diff
@@ -17,15 +17,20 @@ Item {
 	property bool loadersActive
 
 	width: parent.width
-	height: bottomLoader.y + bottomLoader.height
+	height: bottomLoader.y
+			+ (bottomLoader.status === Loader.Ready ? bottomLoader.height : -bottomLoader.anchors.topMargin)
+			+ Theme.geometry_monitorWidget_verticalMargin
```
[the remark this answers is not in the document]
blammit: It was a bit simpler to start with a non-Column layout at the time, since there was some overlapping in the items. But it's a bit easier to read with a Column. Have updated to use this.

components/ValueRange.qml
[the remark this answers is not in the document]
blammit: Added a comment to clarify.

blammit: Added a comment to clarify.

data/DcInputs.qml
blammit: Yeah, that's true. Added back to the Brief page and now it passes it to the panel.

blammit:
> I tested this and on my side the AC Loads gauge is not moving anymore. The Grid gauge looks nice now.
> 
> Top this build, bottom current main build.

Maybe it's something to do with the differing min/max range depending on the connected AC input.

Could you add the VRM portal permissions so I could take a look?

blammit:
> Calling the side panel both side panel and brief monitor panel feels confusing

Agreed, I've renamed it to side panel for consistency.

> Side panel with only 1 or 2 items looks a bit sparse, wonder what are the common configurations

I've adjusted the margins and layouts a bit so that things are less squashed in general.

> AC Loads graph resets and starts from the beginning every time you visit the Overview page and come back. Alternator graph doesn't
> Generator range resets every time you visit the Overview page and come back

These should be fixed to not reset now, unless you open/close the monitor panel. That probably causes it to reset due to the geometry change.

mr-manuel:
@blammit sure, which e-Mail should I add?

blammit:
> @blammit sure, which e-Mail should I add?

victron.vrm@qinetic.com.au, thanks!

jpetrell:
Otherwise looks good, but with the latest changes the solar yield graph has disappeared somewhere. Debugging now.

image

jpetrell:
On the device I often see the side panel quantity label increasing but gauge length decreasing, and vice versa. The gauge and label update in sync, but in different direction. Tested with Tested with Boat/Motorhome demo 1.

>   Generator range resets every time you visit the Overview page and come back

AC Loads graph no longer resets when visiting the Overview page, but this still happens. ^^

blammit:
@mr-manuel I checked your installation. Looking in Settings > Display and Language > Minimum and maximum gauge ranges, the "Maximum power: when AC in 0 connected" is 8990. However, there is a bug in gui-v2 here: it should refer to "Maximum current", not "Maximum power". Fixed here: https://github.com/victronenergy/gui-v2/pull/1042

Have you edited this value manually? If so, if you change it to e.g. 10 amps, then you will see a more obvious fluctuations in the AC load graph in the Brief side panel. The problem was that 8990 was a very high maximum for this graph.

mr-manuel:
Thanks, the missleading description was the issue.

CMakeLists.txt
```diff
@@ -216,6 +216,7 @@ set (VENUS_QML_MODULE_SOURCES
     components/ArcGauge.qml
     components/ArcGaugeQuantityRow.qml
     components/AsymmetricRoundedRectangle.qml
+    components/BarGauge.qml
```
chriadam, line 220: ooi why the rename?  The commit message says what was done, but not why

components/ThreePhaseBarGauge.qml
```diff
@@ -26,18 +26,18 @@ Flow {
 	Repeater {
 		id: phaseRepeater
 
-		delegate: VerticalGauge {
+		delegate: BarGauge {
```
chriadam: It's weird that BarGauge isn't a type provided by the Victron.Gauges module - what is in that module, is it just JS code / helper methods?

chriadam: (I think that it's good that all of the visual components are provided by Victron.VenusOS module, btw.)

data/mock/AcInputsImpl.qml
```diff
@@ -97,10 +97,11 @@ QtObject {
 
 				Component.onCompleted: {
 					_deviceInstance.setValue(input.index)
+					const inputConnected = !!input.modelData["connected"]
```
chriadam, line 100: should this be `!!input.modelData["connected"] || !!input.connected`?  or will the value from the model always precede the value of the input, so e.g. if it transitions to false we need to ignore the value of input.connected altogether?  or am I altogether confused?

CMakeLists.txt
chriadam: Ah, I see, later on it supports also horizontal mode.

pages/BriefMonitorPanel.qml
```diff
@@ -24,56 +24,47 @@ Column {
...
-				id: generatorQuantityLabel
-
-				anchors.top: generatorHeader.bottom
-				dataObject: Global.acInputs.generatorInput
-				font.pixelSize: Theme.font_briefPage_quantityLabel_size
-			}
+	// In most cases there is only 1 generator, so don't worry about other ones here.
+	BriefMonitorWidget {
+		id: generatorWidget
+		title: Global.generators.model.count > 1 ? Global.generators.model.firstObject.name : CommonWords.generator
```
chriadam: should it be >= 1 here?

data/DcInputs.qml
```diff
@@ -137,6 +137,19 @@ QtObject {
 		}
 	}
 
+	function inputTypeIcon(type) {
```
chriadam, line 140: why move this method into JS?

pages/BriefMonitorPanel.qml
```diff
@@ -61,163 +61,146 @@ Column {
...
+			threshold: {
+				// For a system with a min current below zero (i.e. it sometimes exports), the
+				// threshold is the 0 amp mark within the min/max range.
+				if (acInputGraphRange.minimumCurrent < 0) {
+					const range = acInputGraphRange.maximumCurrent - acInputGraphRange.minimumCurrent
+					if (range !== 0) {
+						return Math.abs(acInputGraphRange.minimumCurrent) / range
 					}
 				}
+				// For a system that only imports, the threshold is the halfway point.
```
chriadam: Why 0.5?  I thought it would be 0.8 i.e. warning threshold percentage?

pages/BriefMonitorPanel.qml
```diff
@@ -61,163 +61,146 @@ Column {
...
+				: qsTrId("brief_dc_input")
+		icon.source: Global.dcInputs.model.count === 1
+				? Global.dcInputs.inputTypeIcon(Global.dcInputs.model.firstObject.inputType)
+				: Global.dcInputs.multiInputIcon
+		active: Global.dcInputs.model.count > 0
+		visible: active
+		quantityLabel.dataObject: Global.dcInputs
+		sideComponent: LoadGraph {
+			active: root.animationEnabled
+			threshold: 0.5
```
chriadam: in this case (i.e. inputs), why is there any threshold at all?

components/LoadGraph.qml
```diff
@@ -62,44 +60,40 @@ Item {
...
-			strokeColor: Theme.color_briefPage_sidePanel_loadGraph_warning_strokeColor
+			strokeColor: aboveThresholdFillColor
 			offsetFraction: root.offsetFraction
 			fillGradient: LinearGradient {
 				x1: 0; y1: 0
 				x2: 0; y2: height
-				GradientStop { position: 0; color: Theme.color_briefPage_sidePanel_loadGraph_warning_gradientColor1 }
-				GradientStop { position: 1; color: Theme.color_briefPage_sidePanel_loadGraph_warning_gradientColor2 }
+				GradientStop { position: 0; color: aboveThresholdFillColor }
+				GradientStop { position: 1; color: "transparent" }
```
chriadam, line 69: In the case where the above threshold fill color is equal to the below threshold fill color, this can result in a strange looking graph:

victron-loadgraph-color-threshold

Also, I think there's a typo in commit message, says LoadRange but should be LoadGraph

components/ThreePhaseBarGauge.qml
```diff
@@ -15,35 +15,59 @@ Flow {
...
-				minimumValue: root.minimumValue
-				maximumValue: root.maximumValue
+		delegate: Item {
+			width: root.orientation === Qt.Vertical ? root.width : root._delegateLength
+			height: root.orientation === Qt.Vertical ? root._delegateLength : root.height
+
+			Label {
+				id: phaseLabel
+
+				anchors.verticalCenter: parent.verticalCenter
```
chriadam, line 40: Not sure why (maybe to do with numeric character descents in this font?) but the numbers look like they are baselined to the bottom edge of the bar graph, rather than vertically centered.
victron-side-panel-multi-phase

chriadam:
In general, every installation I looked at either had graphs which didn't show up (I guess because the values were too small given the specified max / range issues) or the graph was pinned at 100%.  J-cal was the exception, its load graph moved a tiny bit.  But the code LGTM, a few minor comments only.  The cleanups are very nice, it's far more maintainable now, nice work!

components/ThreePhaseBarGauge.qml
blammit: Yes, it's basically a singleton with the utility functions.

data/mock/AcInputsImpl.qml
blammit: This is the initial setup, so it should just take the value from the initial configuration, and ignore the value of `input.connected` (since that hasn't been set up yet).

data/DcInputs.qml
blammit: Seemed to make sense to group it with other DC-related functions.

pages/BriefMonitorPanel.qml
blammit: For AC inputs we don't have a warning level as it's good to have more power imported.

chriadam: Then 1.0 instead of 0.5?

pages/BriefMonitorPanel.qml
blammit: Seemed nice to have a mid-point in the graph, but I guess it looks odd. Perhaps we could remove it altogether.

components/LoadGraph.qml
blammit: Removed the mid-point threshold so this should be better now.

Fixed the commit message.

pages/BriefMonitorPanel.qml
blammit: Yeah you're right, let's remove it. Done now

pages/BriefMonitorPanel.qml
blammit: Maybe 0? Then we don't have a line at the top.

Changed now to zero.

components/ThreePhaseBarGauge.qml
blammit: Yeah, not sure of the best way to fix this. I tried using the baselineOffset as the label height, but there is some padding above the digit in that case, so that also looks off.

pages/BriefMonitorPanel.qml
blammit: Ah, should be == 1. Fixed now, thanks.

CMakeLists.txt
blammit: Updated the commit msg.

jpetrell:
Potentially getting overlapping labels now also on this branch
image

Here is one issue with Thai:
image

While the string is too long and should be fixed in translations (I guess should ping Dirk-Jan after merging) looks like the label dimensions also need adjustment.

Could add translation hint suggesting to keep the string short. This still fits:
image

Alternatively could adjust layout so the graph and title never overlapped, but hope it don't come down to that as it the nice graphs quite flat.

Also, would maybe always force left alignment even for the RTL scripts like Arabic:
image

blammit:
> Potentially getting overlapping labels now also on this branch

Thanks, I've updated the side panel to always put the graph below the title, which avoids the overlap. It's a shame that means the graph can't be taller, but we can't have both.

Looks like this now:

image
image

jpetrell (approved):

PR #1039, opened 2024-05-08
Avoid issue opening sub-pages in expanded mode

### LINKED ISSUE #1021: Sometimes settings page opens in expanded mode
The Brief, Overview and Levels pages can open in expanded mode, but usually not other page like settings pages. But if you happen to navigate to a new page just when the expanded mode is being activated you may end with a setting page that lacks navigation bar and status bar (and its actions).

Not a major issue, but looks broken and have seen it maybe dozen times so would be nice to fix to improve overall quality.

jpetrell:
If the user taps Overview page widget or navigation bar during animateNavBarOut animation the page opens without the status bar and navigation bar.

Happens quite rarely so kept the solution simple. Might have been better to start rewinding immediately from the expanded mode, but that is not so trivial with animators where you always have to specify the starting point of the animation. Also simultaneous page opening and collapsing animations might have looked funky.

Fixes #1021.

### DIFF
--- a/ApplicationContent.qml
+++ b/ApplicationContent.qml
@@ -57,7 +57,8 @@ Item {
 		anchors.fill: parent
 
 		onPressed: function(mouse) {
-			mouse.accepted = false
+			// block touch during navigation bar fadeout
+			mouse.accepted = mainView.navBarAnimatingOut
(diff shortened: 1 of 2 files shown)


blammit (approved):
Nice to get this fixed, thanks!

PR #1043, opened 2024-05-10
Min max settings updates

blammit:


components/listitems/ListTextField.qml
```diff
@@ -93,6 +112,20 @@ ListItem {
...
+				right: parent.right
+				verticalCenter: parent.verticalCenter
+			}
+			text: root.suffix
+			font: textField.font
+			color: Theme.color_font_secondary
+			leftPadding: textField.leftPadding / 2
+			rightPadding: textField.leftPadding
+		}
+
```
jpetrell, line 128: While it is good to visually separate the editable value and the suffix feels there is bit too much space between the two when compared to quantity labels and spinboxes with suffixes.

image

components/listitems/ListTextField.qml
```diff
@@ -75,6 +76,8 @@ ListItem {
 				Math.min(implicitWidth + leftPadding + rightPadding, Theme.geometry_listItem_textField_maximumWidth))
 		enabled: root.enabled
 		text: dataItem.isValid ? dataItem.value : ""
+		rightPadding: suffixLabel.implicitWidth
```
jpetrell, line 79: rightPadding grows to suffixLabel.leftPadding + suffixLabel.rightPadding i.e. to textField.leftPadding*1.5 in case there is no suffix

jpetrell (commented):
- Wonder if there is an use case to not enforce precision for number field e.g. support setting decimals to undefined. I guess not in Victron scope
- If dataItem is not valid should we show "--"?
- You can still define placeholder, but it is no longer centered on the field but on the space minus the suffix label, which looks a bit weird. Showing "--" instead of empty in invalid case would make sure the placeholder is never shown on the number field
- Typing really big value results in strings like "2.232323232323e+37". Could maybe just limit the value to n characters (decimal digits included) as there is currently no use case to input large values

pages/settings/PageSettingsDisplayMinMax.qml
```diff
@@ -94,11 +94,15 @@ Page {
...
 							dataItem.uid: Global.systemSettings.serviceUid + "/Settings/Gui/Gauges/Ac/In/%1/Current/Min".arg(index)
+							suffix: Units.defaultUnitString(VenusOS.Units_Amp)
+							decimals: 1
 						}
 						ListNumberField {
 							id: acInputMaxCurrent
 							text: CommonWords.maximum_current
 							dataItem.uid: Global.systemSettings.serviceUid + "/Settings/Gui/Gauges/Ac/In/%1/Current/Max".arg(index)
+							suffix: Units.defaultUnitString(VenusOS.Units_Amp)
+							decimals: 1
```
jpetrell, line 105: Not biggie, but this duplicates default precision for amperes, which is already 1? Could write here Units.defaultUnitPrecision(VenusOS.Units_Amp) instead, though I guess unlikely to ever change.

Wonder if this should be ListQuantityField with API aligned with ListQuantityItem with unit and precision instead of suffic and decimals, but this works well and is only used in one page so maybe not overoptimize it any further.

components/listitems/ListTextField.qml
blammit: Yeah it does look too wide; have removed the left padding so now it looks the same as the spinbox, to be consistent.

components/listitems/ListTextField.qml
blammit: Fixed now. Right padding is only set if suffix is set.

pages/settings/PageSettingsDisplayMinMax.qml
blammit: Good idea, yes ListQuantityField is much nicer. Updated to add this, and get the default unit and precision from Units.

blammit:
> Wonder if there is an use case to not enforce precision for number field e.g. support setting decimals to undefined. I guess not in Victron scope

I've renamed ListNumberField to ListQuantityField, which always has a precision. We can add ListNumberField later on if needed, without a precision.

> If dataItem is not valid should we show "--"?

Done

> You can still define placeholder, but it is no longer centered on the field but on the space minus the suffix label, which looks a bit weird. Showing "--" instead of empty in invalid case would make sure the placeholder is never shown on the number field

Done

> Typing really big value results in strings like "2.232323232323e+37". Could maybe just limit the value to n characters (decimal digits included) as there is currently no use case to input large values

I haven't done this as we don't have a fixed limit to use here. Maybe it's okay if the user decides to use a big number, and the backend can decide what should be done with it.

jpetrell (approved):

jpetrell:
>> Typing really big value results in strings like "2.232323232323e+37"
> Maybe it's okay if the user decides to use a big number

Ah didn't realize "e" and "+" were allowed characters, and you could still modify such values. Then there is no problem, no need to artificially limit the float range.

mr-manuel:
I don't think that this is the expected behaviour?

venus-os-max-values

blammit:
> I don't think that this is the expected behaviour?
> 

@mr-manuel Thanks for testing! I can't reproduce this, even on your installation, but I suspect it's related to number representations in different locales. Could you see if https://github.com/victronenergy/gui-v2/pull/1067 fix this for you.

PR #1044, opened 2024-05-10
Ensure monospace font is loaded

blammit:
On Mac desktop it fails to find the font by name if it does not match exactly.


Main.qml
```diff
@@ -152,6 +152,7 @@ Window {
 
 	FontLoader {
 		source: "qrc:/fonts/MuseoSans-500-monospaced-digits.otf"
+		Component.onCompleted: Global.monospacedFontLoader = fontLoader
```
jpetrell, line 155: You're passing the same font to both fontLoader and monospacedFontLoader?

Also, why not just expose the font family string, i.e. Global.fontFamily and Global.monospacedFontFamily

blammit: > You're passing the same font to both fontLoader and monospacedFontLoader?

🤦 Oops..

And yes the font family makes more sense, but it looks like the monospace font family is also "Museo Sans", same as the default? I tried this on Linux desktop as well. Not sure what's going on, but needs some more investigation.

jpetrell (approved):
Code looks good, good idea to move the FontLoaders to Global. Didn't see regression when running on Linux.

PR #1046, opened 2024-05-13
Solar gauge/overview: show amps when possible 

blammit:


data/mqtt/SystemImpl.qml
```diff
@@ -6,16 +6,4 @@
...
-		model: [
-			"mqtt/system/0/Ac/PvOnGrid",
-			"mqtt/system/0/Ac/PvOnGenset",
-			"mqtt/system/0/Ac/PvOnOutput"
-		]
-	}
-
-	property SystemData systemData: SystemData {}
-}
+SystemData {}
```
jpetrell, line 9: Feels odd to have identical code file in two different folders (also happens with EssImpl and SystemSettingsImpl), but I guess we want to reserve the possibility to write MQTT or DBus-specific code paths later if a need arises. And for most other device models the table model parametrisation is the only difference between MQTT and DBus models, or in RelaysIMpl case there is only one string difference.

Could try to define base DataManager and model components to avoid the code duplication between the sources folders (mqtt, dbus, mock), but overall the code there is so simple that maybe it would confuse more than help.

data/common/PvMonitor.qml
```diff
@@ -73,12 +54,22 @@ Instantiator {
...
 				}
+
+				let _maxPhaseCount = 0
+				for (let i = 0; i < count; ++i) {
+					const acPvDelegate = root.objectAt(i)
+					if (!!acPvDelegate) {
+						_maxPhaseCount = Math.max(_maxPhaseCount, acPvDelegate.vePhaseCount.value || 0)
+					}
+				}
+				root.maxPhaseCount = _maxPhaseCount
```
jpetrell, line 65: Could check in the beginning if `value != maxPhaseCount` and only if they differ than do whole recalculation.

jpetrell (approved):
LGTM

data/mqtt/SystemImpl.qml
blammit: Agreed, those should be reworked. At one point we couldn't use VeQuickItem with the mock backend, and not in wasm either, so it led to a whole bunch of abstractions that are no longer necessary. So a whole bunch of the `data/` code can be reworked, and a lot of the VeQuickItems can just be used for all backends.

data/common/PvMonitor.qml
blammit: Ah yes, thanks, updated.

Issue #1019, opened 2024-04-30
Higher precision requested for the battery voltages


jpetrell:
[Reported](https://community.victronenergy.com/questions/270759/venus-os-v3401-available-for-testing.html) by the customers: 
- "The Battery details display on the Touch 50 and Cerbo-s GX only shows cell voltages to one decimal place. The current released version shows to 3 decimal places. We find this fine resolution invaluable when a customer is recovering a flat battery: both when the battery is at 0%SOC and being changed and then when it is full charged to determine if any cells are damaged. We typically operate aroun 0.003V difference between low and high. lease can we have this high definition resumed."

![image](https://github.com/victronenergy/gui-v2/assets/2203667/ef0e2749-4ccc-4e85-a620-ef0ef94c03f6)

- "In GUI-V2 the min and max battery voltage on the BMS is shown with one digit precision. That is not really useful on LiFePo4 batteries. The old version had 3 digits available. Please update this to 3-digits again.)"
![image](https://github.com/victronenergy/gui-v2/assets/2203667/47e20a91-8e81-4722-9840-a1d5328ca7d7)

chriadam:
@jepefe raising this for your consideration

jepefe:
@chriadam @jpetrell Please, set these voltages to 3 decimals resolution for this product.

jpetrell:
> @chriadam @jpetrell Please, set these voltages to 3 decimals resolution for this product.

Ideally would avoid product-specific variations, but ok. I don't know what battery the first user was using, but I can ping him on the community channel. The second customer talks about LiFePo4 batteries in plural, is there a specific product ID I should use or match for a family of devices?

mr-manuel:
In GUIv1 all batteries displayed the cell voltages with 3 decimals. I don't think that it should be different for some batteries. Better would be that for all batteries there are 3 decimals.

jepefe:
> Ideally would avoid product-specific variations, but ok. I don't know what battery the first user was using, but I can ping him on the community channel. The second customer talks about LiFePo4 batteries in plural, is there a specific product ID I should use or match for a family of devices?

As discussed on Slack, as a rule of thumb:
DC Voltages: 2 decimals.
AC Voltages: No decimals.
This specific case which is cell voltage: 3 decimals.

So no product specific but measurement specific.

PR #1048, opened 2024-05-13
Improve DC voltage precision semantics

chriadam:


pages/settings/devicelist/battery/PageLynxIonSystem.qml
```diff
@@ -42,8 +42,8 @@ Page {
 				//% "Min/max cell voltage"
 				text: qsTrId("lynxionsystem_min_max_cell_voltage")
 				textModel: [
-					{ value: minCellVoltage.value, unit: VenusOS.Units_Volt },
-					{ value: maxCellVoltage.value, unit: VenusOS.Units_Volt },
+					{ value: minCellVoltage.value, unit: VenusOS.Units_Volt, precision: 3 },
+					{ value: maxCellVoltage.value, unit: VenusOS.Units_Volt, precision: 3 },
```
jpetrell, line 46: These lines could I guess be in the other commit that covers the special 3 dec precision case.

jpetrell (commented):
LGTM Didn't find fault with the changes even if there are still a few DC voltages not yet covered:
- "Charged voltage" DC voltage in PageBatterySettingsBattery.qml has 1 decimal precision
- "Maximum charge voltage" in PageSettingsDvcc uses 1 decimal precision
- "Min/Max Voltage" items in SolarDetailBox use default 1 precision
- "Battery voltage" in PageGeneratorConditions.qml uses 1 precision
- PageDcMeterAlarms displays DC voltages, but doesn't define precision
- Battery voltage in PageHub4Debug ESS settings just passes value with no precision defined

> Note that the default precision for Volts is one decimal; this is correct for AC voltages but not DC voltages.

Jesus wrote [here](https://github.com/victronenergy/gui-v2/issues/1019) that "AC Voltages: No decimals."?

Some other observations:

- Due to endless number of battery voltages sprinkled around the settings pages feels like the DC voltage should be the default one to reduce the amount of voltage overrides

- Almost feels like there should be Units_AC_Volt and Units_DC_Volt in units.cpp to avoid having to duplicate precision 2 in two dozen places.

- Could define enum Units.PrecisionDefault that had value -1to improve code readability

components/widgets/BatteryWidget.qml
```diff
@@ -170,6 +170,7 @@ OverviewWidget {
 
 			value: batteryData.voltage
 			unit: VenusOS.Units_Volt
+			precision: 2
```
jpetrell, line 173: Didn't add overflow protection to BatteryWidget voltage label in #1012 as that didn't seem likely, but now with the increased precision that may very well happen. :thinking:

image

chriadam:
> Awesome! Overall LGTM, didn't find fault with the changes even if there are still a few DC voltages not yet covered:
> 
> * "Charged voltage" DC voltage in PageBatterySettingsBattery.qml has 1 decimal precision
> * "Maximum charge voltage" in PageSettingsDvcc uses 1 decimal precision
> * "Min/Max Voltage" items in SolarDetailBox use default 1 precision
> * "Battery voltage" in PageGeneratorConditions.qml uses 1 precision
> * PageDcMeterAlarms displays DC voltages, but doesn't define precision
> * Battery voltage in PageHub4Debug ESS settings just passes value with no precision defined
> 
> > Note that the default precision for Volts is one decimal; this is correct for AC voltages but not DC voltages.
> 
> Jesus wrote [here](https://github.com/victronenergy/gui-v2/issues/1019) that "AC Voltages: No decimals."?
> 
> Some other observations:
> 
> * Due to endless number of battery voltages sprinkled around the settings pages feels like the DC voltage should be the default one to reduce the amount of voltage overrides
> * Almost feels like there should be Units_AC_Volt and Units_DC_Volt in units.cpp to avoid having to duplicate precision 2 in two dozen places.
> * Could define enum Units.PrecisionDefault that had value -1to improve code readability

Good catch for the missed places, I will fix those.

Interesting point about the different units (AC_Volt vs DC_Volt) - I can prototype that.  I agree it's better than having to remember to add precision overrides everywhere (but those are still required for the 3-decimal cell voltages, I guess).

I will add the PrecisionDefault as you suggest, as -1 does seem a bit magic, you're right.

Regarding the 1 vs 0 decimals for AC, I checked the slack conversation which Jesús mentioned in the issue, and in it Matthijs wrote "two decimals is for DC voltage, for AC voltage its just one" and Jesús wrote "For AC we use only one decimal place, for both current and voltage".  Perhaps I was looking at the wrong slack convo though - not sure.

components/widgets/BatteryWidget.qml
chriadam: Damn :-(  Hmm.  I am loath to change the design :-/  Perhaps stacking the three values vertically on the RHS is a possible solution, but would need Serj's input...
Maybe we can just set precision to 1 for this Battery widget only?  (But the image looks like the problem occurs even with 1 decimal of precision, and 0 decimals seems bad.)

jpetrell:
>  I checked the slack conversation which Jesús mentioned in the issue

Seems that discussion occured August last year? @jepefe commented on few weeks ago in one-to-one Slack chat that "For AC Voltage, best do not use any decimal at all, could be the case that for some specific case we need one decimal, but I can't think about any now.". And then also documented the voltage precision rules in the GitHub [here](https://github.com/victronenergy/gui-v2/issues/1019#issuecomment-2092539880).

> DC Voltages: 2 decimals.
> AC Voltages: No decimals.
> This specific case which is cell voltage: 3 decimals.

components/widgets/BatteryWidget.qml
jpetrell: The bottom position looks visually good and used to fit in most cases (7" layout, usual voltage+ampere+watts range), though not sure with the increased precision any more. Columns on the right would solve this case, but potentially introduce others (trying to fit 8 dynamic labels here in small fixed dimensions).

In general should avoid font sizes that require dynamic font fitting, but IMO making the font smaller results in sparse layout that doesn't look very good and this is one of the "Hero" views which should look nice.

Maybe just extent the font fitting to cover the voltage label? Tested that this should work:

```
-       readonly property bool _useSmallFont: root.width/2 - 2*batteryPowerDisplay.anchors.rightMargin
-                                                                                 < quantityLabelWidth(batteryCurrentDisplay.valueText, batteryCurrentDisplay.unitText)/2
-                                                                                 + quantityLabelWidth(batteryPowerDisplay.valueText, batteryPowerDisplay.unitText)
+       readonly property bool _useSmallFont: !quantityLabelFits(batteryVoltageDisplay) || !quantityLabelFits(batteryPowerDisplay)
+
+       function quantityLabelFits(label) {
+               return root.width/2 - 2*Theme.geometry_overviewPage_widget_content_horizontalMargin
+                               > quantityLabelWidth(batteryCurrentDisplay.valueText, batteryCurrentDisplay.unitText)/2
+                               + quantityLabelWidth(label.valueText, label.unitText)
+       }
```

chriadam:
Ah great, I wasn't aware of the other slack convo.  Sounds good, I'll set the Units_Volt_AC default precision to zero, and Units_Volt_DC to two, and then add overrides (to three) just for the cell voltage labels.

components/widgets/BatteryWidget.qml
chriadam: Yep, nice idea, will do this.

pages/solar/SolarDeviceListPage.qml
```diff
@@ -112,7 +112,7 @@ Page {
 				  : CommonWords.pv_inverter
 			quantityTitleModel: [
 				{ text: chargerMode ? CommonWords.yield_today : CommonWords.energy, unit: VenusOS.Units_Energy_KiloWattHour },
-				{ text: CommonWords.voltage, unit: VenusOS.Units_Volt },
+				{ text: CommonWords.voltage, unit: chargerMode ? VenusOS.Units_Volt_DC : VenusOS.Units_Volt_AC },
```
chriadam, line 115: not sure about this - I assume that the inverter voltage is output and therefore AC, but I might be wrong?

jpetrell: Rein commented that he expects those values to be input voltages (thus DC), but that he isn't sure either and that Jesus knows better.

PR #1049, opened 2024-05-13
Improve UI for AC in/out when feeding back to grid

blammit:


components/ThreePhaseDisplay.qml
```diff
@@ -38,7 +38,7 @@ Flow {
 			readonly property color textColor: {
 				const feedingToGrid = root.inputMode
 						&& (model.power || 0) < 0
-						&& Global.systemSettings.essFeedbackToGridEnabled()
+						&& Global.systemSettings.essFeedbackToGridEnabled
```
jpetrell, line 41: :partying_face:

jpetrell:
Getting text overlap, but I guess this is unrelated from these changes and will be sorted after https://github.com/victronenergy/gui-v2/pull/1037 is merged:
image

data/AcInputs.qml
```diff
@@ -50,6 +50,12 @@ QtObject {
 		}
 	}
 
+	function clampMeasurement(value) {
+		// If feed-in to grid is disabled, any values below zero are just noise; so in that case,
+		// user zero as the minimum value.
```
jpetrell, line 55: My first reaction is that ideally the noise would be fixed at the source, this kind of clamping potentially hides real issues. But I guess real-world measurements have noise and filtering is warranted.

jpetrell (commented):
- Nice! With the minus character gone the layouts look much cleaner :ok_hand: 

- With the minus character gone can we increase the horizontal space for the center gauge labels?

- When testing with the mock backend Shore AC input shows the arrow icon on both the left gauges and on the side panel, but generator AC input only on the left gauges. Though not sure if generators can even have feeding back

- Had Grid showing value "--W", the arrow was blue and pointing to the left direction, which feels weird as usually you either see green left arrow or blue right arrow. Could maybe hide the arrow if no electricity is flowing in either direction or show blue right arrow

image

blammit:
> With the minus character gone can we increase the horizontal space for the center gauge labels?

I don't know about that, because even with the minus gone, it could potentially look like this (now sure how realistic this is):

image


> When testing with the mock backend Shore AC input shows the arrow icon on both the left gauges and on the side panel, but generator AC input only on the left gauges. Though not sure if generators can even have feeding back

Good point, checked with Jesús and he said this can be ignored for generators.

> Had Grid showing value "--W", the arrow was blue and pointing to the left direction, which feels weird as usually you either see green left arrow or blue right arrow. Could maybe hide the arrow if no electricity is flowing in either direction or show blue right arrow

To avoid hiding the icon and changing the layout when that happens, I've made the icon grey in this case. What do you think?

image
image

jpetrell (approved):

PR #1051, opened 2024-05-16
Improve rendering of BarGauge

### LINKED ISSUE #1003: Tank gauges sometimes trail behind the actual remaining capacity
Customer [report](https://community.victronenergy.com/questions/279997/venus-os-v3408-available-for-testing.html): "Randomly the fill bar does not appear on a tank or several tanks, note black water at 27%. This has occurred viewing gui2 through VRM and on the Cerbo GX Touch. Screenshots taken a few minutes apart."

image
image
image

Also [reported](https://community.victronenergy.com/questions/279997/venus-os-v3408-available-for-testing.html?childToView=283848#answer-283848) with V3.40~13: "V3.40~13is still throwing a few inconsistent bar charts for the levels on both web view and the Cerbo touch 50 but improved from ~8.".

chriadam:
- Always calculate the _nextPos on _value change, regardless of item visibility
- If animating to _nextPos, ensure we stop any current animation and set the "from" value appropriate, to avoid "over animating".

Fixes #1003

### DIFF
--- a/components/BarGauge.qml
+++ b/components/BarGauge.qml
@@ -19,26 +19,33 @@ Rectangle {
 	property int orientation: Qt.Vertical
 	property bool animationEnabled
 
-	readonly property real _value: isNaN(value) ? 0 : Math.min(1.0, Math.max(0, value))
-
 	color: backgroundColor
 	width: orientation === Qt.Vertical ? Theme.geometry_barGauge_vertical_width_large : parent.width
 	height: orientation === Qt.Vertical ? parent.height : Theme.geometry_barGauge_horizontal_height
 	radius: orientation === Qt.Vertical ? width / 2 : height / 2
 
-	// Only update the nextPos when visible and width/height have been initialized.
-	readonly property real _nextPos: visible && width !== Infinity && height !== Infinity
+	readonly property real _value: isNaN(value) ? 0 : Math.min(1.0, Math.max(0, value))
+
+	// Only update the nextPos when the width/height have been initialized.
+	readonly property real _nextPos: (width !== Infinity && height !== Infinity)
 			? orientation === Qt.Vertical
 				? height - (height * _value)    // slide in from bottom to top
 				: -width + (width * _value)     // slide in from left to right
 			: 0
 
-	on_NextPosChanged: {
+	on_NextPosChanged: if (visible) _updateGauge()
+	onVisibleChanged: if (visible) _updateGauge()
+
+	function _updateGauge() {
 		if (animationEnabled) {
 			const animator = orientation === Qt.Vertical ? yAnimator : xAnimator
-			animator.to = _nextPos
-			animator.start()
-		} else if (visible) {
+			const currValue = orientation === Qt.Vertical ? fgRect.y : fgRect.x
+			if (!animator.running && currValue !== _nextPos) {
+				animator.from = currValue
+				animator.to = _nextPos
+				animator.start()
+			}
+		} else {
 			if (orientation === Qt.Vertical) {
 				fgRect.y = _nextPos
 			} else {
@@ -55,13 +62,16 @@ Rectangle {
 		height: bgRect.height
 		radius: bgRect.radius
 		color: "black" // opacity mask, not visible.
+		z: 1
 	}
 
 	Item {
 		id: sourceItem
+		layer.enabled: true
 		visible: false
 		width: parent.width
 		height: parent.height
+		z: 2
 
 		Rectangle {
 			id: fgRect
@@ -77,12 +87,14 @@ Rectangle {
 				target: fgRect
 				easing.type: Easing.InOutQuad
 				duration: Theme.animation_briefPage_sidePanel_sliderValueChange_duration
+				onRunningChanged: if (!running) Qt.callLater(_updateGauge) // if _nextPos changed during previous animation
 			}
 			YAnimator {
 				id: yAnimator
 				target: fgRect
 				easing.type: Easing.InOutQuad
 				duration: Theme.animation_briefPage_sidePanel_sliderValueChange_duration
+				onRunningChanged: if (!running) Qt.callLater(_updateGauge) // if _nextPos changed during previous animation
 			}
 		}
 	}
@@ -93,5 +105,7 @@ Rectangle {
 		maskEnabled: true
 		maskSource: maskRect
 		source: sourceItem
+		z: 3
 	}
+
 }
--- a/components/BarGauge.qml
+++ b/components/BarGauge.qml
@@ -8,59 +8,18 @@ import QtQuick.Effects as Effects
 import Victron.VenusOS
 import Victron.Gauges
 
-Rectangle {
-	id: bgRect
+BarGaugeBase {
+	id: root
 
-	property int valueType: VenusOS.Gauges_ValueType_NeutralPercentage
-	readonly property int valueStatus: Gauges.getValueStatus(_value * 100, valueType)
-	property color foregroundColor: Theme.statusColorValue(valueStatus)
-	property color backgroundColor: Theme.statusColorValue(valueStatus, true)
-	property real value: 0.0
-	property int orientation: Qt.Vertical
-	property bool animationEnabled
-
-	color: backgroundColor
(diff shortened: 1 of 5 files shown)


components/BarGauge.qml
```diff
@@ -19,26 +19,31 @@ Rectangle {
...
+	// Only update the nextPos when the width/height have been initialized.
+	readonly property real _nextPos: (width !== Infinity && height !== Infinity)
 			? orientation === Qt.Vertical
 				? height - (height * _value)    // slide in from bottom to top
 				: -width + (width * _value)     // slide in from left to right
 			: 0
 
-	on_NextPosChanged: {
+	on_NextPosChanged: if (visible) _updateGauge()
+	onVisibleChanged: if (visible) _updateGauge()
```
jpetrell, line 37: Nice and simple. :ok_hand:  :blush:

jpetrell:
I am still seeing some funky stuff.

Boat/Motorhome demo 1: Quantity label showing 68%, gauge 0%:
DSC_7776

Boat/Motorhome demo 2: Quantity label showing 67%, Gauge showing first ~75% in expanded mode, than after collapsing ~55%:
https://github.com/victronenergy/gui-v2/assets/2203667/74337a38-667e-4a1b-bd74-cab49ef579ca

chriadam:
Thanks, will investigate those ones

chriadam:
Setting `layer.enabled: true` in the `source` item seemed to resolve the issue for me, but more testing required.
I'm currently writing a "clipping" BarGauge implementation which doesn't use any MultiEffect, just in case...

chriadam:
> Setting `layer.enabled: true` in the `source` item seemed to resolve the issue for me, but more testing required.

Nope, still broken.

chriadam:
Updated with a new clipping bar gauge.  It only works for the Tank gauges, and not for other cases (e.g. overview page etc) because it requires a fully opaque "surface colour" to be specified which allows us to draw the rounded edge properly.

Needs more testing, as reproducing this issue can be a bit hit and miss, but I think it's ready for review.

components/ExpandedTanksView.qml
```diff
@@ -55,6 +55,7 @@ Rectangle {
 				totalRemaining: model.device.remaining
 
 				gauge: TankGauge {
+					surfaceColor: root.color
```
chriadam, line 58: this isn't working properly for some reason, will fix.

chriadam: removed.

jpetrell:
> The opacity mask shader effect sometimes didn't render properly on device

Great catch! :muscle:  There has also been discussion about gauges moving out of sync with the value labels in Bea's side panel gauge PRs. At the time Bea tried to investigate the issue, but had no luck. I thought the sync issue was related to labels showing watts and gauges being based on currents, but maybe we are talking about the same issue here? For one I vaguely remember only seeing the sync issues on GX devices.

jpetrell:
Could maybe refactor so BarGauge and ClippingBarGauge share the same base component, now this duplicates fair share of code?

This fixes tanks, but potentially other gauges are also affected. Ideally there was one fix that helps all the cases. Should we file a bug to Qt about broken MultiEffect?

If MultiEffect is culprit could try doing masks with just ShaderEffect to see if helps. Used to be simple in Qt 5:

```
layer.enabled: true
layer.samplerName: "maskSource"
layer.effect: ShaderEffect {
    property var colorSource: insides
    fragmentShader: "
        uniform lowp sampler2D colorSource;
        uniform lowp sampler2D maskSource;
        uniform lowp float qt_Opacity;
        varying highp vec2 qt_TexCoord0;
        void main() {
            gl_FragColor = texture2D(colorSource, qt_TexCoord0) * texture2D(maskSource, qt_TexCoord0).a * qt_Opacity;
        }"
}
```

Didn't have Qt 6 version lying around, so still need to apply [small syntax changes](https://www.basyskom.de/en/2021/qt6-how-to-port-shadereffects/) Qt 6 brought and move the shader to a (ugh) separate qsb file.

chriadam:
I created my own OpacityMaskEffect: 
[opacitymaskeffect-bargauge.diff.txt](https://github.com/victronenergy/gui-v2/files/15372761/opacitymaskeffect-bargauge.diff.txt)

But it doesn't fix the issue - it just doesn't render, sometimes.  I suspect the problem is the old Mali400 drivers, it has a few issues.

Note that the issue you noted above: "gauges moving out of sync with the value labels" is likely the "double animation" issue which I've already fixed.  The MultiEffect / OpacityMaskEffect issue I'm seeing is that the bar doesn't get rendered at all (well, the foreground part of the bar) in some cases.

I will create a base type which encapsulates all of the logic and properties.
In future, once we port the entire UI over to using opaque colours, we can simply use the clipping gauge for all cases, I think.

chriadam:
Done.  FYI the PR to convert over to fully opaque colours was https://github.com/victronenergy/gui-v2/pull/689 but it's quite old and I'm certain it will need to be updated now, if we want to go that way.

jpetrell (approved):
LGTM, didn't see any issues with the tanks anymore.

> But it doesn't fix the issue - it just doesn't render, sometimes. I suspect the problem is the old Mali400 drivers, it has a few issues.

Thanks for testing. :pray: 

> "gauges moving out of sync with the value labels" is likely the "double animation" issue which I've already fixed

Cool! There is still some other issue lurking somewhere as still seeing gauge updating differently to value on GX device:
https://github.com/victronenergy/gui-v2/assets/2203667/989aeb5b-b9fc-4b59-bcef-54fe0dfb2df5

Doesn't reproduce with the mock backend, where gauge and quantity label always update in sync.

> I will create a base type which encapsulates all of the logic and properties.

Looks good. :+1: 

Could add comment somewhere why we need to BarGauge implementations.

chriadam:
Hmm, interesting that the load graph matches the bar gauge, but the label is completely different to both of them...  I will take another look tomorrow.

chriadam:
I took a look into the side panel thing, and the reason is that the label shows the current, whereas the bar (and graph) shows the ratio.

e.g.
```
qml: XXXXXXXXXXX acLR current = 0.140000000596, maxCurrent = 2.154, ratio = 0.06499535775116064, label: 0.1
qml: XXXXXXXXXXX acLR current = 0.370000004768, maxCurrent = 2.154, ratio = 0.17177344696750232, label: 0.4
qml: XXXXXXXXXXX acLR current = 0.259999990463, maxCurrent = 2.154, ratio = 0.12070565945357474, label: 0.3
qml: XXXXXXXXXXX acLR current = 0.449999988079, maxCurrent = 2.154, ratio = 0.20891364349071495, label: 0.4
qml: XXXXXXXXXXX acLR current = 0.379999995232, maxCurrent = 2.154, ratio = 0.17641596807428042, label: 0.4
qml: XXXXXXXXXXX acLR current = 0.469999998808, maxCurrent = 2.154, ratio = 0.21819869953946147, label: 0.5
qml: XXXXXXXXXXX acLR current = 0.40000000596, maxCurrent = 2.154, ratio = 0.1857010241225627, label: 0.4
qml: XXXXXXXXXXX acLR current = 0.819999992847, maxCurrent = 2.154, ratio = 0.3806870904582173, label: 0.8
qml: XXXXXXXXXXX acLR current = 0.52999997139, maxCurrent = 2.154, ratio = 0.24605384001392758, label: 0.5
qml: XXXXXXXXXXX acLR current = 0.439999997616, maxCurrent = 2.154, ratio = 0.20427112238440112, label: 0.4
qml: XXXXXXXXXXX acLR current = 0.490000009537, maxCurrent = 2.154, ratio = 0.227483755588208, label: 0.5
qml: XXXXXXXXXXX acLR current = 0.689999997616, maxCurrent = 2.154, ratio = 0.320334260731662, label: 0.7
```

So they can "get out of sync" if the maxCurrent suddenly changes (due dynamic ranging?  not sure).

chriadam:
Everyone happy for me to merge this one as-is?

blammit:
> I took a look into the side panel thing, and the reason is that the label shows the current, whereas the bar (and graph) shows the ratio.
> 
> So they can "get out of sync" if the maxCurrent suddenly changes (due dynamic ranging? not sure).

Yes that's right, the max current could change due to dynamic ranging or if the AC input gauge changes between import-only and import-export.

PR #1053, opened 2024-05-16
Set the isMulti property correctly for inverter/chargers

### LINKED ISSUE #1040: No way to select "Inverter only" or "Charger only" options from Inverter / Charger mode setting
[Reported](https://community.victronenergy.com/questions/279997/venus-os-v3408-available-for-testing.html?childToView=283848#answer-283848) by Kevin. "gui-v2 dialog that selects inverter mode is showing Charger only and Inverter only grayed out and not selectable. The only selections permitted are off and on. My system uses a Multi Compact 120v unit. All 4 selections are shown and function in gui-v1 and also in GuiMods. Tested this without any third party mods just to be sure there is no interaction."

Looking at the  Inverter / Charger mode dialog code the items "Inverter only" or "Charger only" should be enabled if isMulti parameter is true, but the parameter is never defined.

chriadam:
A pure inverter doesn't have AC inputs, only outputs. An inverter/charger can have AC inputs, and is considered multi. A multi device can be set explicitly to be in either inverter or charger mode via the appropriate dialog, but the mode settings are enabled or disabled based on the isMulti condition.

Fixes #1040

### DIFF
--- a/components/InverterModeDialogLauncher.qml
+++ b/components/InverterModeDialogLauncher.qml
@@ -15,12 +15,18 @@ QtObject {
 			? Global.inverterChargers.inverterChargerModeToText(_modeItem.value)
 			: Global.inverterChargers.inverterModeToText(_modeItem.value)
 
+	readonly property bool isMulti: _numberOfAcInputs.value === undefined ? false : (_numberOfAcInputs.value > 0)
+
 	property VeQuickItem _isInverterChargerItem: VeQuickItem {
 		uid: root.serviceUid + "/IsInverterCharger"
 	}
 
 	property VeQuickItem _modeItem: VeQuickItem {
 		uid: root.serviceUid + "/Mode"
+	}
+
+	readonly property VeQuickItem _numberOfAcInputs: VeQuickItem {
+		uid: root.serviceUid + "/Ac/NumberOfAcInputs"
 	}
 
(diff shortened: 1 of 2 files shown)


components/InverterModeDialogLauncher.qml
```diff
@@ -23,6 +25,10 @@ QtObject {
 		uid: root.serviceUid + "/Mode"
 	}
 
+	readonly property VeQuickItem _numberOfAcInputs: VeQuickItem {
+		uid: root.serviceUid + "/Ac/NumberOfAcInputs"
```
chriadam, line 29: might this result in a non-existent path in some cases?  If so, how can I avoid that?

jpetrell (approved):
LGTM. Works nicely on GX device running demo modes.

Could also fix the current mock multi devices Quattro 48/5000/70-2x100 and MultiPlus 12/3000/120-5 to show Inverter-only and Charger-only options. Currently, the mock backend sets MultiPlus 12/3000/120-5 mode to InverterCharger_Mode_InverterOnly, but that is disabled in the inverter/charger mode dialog (maybe "/Ac/NumberOfAcInputs" hasn't been defined by the mocking?).

jpetrell:
Good to get this to the release so merging already.

PR #1057, opened 2024-05-16
Switch default font

jpetrell:


blammit (approved):

blammit:
@jpetrell minor thing, noticed this reference is still present:

`src/language.cpp:    // On wasm, the scripts for these languages cannot be rendered by the default Museo font.`

PR #1058, opened 2024-05-16
Clean up couple of boot warnings

jpetrell:
```
file:///data/home/root/install/bin/Victron/VenusOS/data/common/AcInputPhaseModel.qml:18: TypeError: Cannot call method 'clampMeasurement' of undefined
file:///data/home/root/install/bin/Victron/VenusOS/data/common/AcInputPhaseModel.qml:18: TypeError: Cannot call method 'clampMeasurement' of undefined
```

### DIFF
--- a/data/common/AcInputPhaseModel.qml
+++ b/data/common/AcInputPhaseModel.qml
@@ -15,7 +15,10 @@ ListModel {
 		if (phaseIndex < 0 || phaseIndex >= count) {
 			return
 		}
-		const v = propertyValue === undefined ? NaN : Global.acInputs.clampMeasurement(propertyValue)
+
+		const v = propertyValue === undefined ? NaN
+				: (Global.acInputs ? Global.acInputs.clampMeasurement(propertyValue) : propertyValue)
+
 		setProperty(phaseIndex, propertyName, v)
 		phaseValueChanged(phaseIndex, propertyName, v)
 	}


data/common/AcInputPhaseModel.qml
```diff
@@ -15,7 +15,11 @@ ListModel {
 		if (phaseIndex < 0 || phaseIndex >= count) {
 			return
 		}
-		const v = propertyValue === undefined ? NaN : Global.acInputs.clampMeasurement(propertyValue)
+
+		const v = propertyValue === undefined ? NaN
+											  : Global.acInputs ? Global.acInputs.clampMeasurement(propertyValue)
+																: propertyValue
```
jpetrell, line 21: Looks funny here, but on Qt Creator
image

blammit: I would just ignore the default Qt Creator indentation and left-align like:

```
const v = propertyValue === undefined ? NaN
        : (Global.acInputs ? Global.acInputs.clampMeasurement(propertyValue) : propertyValue)
```

blammit (approved):
Minor comment, LGTM. Thanks!

data/common/AcInputPhaseModel.qml
jpetrell: Thanks! Much nicer

PR #1065, opened 2024-05-16
Use 0 as mid-point for AC in graph when feed-in is enabled

### LINKED ISSUE #1060: Brief page: Side panel grid graph
Shouldn't the graph for the grid start in the middle, if grid feed in is enabled?

grafik

Like this:

grafik

blammit:
Fixes #1060

### DIFF
--- a/pages/BriefSidePanel.qml
+++ b/pages/BriefSidePanel.qml
@@ -137,18 +137,11 @@ exported power v  0.4 |   /
 			aboveThresholdFillColor: Theme.color_blue   // warning color is not needed for inputs
 			belowThresholdFillColor: _graphShowsFeedIn ? Theme.color_green : Theme.color_blue
 			initialModelValue: 0
-			threshold: {
-				// For a system with a min current below zero (i.e. it sometimes exports), the
-				// threshold is the 0 amp mark within the min/max range.
-				if (acInputGraphRange.minimumCurrent < 0) {
-					const range = acInputGraphRange.maximumCurrent - acInputGraphRange.minimumCurrent
-					if (range !== 0) {
-						return Math.abs(acInputGraphRange.minimumCurrent) / range
-					}
-				}
-				// For a system that only imports, no threshold is required.
-				return 0
-			}
+
+			// For a system that only imports, no threshold is required.
+			// For a system that sometimes exports (i.e. can have values below zero), the threshold
+			// is the mid-point, which should be zero.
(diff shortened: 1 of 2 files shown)


pages/BriefSidePanel.qml
```diff
@@ -166,7 +166,15 @@ exported power v  0.4 |   /
 				id: acInputGraphRange
 
 				phaseModel: root.visible ? Global.acInputs.activeInput.phases : null
-				minimumCurrent: Global.acInputs.activeInputInfo.minimumCurrent
+				minimumCurrent: {
+					if (Global.acInputs.activeInputInfo.minimumCurrent < 0
+							&& Global.acInputs.activeInputInfo.maximumCurrent > 0) {
+						// If the values may go below zero, then ensure that zero is the vertical
+						// mid-point of the graph.
+						return Global.acInputs.activeInputInfo.maximumCurrent * -1
```
chriadam, line 174: what if the magnitude of minimumCurrent is far greater than that of maximumCurrent?  e.g. if minimumCurrent is -500 but maximumCurrent is 200?

blammit: Oh, right -- have fixed this to use max(abs(min_current), max_current).

chriadam: sounds good - the only issue that zero will no longer be the mid-point, but maybe that's not a problem?

chriadam: Ah, I see, now you use that abs value as the max and min value.  Nice.

chriadam (approved):

mr-manuel:
Somehow nothing changed on my end.

grafik

mr-manuel:
I also think, that the graph doesn't have to be symmetrical. It can also be that the min is `-10A` and the max `30A`. This would result in the 0 mid-point to be at 1/4 from the bottom. Maybe this should be already like this?

blammit:
> I also think, that the graph doesn't have to be symmetrical. It can also be that the min is -10A and the max 30A. This would result in the 0 mid-point to be at 1/4 from the bottom. Maybe this should be already like this?

That's how it should be working at the moment in `main`, yes.

I recall talking to Jesús about this previously, and he requested that 0 be the mid-point.

mr-manuel:
I think I now understood the problem better. The mid-point is correct, but the graph fill is not. The graph fill should always be towards 0 and not the bottom. See https://github.com/victronenergy/gui-v2/issues/1060#issue-2300922125

jpetrell:
I'll get these improvements merged for the release, the graph fill can be fixed separately.

PR #1067, opened 2024-05-17
Use DoubleValidator decimals to enforce min/max range precision

blammit:
This prevents the user from entering values with a precision that does not match the specified number of decimals.

### DIFF
--- a/components/listitems/ListQuantityField.qml
+++ b/components/listitems/ListQuantityField.qml
@@ -14,15 +14,7 @@ ListTextField {
 	property int decimals: Units.defaultUnitPrecision(unit)
 
 	suffix: Units.defaultUnitString(unit)
-	textField.validator: DoubleValidator {}
+	textField.validator: DoubleValidator { decimals: root.decimals }
 	textField.inputMethodHints: Qt.ImhDigitsOnly
-	textField.text: tryAcceptInput(value)
-
-	tryAcceptInput: function(inputText) {
-		const n = Number(inputText)
-		if (isNaN(n)) {
-			return "--"
-		}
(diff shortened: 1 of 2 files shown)


mr-manuel:
Unfortunately no change.

mr-manuel:
I discovered, that it does not work anymore since the unis is behind the value. In the version without the unit `A` behind the value the problem does not exist.

mr-manuel:
After more tests I found out that it has to do with the requested website language (not browser language itself). If I'm switching from `de-de` to `en-us` the problem disappeared. After switching back to `de-de` the problem appeared again. The webpage with the wasm build has to be refreshed after the change.

grafik

jpetrell (approved):

jpetrell:
Doesn't fix @mr-manuel's issue, but overall the improvement seems nice.

mr-manuel:
Adding reference for better understanding: https://github.com/victronenergy/gui-v2/pull/1043#issuecomment-2114044833

mr-manuel:
I want also to add, that not only the display is wrong after entering the value, but also that the value is not saved that way.

blammit:
> After more tests I found out that it has to do with the requested website language (not browser language itself). If I'm switching from de-de to en-us the problem disappeared. After switching back to de-de the problem appeared again. The webpage with the wasm build has to be refreshed after the change.

@mr-manuel Thanks! I looked into this further, and the problem is that the number is not formatted according to the application locale. I made some fixes for this, but it requires further investigation/discussion overall in terms of locale and number formatting. Have filed https://github.com/victronenergy/gui-v2/issues/1108 for this.

PR #1074, opened 2024-05-18
Fix bug in spin box for Generator ac load limits

DanielMcInnes:


components/dialogs/NumberSelectorDialog.qml
```diff
@@ -64,6 +64,7 @@ ModalDialog {
 				from: root.from * root._multiplier()
 				to: root.to * root._multiplier()
 				stepSize: root.stepSize * root._multiplier()
+				value: root.value * root._multiplier()
```
blammit, line 67: Is this needed? The `onAboutToShow` handler sets this `value` as well so that would overwrite this binding.

jpetrell:
Seems other Generator conditions are affected as well, on my device settings which display different value on the list item compared to the spin box:

Battery SOC
- Stop when Battery SOC is higher than (90% vs. 81%)
- Stop value during quiet hours (90% vs. 91%)
- Start value during quiet hours (90% vs. 89%)

Battery current
- Start when Battery current is higher than (10.5A vs. 6.0A)
- Start value during quiet hours (20.5A vs. 16.0A)
- Stop value during quiet hours (15.5A vs. 1.0A)

Battery voltage
- Start when Battery voltage is lower than (11.5V vs. 11.0V)
- Start value during quiet hours (11.9V vs. 11.0V)
- Stop when Battery voltage is higher than (12.4V vs. 12.0V)
- Stop value during quiet hours (12.4V vs. 12.0V)

AC load (says AC output on the page, should be aligned)
- Start when power is higher than (1600W vs. 805W)
- Start value during quiet hours (1900W vs. 1210W)

components/dialogs/NumberSelectorDialog.qml
DanielMcInnes: Yes, it's needed. 

When we create a NumberSelectorDialog with 'value: 700', and 'spinBox.from: 500', 'spinBox.to: 1000', 
1. 'NumberSelectorDialog.value' gets initialized to 700. All good.
2. 'spinBox.value' gets initialized with a default value of 0. I think because 'spinBox.from' is set to 500, value gets updated to meet this minimumValue.
3. 'spinBox.onValueChanged()' gets triggered, with the new value of 500. This then updates root.value to 500, which is where the problem lies.
4. 'onAboutToShow()' gets called, which sets spinBox.value to our (incorrect) value of 500.

blammit: > 'onAboutToShow()' gets called, which sets spinBox.value to our (incorrect) value of 500.

Are you saying onAboutToShow() shouldn't set the value? If so then should that line be removed? It just seems odd to set the initial value in two places (both onAboutToShow() and in this binding).

DanielMcInnes: Oh right, I see what you mean, it does seem odd. I've removed setting the initial value in `onAboutToShow()`.

blammit: Yeah, I can't recall now why that initial value is set to a rounded number. My first thought was that it would cause the displayed value to be different from the data value.  But @jpetrell I see that you added that in 365ee5786f56946ee94b36d4c4b892ed1f05a2e2 for a bug fix, maybe you can comment?

blammit (commented):
I'm seeing this when clicking on the spinbox in Settings > Test:

```qrc:/qt/qml/Victron/VenusOS/components/dialogs/NumberSelectorDialog.qml:50:13: QML SpinBox: Binding loop detected for property "value"```

**Comment by DanielMcInnes** [comment]
> I'm seeing this when clicking on the spinbox in Settings > Test:
> 
> `qrc:/qt/qml/Victron/VenusOS/components/dialogs/NumberSelectorDialog.qml:50:13: QML SpinBox: Binding loop detected for property "value"`

Fixed now

**Comment by DanielMcInnes** [comment]
> Seems other Generator conditions are affected as well, on my device settings which display different value on the list item compared to the spin box:
> 
> Battery SOC
> 
>     * Stop when Battery SOC is higher than (90% vs. 81%)
> 
>     * Stop value during quiet hours (90% vs. 91%)
> 
>     * Start value during quiet hours (90% vs. 89%)
> 
> 
> Battery current
> 
>     * Start when Battery current is higher than (10.5A vs. 6.0A)
> 
>     * Start value during quiet hours (20.5A vs. 16.0A)
> 
>     * Stop value during quiet hours (15.5A vs. 1.0A)
> 
> 
> Battery voltage
> 
>     * Start when Battery voltage is lower than (11.5V vs. 11.0V)
> 
>     * Start value during quiet hours (11.9V vs. 11.0V)
> 
>     * Stop when Battery voltage is higher than (12.4V vs. 12.0V)
> 
>     * Stop value during quiet hours (12.4V vs. 12.0V)
> 
> 
> AC load (says AC output on the page, should be aligned)
> 
>     * Start when power is higher than (1600W vs. 805W)
> 
>     * Start value during quiet hours (1900W vs. 1210W)

Hi @jpetrell, could you retest this please? It's behaving nicely in my setup

**Comment by blammit** [comment]
@DanielMcInnes should this be taken out of draft?

**Comment by jpetrell** [comment]
> Hi @jpetrell, could you retest this please? It's behaving nicely in my setup

Still seeing on Battery SOC conditions Start value during quiet hours (90%  vs. 89%) and Stop value during quiet hours (90% vs. 91%). Seems tricky to verify if correct values may appear on one configuration, but be broken on others so suspect other fields may still have issues also.

File: pages/settings/GeneratorCondition.qml [hunk 2]
```diff
@@ -82,8 +83,10 @@ ListNavigationItem {
 						dataItem.uid: bindPrefix + "/QuietHoursStartValue"
 						suffix: root.unit
 						decimals: root.decimals
-						from: quietHoursStopValue.dataItem.isValid && root.startValueIsGreater ? quietHoursStopValue.value + stepSize : 0
-						to: quietHoursStopValue.dataItem.isValid && !root.startValueIsGreater ? quietHoursStopValue.value - stepSize : 100
+						stepSize: root.stepSize
+						from: quietHoursStopValue.dataItem.isValid && root.startValueIsGreater ? quietHoursStopValue.value + stepSize : dataItem.defaultMin
+						to: quietHoursStopValue.dataItem.isValid && !root.startValueIsGreater ? quietHoursStopValue.value - stepSize : dataItem.defaultMax
+						onValueChanged: console.log(dataItem.uid, value, from, to)
```
jpetrell, line 89: Drop console.log

components/dialogs/NumberSelectorDialog.qml
```diff
@@ -26,7 +26,7 @@ ModalDialog {
 	}
 
 	onAboutToShow: {
-		spinBox.value = Math.round(value * _multiplier())
+		spinBox.initializeValue(Math.round(value * _multiplier()))
```
jpetrell, line 29: DateSelector, TImeSelector and PageTankShape have bunch of SpinBox {} in use. Since they don't call initializeValue(() does this mean they are never initialized?

jpetrell: Personally would try avoid requiring imperative initialization functions.

Could we do

```
id: root
onValueChanged: {
    updateAllowed = false
    spinBox.value = value
    updateAllowed = true
}
...
SpinBox {
    onValueChanged: if (updateAllowed) root.value = Number(value / root._multiplier())
}
```

DanielMcInnes:
> @DanielMcInnes should this be taken out of draft?

Done

components/dialogs/NumberSelectorDialog.qml
DanielMcInnes: > Personally would try avoid requiring imperative initialization functions.
> 
> Could we do
> 
> ```
> id: root
> onValueChanged: {
>     updateAllowed = false
>     spinBox.value = value
>     updateAllowed = true
> }
> ...
> SpinBox {
>     onValueChanged: if (updateAllowed) root.value = Number(value / root._multiplier())
> }
> ```

I tried this, and it doesn't work, as root.onValueChanged() gets called before the spinbox is properly constructed, and the changes are lost. I got rid of the imperative initialization function, and do it all in `spinBox.onCompleted`

pages/settings/GeneratorCondition.qml
DanielMcInnes: done

components/dialogs/NumberSelectorDialog.qml
DanielMcInnes: > DateSelector, TImeSelector and PageTankShape have bunch of SpinBox {} in use. Since they don't call initializeValue(() does this mean they are never initialized?

I see your point, it does seem to imply that those spin boxes are never initialized. Got rid of the imperative initialization function.

DanielMcInnes:
> @DanielMcInnes should this be taken out of draft?

done

jpetrell:
Something funky still going on, the battery conditions do not always match between the list item and the spin box dialog.

components/listitems/ListSpinBox.qml
```diff
@@ -29,6 +30,13 @@ ListButton {
 
 	onClicked: Global.dialogLayer.open(numberSelectorComponent, {value: value})
 
+	RangeModel {
+		id: rangeModel
+		minimumValue: root.from
+		maximumValue: root.to
+		value: !dataItem.isValid ? 0 : dataItem.value
```
jpetrell, line 37: Personally I find value: `dataItem.isvalid ? dataItem.value : 0` more readable.

jpetrell (approved):
- LGTM Tested, didn't see issues with the spin box using Generator conditions sub-pages
- Bit unfortunate we need to work-around the whole thing by introducing a new model class
- Nitpick: would be good to draw spin box increment / decrement buttons as disabled when at the end of the range
- There are few plain SpinBox {} elements that don't use ListSpinBox, wonder if the issue could happen there also
- I guess QML bindings don't work well with such range limits as it is hard to control the order in which valid from, to and value properties are set. Qt Quick Controls SpinBox would really need some kind of `enforceRange` option that you could use to choose when the range limits start to be applied, e.g. by byinding `enforceRange` to bunch of isValid properties so that only after each property has been successfully initialized would the value range be enforced.

components/listitems/ListSpinBox.qml
DanielMcInnes: I agree, done

DanielMcInnes:
> Something funky still going on, the battery conditions do not always match between the list item and the spin box dialog.

Fixed now.

blammit (approved):
Is `RangeModel` just doing a similar thing to `Utils.scaleNumber()`? The solution seems a bit heavyweight for this, but I suppose using the same solution as gui-v1 should be good too.

Great to have the spinbox issue fixed, that has been bugging me.

DanielMcInnes:
> Is `RangeModel` just doing a similar thing to `Utils.scaleNumber()`? The solution seems a bit heavyweight for this, but I suppose using the same solution as gui-v1 should be good too.
> 
> Great to have the spinbox issue fixed, that has been bugging me.

Yes, it's the same basic thing

PR #1114, opened 2024-05-20
Extend text field to fit longer placeholder text

jpetrell:
The default size of text field is too small for some placeholder text translations. Truncation was added earlier, but it is better to show the string in full instead (if it still fits in the maximum width limits).

Contributes to #995.

### DIFF
--- a/components/listitems/ListTextField.qml
+++ b/components/listitems/ListTextField.qml
@@ -58,9 +58,9 @@ ListItem {
 		property string _textWhenFocused
 		property bool _accepted
 
-		width: Math.max(
-				Theme.geometry_listItem_textField_minimumWidth,
-				Math.min(implicitWidth + leftPadding + rightPadding, Theme.geometry_listItem_textField_maximumWidth))
+		width: Math.max(Theme.geometry_listItem_textField_minimumWidth,
+						Math.min(Theme.geometry_listItem_textField_maximumWidth,
+								 implicitWidth + leftPadding + rightPadding))
 		enabled: root.enabled
 		text: dataItem.isValid ? dataItem.value : ""
 		rightPadding: suffixLabel.text.length ? suffixLabel.implicitWidth : leftPadding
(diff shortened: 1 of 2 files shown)


components/listitems/ListTextField.qml
```diff
@@ -71,9 +71,15 @@ ListItem {
 			keyEvent.accepted = false
 		}
 
-		width: Math.max(
-				Theme.geometry_listItem_textField_minimumWidth,
-				Math.min(implicitWidth + leftPadding + rightPadding, Theme.geometry_listItem_textField_maximumWidth))
+		width: Math.max(Theme.geometry_listItem_textField_minimumWidth,
```
blammit, line 74: Could add some comments or break up the binding with named variables, to make it easier to understand?

Also, it looks like part of the bug is that the placeholder text (in controls/TextField.qml) is not being sized to the same width as the text field, so that should be fixed too.

jpetrell: > placeholder text (in controls/TextField.qml)

Oh there it is! Assumed placeholder text ws handled by Quick Controls.

> Could add some comments or break up the binding with named variables, to make it easier to understand?

Got the binding much simpler now. I assume min(maximumSize, max(minimumSize, implicitSize) pattern doesn't need much explanation.

> it looks like part of the bug is that the placeholder text (in controls/TextField.qml) is not being sized to the same width as the text field

Not sure what issue you mean? To my eye placeholder seems to anchor correctly to the available space, truncating with elide if needed.

blammit: > Not sure what issue you mean? To my eye placeholder seems to anchor correctly to the available space, truncating with elide if needed.

Never mind, I was looking at the original bug description, but I guess the screenshot was outdated. LGTM :)

blammit (approved):

jpetrell:
Rebased.

PR #1116, opened 2024-05-20
Draw with increased precision if upscaling is required

### LINKED ISSUE #1020: Icons and gauges look  fuzzy when opening GUI v2 on the browser on larger displays
Filing this one so we don't forget. As part of #900 upscaling gui v2 (wasm) was enabled to avoid the common complaint of gui v2 opening smaller than the popup space that VRM website reserves for it. The downside of upscaling is that icons and gauges that paint to a shader texture now look fuzzy. Earlier quantity labels were fuzzy also, but that got fixed by #941 once we stopped drawing them on a bitmap.

Occurs on large tablets like iPad Pro in landscape orientation, and on some desktop displays (depending on the desktop GUI scaling ratio).

Here is an example (difficult to see unless opened in full-screen):
Group 461

jpetrell:
image

Fixes #1020.

### DIFF
--- a/Main.qml
+++ b/Main.qml
@@ -17,10 +17,11 @@ Window {
 	title: qsTrId("venus_os_gui")
 	color: Global.allPagesLoaded && !!guiLoader.item ? guiLoader.item.mainView.backgroundColor : Theme.color_page_background
 
-	width: Qt.platform.os != "wasm" ? Theme.geometry_screen_width : Screen.width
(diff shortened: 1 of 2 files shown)


chriadam (approved):
Nice

mr-manuel:
The icons does not seem sharp to me or is this only for a few icons? It's better then before but still not sharp.

grafik

Anyway, most of the users have a small screen.

PR #1117, opened 2024-05-21
Consistently format numbers according to application locale 

### LINKED ISSUE #1108: Numbers are inconsistently formatted in non-en locales
gui-v2 should format numbers consistently in non-en locales.

For example, in Chrome if you set the "Preferred language" to German, then when running gui-v2 in wasm, then the Brief and Overview pages show quantities with the '.' period as a decimal separator:

image

However, when opening an Inverter/Charger control card, the current limit spinbox uses a comma (`,`) as the decimal separator:

image

gui-v2 should just use one or the other instead of mixing them. Presumably it should always respect the formatting for the system locale.

blammit:
Set application locale and use locale-specific number formatting

Call QLocale::setDefault() when the current language is applied. This allows Qt.locale() to return the locale for the current language, and also this locale will be used as the default locale, e.g. when formatting date/times and floating point numbers for spinboxes and text fields.

In QML, always use toLocaleString() instead of toFixed() to format numbers. The former respects the decimal separator for the locale, while the latter always uses a period as the decimal separator.

In C++, use QLocale::toString() to use the correct locale for number formatting.

Fixes #1108

### DIFF
--- a/Global.qml
+++ b/Global.qml
@@ -14,7 +14,6 @@ QtObject {
 	property var mainView
 	property var mockDataSimulator    // only valid when mock mode is active
 	property var dataManager
-	property var locale: Qt.locale()  // TODO: read from settings
 	property VeQItemTableModel dataServiceModel: null
 	property var firmwareUpdate
 
@@ -88,7 +87,6 @@ QtObject {
 		mainView = null
 		mockDataSimulator = null
 		dataManager = null
-		locale = Qt.locale()
 		dataServiceModel = null
 		firmwareUpdate = null
 		inputPanel = null
--- a/components/listitems/ListQuantityField.qml
+++ b/components/listitems/ListQuantityField.qml
@@ -14,7 +14,10 @@ ListTextField {
 	property int decimals: Units.defaultUnitPrecision(unit)
 
 	suffix: Units.defaultUnitString(unit)
-	textField.validator: DoubleValidator { decimals: root.decimals }
-	textField.inputMethodHints: Qt.ImhDigitsOnly
-	textField.text: Number(value).toFixed(decimals)
+	textField.validator: DoubleValidator {
+		decimals: root.decimals
+		locale: Units.numberFormattingLocaleName
+	}
+	textField.inputMethodHints: Qt.ImhFormattedNumbersOnly
+	textField.text: Units.formatNumber(root.value, root.decimals)
 }
--- a/pages/settings/PageGps.qml
+++ b/pages/settings/PageGps.qml
@@ -19,14 +19,17 @@ Page {
 
 		switch (fmt) {
 		case VenusOS.GpsData_Format_DecimalDegrees: // e.g. 52.34489
-			return val.toFixed(6)
+			return Units.formatNumber(val, 6)
 		case VenusOS.GpsData_Format_DegreesMinutes: // e.g. 52° 20.693 N
-			return "%1° %2 %3".arg(Math.floor(degrees).toFixed()).arg(minutes.toFixed(4)).arg(direction)
+			return "%1° %2 %3"
+				.arg(Units.formatNumber(Math.floor(degrees)))
+				.arg(Units.formatNumber(minutes, 4))
+				.arg(direction)
 		default: // VenusOS.GpsData_Format_DegreesMinutesSeconds e.g. 52° 20' 41.6" N
 			return "%1° %2' %3\" %4"
-					.arg(Math.floor(degrees).toFixed())
-					.arg(Math.floor(minutes).toFixed())
-					.arg(seconds.toFixed(1))
+					.arg(Units.formatNumber(Math.floor(degrees)))
+					.arg(Units.formatNumber(Math.floor(minutes)))
+					.arg(Units.formatNumber(seconds, 1))
 					.arg(direction)
 		}
 	}
@@ -73,19 +76,19 @@ Page {
 					if (speedUnit.value === "km/h") {
 						//: GPS speed data, in kilometers per hour
 						//% "%1 km/h"
-						return qsTrId("settings_gps_speed_kmh").arg((dataItem.value * 3.6).toFixed(1))
+						return qsTrId("settings_gps_speed_kmh").arg(Units.formatNumber(dataItem.value * 3.6, 1))
 					} else if (speedUnit.value === "mph") {
 						//: GPS speed data, in miles per hour
 						//% "%1 mph"
-						return qsTrId("settings_gps_speed_mph").arg((dataItem.value * 2.236936).toFixed(1))
+						return qsTrId("settings_gps_speed_mph").arg(Units.formatNumber(dataItem.value * 2.236936, 1))
 					} else if (speedUnit.value === "kt") {
 						//: GPS speed data, in knots
 						//% "%1 kt"
-						return qsTrId("settings_gps_speed_kt").arg((dataItem.value * (3600/1852)).toFixed(1))
+						return qsTrId("settings_gps_speed_kt").arg(Units.formatNumber(dataItem.value * (3600/1852), 1))
 					} else {
 						//: GPS speed data, in meters per second
 						//% "%1 m/s"
-						return qsTrId("settings_gps_speed_ms").arg(dataItem.value.toFixed(2))
+						return qsTrId("settings_gps_speed_ms").arg(Units.formatNumber(dataItem.value, 2))
 					}
 				}
 			}
@@ -94,7 +97,7 @@ Page {
 				//% "Course"
 				text: qsTrId("settings_gps_course")
 				dataItem.uid: bindPrefix + "/Course"
-				secondaryText: dataItem.isValid ? "%1°".arg(dataItem.value.toFixed(1)) : ""
+				secondaryText: dataItem.isValid ? "%1°".arg(Units.formatNumber(dataItem.value, 1)) : ""
 			}
 
 			ListTextItem {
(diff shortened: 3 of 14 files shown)


blammit:
Need to update to set the application locale.

blammit:
> Need to update to set the application locale.

Updated, ready for review.

blammit:
There are still some issues here. Will close for now.

blammit:
Ready for review now. Basically, the number formatting should always respect the locale for the selected language, regardless of the system or browser setting.

To test:
- Set the preferred Chrome language to e.g. German
- Verify the comma is used as a decimal separator (rather than a period) in the Brief and Overview pages, and also in Settings > Debug > Test (e.g. in range slider values, spinbox popup, ListQuantityField)

mr-manuel:
Please see https://github.com/victronenergy/gui-v2/issues/1108#issuecomment-2123857352

Global.qml
```diff
@@ -14,7 +14,7 @@ QtObject {
 	property var mainView
 	property var mockDataSimulator    // only valid when mock mode is active
 	property var dataManager
-	property var locale: Qt.locale()  // TODO: read from settings
+	property var locale: Qt.locale(Language.localeName)
```
jpetrell, line 17: Initial reaction was that shouldn't Qt fetch the currently active locale by default, seeing QLocale::setDefault(() in language.cpp, but I guess this one is here to force update on language switch.

jpetrell: Why not just use Language.locale? Seems Global.locale is used only in one place.

Also, Language could be renamed to Locale since this it handles now more than just language. Though I guess Locale name is reserved by Qt so dunno. Maybe not worth the effort.

jpetrell:
This is a controversial change, some will like it, others won't.

- Some Arabic users may prefer Latin numerals. Also Arabic numbers are not currently monospaced. Would maybe use Latin numbers always for now.
- Group separators help with readability, but mess up with quantity label alignments we get from the monospacing. Similar to decimal point, but we cannot avoid decimal point. Group separators we could avoid.
- Some European language users are used to English decimal point. Not saying we should necessarily add a setting, just something that community may raise.

image

tests/units/tst_units.qml
```diff
@@ -186,7 +185,7 @@ TestCase {
 
 		// Keep the scale when going 10% below the threshold
 		expect(unit, 9.5675, "9.568", "kWh", true) // hysteresis = true
-		expect(unit, 8.967, "8967", "Wh", true) // hysteresis = true
+		expect(unit, 8.967, "8,967", "Wh", true) // hysteresis = true
```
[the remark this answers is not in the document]
jpetrell: Ran the tst_units with different languages. With simple decimal point and group separator replacement (see branch https://github.com/victronenergy/gui-v2/tree/jpetrell/number-formatting) these languages pass the test English, Czech, Danish, German, French, Italian, Dutch, Russian, Romanian, Swedish, Thai, Turkish, Ukrainian, Chinese, 

- Arabic fails due to using Arabic numbers
- Spanish returns "." and Poland white space as the group separator, but neither apply it in the number formatting

Tested with

Units::Units(QObject *parent)
    : QObject(parent)
{
+    QLocale locale("zh_CN");
+    QLocale::setDefault(locale);
}

Could query all languages from LanguageModel and expose setDefaultLocale() from C++ to running the test with all the supported languages. We are not likely to test every code change with all the locales so automation cloud play an useful part here.

mr-manuel:
The input now works, but as jpetrell mentioned now there is a separator every 3 digits. Maybe it's better to fix number formatting to english locale?

For the PR test results see https://github.com/victronenergy/gui-v2/issues/1108#issuecomment-2124346470

blammit:
@jpetrell sounds like we need to discuss this further first, let's do that.

blammit:
@jpetrell @mr-manuel 

I've updated this to use en_US for all number formatting, as discussed. This should avoid all the issues mentioned by @jpetrell here and also fixes the original issues, i.e. gui-v2 should now use the same formatting for all numbers, instead of a an odd mixture of multiple locales.

pages/settings/PageGps.qml
```diff
@@ -73,19 +76,19 @@ Page {
 					if (speedUnit.value === "km/h") {
 						//: GPS speed data, in kilometers per hour
 						//% "%1 km/h"
-						return qsTrId("settings_gps_speed_kmh").arg((dataItem.value * 3.6).toFixed(1))
+						return qsTrId("settings_gps_speed_kmh").arg(Global.formatNumber(dataItem.value * 3.6, 1))
 					} else if (speedUnit.value === "mph") {
 						//: GPS speed data, in miles per hour
 						//% "%1 mph"
-						return qsTrId("settings_gps_speed_mph").arg((dataItem.value * 2.236936).toFixed(1))
+						return qsTrId("settings_gps_speed_mph").arg(Global.formatNumber(dataItem.value * 2.236936, 1))
```
jpetrell, line 83: C++ formatting used to be in units.cpp and JavaScript formatting (e.g. date&time formatting) functions in Utils.js, having formatting in third place Global feels wrong.

jpetrell:
Overall LGTM
- I am getting test failures running tst_units due to included group separators
- Still not sure about group separators, but guess we can include them and see what the feedback is

pages/settings/PageGps.qml
blammit: Ah yes, let's move this to `Units`.

blammit:
> Overall LGTM

> I am getting test failures running tst_units due to included group separators
> Still not sure about group separators, but guess we can include them and see what the feedback is

Ah, I forgot we would get that with en_US. I'll change it to use the c locale.

tests/units/tst_units.qml
```diff
@@ -86,8 +86,7 @@ TestCase {
 	}
 
 	function test_precisionOne() {
-		var units = [VenusOS.Units_Volt_DC,
-					 VenusOS.Units_VoltAmpere,
+		var units = [VenusOS.Units_VoltAmpere,
```
jpetrell, line 89: This removes tests, would be nice to have test_voltages() test (out of scope here) that tested Volt_DC and Volt_AC.

mr-manuel:
Looks good now. Is the grouping wanted? It was not there before this change.

grafik

jpetrell:
> I'll change it to use the c locale.

Seems that you need to expose that one from C++ by returning QLocale::c() (singleton Language.numberLocale maybe).

tests/units/tst_units.qml
blammit: Very true 😅 I've updated this to add `test_precisionTwo()` for this case.

blammit:
> Seems that you need to expose that one from C++ by returning QLocale::c() (singleton Language.numberLocale maybe).

I've added `Units.toString()` to do the number formatting, and removed the `Language` changes.

Ideally we would have a `Numbers` singleton or something, or some generic C++ utils singleton, to provide `Numbers.toString()` and `Numbers.sumRealNumbers()` and maybe some of the functions from `Utils.js`.

src/units.cpp
```diff
@@ -33,6 +33,12 @@ Unit::Type unitToVeUnit(Victron::VenusOS::Enums::Units_Type unit)
 	}
 }
 
+const QLocale *formattingLocale()
```
jpetrell, line 36: Strictly speaking not formatting locale for anything but specifically for numbers so could also be numberFormattingLocale, but maybe that is implicit in Units scope. Also Units.toString() is not self-evident to me. When I think about it more yeah Units is about number value formatting so makes sense, but still would consider renaming to toString() to e.g. formatNumber() to make what is being formatted clear and to align with e.g. time formatting formatAsHHMM(). Even getDisplayText() could be formatQuantity(). But up to you, the existing names are good too. :blush:

components/listitems/ListQuantityField.qml
```diff
@@ -14,7 +14,10 @@ ListTextField {
 	property int decimals: Units.defaultUnitPrecision(unit)
 
 	suffix: Units.defaultUnitString(unit)
-	textField.validator: DoubleValidator { decimals: root.decimals }
+	textField.validator: DoubleValidator {
+		decimals: root.decimals
+		locale: Units.formattingLocaleName
+	}
 	textField.inputMethodHints: Qt.ImhDigitsOnly
```
jpetrell, line 21: Shouldn't this be ImhFormattedNumbersOnly i.e. "Only number input is allowed. This includes decimal point and minus sign."? Even if usually keyboards don't support digit-only so you get the same keyboard as with ImhDigitsOnly.

jpetrell:
Started wondering could there be that localised virtual keyboard (Qt VKB or Android/iOS) is missing period ".", but only provides  ",". Tested that typing "40,0" becomes "400". But seems Android keyboard always provides both "," and "." (tested Arabic, English, Finnish, German, Chinese). On GX device Qt VKB only provides "." (tested German, Swedish, Danish that use comma as the decimal point like majority of European languages do), but that should be fine.

jpetrell (approved):
Added few comments, but overall LGTM. Would at least rename DigitsOnly to ImhFormattedNumbersOnly to reflect that QuantityField deals with floats and not integers.

src/units.cpp
blammit: Yeah maybe if we had a `Numbers` singleton then it could be `Numbers.toString()`. But it's confusing to use `Units.toString()` for this. Updated to be `Units.formatNumber()` instead.

components/listitems/ListQuantityField.qml
blammit: Good catch, thanks! Updated.

src/units.cpp
blammit: Would be nice to rename getDisplayText() too. Let's do it separately.

PR #1119, opened 2024-05-22
Don't allow flicking beyond GradientListView bounds

blammit:
It's disconcerting when a flick makes the view jump around. Restrict the return-to-bounds behavior to dragging.

This is the same as the behavior allowed in the control cards view.


chriadam (approved):

blammit:
Hm, seems this is not respected in wasm.

blammit:
> Hm, seems this is not respected in wasm.

Never mind, I was checking the wrong behavior.

Issue #1017, opened 2024-04-30
Tank gauge drawn outside the container


jpetrell:
[Reported](https://community.victronenergy.com/questions/270759/venus-os-v3401-available-for-testing.html) by the customer. Looks like the item is trying to horizontally center two gauges with the right one being invisible.

![image](https://github.com/victronenergy/gui-v2/assets/2203667/78717620-096a-48de-9f9a-723fedc45240)

jpetrell:
Reviewed the code in detail, not sure how the issue can reproduce with the latest code. If `mergeTanks=true` and `count=2` should be seeing two gauges on the LPG-Right gauge group item and the gauge width should be smaller. If `mergeTanks=false` and `count=2` there should be only one gauge per tank item with each gauge horizontally centered on the tank item background.

Asked the community member if they can still reproduce with the latest v3.40~13 release.

jpetrell:
Still [reproduces](https://community.victronenergy.com/questions/279997/venus-os-v3408-available-for-testing.html?childToView=285923#answer-285923) to the customer (wizwaretech): "I see no post for 3.40~15, so I'll put this here. The same graph display issues exist, intermittently"

![image](https://github.com/victronenergy/gui-v2/assets/2203667/4fbe70d3-723c-472e-9355-38dc6813e91f)

jpetrell:
[Customer](https://community.victronenergy.com/comments/286686/view.html): "I have to say that the bars on the level display are still showing some issues, see LPG Bow and LPG Stern. These are Bluetooth Mopeka sensors and when I opened the level display the LPG Stern sender had disconnected and when it reconnected whilst the level display was open and the display had to change from 4 to 5 tanks, the bar became displaced."
![image](https://github.com/victronenergy/gui-v2/assets/2203667/4b55e305-8682-4c22-8573-54a9d5cce4dd)

chriadam:
I tried to repro using mock backend, adding a "flaky" tanks config.  Whether the tank was removed and re-added, or whether the tank merely set its device instance to -1 and back again, I could not reproduce the case mentioned in this report.

I tried both with my recent clipping-bar-gauge commits, and without.  Without the clipping bar gauge commits, I saw the previous issue (the double-animation and non-rendered foreground), but not the issue described here in this report.

[...]

PR #1122, opened 2024-05-22
Be more explicit about tank gauge delegate sizing

chriadam:
In TankGaugeGroup there is a Row populated by a Repeater whose delegate is a Loader.  The geometry of the Loader isn't explicitly defined, so will be inherited from its item if the Loader is active.  If a particular Loader is active, its geometry will be set; if that Loader then becomes inactive, it's possible that the geometry of that loader is being "stuck" at the old item's width, rather than being reset to zero.

This commit ensures that the width of the loader is explicitly set to the width of the item if active, otherwise zero.

It also makes various spacings and data bindings much more explicit for maintainability purposes.

A mock config is added which was intended to test the error case, although I was unable to reproduce using the mock backend.

Contributes to #1017

### DIFF
--- a/components/TankGaugeGroup.qml
+++ b/components/TankGaugeGroup.qml
@@ -33,25 +33,29 @@ TankItem {
 	gauge: Row {
 		id: subgauges // contains 1 or more gauges of a single type
 
-		spacing: Theme.geometry_levelsPage_subgauges_spacing
+		height: parent.height
+		spacing: root.mergeTanks ? Theme.geometry_levelsPage_subgauges_spacing : 0
 
 		Repeater {
 			model: root.gaugeTanks
 			delegate: Loader {
 				active: model.index === 0 || root.mergeTanks
-				sourceComponent: TankGauge {
-					expanded: root.expanded
-					animationEnabled: root.animationEnabled
-					width: {
+				width: {
+					if (active) {
 						const availableSpace = root.width - 2*Theme.geometry_levelsPage_subgauges_horizontalMargin
 						if (root.mergeTanks) {
 							return (availableSpace - (gaugeTanks.count - 1) * subgauges.spacing)/gaugeTanks.count
 						} else {
 							return availableSpace
 						}
+					} else {
+						return 0
 					}
-
-					height: subgauges.height
+				}
+				height: subgauges.height
+				sourceComponent: TankGauge {
+					expanded: root.expanded
+					animationEnabled: root.animationEnabled
 					valueType: root.tankProperties.valueType
 					value: (root.mergeTanks ? model.device.level : root.level) / 100
 					isGrouped: root.mergeTanks


chriadam:
I haven't verified that this fixes the issue.  I haven't been able to repro the original issue.

components/TankGaugeGroup.qml
```diff
@@ -33,29 +34,41 @@ TankItem {
...
 
 		Repeater {
 			model: root.gaugeTanks
 			delegate: Loader {
-				active: model.index === 0 || root.mergeTanks
+				id: tankGaugeLoader
+				property int gaugeIndex: model.index
+				property real gaugeLevel: model.device.level
+				active: root.mergeTanks || gaugeIndex === root.delegateIndex
+				width: active && item ? item.width : 0
```
chriadam, line 46: I think this line, width binding, is the real fix for the issue (although it's just a guess, since I can't repro the issue)

jpetrell (approved):
LGTM. These changes should fix the issue in reasonably high likelihood. I dislike the increased complexity a bit, would be nice to understand in detail what is going wrong, but we cannot really if we don't find a way to reproduce. Overall the way the model is used here in nested manner is a bit funky, usually you would update the model with right roles instead of constructing bunch of duplicated empty delegates. But it works, doesn't make sense to rewrite now.

FYI Getting empty gauges here also, but looks like the issue is unrelated but something that is already in main. Can be reproduced on desktop with the mock backend quite easily, usually seeing the empty gauges when going through configs 3.-6.

image

components/TankGaugeGroup.qml
blammit: Could you separate the bug-specific fixes into an individual commit?

components/TankGaugeGroup.qml
```diff
@@ -33,29 +34,41 @@ TankItem {
...
+							// is some sort of race condition / property binding evaluation
+							// order issue which hits this codepath; see #1017
+							return 0
 						}
 					}
 
 					height: subgauges.height
 					valueType: root.tankProperties.valueType
-					value: (root.mergeTanks ? model.device.level : root.level) / 100
+					value: tankGaugeLoader.gaugeLevel / 100
```
blammit, line 68: This is an unrelated change?

data/mock/TanksImpl.qml
```diff
@@ -108,11 +109,45 @@ QtObject {
...
+		function onSetTanksFlaky(flaky) {
+			root.flakeTank.testingFlakyTanks = flaky
+		}
 	}
 
+	property Timer flakeTank: Timer {
+		running: Global.mockDataSimulator.timersActive && testingFlakyTanks
+		interval: 4 * 1000
+		repeat: true
+		property bool removeFlaky: false // whether to remove, or just invalidate, the flaky tank.
```
blammit, line 122: Could you add some comments about what a "flaky" tank means?

components/TankGaugeGroup.qml
```diff
@@ -33,29 +34,41 @@ TankItem {
...
 						const availableSpace = root.width - 2*Theme.geometry_levelsPage_subgauges_horizontalMargin
 						if (root.mergeTanks) {
 							return (availableSpace - (gaugeTanks.count - 1) * subgauges.spacing)/gaugeTanks.count
-						} else {
+						} else if (tankGaugeLoader.gaugeIndex === root.delegateIndex) {
 							return availableSpace
+						} else {
+							// shouldn't be necessary or useful, but just in case there
+							// is some sort of race condition / property binding evaluation
+							// order issue which hits this codepath; see #1017
```
blammit, line 61: I'm not sure about adding this if we don't know whether it fixes the issue, and it creates more code paths we need to maintain.

pages/TanksTab.qml
```diff
@@ -100,6 +100,7 @@ ListView {
 					totalRemaining: mergeTanks ? gaugeTanks.totalRemaining : model.device.remaining
 					totalCapacity: mergeTanks ? gaugeTanks.totalCapacity : model.device.capacity
 
+					delegateIndex: model.index
```
blammit, line 103: There are a few delegates floating around here, so might need a more specific name than `delegateIndex`

components/TankGaugeGroup.qml
```diff
@@ -33,29 +34,41 @@ TankItem {
...
+		spacing: root.mergeTanks ? Theme.geometry_levelsPage_subgauges_spacing : 0
 
 		Repeater {
 			model: root.gaugeTanks
 			delegate: Loader {
-				active: model.index === 0 || root.mergeTanks
+				id: tankGaugeLoader
+				property int gaugeIndex: model.index
+				property real gaugeLevel: model.device.level
+				active: root.mergeTanks || gaugeIndex === root.delegateIndex
```
blammit, line 45: There are a few delegates floating around here, so might need a more specific name than delegateIndex.

blammit: What was this `active: root.mergeTanks || gaugeIndex === root.delegateIndex` change intended to do? The index of the tank within its group doesn't need to match the index of the group itself. These two indexes are from different models.

blammit:
I'm seeing this with configuration 5 in mock mode:

image

blammit:
> FYI Getting empty gauges here also, but looks like the issue is unrelated but something that is already in main. Can be reproduced on desktop with the mock backend quite easily, usually seeing the empty gauges when going through configs 3.-6.

I don't see this in main, only with this PR. What do you see @chriadam ?

jpetrell:
> I don't see this in main, only with this PR. What do you see @chriadam ?

That may be thrue, not seeing it on main either now. Must have been confused with the branches.

chriadam:
You're right, I could repro the issue you saw wiith my original PR.  Reduced to just the bugfix now.  I will investigate the cleanup next, but it can be separate.

chriadam:
Hrm, still repro the issue with my smaller PR.  It might be the height binding change... Fix investigate and fix.

chriadam:
Fixed.

data/mock/TanksImpl.qml
chriadam: Removed the "flaky" mock since it didn't reproduce the issue anyway.  Flaky just meant that it popped in and out of existence.  I was trying to reproduce the case in the report/issue, but couldn't.

components/TankGaugeGroup.qml
chriadam: Yes, it was cleanup.  That's now in #1138 instead.

components/TankGaugeGroup.qml
chriadam: Yes, it's not needed, removed.

components/TankGaugeGroup.qml
chriadam: I now use "groupIndex" as it's clearer.  And yes, the index of the tank now needs to match the index of the group, unless mergeTanks is true.  That is, I draw the "correct" tank rather than the zeroth tank.  Added comments etc.  But the changes are now in #1138 instead of this PR, as this PR now contains only the bugfix.

jpetrell (approved):

blammit (approved):

components/TankGaugeGroup.qml
blammit: Ah -- I see what you mean now, yep

PR #1123, opened 2024-05-22
Work around crash in bar gauge animation

chriadam:
Related to YAnimator semantics.  See QTBUG-125571. Also clean up the implementations a bit.

### DIFF
--- a/components/BarGaugeBase.qml
+++ b/components/BarGaugeBase.qml
@@ -12,6 +12,7 @@ Rectangle {
 
 	property int valueType: VenusOS.Gauges_ValueType_NeutralPercentage
 	readonly property int valueStatus: Gauges.getValueStatus(_value * 100, valueType)
+	property alias foregroundParent: fgRect.parent
 	property color foregroundColor: Theme.statusColorValue(valueStatus)
 	property color backgroundColor: Theme.statusColorValue(valueStatus, true)
 	property color surfaceColor: Theme.color_levelsPage_gauge_separatorBarColor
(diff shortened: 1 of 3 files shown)


components/BarGaugeBase.qml
```diff
@@ -33,11 +34,16 @@ Rectangle {
 				: -width + (width * _value)     // slide in from left to right
 			: 0
 
-	on_NextPosChanged: if (visible) _updateGauge()
-	onVisibleChanged: if (visible) _updateGauge()
+	on_NextPosChanged: if (visible) _updateGauge(bgRect)
+	onVisibleChanged: if (visible) _updateGauge(bgRect)
```
jpetrell, line 38: Personally prefer to avoid abbreviations in variable names, but no biggie. :smile:

components/BarGaugeBase.qml
```diff
@@ -33,11 +34,16 @@ Rectangle {
...
 
-	function _updateGauge() {
-		if (animationEnabled) {
+	function _updateGauge(scopeObject) {
+		if (scopeObject == null) {
```
jpetrell, line 41: Could just check bgRect == null here as that seems to be the scopeObject always.

jpetrell (approved):
LGTM. Some minor comments, didn't see issues on device.

Sorry I thought I approved this already, must just have forgotten to press the button.

components/BarGaugeBase.qml
chriadam: I was worried that attempting to access "bgRect" directly in that scope will attempt to look up the id from the component scope object and therefore crash again.  But, I tested your suggested change, and it seemed to work just fine, and it's definitely better than the passed-object-reference way.  I will update the PR.

jpetrell:
Still LGTM.

PR #1129, opened 2024-05-22
Adjustments to the Overview box sizing

### LINKED ISSUE #1127: No margin between a medium and 3 small widgets on the Overview page on 5" variant
image

jpetrell:
- Reduce the first widget size if there 4 input widgets. Fixes #1127.
- Simplify Overview widget size category calculations
  * This makes the implementation closer to the design that uses Figma autolayouts
  * Calculating the sizes helps align box spacing between different size categories and between the compact/expanded modes.
  * Aligning spacing between all layout combinations (M+3xXS, L+2xS) is not fully possible unless we make the box heights dynamic so calculating the dimensions from tightest combinations (4xS, 5xXS) to quarantee at least Theme.geometry_overviewPage_widget_spacing spacing

### DIFF
--- a/pages/OverviewPage.qml
+++ b/pages/OverviewPage.qml
@@ -166,18 +166,10 @@ SwipeViewPage {
 		const compactPageHeight = Theme.geometry_screen_height
 				- Theme.geometry_statusBar_height
 				- Theme.geometry_navigationBar_height
-
-		if (compactPageHeight !== Theme.geometry_overviewPage_widget_compact_xl_height) {
-			console.log("Warning: theme constants need to be updated.")
-		}
 		const compactWidgetsTopMargin = Math.max(0, (compactPageHeight - compactWidgetHeights) / Math.max(1, widgets.length - 1))
-
 		const expandedPageHeight = Theme.geometry_screen_height
 				- Theme.geometry_statusBar_height
 				- Theme.geometry_overviewPage_layout_expanded_bottomMargin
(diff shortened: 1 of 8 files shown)


pages/OverviewPage.qml
```diff
@@ -103,13 +103,13 @@ SwipeViewPage {
 			case 4:
 				// Only one of the widgets can have L size, and the other ones use a reduced size.
 				if (widget === firstLargeWidget) {
-					widget.size = VenusOS.OverviewWidget_Size_L
+					widget.size = VenusOS.OverviewWidget_Size_M
 				} else if (firstLargeWidget != null) {
 					// There is a large widget, so use M or XS size to fit around it
-					widget.size = _leftWidgets.length == 3 ? VenusOS.OverviewWidget_Size_M : VenusOS.OverviewWidget_Size_XS
+					widget.size = VenusOS.OverviewWidget_Size_XS
```
jpetrell, line 109: length === 4 always here

blammit (approved):
Great to get rid of the hardcoded sizing. I'm wondering if we should add some padding between the label and the 3-phase number display here, but not a big deal.

image

jpetrell:
> I'm wondering if we should add some padding between the label and the 3-phase number display here, but not a big deal.

Good catch. Adjusted three phase display font size a bit.

I am tempted to increase medium box size, but then 3xM layout would break.

PR #1135, opened 2024-05-23
Reduce flick decelaration to make flicks less sticky

### LINKED ISSUE #1124: Flicking with GX devices feels sticky
The current maximum flick velocity is 2500 pixels/second with flick deceleration 5000 pixels/second, meaning the system will grawl to stop in 0.5 seconds. Would consider 3500 or 2500 pixels/second decelaration.

The values are platform-specific so need to be careful of not making e.g. flicking on the wasm build worse.

jpetrell:
- GX device, WASM build and Linux desktop already had the same flicking parameters
- Keep maximumFlickVelocity, but reduce decelaration from 5000 to 2500
- Could go even further, lightweight flicks are pretty standard in the industry

Fixes #1124.

### DIFF
--- a/components/GradientListView.qml
+++ b/components/GradientListView.qml
@@ -24,6 +24,9 @@ ListView {
 		anchors.bottom: root.bottom
 	}
 
+	maximumFlickVelocity: Theme.geometry_flickable_maximumFlickVelocity
+	flickDeceleration: Theme.geometry_flickable_flickDeceleration
+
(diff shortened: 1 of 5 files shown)


blammit (approved):
Should it be done in the Control cards list view as well? And the Tanks tab?

jpetrell:
> Should it be done in the Control cards list view as well? And the Tanks tab?

Yes. Done.

PR #1136, opened 2024-05-23
Open VeBus settings if AC input widget is not a grid meter

### LINKED ISSUE #1026: Tapping AC input box on the Overview page opens empty settings page
When viewing the overview page and then clicking on the grid meter all values on the page are empty. I'm using the internal meter of the MultiPlus II as grid meter.

venus-os-grid

jpetrell:
Fixes #1026.

### DIFF
--- a/components/widgets/AcInputWidget.qml
+++ b/components/widgets/AcInputWidget.qml
@@ -32,6 +32,19 @@ AcWidget {
 	}
 
 	onClicked: {
+		if (root.input.serviceType === "vebus") {
+			const deviceIndex = Global.inverterChargers.veBusDevices.indexOf(root.input.serviceUid)
+			if (deviceIndex >= 0) {
+				const veBusDevice = Global.inverterChargers.veBusDevices.deviceAt(deviceIndex)
+				Global.pageManager.pushPage( "/pages/vebusdevice/PageVeBus.qml", {
+					"title": veBusDevice.name,
+					"veBusDevice": veBusDevice
+				})
+			}
+			return
+		}
+
+		// Assume this is on a grid/genset service
 		Global.pageManager.pushPage("/pages/settings/devicelist/ac-in/PageAcIn.qml", {
 			"title": root.input.name,
 			"bindPrefix": root.input.serviceUid
--- a/data/common/AcInputSettingsModel.qml
+++ b/data/common/AcInputSettingsModel.qml
@@ -1,0 +1,41 @@
+/*
+** Copyright (C) 2024 Victron Energy B.V.
+** See LICENSE.txt for license information.
+*/
+
+import QtQuick
+import Victron.VenusOS
+
+ListModel {
+	id: root
+
+	property string serviceUid
+	property int numberOfAcInputs
+
+	property Instantiator _acInputSettingsObjects: Instantiator {
+		model: root.numberOfAcInputs || null
+		delegate: AcInputSettings {
+			serviceUid: root.serviceUid
+			inputNumber: model.index + 1
+		}
+
+		onObjectAdded: function(index, settings) {
+			let insertionIndex = root.count
+			for (let i = 0; i < root.count; ++i) {
+				if (settings.inputNumber < root.get(i).inputSettings.inputNumber) {
+					insertionIndex = i
+					break
+				}
+			}
+			root.insert(insertionIndex, { inputSettings: settings })
+		}
+		onObjectRemoved: function(index, settings) {
+			for (let i = 0; i < root.length; ++i) {
+				if (root.get(i).inputSettings.inputNumber === settings.inputNumber) {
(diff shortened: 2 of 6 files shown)


components/widgets/AcInputWidget.qml
```diff
@@ -32,10 +32,17 @@ AcWidget {
 	}
 
 	onClicked: {
-		Global.pageManager.pushPage("/pages/settings/devicelist/ac-in/PageAcIn.qml", {
-			"title": root.input.name,
-			"bindPrefix": root.input.serviceUid
-		})
+		if (root.input.serviceUid.indexOf("com.victronenergy.grid") >= 0 || root.input.serviceUid.indexOf("com.victronenergy.genset") >= 0)  {
```
jpetrell, line 35: > Not sure how to detect when grid meter is available.

Asked Rein if there is a way to detect if there is a grid meter: "think you can rely on the service name starting with com.victronenergy.grid or com.victronenergy.genset (for the generator variant of AC input). systemcalc does it too: # If a grid meter is present we use values from it. If not, we look at the multi".

jpetrell:
> There are some funky little differences between settings now depending on where you access the multi device, trying to untangle that now.

Should expose now the same data whether opened from the Overview or device list pages.

Filed task https://github.com/victronenergy/gui-v2/issues/1137 for aligning small differences in VeBus and InverterCharger layouts.

data/common/AcInputSettingsModel.qml
```diff
@@ -0,0 +1,41 @@
+/*
+** Copyright (C) 2024 Victron Energy B.V.
+** See LICENSE.txt for license information.
+*/
+
+import QtQuick
+import Victron.VenusOS
+
+ListModel {
```
blammit, line 9: Nice to have this refactored out into AcInputSettingsModel :) 

This could be moved into a separate commit?

data/common/ActiveAcInput.qml
```diff
@@ -23,6 +23,8 @@ Device {
 	readonly property alias currentLimit: _acInputService.currentLimit
 	readonly property alias phases: _phases
 
+	readonly property int mode: _mode.value === undefined ? -1 : _mode.value
```
blammit, line 26: I don't think we should add this `mode` property or the `inputSettings` property because they don't make sense for AC inputs. For example, looking at https://github.com/victronenergy/venus/wiki/dbus, `/Ac/NumberOfAcInputs` is only present in `com.victronenergy.multi`, and there is no `/Mode` value for `com.victronenergy.<grid|genset>` services. Ideally the properties in ActiveAcInput.qml are common ones that are generically relevant to AC inputs.

But it shouldn't be necessary to add these properties to try to coerce `ActiveAcInput` to have similar properties to a vebus service, as per comment for `AcInputWidget.qml`.

components/widgets/AcInputWidget.qml
blammit: There is an easier way to do this, without needing to coerce this `ActiveAcInput` object to fit the `veBusDevice` value in `PageVeBus.qml`. Basically, Global.inverterChargers.veBusDevices is a model of all vebus devices, which means you can find the matching device from there, instead of passing this `input` directly.

So e.g. this change should be sufficient to fix the issue, without the other `data/` changes:

```
if (root.input.serviceType === "vebus") {
    const deviceIndex = Global.inverterChargers.veBusDevices.indexOf(root.input.serviceUid)
    if (deviceIndex >= 0) {
        const veBusDevice = Global.inverterChargers.veBusDevices.deviceAt(deviceIndex)
        Global.pageManager.pushPage( "/pages/vebusdevice/PageVeBus.qml", {
            "title": veBusDevice.name,
            "veBusDevice": veBusDevice
        })
    }
    return
}

// Assume this is on a grid/genset service
Global.pageManager.pushPage("/pages/settings/devicelist/ac-in/PageAcIn.qml", {
    "title": root.input.name,
    "bindPrefix": root.input.serviceUid
})
```
    
We do need some more documentation for `data/` to make this kind of thing more clear.

blammit (changes requested):

components/widgets/AcInputWidget.qml
jpetrell: Cool! I'll do that

data/common/AcInputSettingsModel.qml
jpetrell: Separated to own commit.

data/common/ActiveAcInput.qml
jpetrell: No more changes to ActiveAcInput.qml

jpetrell:
> Global.inverterChargers.veBusDevices is a model of all vebus devices, which means you can find the matching device from there

Done.

blammit (approved):

PR #1138, opened 2024-05-24
Draw the gauges for the required tanks in each group

chriadam:
Previously, the tank gauge group would draw the gauge for the zeroth tank regardless of whether that tank is actually the gauge needed to be shown in that group.  It would then replace the value associated with that tank with the value specified in the group.

This commit ensures that each tank gauge group only draws the gauge for the actual tank it represents (in the non-merged case) and can use its level directly.

### DIFF
--- a/components/TankGaugeGroup.qml
+++ b/components/TankGaugeGroup.qml
@@ -10,6 +10,7 @@ import Victron.Gauges
 TankItem {
 	id: root
 
+	property int groupIndex: -1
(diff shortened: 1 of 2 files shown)


chriadam:
This is the maintainability improvement commit.  It's not strictly necessary, but the current way it's done is super confusing to me, whereas this makes things more obvious.

blammit (approved):

jpetrell (approved):

PR #1139, opened 2024-05-27
Deactivate Brief left/right gauges when side panel is expanded

blammit:
When the side panel is fully expanded, set active=false on the Brief left/right gauges.


jpetrell (approved):
- Also deactivate AC load gauge when `active: root.state !== "panelOpened"`? 
- Doesn't seem to regress performance as collapsing is smooth. Opening side panel is a bit choppy, but that seems unrelated to this change

blammit:
> Also deactivate AC load gauge when active: root.state !== "panelOpened"?

Forgot about that one. Done, thanks :)

PR #1141, opened 2024-05-27
ListLabel should be visible by default

### LINKED ISSUE #1140: Min/max gauge section headers are no longer visible
Section headers have disappeared:

image

blammit:
Regression from f04e21956e32cad033640ed05a40b02b4ec5864e

Fixes #1140


jpetrell:
Yikes! Good reminder to always test even the most innocent looking, small changes on the device...

PR #1157, opened 2024-05-29
Clear inverter/charger input settings when object becomes invalid

### LINKED ISSUE #1134: Grid current limit setting is sometimes missing in multi device page
Switch between Boat/Motorhome demo 1 and 2 demo modes on one gui-v2 instance (browser or GX device). Go to Settings > Device list > MultiPlus-II 24/3000/70-32 on the other.

Expected result: Grid current limit spin box is editable.

Actual result: Spin box has no label or value.

image

Seeing following errors on the log:
```
venus-gui-v2.js:8 qrc:/qt/qml/Victron/VenusOS/components/AcInputsCurrentLimits.qml:47: TypeError: Cannot read property 'currentLimit' of null
venus-gui-v2.js:8 qrc:/qt/qml/Victron/VenusOS/components/AcInputsCurrentLimits.qml:43: TypeError: Cannot read property 'inputType' of null
```

blammit:
Otherwise, there are stale objects in the input settings model when the object becomes valid again -- for example, if changing the demo mode causes the inverter/charger to be repopulated with different values.

Fixes #1134

### DIFF
--- a/data/common/InverterCharger.qml
+++ b/data/common/InverterCharger.qml
@@ -80,4 +80,10 @@ Device {
 		}
 		inputSettings.get(inputIndex).inputSettings.setCurrentLimit(currentLimit)
 	}
+
+	onValidChanged: {
+		if (!valid) {
+			inverterCharger.inputSettings.clear()
+		}
+	}
 }


jpetrell (approved):

jpetrell:
LGTM. Quickly went through other models under data/common and checked different views between wasm <-> gx after demo mode changes. Didn't find any issues.

Most model properties are directly bound to VeQuickItems that will update automatically when the backend values update. These global data models already handled valid change:
- AcInDeviceModel, Battery, Charger, DcInput, DcLoad, DigitalInput, EnvironmentInput, EvCharger, Generator, InverterCharger, MeteoDevice, MotorDrive, MultiRsDevice, PulseMeter, PvInverter, Relay, SolarCharger, Tank, UnsupportedDevices

Some models are not cleared like SolarHistoryErrorModel and phases ListModel in PvInverters, but have fixed number of model items (solar history always has four errors /LastError1-4 and the inverter always 3 phase paths L1-3 exposed in DBus/MQTT), so there is no risk of data points becoming stale.

blammit:
Thanks for thoroughly checking the other cases :)

PR #1162, opened 2024-05-29
CAN Profiles: add the High Voltage CAN bus as an option

### LINKED ISSUE #1084: CAN Profiles: add the High Voltage CAN bus as an option
gui-v1 commit: d21f13064fd0c70a04c5e876114720714fc5490c

    The High Voltage CAN bus requires a CAN bus controller with CAN FD support, hence
    only provide this option if the CAN config of the CAN interfaces indicates that it
    is capable of handling it.

jpetrell:
WIP. Still need a way to verify this works.

The High Voltage CAN bus requires a CAN bus controller with CAN FD support, hence
only provide this option if the CAN config of the CAN interfaces indicates that it
is capable of handling it.

gui-v1 commit: d21f13064fd0c70a04c5e876114720714fc5490c

Fixes #1084.

### DIFF
--- a/pages/settings/PageSettingsCanbus.qml
+++ b/pages/settings/PageSettingsCanbus.qml
@@ -45,6 +45,12 @@ Page {
 						return profile !== VenusOS.CanBusProfile_Vecan
 					case VenusOS.CanBusConfig_ForcedCanBusBms:
 						return profile !== VenusOS.CanBusProfile_CanBms500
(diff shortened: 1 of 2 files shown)


chriadam:
I think we should merge this, and ask Rein to test.

jpetrell:
> I think we should merge this, and ask Rein to test.

Verified using Rein's scripts that publish the correct messages on a virtual CAN interface so you don't need any cabling. Merging.

Issue #1010, opened 2024-04-30
Calculate several power totals in systemcalc for gui-v2


ReinvdZee:
Currently gui-v2 sums various input/output values in order to get “total” values to draw the graphs and gauges. For https://github.com/victronenergy/venus/issues/1217 systemcalc calculates these totals:

- DC input power: `/Dc/Charger/Power` + `/Dc/FuelCell/Power` + `/Dc/Alternator/Power`
- AC load power: `/Ac/ConsumptionOnInput/L[1,2,3]/Power` + `/Ac/ConsumptionOnOutput/L[1,2,3]/Power`
- PV AC power: `/Ac/PvOnGrid/L[1,2,3]/Power` + `/Ac/PvOnGenset/L[1,2,3]/Power` + `/Ac/PvOnOutput/L[1,2,3]/Power`

It is more efficient to have systemcalc then as well publish these calculations on dbus, instead of having gui-v2 calculate them again. (Possbly gui-v1 can also use them)

Additional values to consider being published by systemcalc:
- `/Ac/ActiveIn/Power` or such to provide the total power for all phases of the currently active AC in. While grid, genset and vebus provide AC input totals, currently for com.victronenergy.multi there is no total power, so gui-v2 creates a total by summing `/Ac/In/1/L[1,2,3]/P`.
- Some way to get the total remaining/capacity for all tanks of the same type

@blammit

PR #1171, opened 2024-05-30
Performance improvements

chriadam:


chriadam:
With all of these improvements, CPU usage on the OverviewPage is reduced by approximately 5%, but there's still a long way to go.  I found the following suspects from profiling, which I still need to fix (e.g. move into C++):

Gauges.js methods
Utils.scaleNumber()

ValueRange.qml should be reimplemented as a C++ type, valueAsRatio() shows up in trace

AcInputPhaseModel.qml _phaseObjects onValueChanged for power
SystemData.qml _updateConsumptionModel()
PvMonitor.qml updateAcTotals()
DcInputWidget.qml _refreshTotalPower()

BatteryWidget.qml useSmallFont / quantityLabelFits()

Doing: model[root.phaseModelProperty] in e.g. ThreePhaseDisplay.qml is showing up in trace

The FrameAnimation plumbing also shows up in a big way also, so that's going to be my next target, although I suspect it's mostly the effort of updating the positions of the electrons and that probably can't be minimised too much.

chriadam:
Disabling all of the BriefPage side gauge loaders when the BriefPage isn't the current page also reduces CPU usage by about 2% while on OverviewPage, from my testing.  But, it makes navigation between pages sticky/jerky, so best to avoid that heavy-handed change, and instead try to target individual bindings within those side gauges, I guess.

chriadam:
Clipping seems to hurt performance quite badly on CerboGX.  I tried changing some gauges to ClippingBarGauge now that the colours are opaque (in this PR), but it actually hurt performance.  Changing them to CheapBarGauge is far better, performance-wise, and I think for these specific use-cases the performance gain vastly outweighs the very slight difference in look/feel compared to baseline.

chriadam:
I suspect that TextRendering might be a big culprit.  QSG_RENDERER_DEBUG=render shows a very large number of non-opaque nodes which cannot be merged into a small number of batches:

```
$ QSG_RENDERER_DEBUG=render ./venus-gui-v2

...

Renderer::render() QSGAbstractRenderer(0xb218b440) "rebuild: none"
Rendering:
 -> Opaque: 13 nodes in 2 batches...
 -> Alpha: 80 nodes in 37 batches...
 - 0xab167320 [retained] [noclip] [opaque] [  merged]  Nodes:    1  Vertices:     4  Indices:     6  root: 0x0
 - 0xab1a0908 [retained] [noclip] [opaque] [unmerged]  Nodes:   12  Vertices:    48  Indices:     0  root: 0xb2191d38
 - 0xac119290 [retained] [noclip] [ alpha] [  merged]  Nodes:    7  Vertices:   288  Indices:   488  root: 0xb2191d38 opacity: 1
 - 0xab1a0c68 [retained] [noclip] [ alpha] [unmerged]  Nodes:    4  Vertices:    16  Indices:     0  root: 0xb2191d38
 - 0xab1a11a8 [retained] [noclip] [ alpha] [  merged]  Nodes:    1  Vertices:    64  Indices:    96  root: 0xb2191d38 opacity: 1
 - 0xab1a0fc8 [retained] [noclip] [ alpha] [  merged]  Nodes:    1  Vertices:    72  Indices:   108  root: 0xb2191d38 opacity: 1
 - 0xab1a0ef0 [retained] [noclip] [ alpha] [unmerged]  Nodes:    2  Vertices:    82  Indices:     0  root: 0xb2191d38
 - 0xab1a09e0 [retained] [  clip] [ alpha] [  merged]  Nodes:    1  Vertices:     6  Indices:     8  root: 0xb2190f58 opacity: 1
 - 0xab19fd10 [retained] [noclip] [ alpha] [unmerged]  Nodes:    1  Vertices:     4  Indices:     6  root: 0xb2191d38
 - 0xb2183918 [retained] [noclip] [ alpha] [  merged]  Nodes:    3  Vertices:   240  Indices:   404  root: 0xb2191d38 opacity: 1
 - 0xab1a0b90 [retained] [noclip] [ alpha] [unmerged]  Nodes:    3  Vertices:    12  Indices:     0  root: 0xb2191d38
 - 0xab1a20d8 [retained] [noclip] [ alpha] [  merged]  Nodes:    2  Vertices:    60  Indices:    90  root: 0xb2191d38 opacity: 1
 - 0xab1a0e18 [retained] [noclip] [ alpha] [  merged]  Nodes:    1  Vertices:    12  Indices:    18  root: 0xb2191d38 opacity: 1
 - 0xab1a1d78 [retained] [noclip] [ alpha] [  merged]  Nodes:    1  Vertices:     4  Indices:     6  root: 0xb2191d38 opacity: 1
 - 0xaca4f260 [retained] [noclip] [ alpha] [  merged]  Nodes:    4  Vertices:    52  Indices:    78  root: 0xb2191d38 opacity: 1
 - 0xab1a1af0 [retained] [noclip] [ alpha] [  merged]  Nodes:    1  Vertices:     4  Indices:     6  root: 0xb2191d38 opacity: 1
 - 0xab1a1a18 [retained] [noclip] [ alpha] [  merged]  Nodes:    1  Vertices:     4  Indices:     6  root: 0xb2191d38 opacity: 1
 - 0xab19fdc8 [retained] [noclip] [ alpha] [  merged]  Nodes:    2  Vertices:    12  Indices:    18  root: 0xb2191d38 opacity: 1
 - 0xab1a1868 [retained] [noclip] [ alpha] [unmerged]  Nodes:    2  Vertices:     8  Indices:     0  root: 0xb2191d38
 - 0xab1a16b8 [retained] [noclip] [ alpha] [  merged]  Nodes:    3  Vertices:    44  Indices:    66  root: 0xb2191d38 opacity: 1
 - 0xab1a0758 [retained] [noclip] [ alpha] [  merged]  Nodes:    1  Vertices:    12  Indices:    18  root: 0xb2191d38 opacity: 1
 - 0xab1a15e0 [retained] [noclip] [ alpha] [  merged]  Nodes:    1  Vertices:     4  Indices:     6  root: 0xb2191d38 opacity: 1
 - 0xab1a1358 [retained] [noclip] [ alpha] [  merged]  Nodes:    4  Vertices:    32  Indices:    48  root: 0xb2191d38 opacity: 1
 - 0xab1a03f8 [retained] [noclip] [ alpha] [  merged]  Nodes:    3  Vertices:    28  Indices:    42  root: 0xb2191d38 opacity: 1
 - 0xab1a1790 [retained] [noclip] [ alpha] [unmerged]  Nodes:    3  Vertices:    12  Indices:     0  root: 0xb2191d38
 - 0xab1a0320 [retained] [noclip] [ alpha] [  merged]  Nodes:    2  Vertices:    28  Indices:    42  root: 0xb2191d38 opacity: 1
 - 0xab1a1f28 [retained] [noclip] [ alpha] [  merged]  Nodes:    1  Vertices:    16  Indices:    24  root: 0xb2191d38 opacity: 1
 - 0xab1a1e50 [retained] [noclip] [ alpha] [  merged]  Nodes:    1  Vertices:     4  Indices:     6  root: 0xb2191d38 opacity: 1
 - 0xab1a0010 [retained] [noclip] [ alpha] [  merged]  Nodes:    6  Vertices:    72  Indices:   108  root: 0xb2191d38 opacity: 1
 - 0xab1a1430 [retained] [noclip] [ alpha] [  merged]  Nodes:    1  Vertices:    12  Indices:    18  root: 0xb2191d38 opacity: 1
 - 0xab1a2000 [retained] [noclip] [ alpha] [  merged]  Nodes:    1  Vertices:     4  Indices:     6  root: 0xb2191d38 opacity: 1
 - 0xab10a4b0 [retained] [noclip] [ alpha] [  merged]  Nodes:    7  Vertices:   288  Indices:   488  root: 0xb2191d38 opacity: 1
 - 0xab10af00 [retained] [noclip] [ alpha] [unmerged]  Nodes:    1  Vertices:     9  Indices:     0  root: 0xb2191d38
 - 0xab10ddf8 [retained] [noclip] [ alpha] [unmerged]  Nodes:    2  Vertices:     8  Indices:     0  root: 0xb2191d38
 - 0xab1a0ab8 [retained] [noclip] [ alpha] [  merged]  Nodes:    1  Vertices:    40  Indices:    60  root: 0xb2191d38 opacity: 1
 - 0xab19ff38 [retained] [noclip] [ alpha] [  merged]  Nodes:    1  Vertices:    16  Indices:    24  root: 0xb2191d38 opacity: 1
 - 0xab1a0d40 [retained] [noclip] [ alpha] [  merged]  Nodes:    1  Vertices:     4  Indices:     6  root: 0xb2191d38 opacity: 1
 - 0xab19fe80 [retained] [noclip] [ alpha] [unmerged]  Nodes:    2  Vertices:    82  Indices:     0  root: 0xb2191d38
 - 0xab1a1280 [retained] [noclip] [ alpha] [  merged]  Nodes:    1  Vertices:    20  Indices:    30  root: 0x0 opacity: 1
 -> times: build: 0, prepare(opaque/alpha): 0/0, sorting: 0, upload(opaque/alpha): 0/0, record rendering: 10
Renderer::render() QSGAbstractRenderer(0xb218b440) "rebuild: full"
Rendering:
 -> Opaque: 13 nodes in 2 batches...
 -> Alpha: 81 nodes in 38 batches...
 - 0xab1a1280 [  upload] [noclip] [opaque] [  merged]  Nodes:    1  Vertices:     4  Indices:     6  root: 0x0
 - 0xab19fe80 [  upload] [noclip] [opaque] [unmerged]  Nodes:   12  Vertices:    48  Indices:     0  root: 0xb2191d38
 - 0xab1a0d40 [  upload] [noclip] [ alpha] [  merged]  Nodes:    7  Vertices:   288  Indices:   488  root: 0xb2191d38 opacity: 1
 - 0xab19ff38 [  upload] [noclip] [ alpha] [unmerged]  Nodes:    4  Vertices:    16  Indices:     0  root: 0xb2191d38
 - 0xab1a0ab8 [  upload] [noclip] [ alpha] [  merged]  Nodes:    1  Vertices:    64  Indices:    96  root: 0xb2191d38 opacity: 1
 - 0xab10ddf8 [  upload] [noclip] [ alpha] [  merged]  Nodes:    1  Vertices:    72  Indices:   108  root: 0xb2191d38 opacity: 1
 - 0xab10af00 [  upload] [noclip] [ alpha] [unmerged]  Nodes:    2  Vertices:    82  Indices:     0  root: 0xb2191d38
 - 0xab10a4b0 [  upload] [  clip] [ alpha] [  merged]  Nodes:    1  Vertices:     6  Indices:     8  root: 0xb2190f58 opacity: 1
 - 0xab1a2000 [  upload] [noclip] [ alpha] [unmerged]  Nodes:    1  Vertices:     4  Indices:     6  root: 0xb2191d38
 - 0xab1a1430 [  upload] [noclip] [ alpha] [  merged]  Nodes:    3  Vertices:   240  Indices:   404  root: 0xb2191d38 opacity: 1
 - 0xab1a0010 [  upload] [noclip] [ alpha] [unmerged]  Nodes:    3  Vertices:    12  Indices:     0  root: 0xb2191d38
 - 0xab1a1e50 [  upload] [noclip] [ alpha] [  merged]  Nodes:    2  Vertices:    60  Indices:    90  root: 0xb2191d38 opacity: 1
 - 0xab1a1f28 [  upload] [noclip] [ alpha] [  merged]  Nodes:    1  Vertices:    12  Indices:    18  root: 0xb2191d38 opacity: 1
 - 0xab1a0320 [  upload] [noclip] [ alpha] [  merged]  Nodes:    1  Vertices:     4  Indices:     6  root: 0xb2191d38 opacity: 1
 - 0xab1a1790 [  upload] [noclip] [ alpha] [  merged]  Nodes:    4  Vertices:    52  Indices:    78  root: 0xb2191d38 opacity: 1
 - 0xab1a03f8 [  upload] [noclip] [ alpha] [  merged]  Nodes:    1  Vertices:     4  Indices:     6  root: 0xb2191d38 opacity: 1
 - 0xab1a1358 [  upload] [noclip] [ alpha] [  merged]  Nodes:    1  Vertices:     4  Indices:     6  root: 0xb2191d38 opacity: 1
 - 0xab1a15e0 [  upload] [noclip] [ alpha] [  merged]  Nodes:    2  Vertices:    12  Indices:    18  root: 0xb2191d38 opacity: 1
 - 0xab1a0758 [  upload] [noclip] [ alpha] [unmerged]  Nodes:    2  Vertices:     8  Indices:     0  root: 0xb2191d38
 - 0xab1a16b8 [  upload] [noclip] [ alpha] [  merged]  Nodes:    3  Vertices:    44  Indices:    66  root: 0xb2191d38 opacity: 1
 - 0xab1a1868 [  upload] [noclip] [ alpha] [  merged]  Nodes:    1  Vertices:    12  Indices:    18  root: 0xb2191d38 opacity: 1
 - 0xab19fdc8 [  upload] [noclip] [ alpha] [  merged]  Nodes:    1  Vertices:     4  Indices:     6  root: 0xb2191d38 opacity: 1
 - 0xab1a1a18 [  upload] [noclip] [ alpha] [  merged]  Nodes:    4  Vertices:    32  Indices:    48  root: 0xb2191d38 opacity: 1
 - 0xab1a1af0 [  upload] [noclip] [ alpha] [  merged]  Nodes:    3  Vertices:    28  Indices:    42  root: 0xb2191d38 opacity: 1
 - 0xaca4f260 [  upload] [noclip] [ alpha] [unmerged]  Nodes:    3  Vertices:    12  Indices:     0  root: 0xb2191d38
 - 0xab1a1d78 [  upload] [noclip] [ alpha] [  merged]  Nodes:    2  Vertices:    28  Indices:    42  root: 0xb2191d38 opacity: 1
 - 0xab1a0e18 [  upload] [noclip] [ alpha] [  merged]  Nodes:    1  Vertices:    16  Indices:    24  root: 0xb2191d38 opacity: 1
 - 0xab1a20d8 [  upload] [noclip] [ alpha] [  merged]  Nodes:    1  Vertices:     4  Indices:     6  root: 0xb2191d38 opacity: 1
 - 0xab1a0b90 [  upload] [noclip] [ alpha] [  merged]  Nodes:    6  Vertices:    72  Indices:   108  root: 0xb2191d38 opacity: 1
 - 0xb2183918 [  upload] [noclip] [ alpha] [  merged]  Nodes:    1  Vertices:    12  Indices:    18  root: 0xb2191d38 opacity: 1
 - 0xab19fd10 [  upload] [noclip] [ alpha] [  merged]  Nodes:    1  Vertices:     4  Indices:     6  root: 0xb2191d38 opacity: 1
 - 0xab1a09e0 [  upload] [noclip] [ alpha] [  merged]  Nodes:    7  Vertices:   288  Indices:   488  root: 0xb2191d38 opacity: 1
 - 0xab1a0ef0 [  upload] [noclip] [ alpha] [unmerged]  Nodes:    1  Vertices:     9  Indices:     0  root: 0xb2191d38
 - 0xab1a0fc8 [  upload] [noclip] [ alpha] [unmerged]  Nodes:    2  Vertices:     8  Indices:     0  root: 0xb2191d38
 - 0xab1a11a8 [  upload] [noclip] [ alpha] [  merged]  Nodes:    1  Vertices:    40  Indices:    60  root: 0xb2191d38 opacity: 1
 - 0xab1a0c68 [  upload] [noclip] [ alpha] [  merged]  Nodes:    1  Vertices:    16  Indices:    24  root: 0xb2191d38 opacity: 1
 - 0xac119290 [  upload] [noclip] [ alpha] [  merged]  Nodes:    1  Vertices:     4  Indices:     6  root: 0xb2191d38 opacity: 1
 - 0xab1a0908 [  upload] [noclip] [ alpha] [unmerged]  Nodes:    2  Vertices:    82  Indices:     0  root: 0xb2191d38
 - 0xab167320 [  upload] [noclip] [ alpha] [unmerged]  Nodes:    1  Vertices:     4  Indices:     0  root: 0xb2191d38
 - 0xab1a1508 [  upload] [noclip] [ alpha] [  merged]  Nodes:    1  Vertices:    20  Indices:    30  root: 0x0 opacity: 1
 -> times: build: 0, prepare(opaque/alpha): 0/0, sorting: 0, upload(opaque/alpha): 0/0, record rendering: 20
Renderer::render() QSGAbstractRenderer(0xb218b440) "rebuild: partial"
Rendering:
 -> Opaque: 13 nodes in 2 batches...
 -> Alpha: 81 nodes in 37 batches...
 - 0xab1a1280 [retained] [noclip] [opaque] [  merged]  Nodes:    1  Vertices:     4  Indices:     6  root: 0x0
 - 0xab167320 [  upload] [noclip] [opaque] [unmerged]  Nodes:   12  Vertices:    48  Indices:     0  root: 0xb2191d38
 - 0xab1a0908 [  upload] [noclip] [ alpha] [  merged]  Nodes:    7  Vertices:   288  Indices:   488  root: 0xb2191d38 opacity: 1
 - 0xac119290 [  upload] [noclip] [ alpha] [unmerged]  Nodes:    4  Vertices:    16  Indices:     0  root: 0xb2191d38
 - 0xab1a0c68 [  upload] [noclip] [ alpha] [  merged]  Nodes:    1  Vertices:    64  Indices:    96  root: 0xb2191d38 opacity: 1
 - 0xab1a11a8 [  upload] [noclip] [ alpha] [  merged]  Nodes:    1  Vertices:    72  Indices:   108  root: 0xb2191d38 opacity: 1
 - 0xab1a0fc8 [  upload] [noclip] [ alpha] [unmerged]  Nodes:    2  Vertices:    82  Indices:     0  root: 0xb2191d38
 - 0xab1a0ef0 [  upload] [  clip] [ alpha] [  merged]  Nodes:    1  Vertices:     6  Indices:     8  root: 0xb2190f58 opacity: 1
 - 0xab1a09e0 [  upload] [noclip] [ alpha] [unmerged]  Nodes:    1  Vertices:     4  Indices:     6  root: 0xb2191d38
 - 0xab19fd10 [  upload] [noclip] [ alpha] [  merged]  Nodes:    3  Vertices:   240  Indices:   404  root: 0xb2191d38 opacity: 1
 - 0xb2183918 [  upload] [noclip] [ alpha] [unmerged]  Nodes:    3  Vertices:    12  Indices:     0  root: 0xb2191d38
 - 0xab1a0b90 [  upload] [noclip] [ alpha] [  merged]  Nodes:    2  Vertices:    60  Indices:    90  root: 0xb2191d38 opacity: 1
 - 0xab1a20d8 [  upload] [noclip] [ alpha] [  merged]  Nodes:    1  Vertices:    12  Indices:    18  root: 0xb2191d38 opacity: 1
 - 0xab1a0e18 [  upload] [noclip] [ alpha] [  merged]  Nodes:    1  Vertices:     4  Indices:     6  root: 0xb2191d38 opacity: 1
 - 0xab1a1d78 [  upload] [noclip] [ alpha] [  merged]  Nodes:    4  Vertices:    52  Indices:    78  root: 0xb2191d38 opacity: 1
 - 0xaca4f260 [  upload] [noclip] [ alpha] [  merged]  Nodes:    1  Vertices:     4  Indices:     6  root: 0xb2191d38 opacity: 1
 - 0xab1a1af0 [  upload] [noclip] [ alpha] [  merged]  Nodes:    1  Vertices:     4  Indices:     6  root: 0xb2191d38 opacity: 1
 - 0xab1a1a18 [  upload] [noclip] [ alpha] [  merged]  Nodes:    2  Vertices:    12  Indices:    18  root: 0xb2191d38 opacity: 1
 - 0xab19fdc8 [  upload] [noclip] [ alpha] [unmerged]  Nodes:    2  Vertices:     8  Indices:     0  root: 0xb2191d38
 - 0xab1a1868 [  upload] [noclip] [ alpha] [  merged]  Nodes:    3  Vertices:    44  Indices:    66  root: 0xb2191d38 opacity: 1
 - 0xab1a16b8 [  upload] [noclip] [ alpha] [  merged]  Nodes:    1  Vertices:    12  Indices:    18  root: 0xb2191d38 opacity: 1
 - 0xab1a0758 [  upload] [noclip] [ alpha] [  merged]  Nodes:    1  Vertices:     4  Indices:     6  root: 0xb2191d38 opacity: 1
 - 0xab1a15e0 [  upload] [noclip] [ alpha] [  merged]  Nodes:    4  Vertices:    32  Indices:    48  root: 0xb2191d38 opacity: 1
 - 0xab1a1358 [  upload] [noclip] [ alpha] [  merged]  Nodes:    3  Vertices:    28  Indices:    42  root: 0xb2191d38 opacity: 1
 - 0xab1a03f8 [  upload] [noclip] [ alpha] [unmerged]  Nodes:    3  Vertices:    12  Indices:     0  root: 0xb2191d38
 - 0xab1a1790 [  upload] [noclip] [ alpha] [  merged]  Nodes:    2  Vertices:    28  Indices:    42  root: 0xb2191d38 opacity: 1
 - 0xab1a0320 [  upload] [noclip] [ alpha] [  merged]  Nodes:    1  Vertices:    16  Indices:    24  root: 0xb2191d38 opacity: 1
 - 0xab1a1f28 [  upload] [noclip] [ alpha] [  merged]  Nodes:    1  Vertices:     4  Indices:     6  root: 0xb2191d38 opacity: 1
 - 0xab1a1e50 [  upload] [noclip] [ alpha] [  merged]  Nodes:    6  Vertices:    72  Indices:   108  root: 0xb2191d38 opacity: 1
 - 0xab1a0010 [  upload] [noclip] [ alpha] [  merged]  Nodes:    2  Vertices:    16  Indices:    24  root: 0xb2191d38 opacity: 1
 - 0xab1a1430 [  upload] [noclip] [ alpha] [  merged]  Nodes:    7  Vertices:   288  Indices:   488  root: 0xb2191d38 opacity: 1
 - 0xab1a2000 [  upload] [noclip] [ alpha] [unmerged]  Nodes:    1  Vertices:     9  Indices:     0  root: 0xb2191d38
 - 0xab10a4b0 [  upload] [noclip] [ alpha] [unmerged]  Nodes:    2  Vertices:     8  Indices:     0  root: 0xb2191d38
 - 0xab10af00 [  upload] [noclip] [ alpha] [  merged]  Nodes:    1  Vertices:    40  Indices:    60  root: 0xb2191d38 opacity: 1
 - 0xab10ddf8 [  upload] [noclip] [ alpha] [  merged]  Nodes:    1  Vertices:    16  Indices:    24  root: 0xb2191d38 opacity: 1
 - 0xab1a0ab8 [  upload] [noclip] [ alpha] [  merged]  Nodes:    1  Vertices:     4  Indices:     6  root: 0xb2191d38 opacity: 1
 - 0xab19ff38 [  upload] [noclip] [ alpha] [unmerged]  Nodes:    2  Vertices:    82  Indices:     0  root: 0xb2191d38
 - 0xab1a0d40 [  upload] [noclip] [ alpha] [unmerged]  Nodes:    1  Vertices:     4  Indices:     0  root: 0xb2191d38
 - 0xab1a1508 [retained] [noclip] [ alpha] [  merged]  Nodes:    1  Vertices:    20  Indices:    30  root: 0x0 opacity: 1
 -> times: build: 0, prepare(opaque/alpha): 0/0, sorting: 0, upload(opaque/alpha): 0/0, record rendering: 16
```

jpetrell (approved):

data/common/AcInputSystemInfo.qml
```diff
@@ -20,8 +20,14 @@ QtObject {
...
 	readonly property string serviceName: _serviceName.value || "" // e.g. com.victronenergy.vebus.ttyO, com.victronenergy.grid.ttyO
 	readonly property int source: _source.value === undefined ? VenusOS.AcInputs_InputSource_NotAvailable : _source.value
-	readonly property real minimumCurrent: _minimumCurrent.value === undefined ? NaN : Global.acInputs.clampMeasurement(_minimumCurrent.value)
-	readonly property real maximumCurrent: _maximumCurrent.value === undefined ? NaN : Global.acInputs.clampMeasurement(_maximumCurrent.value)
+	readonly property real minimumCurrent: _minimumCurrent.value === undefined ? NaN
+			: _feedbackEnabled ? _minimumCurrent.value
+			: Math.max(0, _minimumCurrent.value)
+	readonly property real maximumCurrent: _maximumCurrent.value === undefined ? NaN
+			: _feedbackEnabled ? _maximumCurrent.value
+			: Math.max(0, _maximumCurrent.value)
```
jpetrell, line 28: This had measurable performance impact?

jpetrell (approved):

components/CircularMultiGauge.qml
```diff
@@ -67,8 +67,8 @@ Item {
 						startAngle: 0
 						endAngle: 270
 						value: loader.value
-						progressColor: Theme.statusColorValue(loader.gaugeStatus)
-						remainderColor: Theme.statusColorValue(loader.gaugeStatus, true)
+						progressColor: Theme.color_darkOk,Theme.statusColorValue(loader.gaugeStatus)
+						remainderColor: Theme.color_darkOk,Theme.statusColorValue(loader.gaugeStatus, true)
```
jpetrell, line 71: Getting do not use comma expressions warnings in Qt Creator. But I guess these are needed to make sure the color changes on light/dark style switch.

components/widgets/WidgetConnector.qml
```diff
@@ -200,7 +200,9 @@ Item {
 
 				delegate: Image {
 					opacity: 0.0
-					source: "qrc:/images/electron.svg"
+					source: (Qt.platform.os !== "linux" || Global.isDesktop) ? "qrc:/images/electron.svg"
+						: Theme.colorScheme === Theme.Dark ? "qrc:/images/electron-dark-8.bmp"
+						: "qrc:/images/electron-light-8.bmp"
```
jpetrell, line 212: Did this have notable performance impact, improvement? If not I would sacrifice bit of performance for nicer electrons.

themes/color/Light.json
```diff
@@ -15,7 +26,7 @@
     "color_critical_background": "#F58B87",
 
     "color_font_primary": "color_gray2",
-    "color_font_secondary": "color_dimGray2",
+    "color_font_secondary": "color_dimGray",
```
jpetrell, line 29: Not a fan of gray labels, not sure the performance benefit of avoiding alpha blending on thin glyphs is worth the visual regression.

jpetrell (approved):

jpetrell (commented):
These looks good:
- Add mock key handler to turn animations on/off
- Reduce the number of electrons drawn on OverviewPage
- Reduce cost of phase text color binding in ThreePhaseDisplay
- Reduce cost of value binding in ElectricalQuantityLabel
- Avoid AcInputs.clampMeasurement()
- SideMultiGauge should only track values while page is visible 

I am surprised to see performance benefits didn't turn out to ugly code (like usually with these optimizations...). :partying_face:  I really like the splitting of big JavaScript bindings to smaller ones. Not only to reduce cost of value bindings, but also because it even made the code easier to read.

These two I am most unsure about:
- Use opaque colors
- Use opaque electrons on GX products 

5% percentage point improvement (and not sure how much of that comes from opaque colors) is not huge so wonder if avoiding alpha blending on fonts and electrons is really worth the visual regression. Especially the gray font coloring stands out. Browser, Ekrano and future devices will be able to handle alpha blending just fine. Could maybe just limit the rough electrons to Cerbo product line, where relative CPU load is often multiple of Ekrano levels.

> I think for these specific use-cases the performance gain vastly outweighs the very slight difference in look/feel compared to baseline

Yeah, for most people they provide equivalent experience (and performance matters). Though we are not seeing here vast total performance differences overall, even if individual component-wise the gains are big.

Code here looks good:
- Use CheapBarGauge in SidePanel and OverviewPage widgets 

But in some cases the rounding seems to be missing (not sure if regression by this PR):
image

components/CircularMultiGauge.qml
chriadam: Yeah :-/  I wonder if there is a better way to do this.  It's common enough that I'd prefer not to have a 5-way ternary in each binding, but maybe it's better than this?

components/widgets/WidgetConnector.qml
chriadam: It didn't have much impact on its own.  I will do all of the perf improvements, and then remove this if it doesn't make a difference.  Keeping it for now, because I'm trying to see: if I reduce the number of non-mergable batches down far enough, is there a "step threshold" where everything gets much better?  But, if it's just random bindings and javascript handlers which we can fix one by one by moving to C++, then this electron replacement can be removed.  Let's see.

themes/color/Light.json
chriadam: As with the electrons thing - it didn't have much impact on its own.  Let's wait until I've finished the other things (JS->C++, etc), and then we can remove this change if it proves unnecessary.

chriadam:
That missing rounding issue is definitely a bug.  I suspect it's if the height of the fg rect is less than the specified radius, then Qt just gives up and draws it as rectangular :-/  I can fix this (anchor horizontal centre and reduce fgRect width by 2 if value = 1; set color = transparent if value = 0).  But, as mentioned, I think for these overview widget gauges we should probably try removing the rounding altogether, as rects with rounding are considered non-opaque by the batcher, I believe.

data/common/AcInputSystemInfo.qml
chriadam: It showed up prominently in the QML analyzer profile before this change.  After this change, it doesn't.  BUT the CPU utilisation reported by top didn't change noticeably.  Shrug.  I think it's just a large number of inefficient bindings and handlers adding up to a lot.  JS functions are really problematic, unfortunately.

themes/color/Light.json
chriadam: Especially since (presumably due to antialiasing) the text glyphs end up being rendered with alpha channel anyway, maybe there is literally zero point in using opaque font colours.  Will investigate this next week.  I also want to set the renderType to Text.NativeText to see if the distance field rendering is more or less expensive than opentype or whatever is used on Linux by default.  The goal is to get maximal batching and merging, and minimise the number of reuploads required.  I suspect (but not completely sure) that text has a big part in this, given what I was seeing from the QSG_RENDERER_DEBUG=render output.

jpetrell:
>  I think for these overview widget gauges we should probably try removing the rounding altogether,
> as rects with rounding are considered non-opaque by the batcher

Small aesthetic like rounding may seem secondary to performance, but suspect we are talking about sub-percentage point improvements to the total ~40% CPU usage here, which is not that useful individual improvement either. And definitely not an issue for newer devices like Ekrano GX. Also, we already pause animations if the CPU load grows too high.

chriadam:
This is now ready for review.  I stripped out the changes which were somewhat controversial (e.g. opaque fonts, electron changes, etc), and also the BatteryWidget label fitting change (even though that one is still expensive and causes electron stutters when that animation is enabled, but it's not necessary to fix for 1.0.0 as per Matthijs instructions to me the other day).

pages/settings/debug/PageDebug.qml
```diff
@@ -48,16 +60,11 @@ Page {
...
-
-				//% "Quit"
-				button.text: qsTrId("settings_page_debug_quit")
-
-				onClicked: Qt.quit()
+			SwitchItem {
+				//% "Pause electron animations"
+				text: qsTrId("settings_page_debug_pause_electron_animations")
+				checked: Global.pauseElectronAnimations
+				onClicked: Global.pauseElectronAnimations = !Global.pauseElectronAnimations
```
jpetrell, line 67: Nice!

src/cpuinfo.cpp
```diff
@@ -51,20 +57,54 @@ CpuInfo::CpuInfo(QObject *parent)
...
+
+CpuInfoWorker::CpuInfoWorker(QObject *parent)
+	: QObject(parent)
+{
+}
+
+#ifdef Q_OS_LINUX
+void CpuInfoWorker::calculateCpuUage(unsigned long long previousBusy , unsigned long long previousIdle)
+#else
+void CpuInfoWorker::calculateCpuUage(unsigned long long, unsigned long long)
```
jpetrell, line 101: Typo

src/clocktime.h
```diff
@@ -33,24 +33,24 @@ class ClockTime : public QObject
...
-	Q_PROPERTY(QString currentDate READ currentDate NOTIFY currentDateChanged)
-	Q_PROPERTY(QString currentTime READ currentTime NOTIFY currentTimeChanged)
-	Q_PROPERTY(QString currentDateTimeUtc READ currentDateTimeUtc NOTIFY currentDateTimeUtcChanged)
+	Q_PROPERTY(int year READ year NOTIFY yearChanged FINAL)
+	Q_PROPERTY(int month READ month NOTIFY monthChanged FINAL)
+	Q_PROPERTY(int day READ day NOTIFY dayChanged FINAL)
+	Q_PROPERTY(int hour READ hour NOTIFY hourChanged FINAL)
+	Q_PROPERTY(int minute READ minute NOTIFY minuteChanged FINAL)
+	Q_PROPERTY(int second READ second NOTIFY secondChanged FINAL)
+	Q_PROPERTY(int msec READ msec NOTIFY msecChanged FINAL)
```
jpetrell, line 46: Nice code quality improvements. :raised_hands:  Overriding base component properties seems wrong, wonder what is the use case here of not having FINAL always enabled. I guess sometimes it provides easy way to workaround an issue if you cannot touch the base component.

components/Gauges.js
```diff
@@ -8,28 +8,6 @@
...
-function getValueStatus(value, valueType) {
-	if (valueType === V.VenusOS.Gauges_ValueType_RisingPercentage) {
-		return statusFromRisingValue(value)
-	}
-	if (valueType === V.VenusOS.Gauges_ValueType_FallingPercentage) {
-		return statusFromFallingValue(value)
-	}
-	return V.Theme.Ok
-}
-
```
jpetrell, line 32: Noticed during my performance work that moving functions to C++ was not always faster, I guess due to context switch.

components/widgets/WidgetConnector.qml
```diff
@@ -215,29 +217,19 @@ Item {
...
-		enabled: root._animated
-		target: overviewPageRootAnimation
-
-		function onUpdate(elapsedTime) {
-			var progress = 1000*elapsedTime/electronAnim.duration
-			progress = progress - Math.floor(progress)
-			pathUpdater.progress = startToEnd ? progress : 1.0 - progress
-		}
-	}
-
```
jpetrell, line 233: :muscle:

jpetrell (approved):
Phew you have done a lot. Lot of good code quality improvements too here, not only performance optimisations. Added some small suggestions, but other than the typo didn't spot issues caused by the PR. I saw few oddities with side panel graphs and side gauges, but they are probably not related to the changes here (will comment more in the Slack).

Was not sure what the visual difference between cheap and expensive gauges were. Is the only difference that the value bar is not rounded? IMO both look good so would just drop the more expensive one. Sorry earlier tought the rounding would be dropped on the track background, which would look worse but IMO cheap gauge doesn't now even look like a visual compromise.

blammit:
> AcInputPhaseModel.qml _phaseObjects onValueChanged for power
> SystemData.qml _updateConsumptionModel()
> PvMonitor.qml updateAcTotals()
> DcInputWidget.qml _refreshTotalPower()

@chriadam https://github.com/victronenergy/gui-v2/issues/1010 is a task for adding exactly these things to systemcalc so that gui-v2 doesn't need to calculate them. Should that have a higher priority?

Global.qml
```diff
@@ -61,6 +62,8 @@ QtObject {
 	property bool isDesktop
 	property real scalingRatio: 1.0
 
+	property bool animationEnabled: true // for mock mode only.
```
blammit, line 65: Does anything set this to `false` in non-mock mode? I can't find it.

Also if it's only for mock mode, can it be in Global.mockDataSimulator or something instead? Or perhaps MockDataSimulator can change the value of `Global.mainView.allowPageAnimations`? It would be really nice to not add things to Global that are only used in mock mode.

components/CheapBarGauge.qml
```diff
@@ -0,0 +1,58 @@
+/*
+** Copyright (C) 2023 Victron Energy B.V.
+** See LICENSE.txt for license information.
```
blammit, line 3: Could you add a comment here about when to use CheapBarGauge vs BarGauge? And something in the commit message?

components/ThreePhaseDisplay.qml
```diff
@@ -35,22 +35,16 @@ Flow {
...
-				Theme.color_darkOk // force capture since Theme.statusColorValue() doesn't.
-				return (valueStatus === Theme.Critical || valueStatus === Theme.Warning)
-					? Theme.statusColorValue(valueStatus)
+			readonly property bool feedingToGrid: Global.systemSettings.essFeedbackToGridEnabled
+					&& root.inputMode
+					&& (model.power || 0) < 0
+			readonly property int valueStatus: feedingToGrid ? Theme.Ok
+					: root.phaseModelProperty ? Gauges.getValueStatus(valueRange.valueAsRatio * 100, root.valueType)
+					: Theme.Ok
+			readonly property bool statusIssue: valueStatus === Theme.Critical || valueStatus === Theme.Warning
```
blammit, line 43: Maybe `criticalOrWarning` instead of `statusIssue`?

data/common/AcInputSystemInfo.qml
blammit: I'm a bit worried that this kind of change makes it harder to maintain -- now it's harder to see why `Math.max(0, _maximumCurrent.value)` is used in various places. Having the `Global.acInputs.clampMeasurement()` function essentially also documents why the clamping is necessary, and that gets lost by moving the implementation to various parts of the code base.

Ideally the minimum would just be clamped by systemcalc anyway. But with this change, it will become confusing that Global.acInputs.clampMeasurement() is used in some cases and not in others, particularly when code gets moved around and gets refactored.

What if we changed Global.qml to use:

```
property AcInputs acInputs
property SystemSettings systemSettings
```

instead of:

```
property var acInputs
property var systemSettings
```

Would this be enough to improve the performance, and drop this commit?

pages/BriefPage.qml
```diff
@@ -227,6 +227,7 @@ SwipeViewPage {
 				minimumValue: !!Global.acInputs.activeInputInfo ? Global.acInputs.activeInputInfo.minimumCurrent : NaN
 				maximumValue: !!Global.acInputs.activeInputInfo ? Global.acInputs.activeInputInfo.maximumCurrent : NaN
 				inputMode: true
+				gaugeVisible: root.isCurrentPage
```
blammit, line 230: I would have hoped we could use `visible` instead of adding another property, but is that not the case?

data/common/AcInputSystemInfo.qml
chriadam: Javascript functions are untyped by definition.  They can return a value of any type, and can resolve arbitrary symbols.  When the binding optimiser encounters any call to a js function, it gives up and instead the entire binding expression is evaluated by the full js context, I believe.

chriadam: I've updated the various places with comments which point to AcInputs.clampMeasurement() purely as documentation.  I agree that it does have a maintenance impact, but the perf impact on device is worse, I think.

src/cpuinfo.cpp
chriadam: thanks, fixed

components/ThreePhaseBarGauge.qml
```diff
@@ -45,28 +46,49 @@ Flow {
...
+				minimumValue: root.minimumValue
+				maximumValue: root.maximumValue
+			}
 
+			Loader {
+				id: gaugeLoader
+				anchors.right: parent.right
+				width: parent.width - (phaseLabel.visible ? phaseLabel.width : 0)
+				height: parent.height
+				sourceComponent: (Qt.platform.os === "linux" && !Global.isDesktop) ? cheapGauge : prettyGauge
```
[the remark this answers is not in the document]
chriadam: good point, done

components/Gauges.js
chriadam: yeah.  it should be faster if called from an optimised binding expression, though.

Global.qml
chriadam: added a commit on top to do this, but honestly it feels even uglier to me.

components/AsymmetricRoundedRectangle.qml
```diff
@@ -18,34 +18,7 @@ Item {
...
-			topMargin: root.height - root.border.width
-			leftMargin: root.roundedSide === VenusOS.AsymmetricRoundedRectangle_RoundedSide_Left ? root.radius : 0
-			rightMargin: root.roundedSide === VenusOS.AsymmetricRoundedRectangle_RoundedSide_Right ? root.radius : 0
-		}
-
-		visible: root.roundedSide !== VenusOS.AsymmetricRoundedRectangle_RoundedSide_All
-
-		color: roundedRect.border.color
-	}
+	// the order matters, because any of the colors may be fully opaque...
```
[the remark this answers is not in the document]
chriadam: yes, clarified

components/SegmentedButtonRow.qml
```diff
@@ -14,6 +14,7 @@ Item {
 	property int fontPixelSize: Theme.font_size_body3
 	property alias model: buttonRepeater.model
 	property int currentIndex
+	property bool flat
```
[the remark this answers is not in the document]
chriadam: removed this change, we don't have any flat segmented button rows any more, so the change was unnecessary

components/CheapBarGauge.qml
chriadam: done

components/ThreePhaseDisplay.qml
chriadam: done

pages/BriefPage.qml
chriadam: I double checked, and it works fine with visible.  I have no idea why I added gaugeVisible.  Anyway, removed, using visible instead.  thanks!

Global.qml
blammit: I was thinking of using a Binding object, but I guess this works too

blammit (approved):
Nice work on tracking down all of these!

data/common/AcInputServiceLoader.qml
```diff
@@ -12,7 +12,10 @@ Loader {
 	property string serviceUid
 	property string serviceType
 
-	readonly property real totalPower: hasTotalPower && item ? Global.acInputs.clampMeasurement(item.power) : NaN
+	readonly property real totalPower: (!hasTotalPower || !item) ? NaN
+			: _feedbackEnabled ? _itemPower
+			: _itemPower > 0 ? _itemPower // see AcInputs.clampMeasurement()
+			: 0.0
```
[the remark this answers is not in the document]
chriadam: good catch, will fix

Global.qml
chriadam: Oh, silly me, that makes far more sense.  Will update.

data/common/AcInputServiceLoader.qml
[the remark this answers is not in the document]
chriadam: done.

Global.qml
chriadam: done.

chriadam:
rebased.

jpetrell:
Tested the PR again, didn't spot any issues. :+1:

PR #1181, opened 2024-06-03
Use 5" inch layout on small mobile screens (wasm)

### LINKED ISSUE #919: Use smaller 5" layout variant on phones (wasm)
Current WASM build uses the larger 7" display 1024:600 layout variant. This is a good choice for large desktop and tablet screens, but the smaller 5" variant might be more appropriate with bigger text and buttons when accessing the website from small phone screens.

jpetrell:
Fixes #919.

### DIFF
--- a/src/theme.cpp
+++ b/src/theme.cpp
@@ -1,0 +1,72 @@
+/*
+** Copyright (C) 2024 Victron Energy B.V.
+** See LICENSE.txt for license information.
+*/
+
+#include "theme.h"
+
+using namespace Victron::VenusOS;
+
+#if defined(VENUS_WEBASSEMBLY_BUILD)
+#include <emscripten.h>
+
+EM_JS(int, getScreenWidth, (), {
+  return Math.min(screen.width, screen.height);
+});
+
+#endif
+
+Theme::Theme(QObject *parent) : QObject(parent)
+{
+#if defined(VENUS_WEBASSEMBLY_BUILD)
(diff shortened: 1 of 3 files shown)


src/theme.cpp
```diff
@@ -0,0 +1,71 @@
...
+#include "theme.h"
+
+using namespace Victron::VenusOS;
+
+#if defined(VENUS_WEBASSEMBLY_BUILD)
+#include <emscripten.h>
+
+EM_JS(int, getScreenWidth, (), {
+  return Math.min(screen.width, screen.height);
+});
```
jpetrell, line 15: Cannot declare EM_JS macro on the header as that will result in duplicate symbols error
as any place that includes the header will have own version of the same global function.

src/theme.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+#endif
+
+Theme::Theme(QObject *parent) : QObject(parent)
+{
+#if defined(VENUS_WEBASSEMBLY_BUILD)
+	setScreenSize(getScreenWidth() >= 480
+				  ? Victron::VenusOS::Theme::SevenInch
+				  : Victron::VenusOS::Theme::FiveInch);
```
jpetrell, line 24: Smartphones have 320 - 480 CSS independent pixel wide screens.

blammit: Could you add this as a source comment please?

jpetrell: Good idea. Done. :+1:

chriadam (approved):

PR #1182, opened 2024-06-03
Suppress Android keyboard if no text field has focus

### LINKED ISSUE #1154: After typing the virtual keyboard always comes up on a tap on Android (wasm)
Open gui-v2 on the local network by pointing Android browser to [your_device_ip_address]/gui-beta. Open virtual keyboard by tapping any text field in gui-v2, e.g. one of the fields in Settings > Display & Languages > Minimum and maximum gauge ranges.

Virtual keyboard should open. Now close the keyboard. Then tap anywhere on the gui-v2 area.

Actual result: The keyboard reopens.

Expected result: The keyboard only reopens if you tap another text field.

When debugging with Chrome remote device debugger `chrome://inspect/#devices` I can see that initially `<body>` element is the active element (document.activeElement ). After focusing the text field a new `<input>` element becomes the active element, which it stays unless you reload the whole page.

https://github.com/victronenergy/gui-v2/assets/2203667/dd10ebac-08d1-40e9-ab97-7c172a2958a5

jpetrell:
Fixes #1154.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -22,6 +22,7 @@
 #include <QQmlEngine>
 #include <QQuickWindow>
 #include <QCommandLineParser>
+#include <QQuickItem>
 #include <QStyleHints>
 
 #include <QtDebug>
@@ -228,6 +229,27 @@ EM_JS(int, getWindowInnerHeight, (), {
   return window.innerHeight;
 });
 
+EM_JS(void, setContentEditable, (bool editable), {
+	// Work-around Qt Android issue where keyboard constantly pops up (see QTBUG-88803)
+	const android = /Android/i.test(navigator.userAgent);
+	if (android) {
+		const inputs = document.querySelectorAll('input[type="text"]');
+		for (let i = 0; i < inputs.length; i++) {
+			const input = inputs[i];
+			const rect = input.getBoundingClientRect();
+
+			// Qt <input> has no identifier so identify using off-screen co-ordinates.
+			if (rect.x === -1000 && rect.y === -1000) {
+				input.style.visibility = editable ? "visible" : "hidden";
+				if (editable) {
+					input.focus();
+				}
+			}
+		}
+	}
+});
+
+
 #endif
 
 int main(int argc, char *argv[])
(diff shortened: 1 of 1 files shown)


jpetrell:
> WIP. Still need to verify the fix end-to-end.

No longer WIP, ready for review.

src/main.cpp
```diff
@@ -291,6 +313,14 @@ int main(int argc, char *argv[])
 
 	QScopedPointer<QObject> object(component.beginCreate(engine.rootContext()));
 	const auto window = qobject_cast<QQuickWindow *>(object.data());
+
+	QObject::connect(window, &QQuickWindow::activeFocusItemChanged, [window] {
+		const bool editable = window->activeFocusItem() != nullptr && (window->activeFocusItem()->flags() & QQuickItem::ItemAcceptsInputMethod);
+#if defined(VENUS_WEBASSEMBLY_BUILD)
```
chriadam, line 319: can the entire signal connection be within the ifdef, rather than just the setContentEditable call?

src/main.cpp
```diff
@@ -228,6 +229,27 @@ EM_JS(int, getWindowInnerHeight, (), {
...
+	const android = /Android/i.test(navigator.userAgent);
+	if (android) {
+		const inputs = document.querySelectorAll('input[type="text"]');
+		for (let i = 0; i < inputs.length; i++) {
+			const input = inputs[i];
+			const rect = input.getBoundingClientRect();
+
+			// Qt <input> has no identifier so identify using off-screen co-ordinates.
+			if (rect.x === -1000 && rect.y === -1000) {
+				input.style.visibility = editable ? "visible" : "hidden";
```
chriadam, line 243: Excellent!  I had not thought of this.  Great work.

chriadam (approved):
Tested, works nicely, one minor comment but LGTM.

src/main.cpp
jpetrell: Monkey patching at its finest. :wine_glass:

src/main.cpp
jpetrell: Fixed.

PR #1184, opened 2024-06-04
Use VeQuickItem::numberValue to reduce binding overheads

chriadam:


chriadam:
This one is risky, and unnecessary - we easily hit the performance requirement (in electrons-paused mode) without this PR.
But, it does seem "nice" to me.  Anyway, not for 1.0.0, but worth considering in future.

components/ThreePhaseBarGauge.qml
```diff
@@ -17,6 +17,7 @@ Flow {
 	property bool inputMode
 	property int orientation: Qt.Vertical
 	property bool animationEnabled
+	property bool inOverviewWidget
```
blammit, line 20: This is not used anywhere?

chriadam: leftovers from when I force pushed the other PR.  This PR is targeted at the other branch :-/

blammit (approved):
Great!

blammit:
We might be able to use nullish coalescing to get the same benefit instead, i.e. change

`readonly property real frequency: _frequency.value === undefined ? NaN : _frequency.value`

to

`readonly property real frequency: _frequency.value ?? NaN`

as long as that still results in one reading of `_frequency.value`.

I think we should start using this in the codebase generally to make things more concise.

chriadam:
Does QML's fast-path JS expression evaluator support nullish coalescing, or does it fall back to full JS context evaluation for expressions containing those?

blammit:
> Does QML's fast-path JS expression evaluator support nullish coalescing, or does it fall back to full JS context evaluation for expressions containing those?

No idea, probably requires hunting through the source to find out.

PR #1187, opened 2024-06-04
Alternator: Engine Temperature: Added support

nmbath:
Added support to display engine temperature.

### DIFF
--- a/pages/settings/devicelist/dc-in/PageAlternatorModel.qml
+++ b/pages/settings/devicelist/dc-in/PageAlternatorModel.qml
@@ -28,7 +28,7 @@ ObjectModel {
 	}
 
 	ListTemperatureItem {
-		text: CommonWords.temperature
+		text:  qsTrId("dc_alternator_temperature")
 		dataItem.uid: root.bindPrefix + "/Dc/0/Temperature"
 		allowed: defaultAllowed && dataItem.isValid
 	}
@@ -85,7 +85,7 @@ ObjectModel {
 	}
 
 	ListTemperatureItem {
-		text: CommonWords.temperature
+		text: qsTrId("alternator_engine_temperature")
 		dataItem.uid: root.bindPrefix + "/Engine/Temperature"
 		allowed: defaultAllowed && dataItem.isValid
 	}
--- a/pages/settings/devicelist/dc-in/PageAlternatorModel.qml
+++ b/pages/settings/devicelist/dc-in/PageAlternatorModel.qml
@@ -31,7 +31,6 @@ ObjectModel {
 		text: CommonWords.temperature
 		dataItem.uid: root.bindPrefix + "/Dc/0/Temperature"
 		allowed: defaultAllowed && dataItem.isValid
-		onHeightChanged: console.trace()
 	}
 
 	ListTextItem {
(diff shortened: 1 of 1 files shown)


pages/settings/devicelist/dc-in/PageAlternatorModel.qml
```diff
@@ -85,6 +85,13 @@ ObjectModel {
 		allowed: defaultAllowed && dataItem.isValid
 	}
 
+	ListTemperatureItem {
+		text: CommonWords.temperature
+		dataItem.uid: root.bindPrefix + "/Engine/Temperature"
+		allowed: defaultAllowed && dataItem.isValid
+		onHeightChanged: console.trace()
```
chriadam, line 92: leftover debugging?

pages/settings/devicelist/dc-in/PageAlternatorModel.qml
```diff
@@ -85,6 +85,13 @@ ObjectModel {
 		allowed: defaultAllowed && dataItem.isValid
 	}
 
+	ListTemperatureItem {
+		text: CommonWords.temperature
```
DanielMcInnes, line 89: The change looks good, but I don't see '.../Engine/Temperature' on the bus (I do see '.../Engine/Speed'). Does this need other customized firmware to work?

nmbath: Yes this is a new dbus path that has been added through the ve.can driver [here](https://github.com/victronenergy/dbus-vecan/commit/9cb67d8121041330f2257d34b6e84a38bd33b2d2)

DanielMcInnes (commented):
lgtm (once the leftover debugging is removed)

nmbath:
I have updated to remove the console.trace().

pages/settings/devicelist/dc-in/PageAlternatorModel.qml
jpetrell: Can this be merged? Has the ve.can driver change been included in the existing releases?

nmbath: is in the main branch of vecan-dbus, however that has not yet been merged into VenusOS beta. It is on the todo list.

DanielMcInnes (approved):

msigman:
I love the idea of displaying the temp on the alternator box. However one suggestion is that alternator temp would likely be more useful than the engine temp for this. Devices like Wakespeed and Zeus have a temperature probe and already make the alternator temp available in the Cerbo.

nmbath:
Alternator temp is already displayed, as long as it has been supplied by the controller.

msigman:
> Alternator temp is already displayed, as long as it has been supplied by the controller.
> 
> In all cases this is displayed on the details page.

Understood, makes sense. I was thinking this was adding the temp display on the Overview tile, similar to the battery temp display.  Now I understand you are adding it in the Details page for the alternator. Great addition! 

A separate suggestion then for consideration would be to add the alternator temp to the Overview tile, like the battery temp. I suppose this should be logged as a new issue/feature request.
IMG_1163

chriadam:
@nmbath please ping me or Daniel when the required version of vecan-dbus has made its way into venus, so we can merge this one.  Thanks!

blammit (changes requested):
Needs an update to the temperature names, as commented.

pages/settings/devicelist/dc-in/PageAlternatorModel.qml
nmbath: @blammit I have differentiated using` sTrId("dc_alternator_temperature")` and `sTrId("alternator_engine_temperature")` however not sure what else is needed to define these strings or where they are defined.

blammit: @nmbath it just needs the Engineering English strings, like this:

```
//% "Alternator temperature"
text:  qsTrId("dc_alternator_temperature")
```

and

```
//% "Engine temperature"
text: qsTrId("alternator_engine_temperature")
```

(If those are the strings that should be displayed.)

nmbath: @blammit Many thanks. Done and pushed an udpate.

blammit (approved):

PR #1189, opened 2024-06-05
MQTT: Require portal id for VRM connections

chriadam:


chriadam:
Actually we should merge the veutil change into gui-v2 branch there first, and then recreate this PR.

jpetrell (approved):

PR #1191, opened 2024-06-06
Invert arc and bar gauges when feeding in to grid

### LINKED ISSUE #1145: Invert gauge bars when feed-in is enabled
image

The confusion arises because when feed-in is enabled, the power/current is negative, so the gauge bars move in the opposite direction. We could use an absolute value for the gauge instead when feed-in is enabled, so that a longer green bar indicates more power being fed back to the grid, rather than less.

blammit:
When feeding power to the grid, use an absolute value for the side arc gauges and bar gauges. This effectively reverses the gauge direction so that negative and positive values have the same value on the gauge, though negative values will be drawn in green.

Fixes #1145


chriadam (approved):
LGTM

mr-manuel:
Somehow all got more confusing right now, at least for me. The title of this PR says something other, than this PR do?
From the changed code, the grid values are absolute (always positive, pointing in the same direction) and 0 is always at the bottom of the gauge. But the issue and title say that the goal should be something like this https://github.com/victronenergy/gui-v2/issues/1145#issuecomment-2132992886?

Do I missunderstand something?

Also the max gauge values are now wrong. `2.2 A` is less then half gauge and `2.9 A` are more then half gauge. The set limit is `-17` and `17`.

venus-os

Screenshot 2024-06-06 084640

chriadam:
As per https://github.com/victronenergy/gui-v2/issues/1145#issuecomment-2134772777 the goal was "The idea is to use absolute AC IN values for the gauge and change the color of the bar to green when values are negative which means feeding back the grid."

The issue you point out with the ranging does look like a bug, however - I don't understand by 2.6A would appear > 50% if the range was [-17, 17].

chriadam:
The issue with out-of-sync animations is also a real issue, by the looks.  I am investigating now.

components/SideMultiGauge.qml
```diff
@@ -89,7 +89,13 @@ Item {
...
-				value: root.visible ? model[root.phaseModelProperty] : root.minimumValue
+
+				// When feeding in to grid, use an absolute value for the gauge. This effectively
+				// reverses the gauge direction so that negative and positive values have the same
+				// value on the gauge, though negative values will be drawn in green.
+				value: root.visible
+					   ? (feedingToGrid ? Math.abs(model[root.phaseModelProperty]) : model[root.phaseModelProperty])
+					   : root.minimumValue
 				minimumValue: root.minimumValue
 				maximumValue: root.maximumValue
```
jpetrell, line 100: UX-wise these already make sense, I can easily see the absolute quantity match the bars now, regardless if it is feeding to the grid or vice versa. :partying_face: 

Just wondering shouldn't minimumValue and maximumValue range somehow be adjusted? If I have minimum value -6.9 and maximum value 20.5 doesn't it mean that the bar only needs maximum absolute value from now on? I.e. the range is from 0 to absolute, and not from -6.9 to 20.5, making going from 0 to 100% distance of 20.5 amperes and not 20.5 - (-6.9) = 27.4 amperes like before.

mr-manuel: I think also only one max gauge value would be the easiest for now.

chriadam: That's a very good point... the ranging suddenly becomes very complicated if the (negative) minimum isn't equal to the (positive) maximum :-/
```
minimumValue: 0
maximumValue: Math.max(Math.abs(root.minimumValue), Math.abs(root.maximumValue))
```
perhaps?

chriadam: This also explains the "56% when at 2.2 amps" issue that MrManuel noted:
```
qml: debugging: value as ratio changed: 0.5635294142891379 from value: 2.1600000858306885, min: -17, max: 17
```
since 50% would be 0A.

chriadam:
Updated with some potential improvements for discussion

chriadam:
Hmm, my improvement is bad, because it flips the ranging depending on feed-in which means the range doesn't match for positive vs negative values.  I think MrManual was correct: having only a single maximum makes the most sense.

mr-manuel:
Max gauge ranges are ok again. Also the sync between the gauges and the values that change is much better!

venus-os

blammit:
Thanks for the discussion and updates! Makes sense to change the min/max values as well. I was wondering if we should only use `maximumValue: Math.abs(root.minimumValue)` for the bars that are feeding in, but I suppose it will be confusing for different bars to have different ranges.

@chriadam could you squash the first two commits together? LGTM to merge.

chriadam:
Done.

chriadam (approved):

Issue #811, opened 2024-02-28
gui-v2 / MQTT on LAN security


mpvader:
Waits for work in Venus OS: https://github.com/victronenergy/venus-private/issues/330

jpetrell:
Security wizard in gui-v1:
![image](https://github.com/victronenergy/gui-v2/assets/2203667/afde489e-0160-4558-b152-89cc38136f26)

jpetrell:
Pending completion of inclusion of the enablers in the Venus platform.

chriadam:
This has been merged to main.  Closing as fixed.

Issue #560, opened 2023-11-22
Add read-only UI mode 


blammit:
Provide a mode to show the UI in read-only mode.

The UI should make it clear that it is in read-only mode. Disable the entire settings menu, but make Device List accessible in a read-only form, and it shouldn't be possible to change settings in the control cards.

This needs design specifications. Can be done after v1.0.0.

chriadam:
This is wanted (especially for installers) but can come after 1.0.0, so marking as low_prio.

mpvader:
This relates to read-only user levels in VRM; and remains to be something post v1.0.0 release.

I checked VRM today, with a user that doesn't have full control, and its good there: gui-v2 is not shown in the menu. While for a full access user it is shown.

Later, we'll probably want to have it also for view-only users in VRM, and then have UX such that its clear to the user that he can't change anything.

mpvader:
From Wiebe: read-only mode has been added to VRM.

If it is read-only, you will get N/portal_id/settings/0/Settings/Network/VrmPortal {"max":2,"min":0,"value":1} (I assume value 2 would mean read/write, 1 means read-only, and 0 means no-access? not sure though).

In the case where the client is connected to a read-only VRM portal, the UI should show to the user that it is in read-only mode (perhaps in the title bar?), and disable / grey-out all of the settings controls.

mpvader:
also from @wiebeytec :

> Currently it works, as in, I can see the readings and controls, but it silently ignores setting values."

mpvader:
Discussed in the dev meeting today: as a next step, gui-v2 team will make a prototype and see how that looks/works

blammit:
I've fixed https://github.com/victronenergy/gui-v2/issues/1419 which updates the control cards to use the same components as the settings pages, so now it will be more straightforward to disable these components if the user does not have write access. However some UI components do not have correct disabled states, so I'm investigating how those should be updated.

blammit:
Have been discussing with Serj about how this should be shown in the UI. There are two things to consider:

- Some overall indication that the UI is in read-only mode (e.g. with a banner in the status bar)
- How controls should look when disabled. Some controls have disabled states, but others do not.

Also note we need a way to determine whether the UI is read-only. Adding blocked label for this as; I believe gui-v2 will need this information from VRM or some other external source.

wiebeytec:
[...]

PR #1196, opened 2024-06-06
Adapt existing settings to the new security model

jpetrell:
Don't merge, requires custom image as Venus side changes have not been integrated yet.

image

Changes made to gui-v1:

General
- New Security Profile setting: Secured, Weak, Unsecured

VRM
- "Logging enabled" on/off switch dropped
- New "VRM portal" setting: Off, Read-only, Full
- "VRM two-way communication" switch dropped

Remote Console
- "Manually reboot the GX device after changing these settings.\n\nFirst time use? Make sure to either set or disable the password check." warning dropped
- "Disable password check" and "Enable password check" buttons dropped
- "Enable on VRM" switch dropped
- "Enable on LAN" switch renamed to "Enable Remote Console"
- "Remote Console on VRM - status" moved at the end of the list

Services:
- "MQTT on LAN (SSL)" switch renamed to "MQTT Access"
- "MQTT on LAN (Plain-text)" switch dropped

Contributes to #811.


pages/settings/PageSettingsLogger.qml
```diff
@@ -34,13 +34,16 @@ Page {
...
+				id: vrmPortalMode
+				//% "VRM portal"
+				text: qsTrId("settings_vrm_portal")
+				dataItem.uid: Global.systemSettings.serviceUid + "/Settings/Network/VrmPortal"
 				optionModel: [
-					{ display: CommonWords.disabled, value: 0 },
-					{ display: CommonWords.enabled,	value: 1 },
+					{ display: CommonWords.off, value: VenusOS.Vrm_PortalMode_Off },
+					//% "Read-only"
+					{ display: qsTrId("settings_vrm_portal_readonly"), value: VenusOS.Vrm_PortalMode_ReadOnly },
```
chriadam, line 44: We don't currently support this, see #560 but I guess since on VRM the button won't show up to launch Gui-v2 unless the user has full access, it's not a problem.

pages/settings/PageSettingsLogger.qml
```diff
@@ -150,12 +159,6 @@ Page {
 				}
 			}
 
-			ListSwitch {
-				//% "VRM two-way communication"
```
chriadam, line 154: ooi what replaces this?  if this setting is disabled, a user can connect to VRM via their browser, but won't see any data (since the device isn't in contact with VRM to provide it).  I hope that we don't show the "launch gui-v2" button in VRM if this setting (or its new replacement) is false?

jpetrell: VRM Portal mode setting on the same page, which need to be set to "Full".

jpetrell:
Adapted to Serj's input:
- In the list with radio buttons, it is good to show input field before interaction happens
- Confirmation dialog could be adjusted a bit, the bullet list to center align
- Moved the confirmation step to the end

Also, dropped "existing connections will be disconnected" warning. Jeroen commented that in most cases gui-v2 should automatically reconnect after the security profile change. The original warning was mainly added to gui-v1, because VNC disconnects completely. Though might be good idea to give warning when user switches from unsecured http to secured, since then the user needs to switch to https to reconnect.

Still needs more testing, Jeroen will build test image for people to comment on.

pages/settings/PageSettingsLogger.qml
```diff
@@ -76,20 +79,26 @@ Page {
 					{ display: qsTrId("settings_1_day"), value: 86400 },
 				]
 				dataItem.uid: Global.systemSettings.serviceUid + "/Settings/Vrmlogger/LogInterval"
-				allowed: !!loggerMode.dataItem.value && loggerMode.dataItem.value > 0
+				allowed: vrmPortalMode.dataItem.isValid && vrmPortalMode.dataItem.value > 0
```
chriadam, line 82: maybe add a comment `// i.e. not Vrm_PortalMode_Off`

chriadam (approved):
Code LGTM, haven't tested on device as I believe it's still waiting for something in venus?

jpetrell:
> Code LGTM, haven't tested on device as I believe it's still waiting for something in venus?

You can install the custom image Jeroen provided using a USB stick.

chriadam:
Obsoleted by https://github.com/victronenergy/gui-v2/pull/1370, closing.

Issue #865, opened 2024-03-15
Get rid of ListRadioButtonGroupNoYes component


jpetrell:
Using a Switch component is better for on/off settings than a radio button group that pushes a new full-screen page with No and Yes options. Still used in Fronius inverter and inverter charger (VE.Bus BMS section) settings pages.

PR #1199, opened 2024-06-07
Show acsystem instead of multi devices in control cards 

blammit:


data/InverterChargers.qml
```diff
@@ -9,7 +9,7 @@ import Victron.VenusOS
 QtObject {
 	id: root
 
-	// The "first" inverter/charger is from one of com.victronenergy.vebus, com.victronenergy.multi
+	// The "first" inverter/charger is from one of com.victronenergy.vebus, com.victronenergy.acsystem
 	// or com.victronenergy.inverter (in that order of preference). If there is more than service
```
chriadam, line 13: unrelated to this PR, but I assume this was meant to read "If there is more than *one* service..."

pages/controlcards/InverterChargerCard.qml
```diff
@@ -53,10 +51,8 @@ ControlCard {
 					enabled: modelData.currentLimitAdjustable
 					button.text: "%1 %2".arg(value).arg(Units.defaultUnitString(VenusOS.Units_Amp))
 					onClicked: {
-						if (!root._currentLimitDialog) {
-							root._currentLimitDialog = currentLimitDialogComponent.createObject(Global.dialogLayer)
-						}
-						root._currentLimitDialog.openDialog(modelData, model.index)
+						Global.dialogLayer.open(currentLimitDialogComponent,
+								{ inputType: modelData.inputType, inputIndex: model.index, value: modelData.currentLimit })
```
chriadam, line 55: When testing this using "Boat/Motorhome demo 1" on qinetic-cerbogx, I found that after setting the value to 13A either via the control cards or via the settings->devicelist->MultiPlus48 page, the value would "stick" for a little while but a few minutes later would be back at 50 amps.  Not sure if this is something done by the demo, or whether there is some validation step which is failing at the systemcalc level and preventing the current limit of 13A from being set, or what?

chriadam: Ah!  Interestingly - when I try to set it via the device list page, I get the following toast notification, but I do NOT get it when I try to set it via the control cards:
victron-current-limit-setting-issue

blammit: Mm. We need to align the different use cases where current limit is changed.

blammit: It turns out that the vebus Device List page should not have been showing the notification and preventing the current limit from being set, because it was checking `CurrentLimitIsAdjustable` for the active input, rather than the input that was being displayed. So perhaps this is something else in the demo that is resetting the current limit.

I've updated the control cards so that it also shows similar warning notifications if the BMS/DMC are enabled and `CurrentLimitIsAdjustable` is `false`.

jpetrell:
These are not directly related to these changes, but expose issues with horizontal control card list that is easier to see with more cards:
- Some odd horizontal scrolling happens when opening dialogs from some of the control cards (third or second last cards, Ac System 0 and My Inverter 0 AC on the mokc system)
- Flick to the beginning takes lands on the edge of ESS card. Expected: Flick to the beginning takes to originX and the originX is set to include left page margin
- The flicking threshold is quite large, need to drag quite much to scroll to the next card. Expected: Small nudge is enough to flick to the next item

https://github.com/victronenergy/gui-v2/assets/2203667/b61bb03b-b8c3-4443-9662-cc9cfe403d6c

pages/vebusdevice/PageVeBusBms.qml
```diff
@@ -14,24 +14,24 @@ Page {
...
 
-			ListRadioButtonGroupNoYes {
+			ListTextItem {
 				//% "BMS Error"
 				text: qsTrId("vebus_device_bms_error")
+				secondaryText: CommonWords.yesOrNo(dataItem.value)
 				dataItem.uid: bindPrefix + "/Bms/Error"
-				enabled: false
-				allowed: bmsType === VenusOS.VeBusDevice_Bms_Type_VeBus
+				allowed: dataItem.value === VenusOS.VeBusDevice_Bms_Type_VeBus
```
jpetrell, line 34: Only one ListRadioButtonGroupNoYes left in pages/settings/PageSettingsFroniusInverter.qml and we can mark https://github.com/victronenergy/gui-v2/issues/865 fixed. :partying_face:

data/InverterChargers.qml
blammit: Fixed

blammit:
> Some odd horizontal scrolling happens when opening dialogs from some of the control cards (third or second last cards, Ac System 0 and My Inverter 0 AC on the mokc system)
> Flick to the beginning takes lands on the edge of ESS card. Expected: Flick to the beginning takes to originX and the originX is set to include left page margin
> The flicking threshold is quite large, need to drag quite much to scroll to the next card. Expected: Small nudge is enough to flick to the next item

I've removed the SnapOneItem and fixed the left margin. The flicking threshold seems okay now but we should test and adjust if needed. I noticed the snapping was also causing some twitchy behaviour so would be good to see how it feels without the snapping.

chriadam (approved):
LGTM.

PR #1200, opened 2024-06-07
Add "logout" navigation bar entry for local-LAN wasm

chriadam:


pages/LogoutPage.qml
```diff
@@ -0,0 +1,20 @@
...
+SwipeViewPage {
+	id: root
+
+	//% "Log Out"
+	navButtonText: qsTrId("nav_logout")
+	navButtonIcon: "qrc:/images/logout.svg"
+	url: "qrc:/qt/qml/Victron/VenusOS/pages/LogoutPage.qml"
+	topLeftButton: VenusOS.StatusBar_LeftButton_ControlsInactive
+
+	onIsCurrentPageChanged: if (isCurrentPage) BackendConnection.logout()
```
chriadam, line 19: maybe we want to show a modal dialog asking whether they really want to logout?

pages/LogoutPage.qml
```diff
@@ -0,0 +1,20 @@
...
+import QtQuick
+import Victron.VenusOS
+import QtQuick.Controls.impl as CP
+
+SwipeViewPage {
+	id: root
+
+	//% "Log Out"
+	navButtonText: qsTrId("nav_logout")
+	navButtonIcon: "qrc:/images/logout.svg"
```
chriadam, line 15: doesn't exist yet

src/backendconnection.cpp
```diff
@@ -450,6 +454,26 @@ void BackendConnection::requestShardFromVrmApi()
 		});
 }
 
+void BackendConnection::logout()
+{
+#if defined(VENUS_WEBASSEMBLY_BUILD)
+	emscripten_run_script("location.href = \'logout.html\'");
```
chriadam, line 460: I couldn't find the actual URL in the "2024 Securing Venus" document, and I forgot to write it down during the meeting... Anyone know what it's supposed to be?

components/SwipePageModel.qml
```diff
@@ -16,6 +17,11 @@ ObjectModel {
 			view: root.view
 		}
 	}
+	readonly property Component logoutPage: Component {
```
chriadam, line 20: this adds the logout button as a Navigation Bar entry - but maybe that's not a good idea?  (too cramped, especially on 5-inch layout?)  Where should it go instead?  Some settings menu?

jpetrell:
Pushed commit to move the logout button to the General settings. In most cases there is no need to log out so it doesn't need to be prominently available. Also in many gui-v2 variants it is not shown: on GX device, when accessed from VRM or when security profile is Unsecured (as then no password).

image

chriadam:
Yep, I agree.  LGTM.

pages/LogoutPage.qml
jpetrell: Sorry missed this one. It was mentioned in the issue description: relative path auth/logout.php. Though I don't have the security image now installed here with me to test.

chriadam:
Closing this, obsoleted by #1196

PR #1203, opened 2024-06-09
Firmware version string fixes

blammit:


jpetrell (commented):
gui-v2 shows following version strings on ESS demo

```
BMV          'v3.08'
Bluesolar    'v1.29'
JLB..        ''
Multiplus..  'v422'
PHello       ''
PV Inverter  'v0.02'
```

In addition Boat/Motorhome demo 1 shows e.g.

```
Oil tank (4) 'v3.0.2V'
```

BMV, Bluetsolar and Oil tank (4) match gui-v1, but gui-v1
- Shows "--" instead of empty for JLB.. and PHello 
- Multiplus 48/3000/35-50 shows "422" instead of "v422"
- PV inverter on input 1 shows Firmware version "2" instead of v0.02

mr-manuel:
Also `v1.3.20240610dev` shows correctly

blammit:
> BMV, Bluetsolar and Oil tank (4) match gui-v1, but gui-v1
> Shows "--" instead of empty for JLB.. and PHello
> Multiplus 48/3000/35-50 shows "422" instead of "v422"
> PV inverter on input 1 shows Firmware version "2" instead of v0.02

The firmware version conversion we used is based on the one for VictronConnect, which seems to be slightly different from the one used to set the `Text` value provided for gui-v2.

I've updated it to return "--" when the firmware version is not present, but from what I can tell, the other differences are intentional. The Multiplus is a vebus device so the version should have a "v" prefix, i.e. "v422" instead of "422". The PV inverter is a generic-type device (connected via USB) so the number should be converted into hex and made into a 3-part number like 0.0.2.

Issue #1093, opened 2024-05-19
"MultiRS" and "charger" settings pages are not yet implemented


DanielMcInnes:
I'm not sure why these are still hooked up to PageNotYetImplemented.qml in DeviceListPage

blammit:
There was a discussion in https://github.com/victronenergy/gui-v2/issues/328 that these were not required for implementation yet. But maybe that has changed? Need to confirm with @mpvader

mpvader:
Hey, yes its changed - or at least about to. Izak is finishing up his implementation of it in gui-v2.

blammit:
@mpvader did you mean gui-v2? I checked with @izak and he said he has implemented updates in gui-v1. I can look at porting those to gui-v2.

mpvader:
Yes, good catch! Amended. Pls do

blammit:
Changes merged.

chriadam:
https://github.com/victronenergy/gui-v2/pull/1292/files fixes this issue according my testing

chriadam:
Fixed.

PR #1204, opened 2024-06-09
Implement settings pages for 'charger' services

blammit:
Show PageAcCharger.qml instead of PageNotYetImplemented.qml when an AC charger is clicked in the Device List.

PageNotYetImplemented.qml can be removed as it is no longer used.

Contributes to #1093


pages/settings/devicelist/PageAcCharger.qml
```diff
@@ -0,0 +1,120 @@
...
+			}
+
+			ListSpinBox {
+				text: Global.acInputs.currentLimitTypeToText()
+				writeAccessLevel: VenusOS.User_AccessType_User
+				allowed: defaultAllowed && dataItem.isValid
+				dataItem.uid: root.bindPrefix + "/Ac/In/CurrentLimit"
+				suffix: Units.defaultUnitString(VenusOS.Units_Amp)
+				stepSize: 0.1
+				decimals: 1
```
jpetrell, line 37: Bit unfortunate to duplicate these default precision values, ideally spin box could fetch the unit string and decimals from the given unit type.

pages/settings/devicelist/PageAcCharger.qml
```diff
@@ -0,0 +1,120 @@
...
+
+				VeQuickItem {
+					id: nrOfOutputs
+					uid: root.bindPrefix + "/NrOfOutputs"
+				}
+
+				Repeater {
+					model: nrOfOutputs.value || 1
+					delegate: ListQuantityGroup {
+						text: "%1 %2".arg(CommonWords.battery).arg(model.index + 1)
```
jpetrell, line 51: Wonder if translation of "Battery" can change based on battery number. Maybe "Battery %1" would be safer.

pages/settings/devicelist/PageAcCharger.qml
```diff
@@ -0,0 +1,120 @@
...
+			}
+
+			ListTextItem {
+				text: CommonWords.state
+				secondaryText: Global.system.systemStateToText(dataItem.value)
+				dataItem.uid: root.bindPrefix + "/State"
+			}
+
+			ListSpinBox {
+				text: Global.acInputs.currentLimitTypeToText()
```
jpetrell, line 31: I would have "Input current limit" or CommonWords.inputCurrentLimit here for clarity.

jpetrell (approved):
The page organisation could be improved, but it aligns with gui-v1 so LGTM. Added some small comments.

pages/settings/devicelist/PageAcCharger.qml
blammit: Yes maybe ListSpinBox should have a 'unit' argument instead of 'suffix'?

pages/settings/devicelist/PageAcCharger.qml
blammit: Done

pages/settings/devicelist/PageAcCharger.qml
blammit: Done

PR #1205, opened 2024-06-09
Show DC flow for Multi RS and Inverter RS

### LINKED ISSUE #1083: show DC flow for Multi RS and Inverter RS
gui-v1 commit: 9e16d73b19154c8638af940beea5ca63965e345b

    On the overviews, correctly animate the flow of energy even
    when there is no VE.Bus Multi in the system, by using new summary
    paths on systemcalc to determine what the system is doing.

blammit:
Animate the flow of energy between the Inverter/Charger and Battery even when there is no VE.Bus inverter.

Use /Dc/InverterCharger/Power rather than the vebus /Dc/0/Power value to animate the flow. The former will be valid even when there is system is using a Multi RS rather than VE.Bus.

Also remove the unused vebus power from HubData.qml, and fix the changes in the mock battery for the power/amp/charging values.

This ports 9e16d73b19154c8638af940beea5ca63965e345b from gui-v1.

Fixes #1083

### DIFF
--- a/data/mock/InverterChargersImpl.qml
+++ b/data/mock/InverterChargersImpl.qml
@@ -488,6 +488,24 @@ QtObject {
 		}
 	}
 
+	//--- totals
+
+	readonly property VeQuickItem inverterChargerPower: VeQuickItem {
+		uid: Global.system.serviceUid + "/Dc/InverterCharger/Power"
+	}
+
+	readonly property int batteryMode: Global.batteries.system.mode
+	onBatteryModeChanged: {
+		// Positive power when battery is charging, negative power when battery is discharging
+		const randomPower = Math.round(100 + (Math.random() * 600))
+		const batteryMode = Global.batteries.system.mode
+		const power = batteryMode === VenusOS.Battery_Mode_Charging ? randomPower
+				: batteryMode === VenusOS.Battery_Mode_Discharging ? randomPower * -1
+				: 0 // Battery_Mode_Idle
+		root.inverterChargerPower.setValue(power)
+	}
+
+
 	Component.onCompleted: {
 		populateInverterChargers()
 		populateInverters()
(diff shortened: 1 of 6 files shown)


data/mock/InverterChargersImpl.qml
```diff
@@ -488,6 +488,24 @@ QtObject {
...
 
+	//--- totals
+
+	readonly property VeQuickItem inverterChargerPower: VeQuickItem {
+		uid: Global.system.serviceUid + "/Dc/InverterCharger/Power"
+	}
+
+	readonly property int batteryMode: Global.batteries.system.mode
+	onBatteryModeChanged: {
+		// Postive power when battery is charging, negative power when battery is discharging
```
jpetrell, line 499: Typo

pages/OverviewPage.qml
```diff
@@ -692,12 +692,11 @@ SwipeViewPage {
...
-		// If vebus power is positive: battery is charging, so energy flows to battery.
-		// If vebus power is negative: battery is discharging, so energy flows to inverter/charger.
+		// If inverter/charger power is positive: battery is charging, so energy flows to battery.
+		// If inverter/charger power is negative: battery is discharging, so energy flows to inverter/charger.
 		animationMode: root.isCurrentPage
-				&& !isNaN(Global.system.veBus.power)
-				&& Math.abs(Global.system.veBus.power) > Theme.geometry_overviewPage_connector_animationPowerThreshold
-						? (Global.system.veBus.power > 0
+				&& Math.abs(Global.inverterChargers.power) > Theme.geometry_overviewPage_connector_animationPowerThreshold
+						? (Global.inverterChargers.power > 0
```
jpetrell, line 699: Odd to expose power value that is just used to determine animation direction. But why not. :smile:

jpetrell (approved):

data/mock/InverterChargersImpl.qml
blammit: Fixed

pages/OverviewPage.qml
blammit: Oh true, maybe the value should only be fetched from the Overview page? I've updated it to do this.

PR #1207, opened 2024-06-10
MQTT: ensure the portal id parameter is set before connecting

### LINKED ISSUE #1185: Ensure that gui-v2 VRM connections don't subscribe to wildcard topics
VRM-launched (or connected) gui-v2 instances should know their portal id, and should never subscribe to wildcard topics.

Double check that this is the case.

chriadam:
Previously, it was set after the connection was initiated, but
before the connection was established (i.e. at the same time the
username and password were provided).

Now, it is provided at the very start of the flow, so that the
producer can emit an error if no portal id is provided for
VRM connections.

Also, align the detection of VRM addresses between gui-v2 and
veutil code.

Fixes #1185

### DIFF
--- a/src/backendconnection.cpp
+++ b/src/backendconnection.cpp
@@ -177,7 +177,6 @@ void BackendConnection::initMqttConnection(const QString &address)
 				// TODO: fetch updated credentials via VRM API if required...
 				producer->setCredentials(m_username, m_password);
 			}
-			producer->setPortalId(m_portalId);
 			producer->continueConnect();
 		}
 	});
@@ -187,6 +186,9 @@ void BackendConnection::initMqttConnection(const QString &address)
 	connect(mqttProducer, &VeQItemMqttProducer::errorChanged,
 		this, &BackendConnection::mqttErrorChanged);
 
+	if (!m_portalId.isEmpty()) {
+		mqttProducer->setPortalId(m_portalId);
+	}
 #if defined(VENUS_WEBASSEMBLY_BUILD)
 	mqttProducer->open(QUrl(address), QMqttClient::MQTT_3_1);
 #else


blammit (approved):
LGTM. Should it refer to task https://github.com/victronenergy/gui-v2/issues/1185?

src/backendconnection.cpp
```diff
@@ -187,6 +186,9 @@ void BackendConnection::initMqttConnection(const QString &address)
 	connect(mqttProducer, &VeQItemMqttProducer::errorChanged,
 		this, &BackendConnection::mqttErrorChanged);
 
+	if (!m_portalId.isEmpty()) {
+		mqttProducer->setPortalId(m_portalId);
```
DanielMcInnes, line 190: Will this ever execute? When `BackendConnection::initMqttConnection(...)` is called, m_portalId will always be empty, no? I think `VeQItemMqttProducer::setPortalId(...)` will only ever get called by `VeQItemMqttProducer::onMessageReceived(...)`

src/backendconnection.cpp
```diff
@@ -187,6 +186,9 @@ void BackendConnection::initMqttConnection(const QString &address)
 	connect(mqttProducer, &VeQItemMqttProducer::errorChanged,
 		this, &BackendConnection::mqttErrorChanged);
 
+	if (!m_portalId.isEmpty()) {
```
DanielMcInnes, line 189: Is this wasm only? Can it be moved into the next #ifdef?

DanielMcInnes (approved):
lgtm, 1 comment

src/backendconnection.cpp
chriadam: not wasm specific

chriadam:
> LGTM. Should it refer to task #1185?

Yes, thanks, done.

jpetrell (approved):
Tested custom wasm on the VRM website using instructions [here](https://github.com/victronenergy/gui-v2/wiki/How-to-build-venus-gui-v2#patch-and-pack-the-webassembly-files) and [here](https://github.com/victronenergy/venus-private/wiki/WASM-and-VRM#how-to-make-vrm-showserve-the-user-a-specific-gui-v2-wasm-file).

Added logging to backendconnection.cpp, the execution order seems correct: Disconnected -> Reconnecting -> AboutToConnect -> setPortalId -> Connecting to MQTT source -> Connecting -> AboutToConnect ->  Connected -> Initializing -> Ready.

Initially got "Transport connection error", but after the refresh the gui-v2 was loaded. Doesn't seem related to these changes so LGTM.

PR #1212, opened 2024-06-10
Always calculate average level of multiple tanks if possible

### LINKED ISSUE #1208: Brief page tank gauge value is calculated wrongly
When there are multiple tanks, the Brief Page seems to calculate an "average of levels" rather than an "aggregate level".

Value used is (96+45)/2 = 71%
Calculation should be: (1926+365)/2800 = 82%

chriadam:
If there are two tanks of the same type but vastly different capacities, the combined average level is not the average of their levels, but instead should be calculated based upon their capacities.

Fixes #1208

### DIFF
--- a/data/Tanks.qml
+++ b/data/Tanks.qml
@@ -151,19 +151,25 @@ QtObject {
 		let totalLevel = NaN
 		let totalRemaining = NaN
 		let totalCapacity = NaN
+		let requireFallback = false
 		for (let i = 0; i < model.count; ++i) {
 			const tank = model.deviceAt(i)
 			totalRemaining = Units.sumRealNumbers(totalRemaining, tank.remaining)
 			totalCapacity = Units.sumRealNumbers(totalCapacity, tank.capacity)
 			totalLevel = Units.sumRealNumbers(totalLevel, tank.level)
+			requireFallback = requireFallback || isNaN(tank.remaining) || isNaN(tank.capacity)
 		}
-		model.averageLevel = isNaN(totalLevel) || model.count === 0 ? NaN : totalLevel / model.count
+
 		model.totalRemaining = totalRemaining
 		model.totalCapacity = totalCapacity
-
-		// If /Level is not available for the tanks, calculate it from /Remaining and /Capacity.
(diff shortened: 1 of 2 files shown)


data/Tanks.qml
```diff
@@ -151,19 +151,25 @@ QtObject {
 		let totalLevel = NaN
 		let totalRemaining = NaN
 		let totalCapacity = NaN
+		let requireFallback = false
 		for (let i = 0; i < model.count; ++i) {
 			const tank = model.deviceAt(i)
 			totalRemaining = Units.sumRealNumbers(totalRemaining, tank.remaining)
 			totalCapacity = Units.sumRealNumbers(totalCapacity, tank.capacity)
 			totalLevel = Units.sumRealNumbers(totalLevel, tank.level)
```
jpetrell, line 159: Is totalLevel unused now?

jpetrell (approved):

data/Tanks.qml
```diff
@@ -151,19 +151,25 @@ QtObject {
...
-
-		// If /Level is not available for the tanks, calculate it from /Remaining and /Capacity.
-		if (isNaN(model.averageLevel) && !isNaN(totalRemaining) && !isNaN(totalCapacity) && totalCapacity > 0) {
+		if (!requireFallback && !isNaN(totalRemaining) && !isNaN(totalCapacity) && totalCapacity > 0) {
+			// if we know all tank capacities and usages,
+			// we can calculate the combined level.
 			model.averageLevel = totalRemaining / totalCapacity * 100
+		} else {
+			// only fall back to a crude average level
+			// if we don't have all of the required information.
```
blammit, line 171: Can you explain what the required information is here?

blammit (commented):
Could you add a mock configuration that shows this as well?

DanielMcInnes (approved):

data/Tanks.qml
chriadam: It's still used, in the fall-back branch

data/Tanks.qml
chriadam: done

chriadam:
> Could you add a mock configuration that shows this as well?

Done

PR #1219, opened 2024-06-11
add support for Quattro-II 2x120V

### LINKED ISSUE #1102: add support for Quattro-II 2x120V
see gui-v1 commit d6aafcd4a4efe679b8ad2f05b777897098194921
Author: Jos Pasop <jpasop@victronenergy.com>
Date:   Wed May 10 10:48:12 2023 +0200

    add support for Quattro-II 2x120

https://github.com/victronenergy/venus-private/issues/243

DanielMcInnes:
Fixes #1102

### DIFF
--- a/data/SystemAc.qml
+++ b/data/SystemAc.qml
@@ -14,11 +14,26 @@ QtObject {
 		readonly property real current: phases.count === 1 ? _firstPhaseCurrent : NaN // multi-phase systems don't have a total current
 		readonly property real maximumCurrent: _maximumCurrent.value === undefined ? NaN : _maximumCurrent.value
 		property real _firstPhaseCurrent: NaN
+		readonly property bool l2AndL1OutSummed: !!_l2L1OutSummed.value
+		onL2AndL1OutSummedChanged: setPhaseCount(phases.count)
+		readonly property VeQuickItem _l2L1OutSummed: VeQuickItem {
+			uid: Global.system.veBus.serviceUid ? Global.system.veBus.serviceUid + "/Ac/State/SplitPhaseL2L1OutSummed" : ""
+		}
 
 		property ListModel phases: ListModel {}
 
 		function setPhaseCount(phaseCount) {
 			reset()
+
+			if (l2AndL1OutSummed) {
+				phases.append({
+					name: "L1 + L2",
+					power: NaN,
+					current: NaN
+				})
+				return
+			}
+
 			for (let i = 0; i < phaseCount; ++i) {
 				phases.append({
 					name: "L" + (i + 1),
@@ -29,6 +44,10 @@ QtObject {
 		}
 
 		function setPhaseData(index, data) {
+			if (l2AndL1OutSummed && index > 0) {  // we only display a single combined value for L1 & L2 when L1 and L2 output are summed.
+				return
+			}
+
 			phases.set(index, data)
 
 			// Update totals for the model.
--- a/data/mock/config/MockDataSimulator.qml
+++ b/data/mock/config/MockDataSimulator.qml
@@ -219,6 +219,13 @@ QtObject {
 			event.accepted = true
 			break
 		}
+		case Qt.Key_S:
+		{
+			Global.system.ac.consumption._l2L1OutSummed.setValue(!Global.system.ac.consumption._l2L1OutSummed.value)
+			event.accepted = true
+			break
+		}
+
 		case Qt.Key_T:
 			root.timersActive = !root.timersActive
 			pageConfigTitle.text = "Timers on: " + root.timersActive
--- a/components/widgets/AcLoadsWidget.qml
+++ b/components/widgets/AcLoadsWidget.qml
@@ -23,4 +23,5 @@ AcWidget {
 		phaseModelProperty: "current"
 		maximumValue: Global.system.ac.consumption.maximumCurrent
 	}
+	extraContentLoader.active: root.phaseCount > 1 || Global.system.ac.consumption.l2AndL1OutSummed
 }
--- a/data/mock/InverterChargersImpl.qml
+++ b/data/mock/InverterChargersImpl.qml
@@ -317,6 +317,10 @@ QtObject {
 				}
 			}
 
+			readonly property VeQuickItem _l2L1OutSummed: VeQuickItem {
+				uid: veBusDevice.serviceUid + "/Ac/State/SplitPhaseL2L1OutSummed"
+			}
+
 			readonly property VeQuickItem _ignoreAcIn1: VeQuickItem {
 				uid: veBusDevice.serviceUid + "/Ac/State/IgnoreAcIn1"
 			}
@@ -401,6 +405,7 @@ QtObject {
 				_chargeState.setValue(VenusOS.VeBusDevice_ChargeState_InitializingCharger)
 				_setChargeState.setValue(VenusOS.VeBusDevice_ChargeState_InitializingCharger)
 				_redetectSystem.setValue(0)
+				_l2L1OutSummed.setValue(1)
 				_ignoreAcIn1.setValue(0)
 				_ignoreAcIn2.setValue(0)
 				_waitingForRelayTest.setValue(0)


data/SystemAc.qml
```diff
@@ -17,8 +17,18 @@ QtObject {
...
+
+			if (l2AndL1OutSummed) {
+				phases.append({
+					name: "L1 + L2",
+					power: NaN,
+					current: NaN
+				})
+				return
+			}
+
```
blammit, line 31: This would produce text like "L1 + L2: --" in the AcLoadsWidget, but should it just be showing "L1 + L2" instead, to reflect how it looks in gui-v1? 

Looking at https://github.com/victronenergy/venus/issues/820, if we reflect the gui-v1 AC loads display, then the AcLoadsWidget needs to show:

- a Label with "L1 + L2" if `/SplitPhaseL2L1OutSummed == 1`
- otherwise, show the ThreePhaseDisplay

Additionally, if the individual phase values are not valid in this split phase mode, then in BriefPage, the AC load gauges should be hidden, but the overall power/amps should still be shown. (This can probably be done by setting the relevant `phaseModel` to `undefined`.)

Also could you add a mock simulator toggle to toggle the value of `/SplitPhaseL2L1OutSummed`?

data/SystemAc.qml
```diff
@@ -17,8 +17,18 @@ QtObject {
 
 		property ListModel phases: ListModel {}
 
-		function setPhaseCount(phaseCount) {
+		function setPhaseCount(phaseCount, l2AndL1OutSummed = false) {
```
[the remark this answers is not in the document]
jpetrell: Oh the path is SplitPhaseL2L1OutSummed so I guess it needs to be like this.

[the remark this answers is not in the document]
DanielMcInnes: but why did they name the path 'SplitPhaseL2L1OutSummed'? It makes no sense, and upsets me

data/SystemAc.qml
DanielMcInnes: Screenshot from 2024-06-12 12-46-46
Screenshot from 2024-06-12 12-46-38

gui-v1 and gui-v2 images attached.

Added a mock simulator toggle 'S' for SplitPhaseL2L1OutSummed

blammit: Should it still be able to show "L3" when L1 and L2 are combined? https://github.com/victronenergy/venus/issues/820 doesn't show that as a use case, but gui-v1 `OverviewAcValues.qml` technically supports the ability to show "L1 + L2" as well as "L3" at the same time.

Looking at this again, I'm not so sure on how the UI should look on the Brief and Overview pages. Maybe we need to double check this with Jesús.

data/mock/config/MockDataSimulator.qml
```diff
@@ -219,6 +219,14 @@ QtObject {
 			event.accepted = true
 			break
 		}
+		case Qt.Key_S:
+		{
+			Global.system.ac.consumption._l2L1OutSummed.setValue(!Global.system.ac.consumption._l2L1OutSummed.value)
+			Global.system.ac.consumption.setPhaseCount(1)
```
blammit, line 225: This shouldn't need to `setPhaseCount()`. In a dbus/mqtt system only the value of `/SplitPhaseL2L1OutSummed` would change.

components/widgets/AcWidget.qml
```diff
@@ -18,6 +18,8 @@ OverviewWidget {
 	extraContentChildren: Loader {
 		id: extraContentLoader
 
+		readonly property bool defaultActive: root.phaseCount > 1
```
blammit, line 21: Would be nicer to put this property in the root object so that `AcLoadsWidget` doesn't need to reach into a sub-component.

data/mock/config/MockDataSimulator.qml
DanielMcInnes: Done

components/widgets/AcWidget.qml
DanielMcInnes: If I move 'defaultActive' to the root object, then I think I would have to rename it to eg. 'extraContentLoaderDefaultActive'. To my eyes, 'extraContentLoader.defaultActive' is more readable than 'extraContentLoaderDefaultActive'. What do you think?

blammit: It's definitely less readable. How about leaving out `defaultActive` and just setting `extraContentLoader.active: phasesCount > 1 || Global.system.ac.consumption.l2AndL1OutSummed` from AcLoadsWidget?

data/SystemAc.qml
```diff
@@ -14,11 +14,25 @@ QtObject {
 		readonly property real current: phases.count === 1 ? _firstPhaseCurrent : NaN // multi-phase systems don't have a total current
 		readonly property real maximumCurrent: _maximumCurrent.value === undefined ? NaN : _maximumCurrent.value
 		property real _firstPhaseCurrent: NaN
+		readonly property bool l2AndL1OutSummed: _l2L1OutSummed.value || false
```
jpetrell, line 17: is `|| false` necessary?

DanielMcInnes: yes, otherwise you get 'Cannot assign "undefined" to bool' error messages

data/SystemAc.qml
DanielMcInnes: Jesus says that 'split phase' will never have 3 phases at output.

components/widgets/AcWidget.qml
DanielMcInnes: updated as suggested

data/SystemAc.qml
blammit: How about `_l2L1OutSummed.value !== 1` or `!!_l2L1OutSummed.value` instead?

data/mock/config/MockDataSimulator.qml
blammit: Pressing 'S' doesn't change the phases in Brief/Overview now -- does it work for you?

data/SystemAc.qml
blammit: Phew, that's good :)

data/mock/config/MockDataSimulator.qml
DanielMcInnes: The phase count needs to change while 'SplitPhaseL2L1OutSummed' is true for the phases model to update with the "L1 + L2" text.  This works fine on a real installation, and most of the time it works in mock mode, but sometimes you need to change configs with the arrow keys to cause a change in phase count, which updates the phases model. I thought this was good enough for mock mode, but I can change it if you would like.

data/SystemAc.qml
DanielMcInnes: updated to `!!_l2L1OutSummed.value`.

data/mock/config/MockDataSimulator.qml
blammit: Okay, so if I go to a single-phase system and switch back to 3-phase system, then the phases update. Okay, I didn't realise you had to do that to make it take effect in mock mode. It'd be nice to avoid that, but if it works fine on a real system then LGTM.

data/SystemAc.qml
```diff
@@ -14,11 +14,25 @@ QtObject {
 		readonly property real current: phases.count === 1 ? _firstPhaseCurrent : NaN // multi-phase systems don't have a total current
 		readonly property real maximumCurrent: _maximumCurrent.value === undefined ? NaN : _maximumCurrent.value
 		property real _firstPhaseCurrent: NaN
+		readonly property bool l2AndL1OutSummed: !!_l2L1OutSummed.value
+		readonly property VeQuickItem _l2L1OutSummed: VeQuickItem {
+			uid: Global.system.veBus.serviceUid ? Global.system.veBus.serviceUid + "/Ac/State/SplitPhaseL2L1OutSummed" : ""
```
blammit, line 19: Should this also call `setPhaseCount()` when the value is set? Otherwise if the value changes, or is initialized after `SystemData.qml` calls `setPhaseCount()`, then `setPhaseCount()` will not get the correct value of `l2AndL1OutSummed`.

DanielMcInnes: yes, I think you are right. Gui-v1 would handle this correctly, and this code would not. Updated.

blammit (approved):

PR #1226, opened 2024-06-13
AcInputs load graph on side panel should be zero-centered

### LINKED ISSUE #1073: Brief page: Side panel grid graph fill
The grid graph should be filled to zero and not to the bottom.

Reference to https://github.com/victronenergy/gui-v2/pull/1065#issuecomment-2117300315

## Current
331313346-3e162f39-dc4d-4d6a-a0b4-99ce8ba27d31

## Expectation
331313847-1aace12c-ad3f-4a8d-a602-26d34f491da1

chriadam:
Grid draw should be shown in blue above the midpoint (zero amps) dotted line, and grid feed-in should be shown in green below the midpoint dotted line.

Fixes #1073

### DIFF
--- a/components/LoadGraph.qml
+++ b/components/LoadGraph.qml
@@ -19,6 +19,7 @@ Item {
 	property color belowThresholdFillColor: Theme.color_blue
 	property color horizontalGradientColor1: Theme.color_briefPage_background
 	property color horizontalGradientColor2: "transparent"
+	property bool zeroCentered
 	property alias active: graphAnimation.running
 
 	signal nextValueRequested()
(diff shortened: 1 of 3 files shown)


mr-manuel:
Looks now much better!

jpetrell:
Code LGTM. For feed-in case I could only see tiny amplitudes out of the graphs both on the GX device and on the mock mode, are you sure the limits are correct?

image

blammit (approved):
Nice, looks great!

chriadam:
> Ah had 10A and -40A limits, maybe the asymmetry there was the reason. In such case would display the limit line on 1/5 of the graph height and not on the exact vertical center to utilize the space efficiently.

I do like the idea of utilising the range limits to determine the zero-line offset, but I asked Matthijs about this in the meeting and he said that for now we should just use the vertical center, that every install will be different and have different expectations, and using the center as the zero line will be clearest.  I might have misunderstood what he meant, of course!

DanielMcInnes:
> I do like the idea of utilising the range limits to determine the zero-line offset, but I asked Matthijs about this in the meeting and he said that for now we should just use the vertical center, that every install will be different and have different expectations, and using the center as the zero line will be clearest. I might have misunderstood what he meant, of course!

That's how I remember it.

jpetrell (approved):

PR #1227, opened 2024-06-13
Genset UI updates

blammit:


blammit:
Forgot to update mock mode to test this. Will push an update to do this, but it can also be tested with venus-docker "n".

components/settings/ListGeneratorError.qml
```diff
@@ -0,0 +1,25 @@
+/*
+** Copyright (C) 2023 Victron Energy B.V.
```
jpetrell, line 2: 2024

pages/settings/PageGeneratorManualStart.qml
```diff
@@ -0,0 +1,73 @@
...
+					uid: root.startStopBindPrefix + "/ManualStartTimer"
+				}
+			}
+
+			ListTimeSelector {
+				//% "Run for (hh:mm)"
+				text: qsTrId("settings_page_relay_generator_run_for_hh_mm")
+				enabled: !manualSwitch.checked && remoteStartMode.value === 1
+				dataItem.uid: root.startStopBindPrefix + "/ManualStartTimer"
+				writeAccessLevel: VenusOS.User_AccessType_User
```
jpetrell, line 58: I can set running time for the generator in the Control cards, but for some reason not here. The running time is displayed here, but the list item is always disabled.

jpetrell: I would avoid duplicate implementation and rather open the nice manual start/stop dialog that is already used in the Control cards page when the user taps Manual start list item.

blammit: That's a great idea. Updated to do this; it ended up needing quite a bit of reorganization, but it's a good thing since it showed that the generator control card didn't check for some conditions like the user access level and the /AutoStart flag (to indicate whether start/stop is possible).

pages/settings/devicelist/ac-in/PageAcInModelGenset.qml
```diff
@@ -34,56 +36,46 @@ ObjectModel {
...
-				}
-				autoStartStopItem.setValue(0)
-				modeItem.setValue(1)
-			} else if (index === index_autoStartStop) {
-				autoStartStopItem.setValue(1)
-				modeItem.setValue(0)
+	ListSwitch {
+		//% "Auto start functionality"
+		text: qsTrId("ac-in-genset_auto_start_functionality")
+		dataItem.uid: root.startStopBindPrefix ? root.startStopBindPrefix + "/AutoStartEnabled" : ""
```
jpetrell, line 42: Auto start setting differs from the control cards equivalent in two aspects
- Doesn't show dialog when toggled off
- Isn't disabled when the generator is running

components/CommonWords.qml
```diff
@@ -240,6 +240,15 @@ QtObject {
 	//% "Manual"
 	readonly property string manual: qsTrId("common_words_manual")
 
+	//% "Manual control"
+	readonly property string manual_control: qsTrId("common_words_manual_control")
+
+	//% "Manual start"
+	readonly property string manual_start: qsTrId("common_words_manual_start")
```
jpetrell, line 247: Wonder if it should be "Manual Start" or "Manual start". Existing mode seems to favor capitalization.

components/GeneratorManualControlButton.qml
```diff
@@ -0,0 +1,105 @@
...
+			 ? _state.value === VenusOS.Generators_State_Running
+			   || _state.value === VenusOS.Generators_State_WarmUp
+			 : _generatorStateBeforeDialogOpen === VenusOS.Generators_State_Running
+
+	text: checked
+			//% "Manual Stop"
+		  ? qsTrId("controlcard_generator_subcard_button_manual_stop")
+			/* stopped */
+			//% "Manual Start"
+		  : qsTrId("controlcard_generator_subcard_button_manual_start")
```
jpetrell, line 37: These don't use CommonWords?

jpetrell (approved):
Somethings could be improved, but didn't see regression so approving and merging this to the next release so we get wider feedback.

pages/settings/devicelist/ac-in/PageAcInModelGenset.qml
blammit: Thanks, filed https://github.com/victronenergy/gui-v2/issues/1239 for this.

components/CommonWords.qml
blammit: Mentioned in https://github.com/victronenergy/gui-v2/issues/1239 as well.

PR #1240, opened 2024-06-16
Change tank/temp error "disconnected" to "open circuit"

### LINKED ISSUE #1238: Clarify analogue sensor error status display
See https://github.com/victronenergy/venus-private/issues/378

Changing this can be done on several ways, but I'd say the best way is to adjust the enum [here](https://github.com/victronenergy/gui-v2/blob/a91cb9cd33f8a9dcf6d465ebc19d1a0a140642a5/src/enums.h#L376) and the call [here](https://github.com/victronenergy/gui-v2/blob/a91cb9cd33f8a9dcf6d465ebc19d1a0a140642a5/data/Tanks.qml#L128) and add a common word for "Open circuit". 

But not sure if that is the preferred way for you too.

blammit:
Port of 403b535e460d54c77474ff4766c703e2d3ade4ba from gui-v1.

When tank sensor is absent, show "Open circuit" instead of "Disconnected". The latter former means the sensor is absent, while the latter means that the dbus service for a device has vanished.

Fixes #1238

### DIFF
--- a/components/CommonWords.qml
+++ b/components/CommonWords.qml
@@ -332,6 +332,9 @@ QtObject {
 	//% "Open"
 	readonly property string open_status: qsTrId("common_words_open_status");
 
+	//% "Open circuit"
+	readonly property string open_circuit: qsTrId("common_words_open_circuit");
+
 	//% "Overall history"
 	readonly property string overall_history: qsTrId("common_words_overall_history")
 
--- a/data/Tanks.qml
+++ b/data/Tanks.qml
@@ -125,8 +125,8 @@ QtObject {
 		switch (status) {
 		case VenusOS.Tank_Status_Ok:
 			return CommonWords.ok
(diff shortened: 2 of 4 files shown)


DanielMcInnes:
Is this still WIP? The gui-v1 changes showed 'open circuit' for temperature sensors as well.

jpetrell:
Yeah disconnected is also used in gui-v3/pages/settings/devicelist/temperature/PageTemperatureSensor.qml

DanielMcInnes:
> Yeah disconnected is also used in gui-v3/pages/settings/devicelist/temperature/PageTemperatureSensor.qml

ugh, I'm blind. Lgtm

DanielMcInnes (approved):

blammit:
Oops, yes, updated PageTemperatureSensor as well.

PR #1242, opened 2024-06-17
Do not show battery 'Details' unnecessarily

### LINKED ISSUE #1201: Battery details page showing empty values for Lynx parallel BMS
Rein earlier commented that the battery details page on v2 lacked some values.

image

blammit:
The BatteryDetails anyItemValid binding doesn't work anymore for determining whether the details list item should be shown, because some items on the Details page require a test that does not simply test the isValid value. So, add individual properties to check whether each detail would be visible on the Details page.

Fixes #1201

### DIFF
--- a/pages/settings/devicelist/battery/BatteryDetails.qml
+++ b/pages/settings/devicelist/battery/BatteryDetails.qml
@@ -10,14 +10,6 @@ QtObject {
 	id: root
 
 	property string bindPrefix
-	readonly property bool anyItemValid: {
-		for (let i = 0; i < _dataItems.length; ++i) {
-			if (_dataItems[i].isValid) {
-				return true
-			}
-		}
-		return false
-	}
 
 	readonly property alias modulesOnline: modulesOnline
 	readonly property alias modulesOffline: modulesOffline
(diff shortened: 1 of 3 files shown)


pages/settings/devicelist/battery/BatteryDetails.qml
```diff
@@ -34,6 +26,22 @@ QtObject {
 	readonly property alias installedCapacity: installedCapacity
 	readonly property alias capacity: capacity
 
+	readonly property bool allowsLowestCellVoltage: minCellVoltage.isValid
+	readonly property bool allowsHighestCellVoltage: maxCellVoltage.isValid
+	readonly property bool allowsMinimumCellTemperature: minCellTemperature.isValid
+	readonly property bool allowsMaximumCellTemperature: maxCellTemperature.isValid
+	readonly property bool allowsBatteryModules: modulesOnline.isValid || modulesOffline.isValid
+	readonly property bool allowsNumberOfModulesBlockingChargeDischarge: nrOfModulesBlockingCharge.isValid || nrOfModulesBlockingDischarge.isValid
+	readonly property bool allowsCapacity: installedCapacity.isValid
```
jpetrell, line 35: I am not sure these intermediate properties clarify things, the bindings don't become shorter, also I like seeing isValid as I then know these are bound to DBus/MQTT values.

blammit: Yeah, they don't become shorter :(

PR #1250, opened 2024-06-18
Commit text fields on focus loss

### LINKED ISSUE #1047: Clearly indicate text field confirmation step
Tap a text field to focus the editor. Edit the text using a keyboard. Continue to the next task at hand.

Actual result: If you don't remember to press enter key on the virtual or physicalkeyboard the changes are not committed. This is inconsistent to other inline components like switches and sliders. And in comparison in dialog cases the commit and cancel steps are clear.

I was surprised when I saw this behavior initially, and I still lose my changes from time to time.

Expected result: Unless the value changes implicitly (e.g. when toggling a switch or moving a slider handle), clear cancel and confirmation buttons are displayed (like how e.g. [gui-v1](https://private-user-images.githubusercontent.com/2203667/328180851-3e5f594f-9a2f-495e-b99e-ff3bb5be50ed.png?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MTU1ODgwNTQsIm5iZiI6MTcxNTU4Nzc1NCwicGF0aCI6Ii8yMjAzNjY3LzMyODE4MDg1MS0zZTVmNTk0Zi05YTJmLTQ5NWUtYjk5ZS1mZjNiYjViZTUwZWQucG5nP1gtQW16LUFsZ29yaXRobT1BV1M0LUhNQUMtU0hBMjU2JlgtQW16LUNyZWRlbnRpYWw9QUtJQVZDT0RZTFNBNTNQUUs0WkElMkYyMDI0MDUxMyUyRnVzLWVhc3QtMSUyRnMzJTJGYXdzNF9yZXF1ZXN0JlgtQW16LURhdGU9MjAyNDA1MTNUMDgwOTE0WiZYLUFtei1FeHBpcmVzPTMwMCZYLUFtei1TaWduYXR1cmU9ZTQwODcwNjJhOTE0YmQxNWNkZWI4ZGFiMzUyYzEyYTRjYzlhNTk3MTJlN2JmYzU3NmNiOWE5ZWY5YmM3NjE0NiZYLUFtei1TaWduZWRIZWFkZXJzPWhvc3QmYWN0b3JfaWQ9MCZrZXlfaWQ9MCZyZXBvX2lkPTAifQ.8GyuCkhVoUZ_hOZzmFdZcTOLixyL4tcFyGul93LYI6M) does) so the user knows they need to manually confirm the action for it to take effect.

This is a suggestion, needs discussion and design before proceeding to any changes.

jpetrell:
WIP. Still need to verify changing the behavior is ok for everyone.

Fixes #1047.

### DIFF
--- a/components/listitems/ListTextField.qml
+++ b/components/listitems/ListTextField.qml
@@ -55,9 +55,6 @@ ListItem {
 	property TextField defaultContent: TextField {
 		id: textField
 
-		property string _textWhenFocused
-		property bool _accepted
-
 		width: Math.max(Theme.geometry_listItem_textField_minimumWidth,
 						Math.min(Theme.geometry_listItem_textField_maximumWidth,
 								 implicitWidth + leftPadding + rightPadding))
@@ -66,35 +63,17 @@ ListItem {
 		rightPadding: suffixLabel.text.length ? suffixLabel.implicitWidth : leftPadding
 		horizontalAlignment: root.suffix ? Text.AlignRight : Text.AlignHCenter
 
-		onAccepted: {
+		function accept() {
 			let newValue = text
-			_accepted = true
 			if (dataItem.uid) {
 				dataItem.setValue(newValue)
 			}
-			textField.focus = false
 			root.accepted(newValue)
 		}
 
+		onAccepted: textField.focus = false
 		onEditingFinished: root.editingFinished()
-
-		onActiveFocusChanged: {
-			if (activeFocus) {
-				_textWhenFocused = text
-				_accepted = false
-			} else if (!_accepted && _textWhenFocused !== text) {
-				text = _textWhenFocused
-				revertedAnimation.to = textField.color
-				revertedAnimation.start()
-			}
-		}
-
-		ColorAnimation on color {
-			id: revertedAnimation
-
-			from: Theme.color_orange
-			duration: 400
-		}
+		onActiveFocusChanged: if (!activeFocus) accept()
 
 		Label {
 			id: suffixLabel


blammit:
I'd be happy to commit when losing focus. But what do we want to do with validation? E.g. now you can enter invalid text for IP addresses:

image

jpetrell:
> But what do we want to do with validation? E.g. now you can enter invalid text for IP addresses

Yeah need to be tested thoroughly. Converted to draft for now. Pinged Serj about the design, though could possibly ask for other people to give feedback too.

DanielMcInnes:
> I'd be happy to commit when losing focus. But what do we want to do with validation? E.g. now you can enter invalid text for IP addresses:
> 
> image

When adding modbus devices, the IP address is not validated on entry, instead it is validated when the "Add device" button is clicked. This matches gui-v1, but is not ideal from a UX perspective. Maybe we stick with the gui-v1 approach for v1.0.0, and improve the UX later?

blammit:
> When adding modbus devices, the IP address is not validated on entry, instead it is validated when the "Add device" button is clicked. This matches gui-v1, but is not ideal from a UX perspective. Maybe we stick with the gui-v1 approach for v1.0.0, and improve the UX later?

I'm happy with that. I think the problem is that we have different validation behaviour on different pages at the moment :( Also we can delay validation when you have a submit-type button like "Add device", but if it is just about changing a value and then back-stepping to the previous settings page, then we probably need to validate when the focus is lost.

But yes, let's keep the current behaviour and fix it across all text fields when we've also worked this out with Serj. I agree that the current use of revert-invalid-text-on-focus-loss in some pages is too restrictive and annoying. Maybe it should instead show a toast notification, or a warning as a caption below the text field. But if the data is invalid, and focus is lost, then ideally the data wouldn't be saved to the backend; and then in that case we need to somehow indicate the data is not saved and that the input is invalid.

blammit:
Obsoleted by https://github.com/victronenergy/gui-v2/pull/1356

PR #1255, opened 2024-06-19
Fix Chinese translations not updating 

### LINKED ISSUE #1234: Chinese displaying translation ID instead of the actual translation for 166 strings
Reported by the [customer](https://community.victronenergy.com/questions/292058/venus-os-v34026-available-for-testing.html?childToView=294132#answer-294132), some translation IDs like common_words_battery and battery_state_of_health don't have entries on the <TS> file.

image
image

Looks like Chinese translations have different set of strings compared to the other languages.

jpetrell:
After that Chinese version no longer has translation IDs showing on the GUI. There are still 166 translations missing from POEditor, but those are at least autofilled during build.

Fixes #1234

### DIFF
--- a/i18n/venus-gui-v2_zh_CN.ts
+++ b/i18n/venus-gui-v2_zh_CN.ts
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!DOCTYPE TS>
-<TS version="2.1" language="zh-CN">
+<TS version="2.1" language="zh_CN">
   <context>
     <name></name>
     <message id="ess_card_minimum_soc_value">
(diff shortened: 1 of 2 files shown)


i18n/venus-gui-v2_zh_CN.ts
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!DOCTYPE TS>
-<TS version="2.1" language="zh-CN">
+<TS version="2.1" language="zh_CN">
```
jpetrell, line 3: Caused warning "../i18n/venus-gui-v2_zh_CN.ts won't be updated: it contains translation but the target language is not recognized" during the build, and Chinese translations failed to update.

blammit:
LGTM :)

PR #1260, opened 2024-06-20
1144 wrong gauge values when some phases consume while others feed in

DanielMcInnes:


blammit:
Should we also be removing `clampMeasurement()` and its uses altogether?

Need to clarify these use cases with Jesús as well.

blammit:
@DanielMcInnes I guess this PR can be closed if the problem is actually to do with the paths used for Multi RS, and we want to keep the clamping, as discussed in the task and in slack.

DanielMcInnes:
> @DanielMcInnes I guess this PR can be closed if the problem is actually to do with the paths used for Multi RS, and we want to keep the clamping, as discussed in the task and in slack.

I thought we did want this change? Maybe I missed a message in slack, I can't find it now. A call?

blammit:
We do want to fix the issue, but not by removing clamping. The bug is only on Multi RS systems (correct?) and it is occurring because the paths that we check for whether feed-in is enabled do not work on those systems.

blammit:
> As discussed in the meeting, we will remove the clamping, so that we display the same values as gui-v1. In future, there may be changes to the backend (systemcalc?) to handle this.

Sounds good, and will you be removing AcInputs.qml `clampMeasurement()` and similar uses as well?

DanielMcInnes:
> > As discussed in the meeting, we will remove the clamping, so that we display the same values as gui-v1. In future, there may be changes to the backend (systemcalc?) to handle this.
> 
> Sounds good, and will you be removing AcInputs.qml `clampMeasurement()` and similar uses as well?

Done

DanielMcInnes:
Testing:
multi component: https://vrm.victronenergy.com/installation/394634/dashboard
grid / genset component: https://vrm.victronenergy.com/installation/68572/dashboard
vebus component: https://vrm.victronenergy.com/installation/238504/dashboard

data/common/AcInputPhaseModel.qml
```diff
@@ -9,33 +9,33 @@ import Victron.VenusOS
 ListModel {
 	id: root
 
-	readonly property bool _feedbackEnabled: Global.systemSettings.essFeedbackToGridEnabled
+	readonly property int totalPower: {
```
blammit, line 12: Was this change deliberate? Now there is a `totalPower` here as well as in the `AcInputPhaseModel` in `ActiveAcInput`, and this one does not use a timer.

(By the way, I think it makes sense to declare the `totalPower` and `firstPhaseCurrent` here instead of in `ActiveAcInput`, but that should also be done in a separate commit as it's not related to the `clampMeasurement()` changes.)

blammit (commented):
Could you squash the two commits together?

Also there's still something funny from the rebase, there are some extra commits here from the ESS fixes.

blammit:
> multi component: https://vrm.victronenergy.com/installation/394634/dashboard

@DanielMcInnes on this one I'm seeing green bars on the Brief page for the AC input, but not on the Overview page:

image
image

data/common/AcInputSystemInfo.qml
```diff
@@ -20,14 +20,8 @@ QtObject {
...
-	readonly property real minimumCurrent: _minimumCurrent.value === undefined ? NaN
-			: _feedbackEnabled ? _minimumCurrent.value
-			: Math.max(0, _minimumCurrent.value) // See AcInputs.clampMeasurement().
-	readonly property real maximumCurrent: _maximumCurrent.value === undefined ? NaN
-			: _feedbackEnabled ? _maximumCurrent.value
-			: Math.max(0, _maximumCurrent.value) // See AcInputs.clampMeasurement().
-
-	readonly property bool _feedbackEnabled: Global.systemSettings.essFeedbackToGridEnabled
+	readonly property real minimumCurrent: _minimumCurrent.value === undefined ? NaN : _minimumCurrent.value
+	readonly property real maximumCurrent: _maximumCurrent.value === undefined ? NaN : _maximumCurrent.value
```
blammit, line 24: I forgot to add, could you change these to use `isValid`, like: ` _minimumCurrent.value.isValid ? _minimumCurrent.value : NaN`

chriadam:
> > multi component: https://vrm.victronenergy.com/installation/394634/dashboard
> 
> @DanielMcInnes on this one I'm seeing green bars on the Brief page for the AC input, but not on the Overview page:
> 
> image image

I agree that the overview widget should have two green bars in this case.
Unrelated question: why is the magnitude of the green bars on the Brief Page the same as the blue bar?  Values were -4, -4, and +18 right?

blammit:
> Unrelated question: why is the magnitude of the green bars on the Brief Page the same as the blue bar? Values were -4, -4, and +18 right?

Those were the watt values, not the current values. The current was -0.8, -0.8, 0.8. So it make sense that all the bars align, because the negative values are shown with an absolute gauge value, so effectively they are 0.8, 0.8, 0.8.

data/common/AcInputPhaseModel.qml
DanielMcInnes: This was not deliberate, I think this was a rebase error. I've moved `totalPower` and `firstPhaseCurrent` as suggested

DanielMcInnes:
> > multi component: https://vrm.victronenergy.com/installation/394634/dashboard
> 
> @DanielMcInnes on this one I'm seeing green bars on the Brief page for the AC input, but not on the Overview page:
> 
> image image

Fixed. I also made a change so that the grid widget displays green text for exporting phases even when essFeedbackToGridEnabled is not configured, which feels like the right thing to do

data/common/AcInputSystemInfo.qml
DanielMcInnes: Done

blammit (approved):
Excellent :)

PR #1263, opened 2024-06-20
Fix bug in AC input power calculation

DanielMcInnes:
Specific to AC inputs on Multi RS systems. Tested OK on RS and non-RS systems.

### DIFF
--- a/data/common/AcInputPhaseModel.qml
+++ b/data/common/AcInputPhaseModel.qml
@@ -9,8 +9,6 @@ import Victron.VenusOS
 ListModel {
 	id: root
 
-	signal phaseValueChanged(int phaseIndex, string propertyName, real propertyValue)
-
 	readonly property bool _feedbackEnabled: Global.systemSettings.essFeedbackToGridEnabled
 
 	property Instantiator _phaseObjects: Instantiator {
@@ -25,7 +23,6 @@ ListModel {
 				onClampedCurrentValueChanged: {
 					if (model.index >= 0 && model.index < root.count) {
 						setProperty(model.index, "current", clampedCurrentValue)
-						root.phaseValueChanged(model.index, "current", clampedCurrentValue)
 					}
 				}
 			}
@@ -39,7 +36,6 @@ ListModel {
 				onClampedPowerValueChanged: {
 					if (model.index >= 0 && model.index < root.count) {
 						setProperty(model.index, "power", clampedPowerValue)
-						root.phaseValueChanged(model.index, "power", clampedPowerValue)
 					}
 				}
 			}
--- a/data/common/ActiveAcInput.qml
+++ b/data/common/ActiveAcInput.qml
@@ -18,7 +18,7 @@ Device {
 	readonly property int source: !!inputInfo ? inputInfo.source : VenusOS.AcInputs_InputSource_NotAvailable
 	readonly property alias gensetStatusCode: _acInputService.gensetStatusCode
 
-	readonly property real power: _acInputService.hasTotalPower ? _acInputService.totalPower : _phases.totalPower
+	readonly property real power: _phases.totalPower
 	readonly property real current: phases.count === 1 ? _phases.firstPhaseCurrent : NaN // multi-phase systems don't have a total current
 	readonly property alias currentLimit: _acInputService.currentLimit
 	readonly property alias phases: _phases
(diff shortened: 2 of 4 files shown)


data/common/ActiveAcInput.qml
```diff
@@ -27,25 +27,15 @@ Device {
...
-						total = Units.sumRealNumbers(total, phase ? phase[propertyName] : NaN)
-					}
-				}
-				totalPower = total
-			} else if (propertyName === "current" && count === 1) {
-				firstPhaseCurrent = propertyValue
+		readonly property int totalPower: {
+			let sum = 0
+			for (let i = 0; i < count; ++i) {
+				sum += get(i).power || 0
```
DanielMcInnes, line 33: The individual phase power values are integers, so there is no need to use 'Units.sumRealNumbers'

blammit: Can you move this calculation to the `multiComponent` in AcInputServiceLoader? The total doesn't need to be calculated for grid/genset/vebus as those services provide total power values. We only need to do it for Multi RS.

blammit: Well I guess it's harder to move there directly due to the model here, but can you make sure this is only calculated for the multi case?

blammit: Or, we could use a timer like gui-v1 to always calculate it manually, and avoid fetching the overall power from AcInputServiceLoader for grid/genset/vebus.

blammit (commented):
Could you remove the `_totalPower` and `power` properties/calculations from AcInputServiceLoader as well? I'm pretty sure this means they are no longer used.

data/common/ActiveAcInput.qml
DanielMcInnes: updated as suggested

DanielMcInnes:
> Could you remove the `_totalPower` and `power` properties/calculations from AcInputServiceLoader as well? I'm pretty sure this means they are no longer used.

done

chriadam:
Looks good, although I'm not a huge fan of the fixed interval timer.  Could you improve the commit message so that it explains what the original bug was, and why/how the commit fixes it?

blammit:
>> Could you remove the _totalPower and power properties/calculations from AcInputServiceLoader as well? I'm pretty sure this means they are no longer used.
> done

@DanielMcInnes there is still some stale code there, can you remove the rest as well? We don't need any power calculations in AcInputServiceLoader anymore, since ActiveAcInput doesn't use them anymore. See:

- `hasTotalPower` property
- vebusComponent: power, _powerValue, _power VeQuickItem
- gridOrGensetComponent: power, _power

data/common/AcInputPhaseModel.qml
```diff
@@ -9,8 +9,6 @@ import Victron.VenusOS
 ListModel {
 	id: root
 
-	signal phaseValueChanged(int phaseIndex, string propertyName, real propertyValue)
-
 	readonly property bool _feedbackEnabled: Global.systemSettings.essFeedbackToGridEnabled
```
blammit, line 12: Is this unused now? And same for the _feedbackEnabled checks elsewhere in this file. If unused they can also be removed.

DanielMcInnes:
> Looks good, although I'm not a huge fan of the fixed interval timer. Could you improve the commit message so that it explains what the original bug was, and why/how the commit fixes it?

Done. Could you have a look at the commit message please, and let me know if you're happy with it?

data/common/AcInputPhaseModel.qml
DanielMcInnes: '_feedbackEnabled' is still used

DanielMcInnes:
> > > Could you remove the _totalPower and power properties/calculations from AcInputServiceLoader as well? I'm pretty sure this means they are no longer used.
> > > done
> 
> @DanielMcInnes there is still some stale code there, can you remove the rest as well? We don't need any power calculations in AcInputServiceLoader anymore, since ActiveAcInput doesn't use them anymore. See:
> 
>     * `hasTotalPower` property
> 
>     * vebusComponent: power, _powerValue, _power VeQuickItem
> 
>     * gridOrGensetComponent: power, _power

Done

blammit:
@DanielMcInnes Regarding the commit message:

- "Fix bug in AC input power calculation" is quite generic -- we refer to AC input power from a number of places. Something like "Fix calculation of total AC input power" would be better as it's more specific.

- I don't think the code sample in the commit message is necessary

- "Now, we iterate over the phases correctly." -> but that's not all? There is also a Timer now, so it should explain why the Timer is added (i.e. to get parity with gui-v1, and have less frequent updates to the UI)

- Can you include the issue number with "Fixes #xx"

data/common/ActiveAcInput.qml
```diff
@@ -27,23 +27,19 @@ Device {
...
+		property real firstPhaseCurrent: count === 1 ? get(0).current : NaN
 
-		onPhaseValueChanged: function(phaseIndex, propertyName, propertyValue) {
-			if (propertyName === "power" && !root._acInputService.hasTotalPower) {
-				let total = NaN
-				for (let i = 0; i < phases.count; ++i) {
-					if (i === phaseIndex) {
-						total = Units.sumRealNumbers(total, propertyValue)
-					} else {
-						const phase = get(phaseIndex)
```
DanielMcInnes, line 40: This is where the bug was.
`const phase = get(phaseIndex)`
should be:
`const phase = get(i)`

DanielMcInnes:
> @DanielMcInnes Regarding the commit message:
> 
>     * "Fix bug in AC input power calculation" is quite generic -- we refer to AC input power from a number of places. Something like "Fix calculation of total AC input power" would be better as it's more specific.
> 
>     * I don't think the code sample in the commit message is necessary
> 
>     * "Now, we iterate over the phases correctly." -> but that's not all? There is also a Timer now, so it should explain why the Timer is added (i.e. to get parity with gui-v1, and have less frequent updates to the UI)
> 
>     * Can you include the issue number with "Fixes #xx"

Thanks Bea, done

blammit (approved):

PR #1272, opened 2024-06-24
 Overview page: Grid widget: confusion when some phases in watts and others in kilowatts

DanielMcInnes:
'Small' AcWidgets no longer hide the total power, as there is enough space to show it.
'ExtraSmall' and 'Small' AcWidgets now display units for each phase, eg 'kW', 'W'.

### DIFF
--- a/components/ThreePhaseDisplay.qml
+++ b/components/ThreePhaseDisplay.qml
@@ -75,7 +75,6 @@ Flow {
 						: 0
 				dataObject: model
 				font.pixelSize: phaseLabel.font.pixelSize
-				unitVisible: root.widgetSize >= VenusOS.OverviewWidget_Size_M
 				valueColor: phaseDelegate.textColor
 				unitColor: valueColor == Theme.color_font_primary
 						   ? Theme.color_font_secondary
--- a/components/QuantityLabel.qml
+++ b/components/QuantityLabel.qml
@@ -14,7 +14,6 @@ Item {
 	property alias font: unitLabel.font
 	property alias valueColor: valueLabel.color
 	property alias unitColor: unitLabel.color
-	property alias unitVisible: unitLabel.visible
 	readonly property alias valueText: valueLabel.text
 	readonly property alias unitText: unitLabel.text
 	property int alignment: Qt.AlignHCenter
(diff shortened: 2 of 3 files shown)


blammit:
Can you improve the commit's first line so it is less generic than "fix bug in X"? E.g. "Show per-unit phase in overview AC widget" or "Change display of units and total quantity in overview AC widget". Also add a reference to the issue number.

DanielMcInnes:
> Can you improve the commit's first line so it is less generic than "fix bug in X"? E.g. "Show per-unit phase in overview AC widget" or "Change display of units and total quantity in overview AC widget". Also add a reference to the issue number.

done

DanielMcInnes:
Serj commented:

Hi, thank you for sharing this. One problem i see is that text may become unreadable on target screen at that dimension (smallest we currently have is 12px) However changing from Museo to Roboto font gives us more vertical space, as Roboto is more condensed. In scenarios where we have multiple energy sources - 5 containers on the left, the totals for 3 phases wont fit, and there can be two sources with 3 phases and 5 blocks to show (super yacht scenario).  I would say, if we can read it on GX target screen then do not mind to go with proposal you shared. Taking in consideration all corner cases.

DanielMcInnes:
I can't test on the device with 5 widgets on the left, but on the desktop, a 5" display looks ok:
image

chriadam (approved):

components/ThreePhaseDisplay.qml
```diff
@@ -75,7 +75,6 @@ Flow {
 						: 0
 				dataObject: model
 				font.pixelSize: phaseLabel.font.pixelSize
-				unitVisible: root.widgetSize >= VenusOS.OverviewWidget_Size_M
```
blammit, line 78: Can you remove `unitVisible` from QuantityLabel? With this change, it will no longer be used.

blammit (approved):
LGTM, minor comment about clean-up.

components/ThreePhaseDisplay.qml
DanielMcInnes: Done

PR #1278, opened 2024-06-27
Fix control card for RS systems

blammit:


components/RsSystemCurrentLimitDetails.qml
```diff
@@ -0,0 +1,53 @@
...
+*/
+
+import QtQuick
+import Victron.VenusOS
+
+QtObject {
+	id: root
+
+	required property string bindPrefix
+	readonly property string title: _numberOfAcInputs.isValid && _numberOfAcInputs.value > 1
```
DanielMcInnes, line 13: Is a typical setup for this: an RS inverter with 2 inputs, the first is the grid, and the second a generator? If so, is the idea that we only allow limiting the grid input? Is it assumed that the grid is always input #1?

blammit: I don't know, sorry. @ReinvdZee would you be able to answer this?

DanielMcInnes (approved):
lgtm. Annoying that the RS systems have different paths / behaviors compared to vebus devices. I wonder if the difference is accidental or intentional / necessary for some reason. Could be worth mentioning at the next meeting.

components/RsSystemCurrentLimitDetails.qml
ReinvdZee: @blammit @DanielMcInnes We recommend connecting the generator to AC in 1 and grid to AC in 2. And we limit AC input current to reduce the impact on low-power generators, so AFAIK, we allow limiting both inputs. 

See also: https://www.victronenergy.com/live/multiplus_faq

PR #1282, opened 2024-06-28
Show notifications on startup if alarm is detected

blammit:
This was not working because NavBar setCurrentPage() was being called before the nav bar model was set.

### DIFF
--- a/ApplicationContent.qml
+++ b/ApplicationContent.qml
@@ -15,6 +15,7 @@ Item {
 
 	readonly property bool _goToNotifications: !!pageManager.navBar && !!Global.notifications
 			&& Global.notifications.alarm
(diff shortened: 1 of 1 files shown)


blammit:
Tested on https://vrm.victronenergy.com/installation/2931/dashboard but can test on any system that has an alarm active.

ApplicationContent.qml
```diff
@@ -15,6 +15,7 @@ Item {
 
 	readonly property bool _goToNotifications: !!pageManager.navBar && !!Global.notifications
 			&& Global.notifications.alarm
+			&& !!pageManager.navBar.model // if alarm is detected at startup, wait until nav bar is ready
```
chriadam, line 18: should it be model.count instead?

DanielMcInnes (approved):

ApplicationContent.qml
blammit: Doesn't need to be, but can add it as an extra precaution.

PR #1287, opened 2024-06-30
Solar quantity table summary

blammit:


pages/solar/SolarChargerPage.qml
```diff
@@ -30,7 +30,7 @@ Page {
 						},
 						{
 							title: CommonWords.yield_today,
-							value: root.solarCharger.dailyHistory(0).yieldKwh || 0,
+							value: solarCharger.dailyHistory(0)?.yieldKwh ?? NaN,
```
blammit, line 33: Note, this is equivalent to:

```
const history = solarCharger.dailyHistory(0)
return history ? history.yieldKwh : NaN
```

DanielMcInnes (approved):

PR #1292, opened 2024-07-02
Fix rendering issue on WASM with AsymmetricRoundedRectangle

chriadam:
In WASM, AsymmetricRoundedRectangle (used in TabBar for example on the Levels page) sometimes shows a line on the edge of the clip item (between the rounded background and the non-rounded background section).  The issue doesn't reproduce on desktop or on device, but it can be resolved by overlapping the rounded background section on top of the normal background section by 1 pixel (can do this now, since colours are opaque).

### DIFF
--- a/components/AsymmetricRoundedRectangle.qml
+++ b/components/AsymmetricRoundedRectangle.qml
@@ -17,6 +17,10 @@ Item {
 	property alias radius: roundedRect.radius
 	property alias border: roundedRect.border
 
+	// ensure that the entire texture is faded equally during animations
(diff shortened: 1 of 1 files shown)


blammit (approved):

DanielMcInnes:
This fixes the line, but now the tab fade in / out looks funny. Main branch isn't perfect wrt tab fade in & out, but these changes make it a bit worse

chriadam:
Thanks, fixed, now use a layer to ensure the clip can't mess with the z order etc.

DanielMcInnes (approved):
lgtm

Issue #1106, opened 2024-05-20
digital inputs: Add type for touch input control


DanielMcInnes:
See gui-v1 commit 0e3c23a0e2ad6b7d58280cf66d7d8f7af5269173

ReinvdZee:
Tested on Cerbo Mk2 with Venus OS V3.40~29

This does not work properly on V2 yet. Expected behavior:
- Set digital input type to "Touch input control"
- Toggle digital input, touch should be disabled and this toast should show up:
![image](https://github.com/victronenergy/gui-v2/assets/38864570/a1cad6d8-ad03-4c35-9023-7808c2bb20ed)
- When touching the display then another toast should show up notifying the user that the touch is disabled
![image](https://github.com/victronenergy/gui-v2/assets/38864570/fcddc54e-b543-406b-bfe9-38425fc8e18c)
- When toggling the input again, this toast should show up
![image](https://github.com/victronenergy/gui-v2/assets/38864570/af190fe6-83c5-404a-9faf-a101888d25ba)

In gui-v2, no toasts show up and the touchscreen always works. Reopening issue

kwindrem:
I would say navigation should not be disabled in the view only mode. Also, disabling changes on one display should not disable it on another (local vs multiple remote instances of gui-v2).

mpvader:
Lets discuss next week

blammit:
Discussed that:

- setting can be changed from wasm and device
- touch input disabling should only have effect on device

Additional features should be delayed to v1.1.0.

ReinvdZee:
Enabling / Disabling touch from gui-v2 is actually not as straightforward as we thought. 

`venus-platform` does create the path `Settings/Gui/TouchEnabled`, but does not handle the touch controller. There's a path just to store the setting. Gui-v1 controls the touch controller directly to enable and disable touch control. 

I don't think we should migrate this to venus-platform since a mouse (or touch) event is checked by the gui and rejected if touch is disabled. If we move this part to venus-platform, it will cause a lot of dbus traffic and decrease the responsiveness of the UI. V2 can probably implement it similar to V1. I'm not sure how that works on the WASM build, but thats for later. 

@jhofstee, @mpvader

DanielMcInnes:
[...]

PR #1293, opened 2024-07-03
1106 digital inputs add type for touch input control

DanielMcInnes:


components/ToastNotification.qml
```diff
@@ -111,6 +111,7 @@ Item {
 	Timer {
 		id: timer
 
+		interval: 0
```
DanielMcInnes, line 114: By default, 'interval' is 1000. So, if we created a 1-second auto-closing toast notification, 'onIntervalChanged' would never be triggered, and the timer would never start.

DanielMcInnes:
To test on a cerbo, change Settings/Gui/TouchEnabled with dbus-spy

components/ToastNotification.qml
chriadam: Does anyone trigger this timer without specifying a specific new interval?  if so, it will trigger immediately upon returning to event loop in that case, rather than having a (default) interval of 1000.

chriadam: Ah, I see, the default is zero in the showToastNotification method

chriadam (dismissed):

ApplicationContent.qml
```diff
@@ -60,9 +60,30 @@ Item {
...
+			}
+		}
+
 		anchors.fill: parent
 
 		onPressed: function(mouse) {
+			if (Global.isGxDevice && !touchEnabled.value) {
+				//% "Touch input disabled"
+				Global.showToastNotification(VenusOS.Notification_Info, qsTrId("application_content_touch_input_disabled"), 1000)
+				mouse.accepted = true
```
chriadam, line 83: This is one way to do it.  Another way might be to set the application's root item's enabled property to false.

chriadam: Ah, I see, you want to show the toast notification every time someone touches the screen.

chriadam (approved):
Ok, LGTM.

ApplicationContent.qml
```diff
@@ -60,9 +60,30 @@ Item {
 
 	MouseArea {
 		id: idleModeMouseArea
+
+		property VeQuickItem touchEnabled: VeQuickItem {
```
blammit, line 64: `touchEnabled` doesn't need to be a property, as the parent is `Item`.

blammit (commented):
@DanielMcInnes is there more to come in fixing this issue? The commit says this fixes #1106 but gui-v2 never changes `Settings/Gui/TouchEnabled`, so touch events continue to work on the device even when an input is set to "Touch input control".

DanielMcInnes:
> `TouchEnabled`

Nothing writes to it in gui-v1, either. I think the `dbus_digitalinputs` process handles reading the digital inputs and writing the value to dbus

ApplicationContent.qml
DanielMcInnes: Updated as suggested

blammit:
> Nothing writes to it in gui-v1, either. I think the dbus_digitalinputs process handles reading the digital inputs and writing the value to dbus

Oh, I see. What do you suggest for testing this via wasm or on device?

DanielMcInnes:
> > Nothing writes to it in gui-v1, either. I think the dbus_digitalinputs process handles reading the digital inputs and writing the value to dbus
> 
> Oh, I see. What do you suggest for testing this via wasm or on device?

It needs to be tested on both wasm builds and device builds, they behave differently. I thought it was good enough to merge as-is. Rein has a cerbo set up with digital input switches, I was going to test it there. dbus_digitalinputs reads the digital inputs, and when the input is configured as 'touch input control', it writes to /Settings/Gui/TouchEnabled. I 'tested' it by modifying /Settings/Gui/TouchEnabled with dbus-spy, and device builds and cerbo builds behave as expected.

blammit:
Okay, that makes sense then, so we need the external dbus_digitalinputs process to be triggered, and that only happens when real digital inputs are plugged in. LGTM then :)

blammit (approved):

Issue #1233, opened 2024-06-13
Empty / invalid dbus service displayed on the device list


ReinvdZee:
Discovered when testing a genset on a GX with no other devices connected. 
Even though there is a dbus service for `vebus` with a lot of empty paths, GUI-v2 should not display it when it is not valid.
Gui-v1 does not display the invalid device in the device list. 

![image](https://github.com/victronenergy/gui-v2/assets/38864570/425083f2-3580-4e1f-90a9-196cc4679491)

This also leads to an empty control card: 
![image](https://github.com/victronenergy/gui-v2/assets/38864570/3888f4cc-a92d-4120-ba06-56eaaaea7295)

Issue appears on [this](https://vrm.victronenergy.com/installation/388348/dashboard) system. 
@blammit, @DanielMcInnes @chriadam You have access, and I will leave the system running

blammit:
@ReinvdZee the Device List assumes that a device is valid if the /DeviceInstance is set, which is incorrect in this case. What would be a good test for validity? How about checking `/DeviceInstance`, `/ProductName` and `/ProductId` are all valid?

ReinvdZee:
@blammit Gui-v1 finds a device valid if `/ProductName` or `/CustomName` is set. That is why in this case it works, because both are not set. However, I think V2 should use a more robust test for validity, so I agree with your suggestion:

**A device is valid if `/DeviceInstance`, `/ProductName` and `/ProductId` are all valid and set.** 

Keep in mind that `/CustomName` should be displayed when set, if not, then `/ProductName` is used.

blammit:
Sounds good, thanks!

blammit:
Discussed that /ProductId should not be mandatory. Just look for /DeviceInstance and /ProductName.

chriadam:
Discussed with Rein that the isValid check is currently, wrong.  It is currently:
```
bool BaseDevice::isValid() const
{
	return !m_serviceUid.isEmpty() && !m_productName.isEmpty() && m_deviceInstance >= 0;
}
```

but it should be something like:
```
bool BaseDevice::isValid() const
{
	return !m_serviceUid.isEmpty() && (!m_productName.isEmpty() || !m_customName.isEmpty()) && m_deviceInstance >= 0;
}
```

chriadam:
And, in the device list, if there is a custom name set, the custom name should be shown rather than the product name.

blammit:
Okay, sounds like this previous conclusion was incorrect:

[...]

PR #1310, opened 2024-07-08
Consider a device valid if it has /CustomName instead of /ProductName

### LINKED ISSUE #1296: MultiRS devices not showing up in Device List
https://vrm.victronenergy.com/installation/394634/dashboard
device list is empty on main (only device in that system is multirs device, but it's invisible / not there?)
It shows up in gui-v1 but not gui-v2?

https://vrm.victronenergy.com/installation/394636/dashboard
device list contains three devices (external meter, BMS, etc) but cannot see the multirs device...
This one seems somehow broken in gui-v1 also?

blammit:
Some valid devices do not set /ProductName, but they do have a valid /CustomName, so consider a device valid if either of these are set.

To check this in mock data, set /CustomName instead of /ProductName where the name is not a proper product name.

Fixes #1296, #1233

### DIFF
--- a/src/basedevicemodel.cpp
+++ b/src/basedevicemodel.cpp
@@ -17,7 +17,7 @@ BaseDevice::BaseDevice(QObject *parent)
 
 bool BaseDevice::isValid() const
 {
-	return !m_serviceUid.isEmpty() && !m_productName.isEmpty() && m_deviceInstance >= 0;
+	return !m_serviceUid.isEmpty() && (!m_productName.isEmpty() || !m_customName.isEmpty()) && m_deviceInstance >= 0;
 }
 
 QString BaseDevice::serviceUid() const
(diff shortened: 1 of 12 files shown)


src/basedevicemodel.cpp
```diff
@@ -84,6 +84,23 @@ void BaseDevice::setProductName(const QString &productName)
...
 
+QString BaseDevice::customName() const
+{
+	return m_customName;
+}
+
+void BaseDevice::setCustomName(const QString &customName)
+{
+	if (m_customName != customName) {
+		const bool prevValid = isValid();
```
DanielMcInnes, line 95: We do this `isValid()` check in 4 different places now. Is it worth adding `bool m_isValid', along with `setValid(...)`? Then, whenever we change one of m_serviceUid, m_productName, m_customName or m_deviceInstance, we could call: 
`setValid(isValid());` instead of what we currently do.

blammit: I'm not sure how that would work with the emitting of `validChanged`. Ideally that would be emitted after `m_customName` is set, because otherwise, there would be a point where `valid = true` but `customName` has not yet been set. So the `validChanged` would still have to be done separately.

blammit: An alternative would be to pass a lambda to some function that does the valid check and then sets `m_customName` and similarly for the other functions.

DanielMcInnes (approved):

PR #1316, opened 2024-07-11
Only show the red 'Silence alarm' button when alarms are active

### LINKED ISSUE #1297: Silence Alarms button on NotificationsPage is showing under wrong conditions
The condition for showing the button should be Global.notifications.alarm (i.e. alarm active) and not Global.notifications.alert (i.e. warning active).

The Silence Alarms button should NOT be shown if the only active notifications are warning/alert notifications.

However, gui-v1 offers a way to acknowledge all alerts, whereas in gui-v2 design only the "Silence Alarms" button does this.
One possibility we discussed in the testing session was: maybe the condition for showing the button should be `Global.notifications.alarm || Global.notifications.alert` but the text should change to "Acknowledge Alerts" and colour should be Orange instead of Red when the condition is Global.notifications.alert?

DanielMcInnes:
Previously, we showed the red 'Silence alarm' button when an alert was active. Now, we shoe an orange 'Acknowledge alerts' button if only alerts are active, or a red 'Silence alarm' button if an alarm is active.

Fixes #1297

### DIFF
--- a/pages/NotificationsPage.qml
+++ b/pages/NotificationsPage.qml
@@ -168,10 +168,10 @@ SwipeViewPage {
 		height: Theme.geometry_notificationsPage_snoozeButton_height
 		radius: Theme.geometry_button_radius
 
-		enabled: !!Global.notifications && Global.notifications.alert && root.isCurrentPage
+		enabled: !!Global.notifications && (Global.notifications.alarm || Global.notifications.alert) && root.isCurrentPage
 		opacity: enabled ? 1 : 0
 		Behavior on opacity { OpacityAnimator { duration: Theme.animation_toastNotification_fade_duration} }
-		backgroundColor: Theme.color_critical_background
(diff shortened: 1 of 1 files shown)


pages/NotificationsPage.qml
```diff
@@ -168,26 +168,30 @@ SwipeViewPage {
...
 			}
 
 			Label {
 				anchors.verticalCenter: parent.verticalCenter
 				font.pixelSize: Theme.font_size_caption
-				//% "Silence alarm"
-				text: qsTrId("silence_alarm")
+				text: Global.notifications.alarm
+						//% "Silence alarm"
+					  ? qsTrId("silence_alarm")
```
chriadam, line 192: I'd prefer a more unique translation id, something like: "notifications_silence_alarm"

chriadam (approved):

PR #1318, opened 2024-07-12
Service uid fixes

blammit:


data/AcInputs.qml
```diff
@@ -30,23 +30,36 @@ QtObject {
...
+
+	readonly property Component _activeAcInputComponent: Component {
+		ActiveAcInput {}
+	}
+
+	// Set activeInput to a valid object when /Ac/ActiveIn/Source is set to a valid source and
+	// triggers input1Info or input2Info to become the active input.
+	onActiveInputInfoChanged: {
+		if (activeInput) {
+			activeInput.destroy()
```
chriadam, line 53: so, we now destroy the previously constructed active input object, and then construct a new one, to avoid the possibility of an object having its serviceUid change?

blammit: Yes, that's correct

data/AcInputs.qml
```diff
@@ -30,23 +30,36 @@ QtObject {
...
+			activeInput = null
```
DanielMcInnes, line 54: I think BriefSidePanel.qml:54 uses Global.acInputs.activeInput.phases without checking whether Global.acInputs.activeInput === null

DanielMcInnes (approved):
lgtm, 1 possible nit

data/AcInputs.qml
blammit: The bottom loader is inactive when Global.acInputs.activeInput is not set, according to the loadersActive binding, so it shouldn't get loaded in this case.

PR #1322, opened 2024-07-12
Update explanation in battery measurements setup page

### LINKED ISSUE #1321: Improving wording about battery configuration
In here, replace the top sentence by "Use this menu to define the battery data shown when clicking the Battery icon on the Overview page. The same selection is also visible on the VRM Portal."

I've left the HTML app out on purpose, too niche, too many people won't know what that means - and there is now no more need to mention that by itself.

image

## background

Per gui-v2 we're adding the option for a user to see all the batteries that she/he has defined in the configs. Which makes the feature complete: we already had the config option, and it was already used by two other UIs, being VRM portal + HTML5 app.

Hence the explanation sentence on the top needs revising. There are more wording improvements possible around this topic, but they are left out of this issue and will be part of menu-redesign project.

DanielMcInnes:
Fixes #1321

### DIFF
--- a/pages/settings/PageSettingsBatteries.qml
+++ b/pages/settings/PageSettingsBatteries.qml
@@ -41,7 +41,7 @@ Page {
 		id: batteryListView
 
 		header: ListLabel {
-			//% "Use this menu to define which battery measurements to see on the VRM Portal and the MFD HTML5 App."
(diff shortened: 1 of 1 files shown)


chriadam (approved):
LGTM.  Might have to sync with @dirkjanfaber as we might have to do something else to pull in the updated translations once those are ready.

Issue #1328, opened 2024-07-15
Show zero decimal places for Watts (e.g. in AC Input overview widget)


chriadam:
Don't show decimals for Watts.
Show decimals for KiloWatts.

Check what the rules is for switching between watts and kilowatts (I think it's based on the available space to draw text, but usually the boundary will be 9999 Watts -> 10 kW, I believe - but double check this).

chriadam:
re: the rules for switching between watts and kilowatts: it's hardcoded to be a 10k boundary.  Code follows:
```
			// Kilo scale has 10k limit, other scales grow by 1k
			qreal multiplier = wasPreviousScale ? 0.9 : 1;
			if (scale == VenusOS::Enums::Units_Scale_Kilo) {
				multiplier = wasPreviousScale ? 9 : 10;
			}

			return qAbs(value) >= multiplier*qPow(10, 3*scale);
```

I guess this was discussed with Serj and it was decided that it made the most sense.  This can be revisited if required.

chriadam:
For the first / main issue, of showing decimals for Watts - actually, it will never show a decimal for Watts unless the value is between -1 and 1.  A code comment says:

```
	// If value is between -1 and 1, but is not zero, show one decimal precision regardless of
	// precision parameter, to avoid showing just '0'.
```

But if you would prefer to just show 0, we can do this for Watts, of course.  I will create a PR which does this, for discussion.

chriadam:
https://github.com/victronenergy/gui-v2/pull/1330 has this PR for discussion.

DanielMcInnes:
@chriadam can this issue be closed now that #1330 is merged?

chriadam:
Yes, thanks, closing as fixed.

PR #1330, opened 2024-07-15
Don't show a decimal for zero-precision values in edge cases

chriadam:
Contributes to #1328

### DIFF
--- a/components/SideMultiGauge.qml
+++ b/components/SideMultiGauge.qml
@@ -47,11 +47,13 @@ Item {
 		delegate: Item {
 			id: gaugeDelegate
 
-			readonly property bool feedingToGrid: root.inputMode
-					&& (model.power || 0) < 0
-
 			width: Theme.geometry_briefPage_edgeGauge_width
 			height: root.height
+
+			// ignore noise values (close to zero)
+			readonly property real modelValue: Math.floor(Math.abs(model[root.phaseModelProperty] || 0)) < 1.0 ? 0.0
+					: model[root.phaseModelProperty]
+			readonly property bool feedingToGrid: root.inputMode && modelValue < 0.0
 
 			SideGauge {
 				id: gauge
@@ -93,7 +95,7 @@ Item {
 				// reverses the gauge direction so that negative and positive values have the same
 				// value on the gauge, though negative values will be drawn in green.
 				value: root.visible
-					   ? (gaugeDelegate.feedingToGrid ? Math.abs(model[root.phaseModelProperty]) : model[root.phaseModelProperty])
+					   ? (gaugeDelegate.feedingToGrid ? Math.abs(gaugeDelegate.modelValue) : gaugeDelegate.modelValue)
 					   : root.minimumValue
 				minimumValue: 0
 				maximumValue: Math.max(Math.abs(root.minimumValue), Math.abs(root.maximumValue))
--- a/data/common/AcInputPhaseModel.qml
+++ b/data/common/AcInputPhaseModel.qml
@@ -9,7 +9,7 @@ import Victron.VenusOS
 ListModel {
 	id: root
 
-	property int totalPower
+	property real totalPower
 	property real firstPhaseCurrent: count === 1 ? get(0).current : NaN
 
 	readonly property Timer _timer: Timer { // timer needed so the display doesn't update too frequently
@@ -18,8 +18,8 @@ ListModel {
 		running: true
 		onTriggered: {
 			let sum = 0
-			for (let i = 0; i < _phases.count; ++i) {
-				sum += _phases.get(i).power || 0
+			for (let i = 0; i < root.count; ++i) {
+				sum += root.get(i).power || 0
 			}
 			root.totalPower  = sum
 		}
--- a/src/units.cpp
+++ b/src/units.cpp
@@ -7,6 +7,8 @@
 
 #include <veutil/qt/unit_conversion.hpp>
 
+#include <QtMath>
+
 namespace {
 
 static const QString DegreesSymbol = QStringLiteral("\u00b0");
@@ -231,6 +233,22 @@ quantityInfo Units::getDisplayTextWithHysteresis(VenusOS::Enums::Units_Type unit
 	quantity.unit = defaultUnitString(unit, formatHints);
 	quantity.scale = VenusOS::Enums::Units_Scale_None;
 
+	// For Watts, floor values with magnitude less than 1.0 W.
+	// This ensures we ignore potential noise values.
+	if (unit == VenusOS::Enums::Units_Watt && qAbs(value) < 1.0) {
+		quantity.number = formattingLocale()->toString(0.0, 'f', 0);
+		return quantity;
+	}
+
+	// For Percentages with zero precision, if the value is between 99% and 99.9%,
+	// always show 99% so that it's clear that it's not completely full.
+	if (unit == VenusOS::Enums::Units_Percentage && (precision == 0 || precision == -1) && value > 99) {
+		quantity.number = value < 99.9
+			? formattingLocale()->toString(99.0, 'f', 0)
+			: formattingLocale()->toString(100.0, 'f', 0);
+		return quantity;
+	}
+
 	qreal scaledValue = value;
 
 	// scale value is the unit of measure is scalable
@@ -292,7 +310,7 @@ quantityInfo Units::getDisplayTextWithHysteresis(VenusOS::Enums::Units_Type unit
 		}
 	}
 
-	auto numberOfDigits = [](int value) {
+	auto numberOfDigits = [](int value) -> int {
 		int digits = 0;
 		while (value) {
 			value /= 10;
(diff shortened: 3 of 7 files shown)


src/units.cpp
```diff
@@ -307,12 +307,12 @@ quantityInfo Units::getDisplayTextWithHysteresis(VenusOS::Enums::Units_Type unit
 	}
 
 	// If value is between -1 and 1, but is not zero, show one decimal precision regardless of
-	// precision parameter, to avoid showing just '0'.
+	// precision parameter, to avoid showing just '0' (but don't do this for Watts).
```
blammit, line 310: If we do this as a rule for Watts then I think it would make sense to always do it for 0 precision, regardless of the unit.

chriadam: You're right, also Matthijs agreed.  Done.

chriadam:
I need to fix the unit tests.

src/units.cpp
```diff
@@ -306,27 +306,15 @@ quantityInfo Units::getDisplayTextWithHysteresis(VenusOS::Enums::Units_Type unit
...
-		int digits = numberOfDigits(scaledValue);
-		if (unit != VenusOS::Enums::Units_Volt_DC || digits > 2) // don't clip precision for values like 53.35 V DC.
-			precision = qMax(0, precision - qMax(0, digits - (precision == 1 ? 2 : 1)));
-
-		const qreal vFixedMultiplier = std::pow(10, precision);
-		int vFixed = qRound(scaledValue * vFixedMultiplier);
-		scaledValue = (1.0*vFixed) / vFixedMultiplier;
-		quantity.number = formattingLocale()->toString(scaledValue, 'f', precision);
-	}
+	// if the value is large (hundreds or thousands) no need to display decimals after the decimal point
```
DanielMcInnes, line 309: on line 283, I get a warning: "Value stored to 'scaled' is never read" - would you mind cleaning that up while you're here?

src/units.cpp
```diff
@@ -306,27 +306,15 @@ quantityInfo Units::getDisplayTextWithHysteresis(VenusOS::Enums::Units_Type unit
...
+	int digits = numberOfDigits(scaledValue);
```
DanielMcInnes, line 310: if scaledValue was 99.9, I think 99.9 gets cast to int and set to 99, which gives you 2 digits. Would it be better to do:

`int digits = numberOfDigits(qRound(scaledValue)) `

... which would give you 3 digits?

If not, then please explicitly cast scaledValue to int to silence a warning.

src/units.cpp
```diff
@@ -306,27 +306,15 @@ quantityInfo Units::getDisplayTextWithHysteresis(VenusOS::Enums::Units_Type unit
...
+	if (unit != VenusOS::Enums::Units_Volt_DC || digits > 2) // don't clip precision for values like 53.35 V DC.
+		precision = qMax(0, precision - qMax(0, digits - (precision == 1 ? 2 : 1)));
```
DanielMcInnes, line 312: curly brackets?
Also, earlier we have:
```	// If kilowatt-hours have not been scaled avoid decimals
	if (quantity.scale == VenusOS::Enums::Units_Scale_None && unit == VenusOS::Enums::Units_Energy_KiloWattHour) {
		precision = 0;
	}
```
Should we be overwriting `precision` for unscaled kWh?

src/units.cpp
chriadam: Will fix.  There are a variety of issues I need to fix, discrepancies between the unit tests the actual output.  Not sure why.  Will try to fix everything.

chriadam: Ah, resolved.  My original PR accidentally clobbered a necessary line.

src/units.cpp
chriadam: All of this is code movement, not changes.

src/units.cpp
chriadam: digits here means digits to the left of the decimal, it's meant to be 2.  I will do the cast.

chriadam: err, I don't see a warning.  Well, I will specify an explicit trailing type in the lambda anyway...

src/units.cpp
chriadam: Updated anyway.

components/AcInputDirectionIcon.qml
```diff
@@ -9,11 +9,15 @@ import Victron.VenusOS
 
 // When ESS feedback to grid is enabled, show an arrow indicating the flow direction.
 CP.ColorImage {
+	property real activeInputPower: (!Global.acInputs.activeInput || !Global.acInputs.activeInput.power) ? 0
```
DanielMcInnes, line 12: maybe `readonly property real _activeInputPower `

src/units.cpp
blammit: Not related to this PR, but this `precision = qMax(0, precision - qMax(0, digits - (precision == 1 ? 2 : 1)));` is difficult to make sense of.

components/SideMultiGauge.qml
```diff
@@ -48,7 +48,7 @@ Item {
 			id: gaugeDelegate
 
 			readonly property bool feedingToGrid: root.inputMode
-					&& (model.power || 0) < 0
+					&& (model.power || 0) <= -0.5 // ignore noise values (negative close to zero)
```
blammit, line 51: Can floor the value instead? `&& Math.floor(model.power || 0) < 0`

src/units.cpp
DanielMcInnes: I still see the warning, it is fixed by:
`int digits = numberOfDigits(static_cast<int>(scaledValue));`

components/SideMultiGauge.qml
DanielMcInnes: maybe `Math.round()` instead of `Math.floor()`
```
> Math.floor(-0.1)
-1
```

blammit: I suggested either round() or floor(), and Matthijs preferred floor(). But maybe that's just for the 0-1 case? Would need to check.

src/units.cpp
chriadam: Ah, I see, the parameter, not the return.  Will fix.  Strange that my compiler didn't warn.

components/AcInputDirectionIcon.qml
chriadam: done

components/AcInputDirectionIcon.qml
```diff
@@ -9,11 +9,15 @@ import Victron.VenusOS
...
-	source: !!Global.acInputs.activeInput
-			? (Global.acInputs.activeInput.power < 0 ? "qrc:/images/icon_to_grid.svg" : "qrc:/images/icon_from_grid.svg")
-			: ""
-	color: !Global.acInputs.activeInput || (Global.acInputs.activeInput.power || 0) === 0 ? Theme.color_background_disabled
-			: Global.acInputs.activeInput.power < 0 ? Theme.color_green
+	source: !Global.acInputs.activeInput ? ""
+			: activeInputPower < 0 ? "qrc:/images/icon_to_grid.svg"
+			: "qrc:/images/icon_from_grid.svg"
+	color: (!Global.acInputs.activeInput || activeInputPower === 0) ? Theme.color_background_disabled
+			: activeInputPower < 0 ? Theme.color_green
```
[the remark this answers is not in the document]
chriadam: Ah, I see.  Will do.

tests/units/tst_units.qml
```diff
@@ -60,8 +60,8 @@ TestCase {
 
 			expect(unit, NaN, "--", unitString)
 			expect(unit, 0, "0", unitString)
-			expect(unit, 0.4, "0.4", unitString)
-			expect(unit, 0.55, "0.6", unitString)
```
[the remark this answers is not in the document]
chriadam: Let's fix that as you suggested (by clamping the value coming from the AC Input `power`) rather than messing with rounding semantics in the units.cpp

components/SideMultiGauge.qml
chriadam: I will do: `(Math.floor(Math.abs(value)) < 1.0 ? 0 : value)`

data/common/AcInputPhaseModel.qml
```diff
@@ -9,7 +9,7 @@ import Victron.VenusOS
 ListModel {
 	id: root
 
-	property int totalPower
+	property real totalPower
```
DanielMcInnes, line 12: unrelated to your changes, but could you change all references to '_phases' to 'root' please?

src/units.cpp
```diff
@@ -306,28 +307,27 @@ quantityInfo Units::getDisplayTextWithHysteresis(VenusOS::Enums::Units_Type unit
...
-		int digits = numberOfDigits(scaledValue);
-		if (unit != VenusOS::Enums::Units_Volt_DC || digits > 2) // don't clip precision for values like 53.35 V DC.
-			precision = qMax(0, precision - qMax(0, digits - (precision == 1 ? 2 : 1)));
-
-		const qreal vFixedMultiplier = std::pow(10, precision);
-		int vFixed = qRound(scaledValue * vFixedMultiplier);
-		scaledValue = (1.0*vFixed) / vFixedMultiplier;
-		quantity.number = formattingLocale()->toString(scaledValue, 'f', precision);
+	int digits = numberOfDigits(static_cast<int>(scaledValue));
+	if (unit != VenusOS::Enums::Units_Volt_DC || digits > 2) {
```
DanielMcInnes, line 314: should this be:
    if (unit != VenusOS::Enums::Units_Volt_DC **&&** digits > 2) { 
?

chriadam: No; the precision adjustment is made to all values if the units are not Units_Volt_DC, and they are made to values with more than two digits if the units are Units_Volt_DC.

data/common/AcInputPhaseModel.qml
chriadam: done

src/units.cpp
chriadam: I will improve the code comment to clarify.

chriadam: done

tests/units/tst_units.qml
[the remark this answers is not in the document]
blammit: That's only for AC input power, but we have power/watts from many other sources too. Assuming we want them to all be floored then it would have to be handled in units.cpp.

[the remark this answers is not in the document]
chriadam: Ah.  Ok, I will update.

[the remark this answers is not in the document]
chriadam: Hrm, but if we do that, then we'd have to do extra flooring etc to update the directionality etc in those cases also (and in that case, we may as well floor/clamp the value itself).  This units stuff only affects the string value, effectively, not the input value.

src/units.cpp
```diff
@@ -227,13 +227,14 @@ quantityInfo Units::getDisplayTextWithHysteresis(VenusOS::Enums::Units_Type unit
 		qty.unit = defaultUnitString(unit, formatHints);
 		return qty;
 	}
+
 	quantityInfo quantity;
 	quantity.unit = defaultUnitString(unit, formatHints);
 	quantity.scale = VenusOS::Enums::Units_Scale_None;
```
DanielMcInnes, line 233: this line is redundant

DanielMcInnes (approved):
lgtm. quantityInfo Units::getDisplayTextWithHysteresis(...) still bothers me, but I don't have any ideas on how ti make it ieasier to understand. These changes improve it.

src/units.cpp
```diff
@@ -227,13 +227,14 @@ quantityInfo Units::getDisplayTextWithHysteresis(VenusOS::Enums::Units_Type unit
...
 
 	qreal scaledValue = value;
 
-	// scale value is the unit of measure is scalable
+	// scale value if the unit of measure is scalable
```
DanielMcInnes, line 237: on line 278: `bool scaled = false;`
'scaled' is set but never used

tests/units/tst_units.qml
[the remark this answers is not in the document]
chriadam: As discussed offline, other cases directionality isn't displayed separately, so it's fine.  Have now added a special case for Watts with magnitude < 1.0 to units.cpp

src/units.cpp
chriadam: done.  I thought I fixed this already but I must have clobbered it after a cherry pick.

blammit (approved):

PR #1331, opened 2024-07-15
Solar history fixes

blammit:


blammit:
Can be tested in mock mode: some trackers have names, and some do not. See this example from the solar overview drilldown:

image

DanielMcInnes (approved):

PR #1332, opened 2024-07-15
Update current & power in AcInputPhaseModel when the value is constan…

### LINKED ISSUE #1326: L2 shows dashes while L1 and L3 don't
On [this system](https://vrm.victronenergy.com/installation/370388/overview), which is a EM24 energy meter, three phase, plus a single RS on L3, gui-v2 shows -- on L2 while there is data for it.

Note that the system shows dashes for both L1 and L2 in case of a grid outage, which happens often there.

Discussion with user of that system: https://victrondevelopment.slack.com/archives/C04RYU0UMDE/p1717966505212219

I'm assuming this is a small typo or bug, hence v1.0.0 milestone. If it gets all complicated then we can move this to a later milestone.

image

DanielMcInnes:
…tly 0

Prior to this change, if the first valid phase value was 0, and it remained at 0, we would display '--' in the ThreePhaseDisplay, even though we had a valid value of '0'.

Fixes #1326

### DIFF
--- a/data/common/AcInputPhaseModel.qml
+++ b/data/common/AcInputPhaseModel.qml
@@ -30,20 +30,18 @@ ListModel {
 		delegate: QtObject {
 			readonly property VeQuickItem _current: VeQuickItem {
 				uid:  Global.system.serviceUid + "/Ac/ActiveIn/L" + (model.index + 1) + "/Current"
-				readonly property real currentValue: !isValid ? NaN : value
-				onCurrentValueChanged: {
+				onValueChanged: {
 					if (model.index >= 0 && model.index < root.count) {
-						setProperty(model.index, "current", currentValue)
+						setProperty(model.index, "current", isValid ? value : NaN)
 					}
 				}
 			}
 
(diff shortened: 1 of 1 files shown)


data/common/AcInputPhaseModel.qml
```diff
@@ -30,20 +30,18 @@ ListModel {
...
-						setProperty(model.index, "current", currentValue)
+						setProperty(model.index, "current", isValid ? value : NaN)
 					}
 				}
 			}
 
 			readonly property VeQuickItem _power: VeQuickItem {
 				uid: Global.system.serviceUid + "/Ac/ActiveIn/L" + (model.index + 1) + "/Power"
-				readonly property real powerValue: !isValid ? NaN : value
-				onPowerValueChanged: {
```
DanielMcInnes, line 44: 'onPowerValueChanged' never gets triggered if the value is always 0.

blammit (approved):
Nice, LGTM. I couldn't repro on the test system as it had non-zero values at the time, but verified the fix in mock mode by forcing the value to 0.

Issue #1035, opened 2024-05-06
Gui-v2 Device list: "Remove disconnected devices" has no function 


ReinvdZee:
Seen on gui-v2 v0.2.18. The device list properly shows connected devices: 

![image](https://github.com/victronenergy/gui-v2/assets/38864570/c5159fdd-6227-4bd0-90ad-bead7a30a1e7)

When disconnecting the device, the button "Remove disconnected devices" pops up. In gui-v1 this button then removes the disconnected devices from the list. However, in Gui-v2, the device already disappeared from the list:

![image](https://github.com/victronenergy/gui-v2/assets/38864570/36813581-6764-4d63-8e69-eea58dbcce3c)

So the button has no function. When pressed, the button disappears. 

I suppose either the disconnected should remain present until the button is pressed, or disconnected devices disappear immediately and the button should not appear at all.

kwindrem:
The way gui-v1 works is handy for debugging a system since the device doesn't just disappear from the device list with no reason as to why, or that it was ever there in the first place.

Having devices that go away automatically is a surprise to the end user (I thought I had solar, where did it go?)

blammit:
There are two separate issues here:

- disconnected devices are not visible in the Device List
- the Device List is rebuilt each time the page is opened, so if a device is disconnected when this page is not visible, it will not appear once the page is opened

chriadam:
#1245 was merged, can this be closed?

blammit:
Yep, closing.

chriadam:
- load cerbo, change between demo modes, so some devices will "vanish"
- those devices should be considered "disconnected" - they should still show up (as disabled/grey) in device list
- then there should be a "remove" button to remove them.

DanielMcInnes:
If we change demo mode from 'boat motorhome demo 1' to 'ESS', gui-v1 shows the old devices (eg. refrigerator) that are no longer connected. On gui-v2, we don't see the disconnected devices in the device list.

DanielMcInnes:
![image](https://github.com/victronenergy/gui-v2/assets/6416318/2abfe3f7-dbd4-4a32-bf38-ac5f12f6d6a4)

blammit:
[...]

PR #1337, opened 2024-07-16
Ensure disconnected devices are visible in Device List

blammit:
In, DeviceListPage.qml, show list entries even if their devices are disconnected.

In aggregatedevicemodel.cpp, consider a device to be disconnected when BaseDevice::isValid() return false (i.e. when it no longer has a device instance, etc.) instead of when it is removed from its source model.

In the long run, the Device List should be generated from com.victronenergy.* or mqtt/* instead of the individual device models, so the sourceModel attributes can be dropped at that point.

Contributes to #1035

### DIFF
--- a/src/aggregatedevicemodel.cpp
+++ b/src/aggregatedevicemodel.cpp
@@ -24,6 +24,11 @@ AggregateDeviceModel::DeviceInfo::~DeviceInfo()
 {
 }
 
+bool AggregateDeviceModel::DeviceInfo::isConnected() const
+{
+	return device && device->isValid();
+}
+
 QString AggregateDeviceModel::DeviceInfo::infoId(BaseDevice *device, BaseDeviceModel *sourceModel)
 {
 	if (!device || !sourceModel) {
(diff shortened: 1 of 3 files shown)


src/aggregatedevicemodel.cpp
```diff
@@ -24,6 +24,12 @@ AggregateDeviceModel::DeviceInfo::~DeviceInfo()
 {
 }
 
+bool AggregateDeviceModel::DeviceInfo::isConnected() const
+{
+	return device && device->isValid();
```
DanielMcInnes, line 29: I notice some member variables without the 'm_' prefix (not created in this change). I prefer member variables with the '_m' prefix, is there some reason why this changed?

src/aggregatedevicemodel.cpp
```diff
@@ -24,6 +24,12 @@ AggregateDeviceModel::DeviceInfo::~DeviceInfo()
...
+}
+
```
DanielMcInnes, line 31: extra whitespace

src/aggregatedevicemodel.cpp
blammit: It's just because DeviceInfo is basically a struct without functions (until I added this one) so it's more convenient to not add accessor functions in this regard. If it got more complex it'd be nicer to have `m_` prefixes, accessor functions, etc.

DanielMcInnes (approved):
lgtm, 1 nit

PR #1341, opened 2024-07-16
Don't show a notification the first time .../Settings/Gui/TouchEnable…

### LINKED ISSUE #1336: "Touch input on" toast notification is sometimes shown at startup

DanielMcInnes:
…d is read

Fixes #1336

### DIFF
--- a/ApplicationContent.qml
+++ b/ApplicationContent.qml
@@ -63,15 +63,21 @@ Item {
 
 		VeQuickItem {
 			id: touchEnabled
+
+			property bool _initialized: false
+
 			uid: BackendConnection.serviceUidForType("settings") + "/Settings/Gui/TouchEnabled"
 			onValueChanged: {
-				Global.showToastNotification(VenusOS.Notification_Info,
-											 (value ?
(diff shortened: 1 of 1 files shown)


ApplicationContent.qml
```diff
@@ -63,15 +63,21 @@ Item {
...
 			uid: BackendConnection.serviceUidForType("settings") + "/Settings/Gui/TouchEnabled"
 			onValueChanged: {
-				Global.showToastNotification(VenusOS.Notification_Info,
-											 (value ?
-												 //% "Touch input on"
-												 qsTrId("application_content_touch_input_on") :
-												 //% "Touch input off"
-												 qsTrId("application_content_touch_input_off")),
-											 3000)
+				if (_initialized) {
```
blammit, line 71: Could you add a comment with something like, "Only show the notification when the value changes, not when the application is loaded"?

blammit (approved):

ApplicationContent.qml
DanielMcInnes: Thanks Bea, done

PR #1346, opened 2024-07-22
Update genset error code path

### LINKED ISSUE #1268: Generator: use new error scheme
To support the new error scheme, update the gui to use `/Error/n/Id` instead of `/ErrorCode`. See related gui-v1 [commit](https://github.com/victronenergy/gui/commit/65e28c113e716603df488e177e1a4b45c0f2d8aa)

chriadam:
Use /Error/0/In instead of /ErrorCode
See 2089e6ae3c828d1ffd4f59d18b18b013f70cb596 in gui (v1)

Fixes #1268


chriadam:
See https://github.com/victronenergy/gui/commit/2089e6ae3c828d1ffd4f59d18b18b013f70cb596

chriadam:
Not yet tested.

DanielMcInnes (approved):

ReinvdZee:
@blammit, @chriadam @DanielMcInnes Can we merge this? its currently blocking releasing the first 3.50 beta.

chriadam:
I'm happy to merge it - the PR is pretty simple.  But I haven't tested on hardware, would be good if someone could verify that it does what it's supposed to do, first.

ReinvdZee:
@chriadam I tested it just now, and we're missing the actual error parsing. I see that I've also not been clear on the issue itself. I will update the issue with some additional information

chriadam:
@ReinvdZee updated - but I'm not sure how to test this.  I tried to connect to https://vrm.victronenergy.com/installation/402872/dashboard but it didn't seem to have a generator as far as I could see.

ReinvdZee:
@chriadam You can use the dse simulation of [venus-docker](https://github.com/victronenergy/venus-docker) to test this. This simulates a genset which a GX device (or a docker, as long as dbus-modbusclient is running) can find. 

I'll try to test it as well now.

pages/settings/devicelist/ac-in/PageAcInModelGenset.qml
```diff
@@ -86,12 +86,34 @@ ObjectModel {
 		}
 	}
 
-	ListFpGensetErrorItem {
-		//% "Genset error code"
+	ListItem {
+		id: gensetErrorItem
+
+		//% "Genset error codes"
```
blammit, line 92: Looking at gui-v1, it appears this should just show "Genset error code" with a single error code, rather than all valid error codes?

ReinvdZee: Yes, this field show show the active error code. I.e. the path `/Error/0/Id` contains either an empty string or some error string like "dse:e-4096". This field show display "No error" when the path is empty and the error string if the path is not empty.

Initially, the idea was that the gui needed to parse this code into a description. However, this will be done by venus-platform, see [this](https://github.com/victronenergy/venus/wiki/dbus#error-codes). For now, just display the error string if the path is not empty.

ApplicationContent.qml
```diff
@@ -18,9 +18,7 @@ Item {
 			&& !!pageManager.navBar.model && pageManager.navBar.model.count > 0 // if alarm is detected at startup, wait until nav bar is ready
 	on_GoToNotificationsChanged: {
 		if (_goToNotifications) {
-			pageManager.popAllPages()
-			mainView.controlsActive = false
-			pageManager.navBar.setCurrentPage("NotificationsPage.qml")
+			notificationLayer.popAndGoToNotifications()
```
DanielMcInnes, line 21: qrc:/qt/qml/Victron/VenusOS/pages/settings/devicelist/ac-in/PageAcInModelGenset.qml:107: TypeError: Property 'popAndGoToNotificationsPage' of object NotificationLayer_QMLTYPE_628_QML_638(0x56005f19f2c0) is not a function

DanielMcInnes:
I committed some changes so that this now shows error codes in the device list page.

Screenshot from 2024-09-09 11-14-31

Clicking on the genset error code takes you to the Notifications page, but I notice that the page is empty. Is this a bug, or do I have to configure the system to raise an alert?

DanielMcInnes:
This seems to cope ok with a silly number of genset errors:
Screenshot from 2024-09-09 11-39-31

data/common/GensetErrorModel.qml
```diff
@@ -0,0 +1,49 @@
...
+*/
+
+import QtQuick
+import Victron.VenusOS
+
+ListModel {
+	id: root
+
+	property string uidPrefix
+	property list<string> errorCodes: ["", "", "", "", "", "", "", ""]
```
DanielMcInnes, line 13: Can we get rid of `property list<string> errorCodes` ? It seems strange to have this information saved twice

DanielMcInnes: @chriadam you can reassign this ticket to me if you want

DanielMcInnes: removed `property list<string> errorCodes`

data/common/GensetErrorModel.qml
```diff
@@ -22,24 +21,13 @@ ListModel {
...
 				root.append({"errorNumber": i, "errorCode": ""})
 			}
 		}
 
-		for (let i = 0; i < root.count; ++i) {
-			const data = get(i)
-			if (data.errorNumber === errorNumber) {
-				root.setProperty(i, "errorCode", errorCode)
-				return
-			} else if (data.errorNumber > errorNumber) {
```
DanielMcInnes: @chriadam I got rid of this path, I didn't think it would ever execute. Just flagging this for your attention in case I am wrong

chriadam: with the new code, I think it's fine, as you initialize all of the values on first call.  so, yep, no worries.

pages/settings/devicelist/ac-in/PageAcInModelGenset.qml
chriadam: As per discussion in meeting, and last comment on the task, we will now show all of the codes (but not the translated strings, as those will be shown separately as notifications).

pages/settings/devicelist/ac-in/PageAcInModelGenset.qml
```diff
@@ -86,12 +86,31 @@ ObjectModel {
...
+			return errorCodes.length ? errorCodes : CommonWords.none_errors
+		}
+
 		allowed: defaultAllowed && dataItem.isValid
-		nrOfPhases: root.nrOfPhases.value || 3
+		enabled: secondaryText !== CommonWords.none_errors
+
+		onClicked: Global.notificationLayer.popAndGoToNotifications()
+
+		property VeQuickItem dataItem: VeQuickItem {
```
blammit, line 108: Seems like this `dataItem` and `errorModel` do not need to be properties, and can just be child items?

data/common/GensetErrorModel.qml
```diff
@@ -0,0 +1,33 @@
+/*
+** Copyright (C) 2023 Victron Energy B.V.
+** See LICENSE.txt for license information.
+*/
+
+import QtQuick
+import Victron.VenusOS
+
+ListModel {
```
blammit, line 9: Maybe it'd be simpler to use VeQItemTableModel instead of populating a ListModel separately? But this seems okay too.

blammit (approved):

pages/settings/devicelist/ac-in/PageAcInModelGenset.qml
DanielMcInnes: updated as suggested

Issue #1269, opened 2024-06-21
Parallel Lynx BMS: Show that Lynx BMS is waiting to connect to a parallel system


ReinvdZee:
See [this](https://github.com/victronenergy/venus/issues/1283) issue, we'd like to show that a Lynx BMS is waiting to connect to the system. Currently, the BMS summary just shows its Soc, voltage and current (0A, because its not connected). 

Instead, the following should be displayed: 

-  ~~Summary (device list entry): "Waiting" + voltage delta + threshold (formatting suggestions welcome)~~
-  ~~BMS state: "Waiting"~~ 

~~**Paths**~~
~~When the BMS is in the waiting state, the path `/State` will be 18, which is a new state dedicated for this situation. 
The voltage delta can be calculated from the physical BMS voltage and the parallel BMS voltage:
Voltage delta = `com.victronenergy.battery.lynxparallelx/Dc/0/Voltage` - `com.victronenergy.battery.socketcan_.../Dc/0/Voltage`~~

~~And since its a delta, should we take the absolute value of it? @mpvader ?~~

~~The voltage threshold will be reported in the path `/ParallelConnectThreshold`, but this must be added to vecan, so the issue is currently blocked.~~

**UPDATE**

We've changed "Waiting" into "Pending" and made some more changes. See https://github.com/victronenergy/venus/issues/1283 and [slack](https://victrondevelopment.slack.com/archives/C0154KFUE1M/p1716554398290749) for the discussions. 

A new state with value 18 is added to the BMS to represent this state. The implementation must be as follows: 

When the path `/State` is 18:
- Set the "State" field to "Pending" 
- Set the summary field to "Pending <battery voltage> <soc>.
- Do **not** translate the "Pending" term on the summary. (Reason for this is to prevent overlap with the BMS name in case of long translations)
- Show the field "System voltage" in the device-page when `/State` is 18. This field contains the voltage of the parallel BMS. To find the parallel BMS the individual BMS is part of, use the `/N2kDeviceInstance` path to compose the parallel BMS service name: `com.victronenergy.battery.lynxparallel<n2kdevice instance>`. The path `/Dc/0/Voltage` gives the parallel BMS voltage. 

Gui-v1 screenshots:

![image](https://github.com/victronenergy/gui-v2/assets/38864570/ab523dc2-6c10-4512-bf81-3e6dd5decaca)
[...]

PR #1347, opened 2024-07-23
Show that Lynx BMS is waiting to connect to a parallel system

DanielMcInnes:
Fixes #1283

### DIFF
--- a/components/CommonWords.qml
+++ b/components/CommonWords.qml
@@ -338,6 +338,9 @@ QtObject {
 	//% "Overall history"
 	readonly property string overall_history: qsTrId("common_words_overall_history")
 
+	//% "Pending"
+	readonly property string pending: qsTrId("common_words_pending")
+
 	//% "Password"
 	readonly property string password: qsTrId("common_words_password")
 
--- a/data/common/Battery.qml
+++ b/data/common/Battery.qml
@@ -17,6 +17,8 @@ Device {
 	readonly property real timeToGo: _timeToGo.value === undefined ? NaN : _timeToGo.value  // in seconds
 	readonly property string icon: !!Global.batteries ? Global.batteries.batteryIcon(battery) : ""
 	readonly property int mode: !!Global.batteries ? Global.batteries.batteryMode(battery) : -1
+	readonly property bool isParallelBms: _numberOfBmses.isValid
+	readonly property int state: _state.isValid ? _state.value : NaN
 
 	readonly property VeQuickItem _stateOfCharge: VeQuickItem {
 		uid: battery.serviceUid + "/Soc"
@@ -40,5 +42,13 @@ Device {
 
 	readonly property VeQuickItem _timeToGo: VeQuickItem {
 		uid: battery.serviceUid + "/TimeToGo"
+	}
+
+	property VeQuickItem _numberOfBmses: VeQuickItem {
+		uid: battery.serviceUid + "/NumberOfBmses"
+	}
+
+	property VeQuickItem _state: VeQuickItem {
+		uid: battery.serviceUid + "/State"
 	}
 }
--- a/pages/settings/devicelist/DeviceListPage.qml
+++ b/pages/settings/devicelist/DeviceListPage.qml
@@ -51,11 +51,17 @@ Page {
 		case "battery":
 			url = "/pages/settings/devicelist/battery/PageBattery.qml"
 			params = { "battery" : device }
-			summary = [
-				Units.getCombinedDisplayText(VenusOS.Units_Percentage, device.stateOfCharge),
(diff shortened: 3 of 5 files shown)


DanielMcInnes:
To test: get the latest venus-docker and run simulation lpp

data/common/Battery.qml
```diff
@@ -17,6 +17,9 @@ Device {
 	readonly property real timeToGo: _timeToGo.value === undefined ? NaN : _timeToGo.value  // in seconds
 	readonly property string icon: !!Global.batteries ? Global.batteries.batteryIcon(battery) : ""
 	readonly property int mode: !!Global.batteries ? Global.batteries.batteryMode(battery) : -1
+	readonly property string numberOfBmses: _numberOfBmses.isValid ? _numberOfBmses.value : "" // eg. "1/2"
```
blammit, line 20: Is `numberOfBmses` needed as a property? It is not used anywhere.

components/dialogs/ModalDialog.qml
```diff
@@ -44,7 +44,7 @@ T.Dialog {
 	rest of the app. Specifying 'width' and 'height' instead of 'implicitWidth' and 'implicitHeight' for ModalDialog
 	makes it scale properly when you shrink the window. See https://bugreports.qt.io/browse/QTBUG-127068
 	*/
-	width: Theme.geometry_modalDialog_width
+	width: Theme.geometry_modalDialog_width // TODO: https://bugreports.qt.io/browse/QTBUG-127068 is fixed in Qt 6.7, we can specify implicitWidth & implicitHeight once we upgrade.
```
blammit, line 47: Unrelated comment? Also maybe this should be a new issue instead of a comment.

pages/settings/devicelist/DeviceListPage.qml
```diff
@@ -51,11 +51,17 @@ Page {
...
 			url = "/pages/settings/devicelist/battery/PageBattery.qml"
 			params = { "battery" : device }
-			summary = [
-				Units.getCombinedDisplayText(VenusOS.Units_Percentage, device.stateOfCharge),
-				Units.getCombinedDisplayText(VenusOS.Units_Volt_DC, device.voltage),
-				Units.getCombinedDisplayText(VenusOS.Units_Amp, device.current),
-			]
+			summary = (!device.isParallelBms && device.state === VenusOS.Battery_State_Pending)
+					? [
+						  "Pending", // intentionally not translated to prevent overlap with the BMS name in case of long translations
```
blammit, line 56: We should still translate the string, and if there is a UI issue that is creating the overlap then we should fix the UI issue.

I tried something like this and couldn't get an overlap:

image

data/common/Battery.qml
DanielMcInnes: It is used by `isParallelBms`, which is used in DeviceListPage and PageBattery

components/dialogs/ModalDialog.qml
DanielMcInnes: removed comment, created https://github.com/victronenergy/gui-v2/issues/1374

pages/settings/devicelist/DeviceListPage.qml
DanielMcInnes: updated as suggested

data/common/Battery.qml
blammit: `isParallelBms` uses `_numberOfBmses` directly instead of this `numberOfBmses` property, so I think we can just have the `_ numberOfBmses` VeQuickItem without the `numberOfBmses` property.

components/CommonWords.qml
```diff
@@ -338,6 +338,9 @@ QtObject {
 	//% "Overall history"
 	readonly property string overall_history: qsTrId("common_words_overall_history")
 
+	//% "Pending"
+	readonly property string pending: qsTrId("common_words_pending")
```
blammit, line 342: I'm not sure we should have generic terms like this in CommonWords that are not attached to any context. But, that's a separate discussion about CommonWords and not related to this PR.

blammit (approved):
Minor comments. If we can remove the `numberOfBmses` property then LGTM

pages/settings/devicelist/DeviceListPage.qml
chriadam: I thought the "don't translate Pending" was a specific requirement?  At least Rein bolded it in #1269

data/common/Battery.qml
DanielMcInnes: whoops, removed

pages/settings/devicelist/DeviceListPage.qml
DanielMcInnes: I think it was a specific requirement for gui-v1, they don't wrap device names so it is more of an issue there. For gui-v2, it should be ok

PR #1349, opened 2024-07-23
Notifications: Don't display 'acknowledge all' after clicking 'silenc…

### LINKED ISSUE #1348: Notifications page: When "silence alarm" is clicked, the "Acknowledge all" button is briefly displayed

DanielMcInnes:
…e alarm'

Fixes #1348


pages/NotificationsPage.qml
```diff
@@ -195,6 +197,12 @@ SwipeViewPage {
 			}
 		}
 
-		onClicked: Global.notifications.acknowledgeAll()
+		onClicked: {
+			// break bindings so that color, text and image don't change while fading out
+			backgroundColor = backgroundColor
```
blammit, line 202: Won't this cause problems the next time an alarm or alert occurs, and the binding is required again?

Would it be easier to create two separate buttons, one for alerts and one for alarms?

DanielMcInnes: Looking at the code, it *should* cause problems, but for some reason it doesn't, and subsequent alarms display correctly. I will update as suggested

DanielMcInnes: Yes it does cause problems. Updated as suggested.

components/NotificationButton.qml
```diff
@@ -0,0 +1,32 @@
...
+Button {
+	id: root
+
+	property alias contentItemChildren: row.children
+
+	anchors {
+		right: parent ? parent.rightSideRow.right : undefined
+		verticalCenter: parent.verticalCenter
+	}
+	parent: !!Global.pageManager ? Global.pageManager.statusBar : undefined
```
DanielMcInnes, line 18: The button used to be parented to `Global.pageManager.statusBar.rightSideRow`, I had to change to to `Global.pageManager.statusBar` and anchor it to the RHS of `Global.pageManager.statusBar.rightSideRow`, as `Row` doesn't re-layout itself in a timely manner when its children's visibility changes.

chriadam: What happens if other buttons (aside from Display Sleep etc) get added to the rightSideRow?

DanielMcInnes: If we want to add other stuff to rightSideRow, we will have to revisit this. It would depend how the new button is supposed to behave.

components/NotificationButton.qml
```diff
@@ -0,0 +1,32 @@
...
+	leftPadding: Theme.geometry_silenceAlarmButton_horizontalPadding
+	rightPadding: Theme.geometry_silenceAlarmButton_horizontalPadding
+	height: Theme.geometry_notificationsPage_snoozeButton_height
+	radius: Theme.geometry_button_radius
+	opacity: enabled ? 1 : 0
+	Behavior on opacity { OpacityAnimator { duration: Theme.animation_toastNotification_fade_duration } }
+
+	contentItem: Row {
```
blammit, line 26: `Button` already supports showing an icon and text side-by-side, so it should be possible to do:

```
NotificationButton {
    icon.source: "qrc:/images/icon_alarm_snooze_24.svg"
    text: "Silence alarm"
}
```

instead of overriding the `contentItem`.

components/NotificationButton.qml
blammit: What about moving the buttons to `StatusBar.qml` and defining the NotificationButton component inline (like how `StatusBarButton` is defined there)? That way the buttons are defined directly within the layout in StatusBar.qml, and we wouldn't need awkward alias properties like Global.pageManager.statusBar.rightSideRow.

blammit (commented):
In mock mode, I pressed Shift+N to generate a new alarm. gui-v2 jumped to the Notifications page and showed the "Silence alarms" button, but then it quickly disappeared and was replaced by "Acknowledge alerts". I think I'd expect "Silence alarms" to take precedence if both alerts and alarms are present.

Also, if I click "Silence alarms" or "Acknowledge alerts", the button jumps slightly to the left before it disappears.

DanielMcInnes:
> In mock mode, I pressed Shift+N to generate a new alarm. gui-v2 jumped to the Notifications page and showed the "Silence alarms" button, but then it quickly disappeared and was replaced by "Acknowledge alerts". I think I'd expect "Silence alarms" to take precedence if both alerts and alarms are present.
> 
> Also, if I click "Silence alarms" or "Acknowledge alerts", the button jumps slightly to the left before it disappears.

Re. the "Silence alarms" button disappearing and being replaced by "Acknowledge alerts", I saw this as well, it is caused by a timer in the mock NotificationsImpl that sets `.../Active` to null. If you turn off the timers by pressing 'T', you shouldn't see it any more.

Re. the button jumping slightly to the left, I don't see this. I'm hoping it is caused by the mock timers deactivating alarms - would you mind disabling the mock timers and checking again?

components/NotificationButton.qml
DanielMcInnes: Updated as suggested.

components/NotificationButton.qml
DanielMcInnes: updated as suggested

DanielMcInnes: This has the side effect of the button text color changing while the button is pressed, as per our default Button behaviour. Looks ok to me.

blammit:
> Re. the button jumping slightly to the left, I don't see this. I'm hoping it is caused by the mock timers deactivating alarms - would you mind disabling the mock timers and checking again?

It still occurs. But if it doesn't occur with real alarms on device/wasm, then it LGTM.

DanielMcInnes:
> > Re. the button jumping slightly to the left, I don't see this. I'm hoping it is caused by the mock timers deactivating alarms - would you mind disabling the mock timers and checking again?
> 
> It still occurs. But if it doesn't occur with real alarms on device/wasm, then it LGTM.

I managed to reproduce this on a cerbo. Fixed now.

components/StatusBar.qml
```diff
@@ -90,6 +112,42 @@ Rectangle {
...
+	NotificationButton {
+		id: alarmButton
+
+		anchors {
+			right: rightSideRow.right
+			verticalCenter: parent.verticalCenter
+		}
+		enabled: notificationButtonsEnabled && !!Global.notifications && Global.notifications.alarm
+		backgroundColor: Theme.color_critical_background
+		display: C.AbstractButton.TextBesideIcon
```
blammit, line 146: `TextBesideIcon` is the default, so this is not needed

components/StatusBar.qml
```diff
@@ -90,6 +112,42 @@ Rectangle {
 			rightMargin: Theme.geometry_statusBar_rightSideRow_horizontalMargin
 			verticalCenter: parent.verticalCenter
 		}
+		width: Math.max(20, implicitWidth)
+	}
+
+	Rectangle {
+		anchors.fill: rightSideRow
+		color: "pink"
+	}
```
blammit, line 121: Leftover debug?

components/StatusBar.qml
```diff
@@ -14,7 +15,8 @@ Rectangle {
 
 	property int leftButton: VenusOS.StatusBar_LeftButton_None
 	property int rightButton: VenusOS.StatusBar_RightButton_None
-	property alias rightSideRow: rightSideRow
+	property bool notificationButtonsEnabled
```
blammit, line 18: I wonder if the value of `notificationButtonsEnabled` can just be set from within StatusBar.qml (e.g. if it could find the current page url somehow to check whether it's the NotificationsPage.qml) to avoid needing to set this from MainView.qml and needing the extra property in SwipePageModel. But no big deal either way.

components/StatusBar.qml
```diff
@@ -5,6 +5,7 @@
 
 import QtQuick
 import QtQuick.Controls as C
+import QtQuick.Controls.impl as CP
```
blammit, line 8: No longer needed?

blammit (approved):
Works well, some minor comments. LGTM

components/StatusBar.qml
DanielMcInnes: removed

components/StatusBar.qml
DanielMcInnes: ugh, whoops.

components/StatusBar.qml
DanielMcInnes: updated as suggested

components/StatusBar.qml
DanielMcInnes: removed

PR #1352, opened 2024-07-24
Run unit tests

DanielMcInnes:


CMakeLists.txt
```diff
@@ -65,11 +80,6 @@ message("Building VenusOS for ${CMAKE_SYSTEM_NAME}")
 message("CMAKE_CXX_FLAGS: ${CMAKE_CXX_FLAGS}")
 message("CMAKE_C_FLAGS: ${CMAKE_C_FLAGS}")
 
-option(VENUS_DESKTOP_BUILD "enable desktop build via cmake -DVENUS_DESKTOP_BUILD=ON" OFF) # Disabled by default
```
DanielMcInnes, line 68: VENUS_DESKTOP_BUILD, VENUS_WEBASSEMBLY_BUILD, and MQTT_WEBSOCKETS_ENABLED were never used for anything

chriadam: MQTT_WEBSOCKETS_ENABLED is used in src/veutil for the mqtt stuff.

chriadam: main.cpp has /main.cpp:259:#if !defined(VENUS_WEBASSEMBLY_BUILD) && !defined(VENUS_DESKTOP_BUILD) etc

DanielMcInnes: The cmake options were never used, i.e. `cmake -DVENUS_WEBASSEMBLY_BUILD=ON` doesn't do anything. The VENUS_WEBASSEMBLY_BUILD preprocessor stuff still gets set the same way, and still works

blammit:
I like having the tests run regularly, but on every build does seem a bit much, as it adds verbose output and a lot more noise to the logs when I rebuild and run. Could this be enabled with a specific flag instead?

It definitely makes sense to run tests whenever a PR is pushed to the repo. It would be nice to do that somehow.

DanielMcInnes:
> I like having the tests run regularly, but on every build does seem a bit much, as it adds verbose output and a lot more noise to the logs when I rebuild and run. Could this be enabled with a specific flag instead?
> 
> It definitely makes sense to run tests whenever a PR is pushed to the repo. It would be nice to do that somehow.

Updated - the unit tests no longer run every build, unless you specify `cmake -DRUN_UNIT_TESTS=ON`
The unit tests are run on the CI machine when you push changes

CMakeLists.txt
chriadam: Ah!  I see

chriadam:
Looks great, thanks.  I haven't had a chance to test this properly yet, though.  I will by Thursday for sure.

CMakeLists.txt
```diff
@@ -1018,6 +1025,16 @@ else()
     )
 endif()
 
+if (("${RUN_UNIT_TESTS}" STREQUAL "ON") AND ("${VENUS_DESKTOP_BUILD}" STREQUAL "ON")) # Don't run unit tests when building wasm or cerbo
```
blammit, line 1028: Should it always run unit tests if RUN_UNIT_TESTS is set, regardless of VENUS_DESKTOP_BUILD? If it's set then it seems like you know what you're doing, regardless of the platform.

blammit (approved):
Great! Nice work on adding the github action as well.

CMakeLists.txt
DanielMcInnes: sure, updated as suggested

PR #1356, opened 2024-07-25
Validate and accept/reject text input when focus is lost

blammit:
Align all validation/accept/reject behavior of text fields, so when text field focus is lost:

- Do text validation if required. If it succeeds, save the input; it it fails, highlight the text field and show a toast notification to indicate why the input is invalid.
- If no validation is required, save the input immediately.

Avoid setting a `validator` on the text field, so that text can be freely entered before the input is accepted. 

Text validation can also be forced by pressing Enter/Return, this is not required to save the input.

### DIFF
--- a/components/ToastNotification.qml
+++ b/components/ToastNotification.qml
@@ -17,6 +17,20 @@ Item {
 	property alias text: label.text
 	property alias autoCloseInterval: timer.interval
 
+	property bool _closed
+
+	function close(immediately) {
+		if (_closed) {
+			return
+		}
+		_closed = true
+		if (immediately) {
+			dismissed()
+		} else {
+			dismiss.dismissClicked = true
+		}
+	}
+
 	signal dismissed()
 
 	implicitWidth: parent ? parent.width : 0
--- a/components/listitems/ListIpAddressField.qml
+++ b/components/listitems/ListIpAddressField.qml
@@ -9,7 +9,35 @@ import Victron.VenusOS
 ListTextField {
 	id: root
 
+	function _isValidIp(text) {
+		// Do a simple IPv4 aaddress validation: 4 groups of 3 digits, separated by a decimal.
+		if (!text.match(/^([0-9]{1,3}\.){3}[0-9]{1,3}$/)) {
+			return false
+		}
+		const groups = text.split(".")
+		for (let i = 0; i < groups.length; ++ i) {
+			const group = parseInt(groups[i])
+			if (group < 0 || group >= 256) {
+				return false
+			}
+		}
+		return true
+	}
+
+	text: CommonWords.ip_address
 	placeholderText: "000.000.000.000"
-	textField.validator: RegularExpressionValidator { regularExpression: /[0-9\.]{1,15}/ }
 	textField.inputMethodHints: Qt.ImhDigitsOnly
+	validateInput: function() {
+		const trimmed = textField.text.trim()
+		if (!_isValidIp(trimmed)) {
+			//% "'%1' is not a valid IP address."
+			return validationResult(VenusOS.InputValidation_Result_Error, qsTrId("ip_address_input_not_valid").arg(trimmed))
+		}
+		return validationResult(VenusOS.InputValidation_Result_OK, "", trimmed)
+	}
+	saveInput: function() {
+		if (dataItem.uid) {
+			dataItem.setValue(textField.text.trim())
+		}
+	}
 }
--- a/components/listitems/ListQuantityField.qml
+++ b/components/listitems/ListQuantityField.qml
@@ -14,10 +14,22 @@ ListTextField {
 	property int decimals: Units.defaultUnitPrecision(unit)
 
 	suffix: Units.defaultUnitString(unit)
-	textField.validator: DoubleValidator {
-		decimals: root.decimals
-		locale: Units.numberFormattingLocaleName
-	}
 	textField.inputMethodHints: Qt.ImhFormattedNumbersOnly
 	textField.text: Units.formatNumber(root.value, root.decimals)
+	validateInput: function() {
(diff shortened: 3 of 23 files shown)


components/listitems/ListIpAddressField.qml
```diff
@@ -9,7 +9,21 @@ import Victron.VenusOS
 ListTextField {
 	id: root
 
+	text: CommonWords.ip_address
 	placeholderText: "000.000.000.000"
-	textField.validator: RegularExpressionValidator { regularExpression: /[0-9\.]{1,15}/ }
-	textField.inputMethodHints: Qt.ImhDigitsOnly
+	textField.inputMethodHints: Qt.ImhFormattedNumbersOnly
```
DanielMcInnes, line 14: nit - This allows us to enter minus signs. Can we keep the old `textField.validator: RegularExpressionValidator { regularExpression: /[0-9\.]{1,15}/ }` along with the new stuff? If it clashes, don't worry about keeping the old stuff.

components/listitems/ListQuantityField.qml
```diff
@@ -14,10 +14,22 @@ ListTextField {
...
 	textField.text: Units.formatNumber(root.value, root.decimals)
+	validateInput: function() {
+		const numberValue = Units.formattedNumberToReal(textField.text)
+		if (isNaN(numberValue)) {
+		   return validationResult(VenusOS.InputValidation_Result_Error, CommonWords.error_nan.arg(textField.text))
+		}
+
+		// In case the user has entered a number with a greater precision than what is supported,
+		// adjust the precision of the displayed number.
+		const formattedNumber = Units.formatNumber(numberValue, root.decimals)
```
DanielMcInnes, line 27: I think we are confusing number of decimal places with precision here. eg. 123.45 has 2 decimals, 5 precision.

```QString Units::formatNumber(qreal number, int precision) const
{
	return formattingLocale()->toString(number, 'f', precision);
}
```


https://doc.qt.io/qt-6/qlocale.html#toString
"The special precision value [QLocale::FloatingPointShortest](https://doc.qt.io/qt-6/qlocale.html#FloatingPointPrecisionOption-enum) selects the shortest representation that, when read as a number, gets back the original floating-point value. Aside from that, any negative precision is ignored in favor of the default, 6."

components/listitems/ListTextField.qml
```diff
@@ -65,35 +136,41 @@ ListItem {
...
 		}
 
-		ColorAnimation on color {
-			id: revertedAnimation
-
-			from: Theme.color_orange
-			duration: 400
+		onActiveFocusChanged: {
+			// When focus is lost and the text was changed, validate and save the text.
+			if (_validateBeforeSaving) {
```
DanielMcInnes, line 169: will this also run when it gains active focus?

pages/settings/PageVrmDeviceInstances.qml
```diff
@@ -236,8 +236,8 @@ Page {
 			return count
 		}
 
-		function updateSortOrder(index) {
-			const data = get(index)
+		function updateSortOrder(modelIndex) {
+			const data = get(modelIndex)
 			const insertionIndex = findInsertionIndex(data.deviceClass, data.name, data.vrmInstance)
 			if (insertionIndex !== modelIndex) {
 				move(index, insertionIndex, 1)
```
DanielMcInnes, line 243: should `move(index, insertionIndex, 1)` be: `move(modelIndex, insertionIndex, 1)` ?

components/listitems/ListTextField.qml
blammit: No, it won't. It will only run when losing focus or pressing Enter/Return.

components/ToastNotification.qml
```diff
@@ -17,6 +17,17 @@ Item {
 	property alias text: label.text
 	property alias autoCloseInterval: timer.interval
 
+	function close(immediately) {
+		if (!root) {
+			return
```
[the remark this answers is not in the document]
blammit: If the object is destroyed then that could happen. It was a bit weird to do it that way though, I've added a `_closed` to check it's not closed more than once.

components/listitems/ListIpAddressField.qml
blammit: I've reverted this to use Qt.ImhDigitsOnly as that still has the decimal point, and then we don't need the regex validator.

Ideally we wouldn't use validators anymore, so that you can freely edit the text input before you accept or take focus away from it (for example, if you wanted to paste in extra text and then edit it to clean it up).

pages/settings/PageVrmDeviceInstances.qml
blammit: Indeed, fixed. Thanks!

components/listitems/ListQuantityField.qml
DanielMcInnes: Never mind, this is fine. The doco goes on to contradict itself, saying precision means the number of digits after the decimal point.

DanielMcInnes (approved):
lgtm

chriadam (approved):

components/listitems/ListIpAddressField.qml
```diff
@@ -9,7 +9,21 @@ import Victron.VenusOS
 ListTextField {
 	id: root
 
+	text: CommonWords.ip_address
 	placeholderText: "000.000.000.000"
-	textField.validator: RegularExpressionValidator { regularExpression: /[0-9\.]{1,15}/ }
 	textField.inputMethodHints: Qt.ImhDigitsOnly
+	validateInput: function() {
+		const trimmed = textField.text.trim()
+		// Do a simple IPv4 aaddress validation: 4 groups of 3 digits, separated by a decimal.
```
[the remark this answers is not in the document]
blammit: Added.

PR #1372, opened 2024-07-26
Fix tank and environment input object cleanup in mock mode

blammit:
Set the deviceInstance property value directly, instead of calling _deviceInstance.setValue(), so that BaseDevice::setDeviceInstance() is called and BaseDevice::validChanged is emitted when the device instance is set to -1. This ensures the clean up handlers in Tank.qml and EnvironmentInput.qml are triggered.

Also clean up Tank.qml to use a single signal handler to check whether the tank needs to be added/removed to/from the tank models.

### DIFF
--- a/data/common/Tank.qml
+++ b/data/common/Tank.qml
@@ -45,7 +45,7 @@ Device {
 	}
 
 	onValidChanged: Qt.callLater(tank._updateModel)
-	onTypeChanged: Qt.callLater(tank._updateModel)
+	onTypeChanged: Qt.callLater(tank._updateModel) // if type changes, move tank to the correct model
 
 	function _updateModel() {
 		if (valid && type >= 0) {
--- a/data/mock/EnvironmentInputsImpl.qml
+++ b/data/mock/EnvironmentInputsImpl.qml
@@ -85,7 +85,7 @@ Item {
 			const hasCreatedObjects = _createdObjects.length > 0
 			while (_createdObjects.length > 1) {
 				let obj = _createdObjects.pop()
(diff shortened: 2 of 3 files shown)


data/common/Tank.qml
```diff
@@ -44,12 +44,10 @@ Device {
...
 
-	onValidChanged: Qt.callLater(tank._updateModel)
-	onTypeChanged: Qt.callLater(tank._updateModel)
-
-	function _updateModel() {
-		if (valid && type >= 0) {
-			if (_tankModel && _tankModel.type !== type) {
+	readonly property bool _deviceAndTypeValid: valid && type >= 0
+	on_DeviceAndTypeValidChanged: {
+		if (_deviceAndTypeValid) {
```
DanielMcInnes, line 49: previously, we would update the tank model if the tank type changed, eg. from water -> petrol. I don't think this will happen any more. Could this cause problems?

chriadam: Do you mean, add back the `|| _tankModel.type !== type`  condition to the if check below?

DanielMcInnes: No, that still wouldn't work, I don't think `on_DeviceAndTypeValidChanged:` would trigger if `type` changed from 2 -> 3.

I just tested a desktop build of main branch, if you change a tank's type, you can see that it has changed on the Levels page. On this branch, if you change a tank's type, then go back to the levels page, it displays the old tank type.

blammit: Ugh, yes. Thanks for spotting this -- have updated to avoid this change.

DanielMcInnes (approved):

PR #1381, opened 2024-08-01
Request user confirmation when changing VRM mode in wasm

### LINKED ISSUE #1378: Ask for confirmation when setting VRM mode to 'off' or 'read-only' when doing that over VRM
This applies to the setting Settings -> VRM Portal -> VRM Mode. 

Implement a popup that says "Are you sure? Changing this setting to read-only or off will lock you out. Please type YES to confirm".

And then have a box where people can type YES. Note that if that is more difficult than whats possible to do in a day; then skip that for now and leave the issue open to do that later.

Preferably the popup is only shown when gui-v2 is in WASM mode and opened via VRM. As far as I know, gui-v2 code knows that (?), and can only show the popup in that situation. Same thing: if more difficult, then fine to do that only later.

Background: when you use GUIv2 over VRM, it's currently very easy to (semi) lock yourself out by changing the new VRM mode to 'off' or 'read-only'. Immediately after setting either mode, it will stop accepting new commands.

blammit:
Changing the VRM mode to Read-only or Off will lock the user out of the VRM portal, so show a confirmation when changing this while connected via VRM.

Fixes #1378

### DIFF
--- a/pages/settings/PageSettingsLogger.qml
+++ b/pages/settings/PageSettingsLogger.qml
@@ -35,9 +35,17 @@ Page {
 
 			ListRadioButtonGroup {
 				id: vrmPortalMode
+
+				function setMode(mode) {
+					dataItem.setValue(mode)
+					Global.pageManager.popPage(root)
+				}
+
 				//% "VRM portal"
 				text: qsTrId("settings_logging_vrm_portal")
 				dataItem.uid: Global.systemSettings.serviceUid + "/Settings/Network/VrmPortal"
+				updateOnClick: false
+				popDestination: undefined   // do not automatically pop page when value is selected
 				optionModel: [
 					{ display: CommonWords.off, value: VenusOS.Vrm_PortalMode_Off },
 					//% "Read-only"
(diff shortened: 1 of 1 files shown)


pages/settings/PageSettingsLogger.qml
```diff
@@ -35,16 +35,52 @@ Page {
...
+				function setMode(mode) {
+					dataItem.setValue(mode)
+					Global.pageManager.popPage(root)
+				}
+
 				//% "VRM portal"
 				text: qsTrId("settings_logging_vrm_portal")
 				dataItem.uid: Global.systemSettings.serviceUid + "/Settings/Network/VrmPortal"
+				updateOnClick: false
+				popDestination: undefined
```
chriadam, line 48: could you had a comment about this, does it mean that we handle transitions manually, or?

pages/settings/PageSettingsLogger.qml
```diff
@@ -35,16 +35,52 @@ Page {
...
+
+					ModalWarningDialog {
+						property int mode
+
+						//% "Are you sure?"
+						title: qsTrId("settings_vrm_portal_mode_confirm_title")
+						//: %1 = translation for 'Read-only', %2 = translation for 'Off'
+						//% "Changing this setting to %1 or %2 will lock you out."
+						description: qsTrId("settings_vrm_portal_mode_confirm_description")
+								.arg(vrmPortalMode.optionModel[1].display)
```
chriadam, line 78: Can we do this sort of translation infixing?  I think it might be tricky in some languages.
It might be simpler to just have a single translation string: "Changing VRM portal mode to Read-Only or Off will lock you out" or similar

blammit: I thought it would be better to substitute, actually, so that the string automatically updates if the option names change. But I suppose the flexibility is not needed; I can just hard-code them.

pages/settings/PageSettingsLogger.qml
blammit: Yes, commented

chriadam (approved):
Code LGTM, and the modal dialog shows correctly when required, in my testing.

However, setting it to Read-Only mode doesn't seem to ... actually be read-only.  I mean, afterwards, the user can still set values and those values appear to propagate to the backend (e.g. can set it to Off or Full afterwards, and that seems to work).  Discussed that it's maybe because the VRM backend currently might not support that feature yet?

blammit:
I will merge this so that it can be tested, but let me know if there is further work to be done.

Issue #1340, opened 2024-07-16
When a battery is created with an invalid serviceUid, it is never destroyed


DanielMcInnes:
`mqtt/system/0/Batteries` is updated regularly, perhaps when voltage / current changes? If  the reported batteries are not already part of the model, new Battery objects are created, and are then only added to the model when they become valid, which never happens in some situations. A few seconds later, `mqtt/system/0/Batteries` is updated again, and more invalid Battery objects are created.

DanielMcInnes:
I have only seen this happen on a simulated system, that reported batteries via mqtt/system/0/Batteries that didn't exist. However, if it did happen on a real system, it would be tricky to track down. I think gui-v2 should be able to cope with something like this without leaking.

blammit:
Real devices are unlikely to have an invalid service uid, so setting this to low priority.

blammit:
@DanielMcInnes  With the changes for https://github.com/victronenergy/gui-v2/issues/1357, all Battery objects generated from `mqtt/system/0/Batteries` are now deleted when when the Battery drilldown page is popped. I think this is sufficient to ensure they are cleaned up; is this okay?

DanielMcInnes:
@blammit , not really. We could be on any page, and every time battery current changes (I saw this happen about every 2 seconds) we create new battery objects. If they never become 'valid', they leak

blammit:
They shouldn't be created anywhere except the Battery list page now. This is a change from before, when they were created as part of the `Global.batteries.model` generation.

DanielMcInnes:
oh right, that should be ok then

PR #1382, opened 2024-08-01
Show all com.victronenergy.battery services in Device List

### LINKED ISSUE #1357: Device List should show all batteries under com.victronenergy.battery.*
https://github.com/victronenergy/gui-v2/issues/1224 changed the battery list to be generated by `com.victronenergy.system/Batteries` instead of `com.victronenergy.battery.*`. However, this affects the batteries shown in the Device List as well. 

We actually need two separate battery lists, like this:

- When clicking on the Overview battery, show a list of all batteries in com.victronenergy.system/Batteries (if there is more than one)
- In the Device List, show all battery services under com.victronenergy.battery.*

blammit:
gui-v2 need two separate battery lists, like this:

- In the Overview Battery drilldown, show all batteries from the list in com.victronenergy.system/Batteries
- In the Device List, show all battery services under com.victronenergy.battery.*

Fix Global.batteries.model to contain all battery services under com.victronenergy.battery.*. The Overview Battery list can be created specifically when requested, when the Battery widget is clicked.

Fixes regression from 5ebe75600bdf7ecacabc480a34fec153642355cf.

Fixes #1357

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -797,6 +797,7 @@ endif()
 # Dbus_QML_MODULE
 set(Dbus_QML_MODULE_SOURCES
     data/dbus/AcSystemDevicesImpl.qml
+    data/dbus/BatteriesImpl.qml
     data/dbus/ChargersImpl.qml
     data/dbus/DBusDataManager.qml
     data/dbus/DcInputsImpl.qml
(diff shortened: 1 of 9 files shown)


DanielMcInnes (approved):

DanielMcInnes:
Did you want to address https://github.com/victronenergy/gui-v2/issues/1340 in this same ticket? No worries if not.

blammit:
Oh, good point. Since there's a separate task I will just make a separate PR for that, thanks.

Issue #686, opened 2024-01-26
WASM: ui doesn't reload (including download) in case GX is updated


mpvader:
(note that this is just based on a user report, I've not tried to reproduce this).

A user, that had the "100% doesn't fit in the brief page bug" in some version prior to Venus OS v3.20~40, thought that it wasn't fixed after updating. But it was, he just had to do a CTRL-F5 before seeing it.

Which sounds to me like something that we indeed need to make some implementation for: users often leave their web pages open. What gui-v2 when running in a web browser should do is by itself reload if the GX device was updated.

mpvader:
Meanwhile, I was also able to reproduce it.

blammit:
Maybe every 10 minutes, gui-v2 can check the currently installed version, and if it doesn’t match the one that was loaded when the app started, then force the app to be reloaded in the browser.

chriadam:
There are a variety of things we can do here.

The first issue is that we need to detect the version update (as suggested, either via polling the firmware version value, or via listening to a change signal for it, and handling specifically).

Once we have detected that version update, we need to trigger a browser page reload.  Some preliminary investigation suggests that we can modify the qtloader.js to handle an application exit code and then trigger reloading the page, which should pull down the new blob from the device.

kwindrem:
If the reload takes more than a second or two, interrupting the GUI could get in the way of a critical operation. After years of "stealing" the computer out from under a mission critical task, Microsoft now asks if you want to update now. You should ask before triggering an operation that could take a while on a slow internet connection.

chriadam:
We could potentially show a modal dialog "Version update detected, reload required" which requires the user to click "OK" which will then trigger the reload, rather than automatically triggering the reload.

kwindrem:
And a "Cancel" button also. Otherwise the effect is the same: a forced reload which could take time. Remember the WASM image is 15+ MB.

mpvader:
In the end it will need to be forced @kwindrem , that is just how it is.

So, if there has to be a cancel button, ok, but then popup should come back 10 minutes later again.

The weirdest things may happen when using an outdated gui versus the running Venus OS version. Ofcourse it will be fine 99% of the time, but not always.

So @chriadam, please implement this:

Text: Version updated detected, reload required

Two buttons:

1. Remind me in 10 minutes
2. OK

And the OK one is the larger one or highlighted one.

kwindrem:
[...]

PR #1386, opened 2024-08-02
Fix regression in page reloading after firmware update

chriadam:
We need to prevent rebuildUi() being triggered as a result of the firmware update being detected, otherwise the dialog layer will be destroyed and so the modal dialog which triggers the reload will be destroyed.

Also, with security profile support included, we need to update the method we use to reload the page when a firmware update is detected to ensure that it matches the same reload process used when a change to the security profile is detected.

### DIFF
--- a/Global.qml
+++ b/Global.qml
@@ -60,6 +60,8 @@ QtObject {
 
 	property string firmwareInstalledBuild // don't clear this on UI reload.  it needs to survive reconnection.
 	property bool firmwareInstalledBuildUpdated // as above.
+	property bool needPageReload: Qt.platform.os == "wasm" && firmwareInstalledBuildUpdated // as above.
+
 	property bool isDesktop
 	property bool isGxDevice: Qt.platform.os === "linux" && !isDesktop
 	property real scalingRatio: 1.0
--- a/Main.qml
+++ b/Main.qml
@@ -69,7 +69,7 @@ Window {
 		onConnectionReadyChanged: {
 			if (connectionReady) {
 				active = true
-			} else if (active) {
+			} else if (active && !Global.needPageReload) {
 				root.rebuildUi()
 				active = false
 			}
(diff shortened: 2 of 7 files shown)


chriadam:
WIP, don't merge.

DanielMcInnes:
To test this, should I have a wasm session running, then update the cerbo firmware via gui-v1?

chriadam:
> To test this, should I have a wasm session running, then update the cerbo firmware via gui-v1?

To repro the original issue: create a wasm build from main, patch-and-pack.sh, scp artifacts to cerbo `/var/www/venus/gui-beta`, reboot device, then point your browser to local cerbo (i.e. `http://cerbo-ip/gui-beta`), then while it's running ssh to cerbo and either change your device opkg feed from release to testing with `/opt/victronenergy/swupdate-scripts/set-feed.sh testing` or change it from testing to release via `/opt/victronenergy/swupdate-scripts/set-feed.sh release` (depending on what feed you're currently using) and then do `/opt/victronenergy/swupdate-scripts/check-updates.sh -update` to install the different firmware image.  You should notice in the browser console log something like "Firmware update detected, reloading page in 10 seconds" but then you will NOT see a modal dialog popup, nor will the page reload.

To test this fix: create a wasm build from this branch, patch-and-pack.sh, scp artifacts to cerbo `/var/www/venus/gui-beta`, reboot device, then point your browser to local cerbo (i.e. `http://cerbo-ip/gui-beta`), then while it's running ssh to cerbo and either change your device opkg feed from release to testing with `/opt/victronenergy/swupdate-scripts/set-feed.sh testing` or change it from testing to release via `/opt/victronenergy/swupdate-scripts/set-feed.sh release` (depending on what feed you're currently using) and then do `/opt/victronenergy/swupdate-scripts/check-updates.sh -update` to install the different firmware image.  You should notice in the browser console log something like: "Firmware update detected, reloading page in 10 seconds" and then you WILL see a modal dialog popup, and then the page will reload automatically.

Note that as described in the issue #686 this doesn't work completely for VRM; only for local wasm connections.

DanielMcInnes (approved):

chriadam:
Rebased.  Merging.

PR #1387, opened 2024-08-05
Add AC charger to inverter/charger overview drilldown

blammit:
Can be tested with any installation with a `charger` service, or with the `na` simulation in venus-docker.

### DIFF
--- a/components/widgets/InverterChargerWidget.qml
+++ b/components/widgets/InverterChargerWidget.qml
@@ -12,9 +12,18 @@ OverviewWidget {
 	onClicked: {
 		if ((Global.inverterChargers.veBusDevices.count
 				+ Global.inverterChargers.inverterDevices.count
+				+ Global.chargers.model.count
 				+ Global.acSystemDevices.model.count) > 1) {
 			Global.pageManager.pushPage("/pages/invertercharger/InverterChargerListPage.qml")
 		} else {
+			// Show page for chargers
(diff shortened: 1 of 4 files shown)


components/widgets/InverterChargerWidget.qml
```diff
@@ -12,9 +12,18 @@ OverviewWidget {
...
 				+ Global.inverterChargers.inverterDevices.count
+				+ Global.chargers.model.count
 				+ Global.acSystemDevices.model.count) > 1) {
 			Global.pageManager.pushPage("/pages/invertercharger/InverterChargerListPage.qml")
 		} else {
+			// Show page for chargers
+			if (Global.chargers.model.count) {
+				const charger = Global.chargers.model.firstObject
+				Global.pageManager.pushPage("/pages/settings/devicelist/PageAcCharger.qml",
+						{ "bindPrefix": charger.serviceUid, "title": charger.name })
```
blammit, line 23: The duplication between here and the page-pushing code in InverterChargerListPage.qml is a bit clunky, but that needs to be resolved as part of a wider effort to provide a globally-accessible way to push pages of specific service types.

blammit: Filed https://github.com/victronenergy/gui-v2/issues/1388 for this.

DanielMcInnes (approved):

PR #1395, opened 2024-08-07
VRM instance setting fixes

blammit:


DanielMcInnes:
Re. the commit message:
"Ensure the reboot prompt is shown if the user changes a VRM instance and navigates back without removing focus from the text field."
... I get a reboot prompt even if I change focus from the text field before navigating back. Is that intentional? If so, maybe the commit message should be changed.

chriadam (approved):
LGTM.

DanielMcInnes:
If I edit a VRM device instance without pressing 'Enter' or changing focus, then click back, should I get the reboot prompt?

DanielMcInnes (commented):
One question, other than that, lgtm

DanielMcInnes (approved):

blammit:
> If I edit a VRM device instance without pressing 'Enter' or changing focus, then click back, should I get the reboot prompt?

Yes, it should remove focus at that point and show the prompt.

DanielMcInnes:
> > If I edit a VRM device instance without pressing 'Enter' or changing focus, then click back, should I get the reboot prompt?
> 
> Yes, it should remove focus at that point and show the prompt.

I don't get the prompt when I do that

blammit:
> I don't get the prompt when I do that

Thanks, I think it was due to an issue where the swap dialog was visible at the time, or if the new value hadn't been saved to the backend yet. I've updated the last patch in this PR to fix this.

PR #1399, opened 2024-08-08
Fix CI build failure

DanielMcInnes:
Revert 'jurplel' from v4 back to v3.

I don't really understand why my original Qt6.6.3 changes caused an issue, they built fine on the dev branch, then failed when I merged them to main.

I created this branch from main, then pushed a whitespace change to trigger a CI build, and saw the build issue (shader tools). I pushed the fix, and the following CI build succeeded.

### DIFF
--- a/.github/workflows/build-wasm.yml
+++ b/.github/workflows/build-wasm.yml
@@ -51,7 +51,7 @@ jobs:
         host: 'linux'
         target: 'desktop'
         arch: 'gcc_64'
-        modules: 'qtwebsockets qt5compat qtshadertools'
+        modules: 'qtwebsockets qtshadertools'
         dir: '/opt/hostedtoolcache'
     - name: Install QT ${{env.QT_VERSION}} linux wasm and cmake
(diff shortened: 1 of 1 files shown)


.github/workflows/build-wasm.yml
```diff
@@ -42,9 +42,20 @@ jobs:
...
+    - name: Install Qt ${{env.QT_VERSION}} linux desktop
+      if: steps.cached_qt_emscripten.outputs.cache-hit != 'true'
+      uses: jurplel/install-qt-action@v3
+      with:
+        aqtversion: '==3.1.*'
+        version: "${{env.QT_VERSION}}"
+        host: 'linux'
+        target: 'desktop'
+        arch: 'gcc_64'
+        modules: 'qtwebsockets qt5compat qtshadertools'
```
chriadam, line 54: do we still need qt5compat?

chriadam (approved):
LGTM

.github/workflows/build-wasm.yml
DanielMcInnes: no we don't, removed

PR #1402, opened 2024-08-08
Fix Security Profile setting password handling

chriadam:


components/listitems/ListRadioButtonGroup.qml
```diff
@@ -135,14 +137,16 @@ ListNavigationItem {
 					Component {
 						id: passwordComponent
 
-						Column {
+						Item {
+							implicitHeight: passwordLabel.height + passwordField.height
```
DanielMcInnes, line 141: can this Item be a Column, and get rid of implicitHeight and passwordField.y ?

components/listitems/ListRadioButtonGroup.qml
```diff
@@ -162,46 +168,44 @@ ListNavigationItem {
 								textField.echoMode: TextInput.Password
 								enabled: radioButton.enabled
 								backgroundRect.color: "transparent"
-								Component.onCompleted: allowed = model.index !== root.currentIndex
+
+								// Setting implicitHeight should not be required, but for some reason
+								// the ListTextField type doesn't have a valid implicitHeight by default
+								// if the height is specified.  So, specify it.
+								implicitHeight: defaultImplicitHeight
```
DanielMcInnes, line 175: I would prefer this be fixed in ListTextField. If it's wrong there, it will cause problems in lots of places

blammit:
@chriadam @DanielMcInnes 

As discussed, I've reworked the password validation UI and API so it is a bit cleaner and easier to use similarly for both 'Access level' and 'Security profile' options. I've kept the original patches in `chriadam/fix-radio-button-passwords-orig` for reference.

I've added a 'Confirm' button next to the password field. The idea is that if you click this (or press Enter/Return) then the password will be validated. 

image

If you back-step, the action is cancelled, and the password is *not* validated. The reason for adding this is that we should have an option to cancel the mode change without validating the password. However, it does mean that this behaves a bit differently from text inputs elsewhere, as those are validated whenever focus is lost, including when you backstep to the previous page. 

What do you think?

blammit:
Oops, there are some issues in the rebase. Will fix and update.

blammit:
> Oops, there are some issues in the rebase. Will fix and update.

Fixed now, ready to review again.

blammit:
Updated the Debug > Test page to show how the password changes work for the combo boxes.

components/Utils.js
```diff
@@ -325,3 +325,12 @@ function simplifiedNetworkType(t) {
...
 	return n ? "0x" + n.toString(16).toUpperCase() : ""
 }
+
+// Returns an object to return for text field validation. The status can be:
+// - InputValidation_Result_Error: the input text is invalid; highlight the input with a red border
+// - InputValidation_Result_OK: the input text is valid, and can be saved, or red highlight can be removed
+// - InputValidation_Result_Unknown: the input text should not be saved, but is not invalid, so
+//   do not highlight with a red border. This is useful when the input string is empty.
+function validationResult(status, errorText = "", adjustedText = undefined) {
+	return { status: status, errorText = errorText, adjustedText = adjustedText }
```
DanielMcInnes, line 335: This syntax looks weird. Is there a reason why it's not: 

`return { status: status, errorText: errorText, adjustedText: adjustedText }` ?

blammit: That looks like a copy-paste error :( Thank you, fixed.

DanielMcInnes (approved):

pages/settings/PageSettingsGeneral.qml
```diff
@@ -187,17 +193,23 @@ Page {
...
 					//% "Superuser"
 					{ display: qsTrId("settings_access_superuser"), value: VenusOS.User_AccessType_SuperUser, readOnly: true },
 					//% "Service"
 					{ display: qsTrId("settings_access_service"), value: VenusOS.User_AccessType_Service, readOnly: true },
 				]
+				validatePassword: (index, password) => {
+					if ((index === 0 || index === 1) && password === "ZZZ") {
+						return Utils.validationResult(VenusOS.InputValidation_Result_OK)
+					}
+					//% "%1: Incorrect password"
```
chriadam, line 210: what is %1 here?

chriadam: Maybe worth adding a context comment so that translators have a better idea about what it is.  I assume it's a number?  (If it's a string, it will need to be translated, and then this translation form probably doesn't work, as individually translated strings can't be injected into larger translations, generally).

blammit: The primary label text on the matching list item, e.g. "User" or "User & Installer" in the Access Level case. We could also just show "Incorrect password" by itself.

components/listitems/ListRadioButtonGroup.qml
chriadam: No.  Column doesn't have an implicitHeight, only a height, and if it's in a Loader, the Column height will be written to fill the Loader (if the loader dimensions are set).  This is the precise reason why I changed the root component item to be an Item - it can have an implicitHeight (which can be used by the Loader to inform its explicit dimensions) as well as a height (which can be set by Loader etc).

pages/settings/PageSettingsGeneral.qml
blammit: Simplified this, it's now just "Incorrect password".

Issue #1389, opened 2024-08-05
Failing to build gui-v2 on macOS


itemir:
Following the instructions on [How to build venus gui v2](https://github.com/victronenergy/gui-v2/wiki/How-to-build-venus-gui-v2#patch-and-pack-the-webassembly-files), specifically *Building for Linux / Max OS X Desktop* on a macOS 14.5 but failing with the following error message:

```
itemir@Ilkers-MacBook-Pro build % ~/Qt/Tools/CMake/CMake.app/Contents/bin/cmake -DNO_CACHEGEN=true -DLOAD_QML_FROM_FILESYSTEM=true -DCMAKE_BUILD_TYPE=MinSizeRel ../
-- The CXX compiler identification is AppleClang 15.0.0.15000309
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
CMake Deprecation Warning at tests/CMakeLists.txt:6 (cmake_minimum_required):
  Compatibility with CMake < 3.5 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value or use a ...<max> suffix to tell
  CMake that the project does not need compatibility with older versions.


-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Success
-- Found Threads: TRUE
-- Performing Test HAVE_STDATOMIC
-- Performing Test HAVE_STDATOMIC - Success
-- Found WrapAtomic: TRUE
-- Found OpenGL: /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX14.5.sdk/System/Library/Frameworks/OpenGL.framework
-- Found WrapOpenGL: TRUE
-- Could NOT find WrapVulkanHeaders (missing: Vulkan_INCLUDE_DIR)
-- Could NOT find WrapVulkanHeaders (missing: Vulkan_INCLUDE_DIR)
-- Could NOT find WrapVulkanHeaders (missing: Vulkan_INCLUDE_DIR)
-- Could NOT find WrapVulkanHeaders (missing: Vulkan_INCLUDE_DIR)
[...]

PR #1403, opened 2024-08-08
Fix qobject cast in units hysteresis handling

chriadam:
Contributes to #1389


DanielMcInnes (approved):
Did this actually work before? i.e. did qobject_cast convert Units* to Units, or did it return a nullptr?

PR #1406, opened 2024-08-09
1323 lynx shunt vecan peukert exponent not adjustable

DanielMcInnes:


components/dialogs/NumberSelectorDialog.qml
```diff
@@ -62,8 +62,8 @@ ModalDialog {
 				textFromValue: function(value, locale) {
 					return Units.formatNumber(value / root._multiplier(), root.decimals) + root.suffix
 				}
-				from: root.from * root._multiplier()
-				to: root.to * root._multiplier()
+				from: Math.max(-2147483648, root.from * root._multiplier())
+				to: Math.min(2147483647, root.to * root._multiplier())
```
blammit, line 66: Can we get these constants from somewhere else instead of hardcoding them?

DanielMcInnes: I can't find these defined by qt anywhere. Added Global.int32Min, int32Max

blammit: It would be nicer to add it to some number-feature-specific singleton instead, but I guess we don't have that at the moment.

components/listitems/ListSpinBox.qml
```diff
@@ -14,8 +14,7 @@ ListButton {
 	property string suffix
 	property int decimals
 	property int from: !isNaN(dataItem.min) ? dataItem.min : 0
-	// Number.MAX_SAFE_INTEGER === 9,007,199,254,740,991, which is less than C++ 64 bit max int 9,223,372,036,854,775,807, so should be safe to use for 'SpinBox.to'
-	property int to: !isNaN(dataItem.max) ? dataItem.max : Number.MAX_SAFE_INTEGER
+	property int to: !isNaN(dataItem.max) ? dataItem.max : 2147483647 / Math.pow(10, decimals) // qml int is a signed 32 bit value
```
blammit, line 17: Remove this hardcoded one as well?

blammit (approved):
One hardcoded value remaining, LGTM

components/listitems/ListSpinBox.qml
DanielMcInnes: whoops, done

PR #1407, opened 2024-08-09
Improve SpinBox scaling performance

### LINKED ISSUE #1371: Spinboxes: UI on Cerbo becomes unresponsive when holding + button (?)
Reported on community:

IMG_2454

DanielMcInnes:
We were doubling the number of calls to SpinBox.increase() / decrease() every 500mS, which meant that after holding the + button down for 10 seconds, there were 1 million calls to increase() every second. The cerbo would become unresponsive shortly after that.

Fixes #1371

### DIFF
--- a/components/controls/SpinBox.qml
+++ b/components/controls/SpinBox.qml
@@ -9,6 +9,10 @@ import QtQuick.Templates as CT
 import QtQuick.Controls.impl as CP
 import Victron.VenusOS
 
+// SpinBox uses a binding to increase 'stepSize' when the user holds a button down for a while. This allows the spin box to quickly change arbitrarily large values.
+// When the button is released, 'stepSize' reverts to its original value.
+// TODO - find a way to do this without exposing the changes to 'stepSize', as it may surprise developers when the value changes unexpectedly.
+
 CT.SpinBox {
 	id: root
 
@@ -17,6 +21,7 @@ CT.SpinBox {
 	property int indicatorImplicitWidth: Theme.geometry_spinBox_indicator_minimumWidth
 	property int orientation: Qt.Horizontal
 	property int _scalingFactor: 1
+	property int _originalStepSize
 
 	signal maxValueReached()
 	signal minValueReached()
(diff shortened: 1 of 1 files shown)


components/controls/SpinBox.qml
```diff
@@ -119,22 +120,27 @@ CT.SpinBox {
...
+			if (running) {
+				_originalStepSize = stepSize
+			} else {
 				_scalingFactor = 1
 			}
 		}
 	}
 
+	Binding {
+		root.stepSize: root._originalStepSize * _scalingFactor
```
blammit, line 132: Works well but would be nice to do this without a Binding, as this changes the value of the public `stepSize` property, which would be confusing if you've set the stepSize and expect it to stay on that value.

DanielMcInnes: It gets restored back to the original value when you release the button, is that good enough? Or is it an issue if it changes temporarily?

blammit: Ideally it wouldn't change temporarily, if we can avoid it.  Could we just stop the repeat timer after a certain number of iterations?

DanielMcInnes: If we did that, we would lose the ability to handle arbitrarily large ranges. We would still need to call `increase()` thousands of times to have useful scaling, which feels wrong. I'll think about it. Maybe `SpinBox` becomes an `Item` that contains a `CT.SpinBox`, with aliases etc.

DanielMcInnes: I can't think of a better way to do it. I looked through the code base and can't see anywhere that a changing stepSize would cause problems / surprises

blammit: Okay cool, let's keep it as-is, but could you just add a TODO comment to find an approach that doesn't modify stepSize? In case in the future we set stepSize somewhere and then we're surprised when it sometimes changes from the set value.

blammit (approved):
Minor comment, LGTM.

components/controls/SpinBox.qml
DanielMcInnes: Added TODO comment

PR #1410, opened 2024-08-14
Add QR code generation support

### LINKED ISSUE #1350: Add QR code generation support to gui-v2
For a third-party vendor (Tailscale) integration into VenusOS, QR code support is needed.  The settings page which will be integrated into gui-v2 must be able to generate a QR code based on some system information.

So, possible solution:
1) import qzxing (Apache 2.0 license) as a git submodule and build it as a dependency within gui-v2
2) ensure that its types are registered into appropriate QML namespace for use within gui-v2 settings pages code
3) ensure that it is built statically for WebAssembly builds.

My understanding is that this is needed for v3.50 release i.e. 1.0.0 milestone.

Should discuss all of these details with Matthijs at the meeting on Thursday to double check that I understood the requirements correctly.

DanielMcInnes:
Fixes #1350

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,6 @@
 [submodule "veutil"]
 	path = src/veutil
 	url = git@github.com:victronenergy/veutil.git
+[submodule "src/qzxing"]
+	path = src/qzxing
+	url = https://github.com/victronenergy/qzxing
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -120,6 +120,10 @@ if(QT_KNOWN_POLICY_QTP0001)
     qt_policy(SET QTP0001 NEW) # >Qt6.5 only. Enabling this policy ensures that your QML module is placed under a default import path, and its types can be found without manual calls to QQmlEngine::addImportPath.
 endif()
 
+set(QZXING_USE_QML ON)
+set(QZXING_USE_ENCODER ON)
+add_subdirectory(src/qzxing/src)
+
 include_directories(src/veutil/inc src .)
 
 set (VENUS_QML_MODULE_SINGLETON_SOURCES # All qml singletons have to be added here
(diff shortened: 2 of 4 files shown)


.gitmodules
```diff
@@ -1,3 +1,6 @@
 [submodule "veutil"]
 	path = src/veutil
 	url = git@github.com:victronenergy/veutil.git
+[submodule "src/qzxing"]
+	path = src/qzxing
+	url = git@github.com:DanielMcInnes/qzxing.git
```
DanielMcInnes, line 6: I had to modify the qzxing project to suit our purposes. If we branch the official qzxing repo, I think it is possible for them to delete our branch, so I think the best place to keep it is in a new victron repo. If @chriadam and @blammit agree, I'll ask someone (Dirk Jan?) to create a new victron repo.

chriadam: What modifications were required?

DanielMcInnes: hardly anything. Just use qt6 intead of qt5, and don't use Qt Multimedia

chriadam: one possibility is: instead of forking it, just add the sources and whatever directly to a module in our CMakeLists.txt, rather than including their CMakeLists.txt as a subproject/subdirectory directly.

DanielMcInnes: I don't like that idea very much. There are 500 lines of cmake code in qzxing, so there is a fair bit going on

DanielMcInnes: I asked Dirk Jan about creating a new repo, he said only Matthijs and Martin have the permissions for that, and they are both on holidays. @chriadam , do you want to merge this as-is, or wait for Matthijs to get back?

chriadam:
A possible alternative to forking might just be to add a patch to our git repo, add the upstream qzxing as a submodule, and just apply the patch as a pre-build step.

.gitmodules
chriadam: I don't think we can merge this.  Let's ask Matthijs which solution suits him (i.e.: 1) fork qzxing under victron git, 2) checkout upstream qzxing as a submodule and apply a patch as a pre-build step in gui-v2 via cmake magic) first.

DanielMcInnes: updated to use victron's forked qzxing repo

chriadam (approved):

PR #1415, opened 2024-08-19
Show custom tank name

DanielMcInnes:
Fix bug in TanksTab. We were enabling the ExpandedTanksView when there was more than 1 tank of a specific type. We should only enable this when there are 'merged' tanks displayed in the delegate.

### DIFF
--- a/pages/TanksTab.qml
+++ b/pages/TanksTab.qml
@@ -108,7 +108,7 @@ ListView {
 
 					PressArea {
 						anchors.fill: parent
-						enabled: tankTypeDelegate.tankModel.count > 1
+						enabled: tankTypeDelegate.mergeTanks
 						radius: gaugeGroup.radius
 						onClicked: {
 							expandedTanksLoader.tankModel = tankTypeDelegate.tankModel
--- a/data/common/Tank.qml
+++ b/data/common/Tank.qml
@@ -91,7 +91,7 @@ Device {
 		// This must be a generic Victron tank sensor, where the product name is always "Tank
 		// sensor" and there is no custom name, so use the tank type to provide a meaningful name.
 		? Gauges.tankProperties(type).name
-		: productName || customName
+		: customName || productName
 
 	description: {
 		if (customName.length > 0) {


DanielMcInnes:
Matthijs asked us to remove the ExpandedTanksView. I think he thought it was useless because we were enabling it when we shouldn't have. If you have 'merged' tanks it is more useful, you can see individual tank capacities / custom names etc.

DanielMcInnes:
Screenshot from 2024-08-19 16-53-42
ExpandedTanksView is now only enabled for 'merged' tanks such as the Fresh Water tanks show here.

components/ExpandedTanksView.qml
```diff
@@ -47,7 +47,7 @@ Rectangle {
 				width: Gauges.width(groupedSubgaugesRepeater.count, Theme.geometry_levelsPage_max_tank_count, root.width)
 				height: Theme.geometry_levelsPage_panel_expanded_height
 
-				header.text: model.device.name || root._tankProperties.name
+				header.text: model.device.customName || model.device.name || root._tankProperties.name
```
blammit, line 50: This won't be needed if the tank name is fixed in the other PR.

pages/TanksTab.qml
```diff
@@ -93,7 +93,7 @@ ListView {
 
 					animationEnabled: root.animationEnabled
 					tankType: tankTypeDelegate.tankType
-					header.text: mergeTanks ? tankProperties.name : model.device.name || tankProperties.name
+					header.text: mergeTanks ? tankProperties.name : model.device.customName || model.device.name || tankProperties.name
```
blammit, line 96: Same here

pages/TanksTab.qml
```diff
@@ -108,7 +108,7 @@ ListView {
 
 					PressArea {
 						anchors.fill: parent
-						enabled: tankTypeDelegate.tankModel.count > 1
+						enabled:tankTypeDelegate.mergeTanks
```
blammit, line 111: Add space after `enabled:`?

blammit (approved):

PR #1417, opened 2024-08-20
Show the custom name on the DeviceList -> details page

### LINKED ISSUE #1365: Devicelist -> tank level details page: show custom name at the top
so not like below screenshot, it shouldn't say GX Tank 140.

image

DanielMcInnes:
Fixes #1365

Matthijs asked for this for tank devices, but I have done it for all devices. I can't think of a reason why you wouldn't want the custom name instead of a more generic name.

### DIFF
--- a/data/common/Tank.qml
+++ b/data/common/Tank.qml
@@ -91,7 +91,7 @@ Device {
 		// This must be a generic Victron tank sensor, where the product name is always "Tank
 		// sensor" and there is no custom name, so use the tank type to provide a meaningful name.
 		? Gauges.tankProperties(type).name
-		: productName || customName
+		: customName || productName
 
 	description: {
 		if (customName.length > 0) {


pages/settings/devicelist/DeviceListPage.qml
```diff
@@ -203,7 +203,7 @@ Page {
 		default:
 			return null
 		}
-		params.title = device.name
+		params.title = device.customName || device.name
```
blammit, line 206: The device name should ideally come from the custom name. I think the bug is in Tank.qml where the device name is generated. This would do the trick:

```
diff --git a/data/common/Tank.qml b/data/common/Tank.qml
index 0d0f9918..1bbb1aa4 100644
--- a/data/common/Tank.qml
+++ b/data/common/Tank.qml
@@ -91,7 +91,7 @@ Device {
                // This must be a generic Victron tank sensor, where the product name is always "Tank
                // sensor" and there is no custom name, so use the tank type to provide a meaningful name.
                ? Gauges.tankProperties(type).name
-               : productName || customName
+               : customName || productName
 
        description: {
                if (customName.length > 0) {
```

DanielMcInnes: updated as suggested

blammit (approved):

PR #1418, opened 2024-08-20
Move the solar widget up to avoid connectors crossing

### LINKED ISSUE #1367: Overview page: move 'solar yield' widget above the DC generator widget
image

We want to minimize the number of overlapping connectors.

DanielMcInnes:
(Overview page)

The solar widget is the only widget on the left side of the overview page that can connect to both of the center widgets (inverter & battery). There is now no situation where the LHS widget connectors cross each other.

Screenshot from 2024-08-20 11-33-45

Fixes #1367

### DIFF
--- a/pages/OverviewPage.qml
+++ b/pages/OverviewPage.qml
@@ -13,15 +13,24 @@ SwipeViewPage {
 	readonly property var _centerWidgets: [inverterChargerWidget, batteryWidget]
 	property var _rightWidgets: []
 
-	// Preferred order for the input widgets on the left hand side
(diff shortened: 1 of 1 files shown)


pages/OverviewPage.qml
```diff
@@ -17,11 +17,11 @@ SwipeViewPage {
 	readonly property var _leftWidgetOrder: [
```
blammit, line 17: Could you add a brief comment for the property to say why the order matters? In case we need to add more widgets later on.

blammit (approved):
Great, I'm relieved it's a simple solution 😁

pages/OverviewPage.qml
DanielMcInnes: Done

PR #1420, opened 2024-08-20
Use 'flat' ListItem instead of ControlValue-type items

### LINKED ISSUE #1419: Use ListItem components in control cards
The control cards use ControlValue.qml and its derived components, which don't currently support the read/write access control like the settings ListItem components. Investigate whether the control cards can use ListItem and derived components instead; this also avoids having a separate set of UI components for the control cards that need to be maintained separately.

blammit:
Add 'flat' property to ListItem. This adjusts the padding and font size slightly to match the appearance of the controls in the control cards, and hides the background rectangle.

This allows ListItem to be used from the control cards, instead of using ControlValue and the components that extend it, allowing the control cards UI to use the same read/write access controls as those in ListItem, and reducing the number of custom controls that need to be maintained.

Fixes #1419


pages/controlcards/SwitchesCard.qml
```diff
@@ -15,27 +15,31 @@ ControlCard {
...
 
-	ListView {
+	Column {
 		id: switchesView
 
 		anchors {
-			top: parent.top
-			topMargin: Theme.geometry_controlCard_mediumItem_height
+			top: root.title.bottom
+			topMargin: Theme.geometry_controlCard_status_bottomMargin
```
DanielMcInnes, line 23: `Theme.geometry_controlCard_status_bottomMargin` is only every used as a topMargin

DanielMcInnes (approved):
1 nit, lgtm

pages/controlcards/SwitchesCard.qml
DanielMcInnes: never mind, I see what this means now

PR #1423, opened 2024-08-22
Sort BaseDeviceModel by device name by default

### LINKED ISSUE #1299: Sort device lists in alphabetical order
As reported:

"In GUI V2, when you click on items in Overview page and get a list of devices, could these be sorted alphabetically.. Examples of Solar and Batteries.. Turning these into an alphabetical list makes life easier."

Some device lists are already sorted alphabetically, but not all of them, so gui-v2 should do this more consistently, and sort all user-visible device lists alphabetically. Currently AggregateDeviceModel can be sorted alphabetically but not BaseDeviceModel, which creates this confusion.

Note the inverter/charger and generator models are special as gui-v2 needs to find the device with the lowest Device Instance in these cases. (Note, the inverter/charger list shown in the overview drilldowns is already sorted alphabetically.)

[Edit: tanks and temp sensors in the Levels page should be sorted as well.]

blammit:
All models that are used for displaying device lists should be sorted by device name. Only the inverter/charger and generator models need to be sorted solely by device instance, as this is used to determine the "first" or "main" inverter/charger and generator.

Fixes #1299

### DIFF
--- a/src/basedevicemodel.cpp
+++ b/src/basedevicemodel.cpp
@@ -165,6 +165,24 @@ void BaseDeviceModel::setModelId(const QString &modelId)
 	}
 }
 
+void BaseDeviceModel::setSortBy(int sortBy)
+{
+	if (count() > 0) {
+		qmlInfo(this) << "Changing sort order after initialization is not supported";
+		return;
+	}
+
+	if (m_sortBy != sortBy) {
+		m_sortBy = sortBy;
+		emit sortByChanged();
+	}
+}
+
+int BaseDeviceModel::sortBy() const
+{
+	return m_sortBy;
+}
+
 QVariant BaseDeviceModel::data(const QModelIndex &index, int role) const
 {
 	const int row = index.row();
@@ -215,6 +233,8 @@ bool BaseDeviceModel::addDevice(BaseDevice *device)
 	m_devices.insert(index, device);
 	emit endInsertRows();
 	emit countChanged();
+
+	connect(device, &BaseDevice::nameChanged, this, &BaseDeviceModel::deviceNameChanged);
 	refreshFirstObject();
 
 	return true;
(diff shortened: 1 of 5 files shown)


src/basedevicemodel.h
```diff
@@ -91,14 +91,23 @@ class BaseDeviceModel : public QAbstractListModel
 		DeviceRole = Qt::UserRole
 	};
 
+	enum SortBy {
+		SortByDeviceName = 0x1,
```
DanielMcInnes, line 95: is 'SortByDeviceName' also in ascending order?

src/basedevicemodel.cpp
```diff
@@ -290,18 +318,76 @@ BaseDevice *BaseDeviceModel::deviceAt(int index) const
...
+			// Move the entry to be immediately before this item.
+			toIndex = i > fromIndex ? i - 1 : i;
+			break;
+		}
+	}
+
+	if (fromIndex != toIndex) {
+		const int destIndex = toIndex > fromIndex ? toIndex + 1 : toIndex;
+		beginMoveRows(QModelIndex(), fromIndex, fromIndex, QModelIndex(), destIndex);
+		m_devices.move(fromIndex, toIndex);
```
DanielMcInnes, line 352: I am not familiar with QVector::move(). If we have 3 devices, `["charlie", "echo", "golf"]`, and "charlie" changes to "zulu", do we end up with:
`[ null, "echo", "golf", "zulu"]`
or:
`[ "echo", "golf", "zulu"]` ?

DanielMcInnes (commented):
Do you think BaseDeviceModel should have unit tests?

DanielMcInnes (approved):

src/basedevicemodel.h
blammit: Yes, also ascending.

PR #1433, opened 2024-09-02
Capitalize "VRM Portal" list item text

blammit:
This aligns with the capitalization used in gui-v1.


blammit:
@dirkjanfaber do I need to do anything with the translations here? Or is it enough to just update the translation source?

DanielMcInnes (approved):

PR #1434, opened 2024-09-02
Show text validation notification in non-error cases

blammit:
If notification text is passed to validationResult() when validating text input, then show this notification regardless of whether the validation succeeds or fails, instead of only showing it when validation fails.

This is useful for showing notification text when the input text is adjusted, for example.

Update the demo components page to demonstrate this.


DanielMcInnes (approved):
lgtm

PR #1436, opened 2024-09-04
Text input validation improvements

blammit:


blammit:
I'm wondering whether `InputValidation_Result_Warning` confuses the API a bit, in terms of adding too many options. An alternative is to add an argument to `Utils.validationResult()` to specify the notification type, but I think it's preferable for ListTextField to automatically show the correct type of notification, than to add another configuration parameter.

DanielMcInnes (approved):
lgtm

DanielMcInnes:
> I'm wondering whether `InputValidation_Result_Warning` confuses the API a bit, in terms of adding too many options. An alternative is to add an argument to `Utils.validationResult()` to specify the notification type, but I think it's preferable for ListTextField to automatically show the correct type of notification, than to add another configuration parameter.

I don't think adding `InputValidation_Result_Warning` makes the API confusing, it seems fine to me

components/Utils.js
```diff
@@ -329,6 +329,7 @@ function toHexFormat(n) {
 // Returns an object to return for text field validation. The status can be:
 // - InputValidation_Result_Error: the input text is invalid; highlight the input with a red border
 // - InputValidation_Result_OK: the input text is valid, and can be saved, or red highlight can be removed
+// - InputValidation_Result_Warning: same as InputValidation_Result_OK, but show a warning (rather than info) notification
```
chriadam, line 332: this comment seems to suggest that InputValidation_Result_OK will cause an Info notification to be shown.  Is that the case?

chriadam: Ah, I see, that was added in 872b47990f04832976e817e822714de6f09f14d1.  Maybe the comment for InputValidation_Result_OK can be expanded to say: "An info toast notification will be raised to inform the user that validation succeeded" or something?

chriadam: Ah, but the OK was changed to Warning in this commit.  So, in the InputValidation_Result_OK case now, there should be no toast notifications raised, I believe.  If this is indeed the case, I would simply change this comment to read "but show a Warning-styled toast notification", perhaps.

chriadam (approved):
One nitpick, aside from that LGTM

components/Utils.js
blammit: If `notificationText` is set, then a notification will still be shown, but it's optional. I've updated the docs to try to make it more clear about how  notifications are handled.

Issue #1445, opened 2024-09-06
No new firmware found spurious message


mpvader:
Easy to reproduce on a real GX device:

![IMG_2906](https://github.com/user-attachments/assets/c8d1ab70-752d-49e5-89bc-0c9df77ea1fe)

low prio since just a bit sloppy / confusing to the user.

chriadam:
Fixed.

PR #1452, opened 2024-09-11
Wait for valid available version value during update check

chriadam:
Previously, we assumed the valid available version value would arrive before the application returned to the event loop.

In fact, it might arrive asynchronously, depending on when the value becomes available in the backend.

This commit ensures that we wait for a valid value after starting an update check (with a timeout in case no valid value ever becomes available).

Fixes issue #1445


DanielMcInnes (approved):
lgtm.

One nit:
FirmwareUpdate.qml:
`property bool checkingForUpdate` -> `readonly property alias checkingForUpdate: _updateCheckTimeout.running`

PR #1459, opened 2024-09-15
Fix text input visibility on wasm mobile

blammit:


src/main.cpp
```diff
@@ -249,6 +249,13 @@ EM_JS(void, setContentEditable, (bool editable), {
 	}
 });
 
+EM_JS(bool, hasNativeVirtualKeyboard, (), {
+	// Not the best way to test for whether gui-v2 is running in a mobile browser that has its own
+	// VKB, but once QTBUG-128406 is fixed, the wasm keyboard handler can be removed.
+	return /Android|iPad|iPhone/i.test(navigator.userAgent)
+		   // Newer iPads use a Macintosh user agent, so need a hacky test here.
+		|| (/Macintosh/i.test(navigator.userAgent) && navigator.maxTouchPoints && navigator.maxTouchPoints > 1);
```
chriadam, line 257: does this not catch e.g. Mac Book Pro with multi touch trackpad?

ApplicationContent.qml
```diff
@@ -128,22 +128,24 @@ Item {
...
+	// Keyboard handling:
+	// - On CerboGX/EkranoGX devices, show the Qt VKB using components/InputPanel.qml.
+	// - On Wasm mobile, the mobile OS shows the native VKB. If in landscape orientation, use
+	//   components/WasmVirtualKeyboardHandler.qml to auto-scroll flickables and ensure the active
+	//   text field is not obscured by the native VKB.
+	// - On Wasm desktop and desktop platforms, no special input handling is required, as the
+	//   hardware keyboard can be used for text input.
+	//
+	// Note this Loader is the top-most layer, to allow the idleModeMouseArea beneath to call
+	// testCloseOnClick() when clicking outside of the focused text field, to auto-close the Qt VKB.
```
DanielMcInnes, line 140: I think `testCloseOnClick()` has been renamed `acceptMouseEvent()`

components/WasmVirtualKeyboardHandler.qml
```diff
@@ -0,0 +1,141 @@
...
+
+// This handler increases the space below a flickable area whenever a text field within the
+// flickable is focused, if the text field is near the bottom of the flickable, and would otherwise
+// be obscured by the native virtual keyboard.
+//
+// This is required on Qt for Wasm on mobile browsers, as Qt.inputMethod does not provide the
+// geometry of the native VKB in this instance (see QTBUG-128406) so it is not possible to resize
+// and scroll flickables appropriately. So, the safest bet is to just scroll enough to show the
+// text field at the very top of the flickable, as the native VKB may almost fill the entire screen
+// on a mobile device in landscape orientation.
```
chriadam, line 17: funnily enough, on my Android phone, in landscape mode it puts the vkb in a bubble window which covers roughly the whole right hand side of the screen, which ends up obscuring everything no matter the scroll position.

anyway, let's ignore this edge case, as you can still type and enter data etc as required in that case, and in most cases your solution works nicely.

chriadam (approved):
LGTM

src/main.cpp
blammit: It probably does, but not sure of a better solution. Worst case is that the flickable will be unnecessarily auto-scrolled, but still be usable.

ApplicationContent.qml
blammit: Fixed, thanks

components/WasmVirtualKeyboardHandler.qml
blammit: Ugh. If it resizes the browser in that case then we could probably use the window innerWidth to adjust things, but would be nice to just use Qt.inputMethod instead, if the Qt bug gets fixed.

ApplicationContent.qml
blammit: Oops, fixed

PR #1471, opened 2024-09-16
Add QR code to Tailscale page

mr-manuel:
Enable QR code on Tailscale page after https://github.com/victronenergy/gui-v2/issues/1350#issuecomment-2330715682 was solved.

### DIFF
--- a/pages/settings/PageSettingsTailscale.qml
+++ b/pages/settings/PageSettingsTailscale.qml
@@ -5,8 +5,7 @@
 
 import QtQuick
 import Victron.VenusOS
-// NOTE: wait for https://github.com/victronenergy/gui-v2/issues/1350 to be merged
-// import QZXing
+import QZXing
 
 Page {
 	id: root
@@ -172,8 +171,7 @@ Page {
 				horizontalAlignment: Text.AlignHCenter
 			}
 
-			// NOTE: wait for https://github.com/victronenergy/gui-v2/issues/1350 to be merged
-			/*Rectangle {
+			Rectangle {
 				id: qrCodeRect
 				color: Theme.color_page_background
 				width: 200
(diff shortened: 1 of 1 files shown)


mr-manuel:
Tested. The QR code seems a bit unsharp on a bigger screen. Should be ok on a normal GX device screen.

Bildschirmfoto am 2024-09-16 um 15 58 50

grafik

DanielMcInnes:
> Tested. The QR code seems a bit unsharp on a bigger screen. Should be ok on a normal GX device screen.
> 
> Bildschirmfoto am 2024-09-16 um 15 58 50
> 
> grafik

@mr-manuel , if you change sourceSize.width and sourceSize.height to match actual width & height, the QR code becomes sharp

DanielMcInnes:
This all works nicely once tailscale is installed via `curl -s https://raw.githubusercontent.com/mr-manuel/venus-os_TailscaleGX/main/download-and-install.sh | bash`. However, if tailscale is not installed, the user is stuck on this screen:

image


The message just says 'Starting...', there is nothing to indicate that it will never start because tailscale is not installed.
How is this meant to work? In the future, will tailscale be installed by default via firmware update?

mr-manuel:
The Tailscale entry should not be shown, if you don‘t have Tailscale installed. This service still being integrated in Venus OS and will be installed by default in the future.

https://github.com/victronenergy/gui-v2/blob/398edf6103f9d41a3aa933a286dc797d66672bce/pages/settings/PageSettingsServices.qml#L29-L42

mr-manuel:
> @mr-manuel , if you change sourceSize.width and sourceSize.height to match actual width & height, the QR code becomes sharp
How do I achieve that, since the GUI is scaled? In the source code the QR code should always be 200x200 but is upscaled if the GUI has more space.

DanielMcInnes:
> > @mr-manuel , if you change sourceSize.width and sourceSize.height to match actual width & height, the QR code becomes sharp
> > How do I achieve that, since the GUI is scaled? In the source code the QR code should always be 200x200 but is upscaled if the GUI has more space.

Just make it 200x200, and it will be fine

mr-manuel:
Fixed, thanks. You can merge it.

DanielMcInnes (approved):

PR #1487, opened 2024-09-18
Don't display heatsink temperature when value not populated

### LINKED ISSUE #1481: DC gensets - minor issues
Heatsink temperature present when value not populated:
image

BMS Settings -> Charge voltage is missing the voltage unit:
image

DanielMcInnes:
Display 'V' suffix for the charge voltage.
Fixes #1481

### DIFF
--- a/components/PageGensetModel.qml
+++ b/components/PageGensetModel.qml
@@ -115,13 +115,13 @@ ObjectModel {
 			return errorCodes.length ? errorCodes : CommonWords.none_errors
 		}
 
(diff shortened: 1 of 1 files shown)


blammit (approved):
Maybe also rename the `dataItem` id to something else, to avoid this problem cropping up again in the future?

DanielMcInnes:
there are other instances of this same issue in this file, I'll need to fix and retest

DanielMcInnes:
> Maybe also rename the `dataItem` id to something else, to avoid this problem cropping up again in the future?

done

PR #1489, opened 2024-09-19
Add support for Essential/Critical Loads    

blammit:
For the Overview page, the system can be configured to show either the combined AC Loads, or separate Essential Loads and (non-essential) AC Loads.

Prior to this change, the AC Loads box was always showing the combined AC loads. That is, it showed the measurements at `com.victronenergy.system/Ac/Consumption`, like this:

image

Now, when the loads are separated into essential and non-essential, the AC Loads box shows the measurements at `com.victronenergy.system/Ac/ConsumptionOnInput`, and the Essential Loads shows those at `com.victronenergy.system/Ac/ConsumptionOnOutput`.

image

### Testing

Essential Loads can be seen on installations like:    

- https://vrm.victronenergy.com/installation/394636/overview
- https://vrm.victronenergy.com/installation/68572/dashboard    
    
The Essential Loads reported by gui-v2 should match the "Critical Loads" shown in the VRM Portal for each installation:

image
    
For mock testing:

- Press 'q' to toggle the Essential Loads on/off
- I've added four mock configurations to test how the Essential Loads, AC Loads and EVCS widgets are connected in different configurations. 

The EVCS widget's connectors are now more complex, as it can connect to Inverter/Charger, AC Loads or Essential Loads depending on whether the AC loads are combined, and also depending on the "Position" configuration of the EV chargers on the system. For example, this is how it looks when there are EV chargers with both "AC in" and "AC out" positions:

image

(See the last four mock configurations in BriefAndOverviewPageConfig.qml for how the `Position` affects the widget connections.)


blammit:
Made some minor API updates but the functionality remains the same.

components/ObjectAcConnection.qml
```diff
@@ -0,0 +1,68 @@
...
+*/
+
+import QtQuick
+import Victron.VenusOS
+
+QtObject {
+	id: root
+
+	property string bindPrefix
+	property string powerKey: "Power"
```
DanielMcInnes, line 13: I don't see powerKey ever get changed, is this necessary?

data/EvChargers.qml
```diff
@@ -151,5 +165,10 @@ QtObject {
 		}
 	}
 
+	readonly property Timer _measurementUpdates: Timer {
+		interval: 1000
+		onTriggered: root._doUpdateTotals()
```
chriadam, line 170: is there some way this can be done reactively rather than once per second?  or, is there a specific reason to coalesce change signals from the previous 1 second?

DanielMcInnes:
For the attached image, what does this mean? Is the EVCS an AC Load, an Essential Load, or both? If both, does that mean the EVCS load gets counted twice?

image

data/EvChargers.qml
chriadam: Ah, I see - there' s a comment explaining this in ObjectAcConnection.qml.  I guess it's the same reason for here?

data/EvChargers.qml
```diff
@@ -13,6 +13,11 @@ QtObject {
 	property real current: NaN
 	property real energy: NaN
 
+	property int acInputPositionCount   // Chargers with input position, i.e. connected to (non-essential) AC loads
```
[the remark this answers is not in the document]
chriadam: I actually prefer the current variable name (with the comment) as it shows that the EV charger is considered an "ac input" due to its position setting.

data/EvChargers.qml
```diff
@@ -13,6 +13,11 @@ QtObject {
...
+	property real acInputPositionPower: NaN // The total power for chargers with an input position
+	property int acOutputPositionCount  // Chargers with output position, i.e. connected to Essential loads
+	property real acOutputPositionPower: NaN // The total power for chargers with an output position
```
chriadam, line 19: So, one question I have: I think I understand these two cases above (i.e. if position is "input" then there will be a connector going from AC Loads to the EV charger; and if position is "output" then there will be a connector going from Essential AC Loads to the EV charger).

But in which case will the connector go directly from the inverter to the EV Charger?

data/SystemLoad.qml
```diff
@@ -10,11 +10,10 @@ QtObject {
 	id: root
 
 	required property string systemServiceUid
+	required property bool splitAcLoads
```
chriadam, line 13: nitpick: split is a verb, so this property name is suboptimal.  But `acLoadsAreSplit` is a bit verbose / unwieldy, so maybe `splitAcLoads` is better after all..

data/EvChargers.qml
chriadam: Ah, is it where "splitAcLoads" is false, so the system doesn't have any essential loads?

chriadam (approved):
Nitpicks, but LGTM.  Great work.

components/ObjectAcConnection.qml
blammit: Not needed, have removed them

data/EvChargers.qml
blammit: Yes, that's the same reason. We should do the same for the PV totals ... will do that separately.

components/ObjectAcConnection.qml
```diff
@@ -0,0 +1,68 @@
...
+	}
+	readonly property VeQuickItem powerL3: VeQuickItem {
+		uid: bindPrefix + "/L3/" + powerKey
+		onValueChanged: root.phases.setValue(2, PhaseModel.PowerRole, value)
+	}
+	readonly property VeQuickItem currentL1: VeQuickItem {
+		uid: bindPrefix + "/L1/" + currentKey
+		onValueChanged: root.phases.setValue(0, PhaseModel.CurrentRole, value)
+	}
+	readonly property VeQuickItem currentL2: VeQuickItem {
```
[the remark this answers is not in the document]
blammit: Only in this file. I've renamed them to start with an underscore.

components/widgets/WidgetConnectorPath.qml
```diff
@@ -42,7 +43,7 @@ Item {
 
 	function reloadPathLayout() {
```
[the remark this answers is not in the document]
blammit: They have to be reset imperatively, otherwise there are too many bindings that update the paths when the overview page is redrawn, and also sometimes the paths don't get recalculated properly ☹️

data/EvChargers.qml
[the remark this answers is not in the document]
blammit: Yes it's related to whether it is considered as AC input or AC output according to its position, and then that could be interpreted by the Overview page as whether it's going to the AC Loads (output) or Essential Loads (input). I think it's cleaner to separate those two things (the "position" and how that appears on the Overview page).

src/phasemodel.cpp
```diff
@@ -0,0 +1,128 @@
...
+
+	endResetModel();
+
+	if (prevModelCount != count()) {
+		emit countChanged();
+	}
+}
+
+QHash<int, QByteArray> PhaseModel::roleNames() const
+{
```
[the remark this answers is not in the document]
blammit: Thanks, done

data/EvChargers.qml
blammit: Yes, that's correct.

data/SystemLoad.qml
blammit: Yeah, it is a verb but I'm not sure of a better name :(

DanielMcInnes (approved):
lgtm

blammit:
> or the attached image, what does this mean? Is the EVCS an AC Load, an Essential Load, or both? If both, does that mean the EVCS load gets counted twice?

This means there are multiple EV chargers, and some of them have the "AC in" position (connect to AC Loads) while others have the "AC out" position (connect to Essential Loads). So the load should only get counted once.

> Is EVCS considered to only ever be a load? I can see this getting more complicated when 'vehicle - to - load' functionality (i.e. the car battery supplies power to the house) becomes more common.

I don't know the answer to that 🤔 

> Also, the connectors circled in pink look a bit funny

This fix is more complicated than I thought. Will need a separate PR for this.

blammit:
There's one more configuration I forgot to add support for -- when there is an AC-in load, but `HasAcOutSystem=0`, so you should still only see the AC Loads in this case, without Essential Loads. Will update the PR to do this.

blammit:
> There's one more configuration I forgot to add support for -- when there is an AC-in load, but HasAcOutSystem=0, so you should still only see the AC Loads in this case, without Essential Loads. Will update the PR to do this.

Done

data/SystemLoad.qml
blammit: Changed this to `showInputLoads` to align with more recent gui-v1 updates that uses this naming.

Issue #1480, opened 2024-09-17
Reconnect does not work


mr-manuel:
When you have opened the WASM build and reboot the GX device, the GUI does not reconnect and the page need to be refreshed.

In the developer console of the browser you see that two retries are made and after an error is thrown.

```
BackendConnection state: Victron::VenusOS::BackendConnection::Reconnecting (2) venus-gui-v2.js:8:18090
Firefox cannot establish a connecton to the server at ws://venus.local/websocket-mqtt. venus-gui-v2.js:8:243746
```

![Bildschirmfoto am 2024-09-17 um 11 01 00](https://github.com/user-attachments/assets/d342b242-c5d7-400f-afbb-5909e3d51e65)

chriadam:
Fixed.  Please let me know if you still see this issue when using v0.3.12.

mr-manuel:
Can confirm that this is fixed.

PR #1497, opened 2024-09-23
Fix bugs in reconnection and page reload due to firmware update

chriadam:
Contributes to issue #1480

### DIFF
--- a/src/backendconnection.cpp
+++ b/src/backendconnection.cpp
@@ -46,9 +46,8 @@ BackendConnection::State BackendConnection::state() const
 
 void BackendConnection::setState(State backendConnectionState)
 {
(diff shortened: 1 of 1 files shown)


DanielMcInnes:
is there an issue associated with this PR?

chriadam:
Yes, check commit message, it's #1480

DanielMcInnes:
posting here, as I can't post comments in the veutil changes:

ve_qitems_mqtt.hpp: 223:  WebSocketDevice now has 3 boolean member variables, mConnecting, mConnected, and mClosing. Can only 1 of these be true at a time? Feels like they should be replaced with a single 'state' variable, some kind of enum, with 3 possible values, Connecting, Connected, and Closing.

DanielMcInnes (approved):

PR #1501, opened 2024-09-24
generators: add option to skip warm up when start is triggered by inverter overload

ReinvdZee:
Also fixes a small bug where the "Control Status" field shows "Stopped" instead of the condition that started the generator. The dbus field `/RunningByConditionCode` must be used to parse the condition name instead of `/RunningByCondition`.

### DIFF
--- a/pages/settings/GeneratorCondition.qml
+++ b/pages/settings/GeneratorCondition.qml
@@ -14,6 +14,7 @@ ListNavigationItem {
 	property string timeUnit: "s"
 	property int decimals: 1
 	property bool startValueIsGreater: true
+	property bool supportsWarmup: false
 	property string name: text
 
 	//% "Use %1 value to start/stop"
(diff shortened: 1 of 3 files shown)


pages/settings/GeneratorCondition.qml
```diff
@@ -131,6 +132,13 @@ ListNavigationItem {
 						dataItem.uid: bindPrefix + "/StopTimer"
 						suffix: root.timeUnit
 					}
+
+					ListSwitch {
+						//% "Skip generator warm-up"
```
ReinvdZee, line 137: Should we turn this into a `CommonWords.something`?

DanielMcInnes: No, we only put things in CommonWords when they are repeated

DanielMcInnes (approved):
lgtm

PR #1504, opened 2024-09-25
Migrate remaining Bluetooth bits that rely on Connman to Venus platform

### LINKED ISSUE #931: Migrate remaining Bluetooth bits that rely on Connman to Venus platform
Wi-Fi and Ethernet settings are migrated (#618) from Connman to Venus platform APIs to e.g. allow configuring settings remotely with WebAssembly build. There is still one thing that relies on Connman directly, the availability of Bluetooth modem is checked from Connman `Connman.technologyList.indexOf("bluetooth") !== -1`. Discussed with Rein that we could instead make `/Settings/Services/Bluetooth` conditional on the presence of Bluetooth. This would allow removing Connman dependency in both GUI v1 and v2. The Bluetooth availability check is used to show/hide Bluetooth and Bluetooth sensor settings pages.

DanielMcInnes:
Fixes #931


blammit (approved):
Can remove `src/connman-api.h` as well, otherwise LGTM.

So nice to get rid of all that code :)

DanielMcInnes:
> connman-api.h

Removed, thanks Bea

PR #1505, opened 2024-09-25
Show toast notification when root password is changed

### LINKED ISSUE #1457: Show toast notification when root password is changed
Align with gui-v1 behavior, and show a toast notification when the root password is changed in settings.

This should be a simple matter of showing a notification from the `saveInput()` function for the text field.

DanielMcInnes:
Fixes #1457

### DIFF
--- a/pages/settings/PageSettingsGeneral.qml
+++ b/pages/settings/PageSettingsGeneral.qml
@@ -227,6 +227,8 @@ Page {
 						var object = {"SetRootPassword": secondaryText}
 						var json = JSON.stringify(object)
 						securityApi.setValue(json)
+						//% "Root password changed to %1"
+						Global.showToastNotification(VenusOS.Notification_Info, qsTrId("settings_root_password_changed_to").arg(secondaryText), 5000)
 					}
 				}
 			}


pages/settings/PageSettingsGeneral.qml
```diff
@@ -227,6 +227,8 @@ Page {
 						var object = {"SetRootPassword": secondaryText}
 						var json = JSON.stringify(object)
 						securityApi.setValue(json)
+						//% "Root password changed to %1"
+						Global.showToastNotification(VenusOS.Notification_Info, qsTrId("settings_root_password_changed_to").arg(secondaryText), 5000)
```
DanielMcInnes, line 231: This shows the following toast notification. I'm not sure it's a good idea to show the new password, but this is what gui-v1 does.

image

blammit: @jepefe okay if gui-v2 shows the root password, as per the existing behaviour in gui-v1?

jepefe: Is the notifications only shown on the Gui-v2 instance that is changing it, or will be shown on all of the instances?
If is only shown in the current instance, then you can keep the password, otherwise just tell "Root password changed"

DanielMcInnes: Yes, it is only shown on the gui-v2 instance that changes the password.

blammit: It will only show on this one.

blammit (approved):

PR #1512, opened 2024-09-27
Phase model updates

blammit:


data/common/EvCharger.qml
```diff
@@ -9,29 +9,44 @@ import Victron.VenusOS
...
 
-		Component.onCompleted: {
-			const properties = [_phase1Power, _phase2Power, _phase3Power]
-			for (let i = 0; i < properties.length; ++i) {
-				const v = properties[i].value
-				append({ name: "L" + (i + 1), power: v === undefined ? NaN : v })
+		readonly property Instantiator _phases: Instantiator {
+			model: 3
+			delegate: QtObject {
+				required property int index
```
DanielMcInnes, line 37: is this property required? Don't instantiator delegates already have an `index` property? Likewise for PvInverter

DanielMcInnes (approved):

data/common/EvCharger.qml
blammit: They do, but the `required` makes this explicit, instead of using the implicitly-provided `index` in the namespace.

I think it'll be nice to use `required` properties more going forward, to make the property references more explicit within delegates, and also to specify the "mandatory" properties for components. We have to do this carefully as the usage causes implicit properties to be removed altogether within delegates, which might break some existing code.

PR #1516, opened 2024-09-30
AC-in settings page improvements

blammit:


pages/settings/devicelist/ac-in/PageAcInModel.qml
```diff
@@ -12,16 +12,28 @@ ObjectModel {
 	property string bindPrefix
 	property int productId
 
-	readonly property var nrOfPhases: VeQuickItem {
+	readonly property VeQuickItem nrOfPhases: VeQuickItem {
 		uid: root.bindPrefix + "/NrOfPhases"
 	}
 
+	// The selected phase, if applicable (e.g. Multi RS in PV inverter only mode, which is assigned
+	// to a specifc phase).
```
DanielMcInnes, line 20: typo

DanielMcInnes (approved):

pages/settings/devicelist/ac-in/PageAcInModel.qml
blammit: Oh, good point, have fixed this to look at `/NrOfPhases` first, as per the gui-v1 commit.

I assume the `phase` could change, but in that case this should respond and update the phase number list.

PR #1517, opened 2024-09-30
Multi RS fixes

blammit:


components/widgets/AcInputWidget.qml
```diff
@@ -32,8 +32,8 @@ AcWidget {
 	}
 
 	onClicked: {
-		if (root.input.serviceType === "multi") {
```
DanielMcInnes, line 35: so is "multi" not a valid service type? Has this been wrong the whole time? https://github.com/victronenergy/venus/wiki/dbus suggests that "multi" is still valid, but I don't know how up to date that is

blammit: It's still valid, but the backend was reworked, and now we show `multi` services in the UI as a subset of [acsystem](https://github.com/victronenergy/venus/wiki/dbus#acsystem ) services, instead of as standalone services. E.g. in cd6d107cc73ea7cbbf67750e72c7fad749b62bfb the Device List entries were reworked to show `acsystem` services instead of `multi` services.

Basically, an `acsystem` service is a manager for one or more Multi RS `multi` devices. We can use `mqtt/acsystem/Devices` to find the `multi` services for that `acsystem`.

DanielMcInnes (approved):

PR #1521, opened 2024-09-30
Fix broken Chinese translations

DanielMcInnes:


DanielMcInnes:
This is not ready to merge yet, I expect this will impact POEditor, it needs Dirk-Jan's input

dirkjanfaber:
Fairly sure that this indeed will mess things up. There are several language codes for Chinese and plain `zh` isn't one of them. (See https://poeditor.com/docs/languages)

_If_ we want to use the plain `zh`, then the script https://github.com/victronenergy/gui-v2/blob/main/scripts/download_from_poeditor.sh needs to be adjusted too. 

E.g. by applying the following (untested) patch:
```
--- download_from_poeditor.sh~	2024-09-30 18:06:23.787387307 +0000
+++ download_from_poeditor.sh	2024-09-30 18:09:28.163510182 +0000
@@ -15,7 +15,12 @@
   exit 1
 fi

-code=${1/_/-}
+download_code=${code}
+
+if [ "${code}" == "zh" ]
+then
+  download_code="zh-cn"
+fi

 output_file=$2
 echo "### Fetching ${code}, storing in ${output_file}"
@@ -23,7 +28,7 @@
   -d api_token="${POEDITOR_TOKEN}" \
   -d id="674443" \
   -d type="ts" \
-  -d language="${code}" |\
+  -d language="${download_code}" |\
    jq -r '.result.url')

 if [ "${download_url}" == "null" ]
@@ -34,4 +39,9 @@
   curl -X GET ${download_url} --output ${output_file}
 fi

+if [ "${code}" == "zh" ]
+then
+  sed -i -e 's/<TS version="2.1" language="zh-CN">/<TS version="2.1" language="zh">/' ${output_file}
+fi
+
 exit 0
 ```
 
 In my opinion this is a bit of a hack to get things working. But it doesn't feel like the best solution.

**Comment by DanielMcInnes** [comment]
Hi @dirkjanfaber , I have updated the PR. Does this look any better? We don't need to change the filename after all, we just need to change the 'language' tag at the top of i18n/venus-gui-v2_zh_CN.ts. If you give me a poeditor token, I think I can patch & test the upload & download scripts

**Comment by dirkjanfaber** [comment]
Hi @DanielMcInnes. This does look way better. Thanks! 

Gave you the needed API key. If you adjust the file (I believe only the download one needs changing) and update the PR, I'll review and approve it.

**Comment by DanielMcInnes** [comment]
@dirkjanfaber , ready for review

File: scripts/download_from_poeditor.sh [hunk 1]
```diff
@@ -34,4 +34,10 @@ else
   curl -X GET ${download_url} --output ${output_file}
 fi
 
+if [ "${code}" == "zh-CN" ]
+then
+  echo "### POEditor uses zh-CN, Qt doesn't recognize this language code, it uses zh_CN. Patching ${output_file}, replacing zh-CN with zh_CN."
+  sed -Ei 's/<TS version=(.*) language="zh-CN">/<TS version=\1 language="zh_CN">/' "${output_file}"
```
DanielMcInnes, line 40: this should cope with changing versions of TS> Currently at v. 2.1, no idea how often they change

dirkjanfaber: I don't think they change often. Personally I would be more inclined to write that as `'s/(<TS version="[0-9\.]+" language=")zh-CN("[^>]*>)/\1zh_CN\2/'`, also allowing for new attributes to be added. But as I have no reason to believe that is going to happen,  I am fine with the current solution too.

dirkjanfaber (approved):
Looks good to me.

PR #1525, opened 2024-10-01
Avoid unnecessary VeQuickItem::value reads

blammit:
If isValid is true, then the value should be usable.

### DIFF
--- a/components/AcPhase.qml
+++ b/components/AcPhase.qml
@@ -10,14 +10,11 @@ QtObject {
 	id: root
 
 	property string serviceUid
-	readonly property real frequency: _frequency.value === undefined ? NaN : _frequency.value
-	readonly property real current: _current.value === undefined ? NaN : _current.value
-	readonly property real voltage: _voltage.value === undefined ? NaN : _voltage.value
-	readonly property real power: _power.value === undefined ? NaN : _power.value
-	readonly property bool valid: !isNaN(frequency)
-								  && !isNaN(current)
-								  && !isNaN(voltage)
-								  && !isNaN(power)
+	readonly property real frequency: _frequency.isValid ? _frequency.value : NaN
+	readonly property real current: _current.isValid ? _current.value : NaN
+	readonly property real voltage: _voltage.isValid ? _voltage.value : NaN
+	readonly property real power: _power.isValid ? _power.value : NaN
+	readonly property bool valid: _frequency.isValid && _current.isValid && _voltage.isValid && _power.isValid
 
 	readonly property VeQuickItem _frequency: VeQuickItem {
 		uid: serviceUid + "/F"
--- a/components/AcPhase.qml
+++ b/components/AcPhase.qml
@@ -14,7 +14,6 @@ QtObject {
 	readonly property real current: _current.isValid ? _current.value : NaN
 	readonly property real voltage: _voltage.isValid ? _voltage.value : NaN
 	readonly property real power: _power.isValid ? _power.value : NaN
-	readonly property bool valid: _frequency.isValid && _current.isValid && _voltage.isValid && _power.isValid
 
 	readonly property VeQuickItem _frequency: VeQuickItem {
 		uid: serviceUid + "/F"
--- a/data/common/Generator.qml
+++ b/data/common/Generator.qml
@@ -9,11 +9,11 @@ import Victron.VenusOS
 Device {
 	id: generator
 
-	readonly property int state: _state.value === undefined ? -1 : _state.value
+	readonly property int state: _state.isValid ? _state.value : -1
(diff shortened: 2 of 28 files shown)


data/common/Generator.qml
```diff
@@ -9,11 +9,11 @@ import Victron.VenusOS
 Device {
 	id: generator
 
-	readonly property int state: _state.value === undefined ? -1 : _state.value
+	readonly property int state: _state.isValid ? _state.value : -1
 	readonly property bool autoStart: _autoStart.value === 1
-	readonly property int manualStartTimer: _manualStartTimer.value === undefined ? -1 : _manualStartTimer.value
+	readonly property int manualStartTimer: _manualStartTimer.isValid ? _manualStartTimer.value : 0
```
DanielMcInnes, line 14: Intentional behaviour change?

components/AcPhase.qml
```diff
@@ -10,14 +10,11 @@ QtObject {
...
-	readonly property real power: _power.value === undefined ? NaN : _power.value
-	readonly property bool valid: !isNaN(frequency)
-								  && !isNaN(current)
-								  && !isNaN(voltage)
-								  && !isNaN(power)
+	readonly property real frequency: _frequency.isValid ? _frequency.value : NaN
+	readonly property real current: _current.isValid ? _current.value : NaN
+	readonly property real voltage: _voltage.isValid ? _voltage.value : NaN
+	readonly property real power: _power.isValid ? _power.value : NaN
+	readonly property bool valid: _frequency.isValid || _current.isValid || _voltage.isValid || _power.isValid
```
DanielMcInnes, line 17: behaviour change?

data/common/Generator.qml
```diff
@@ -9,11 +9,11 @@ import Victron.VenusOS
...
 	readonly property int runtime: _runtime.value || 0
-	readonly property int runningBy: _runningBy.value === undefined ? -1 : _runningBy.value
+	readonly property int runningBy: _runningBy.isValid ? _runningBy.value : 0
```
DanielMcInnes, line 16: Intentional behaviour change?

data/common/PulseMeter.qml
```diff
@@ -9,7 +9,7 @@ import Victron.VenusOS
 Device {
 	id: pulseMeter
 
-	readonly property real aggregate: _aggregate.value === undefined ? -1 : _aggregate.value
+	readonly property real aggregate: _aggregate.isValid ? _aggregate.value : NaN
```
DanielMcInnes, line 12: Intentional behaviour change?

components/AcPhase.qml
blammit: Yes, I think the original intention was to have valid = "any value is valid", so it should be the desired outcome here

data/common/Generator.qml
blammit: Yes, the -1 was a bit strange for a "timer" value so I changed it to 0.

data/common/Generator.qml
blammit: Yes, similar to the other case, the -1 was a bit strange for a "timer" value so I changed it to 0.

data/common/PulseMeter.qml
blammit: Yes, this should have been NaN as -1 doesn't make sense for the real.

components/AcPhase.qml
chriadam: the original code seems to require "all values are valid" though?

data/common/Generator.qml
chriadam: all uses seem to do: `if (root.generator.manualStartTimer > 0)` so, either 0 or -1 is fine.  I kind of like using -1 for all non-valid integer properties, and NaN for all non-valid real properties, but I have no strong opinion.

chriadam:
Aside from the open question about AcPhase.valid, LGTM.

components/AcPhase.qml
blammit: Oh, sorry. Yes, this is a behaviour change, but if a phase only has some of these values set, it seems it should still be considered valid.

blammit: It seems like this `valid` property is not actually used, so I'll remove it.

data/common/PulseMeter.qml
blammit: Updated the commit message so that this behavioural change is noted.

data/common/Generator.qml
blammit: Updated the commit message so that this behavioural change is noted.

DanielMcInnes (approved):

PR #1526, opened 2024-10-01
Add setting to hide input loads on overview

### LINKED ISSUE #1460: add setting to hide input loads on overview
as per gui-v1 commit 7cecfbf09dd5899bca7a9c76fd16ed12ee55f332

blammit:
This is to be used in systems with AC-coupled PV on the input, or systems with a grid meter used for energy accounting, but with no actual loads on the input side. It hides the AC input side loads on the overview.

As per 7cecfbf09dd5899bca7a9c76fd16ed12ee55f332 from gui-v1.

Fixes #1460

### DIFF
--- a/pages/settings/PageSettingsSystem.qml
+++ b/pages/settings/PageSettingsSystem.qml
@@ -84,6 +84,28 @@ Page {
 			}
 
 			ListRadioButtonGroup {
+				//% "Position of AC loads"
+				text: qsTrId("settings_system_ac_position")
+				dataItem.uid: Global.systemSettings.serviceUid + "/Settings/SystemSetup/HasAcInLoads"
+				optionModel: [
+					{
+						//% "AC input & output"
+						display: qsTrId("settings_system_ac_input_and_output"),
+						//% "Use this option when AC-loads are present on the input of the Inverter/Charger. Use this option if unsure."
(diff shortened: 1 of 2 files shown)


pages/settings/PageSettingsSystem.qml
```diff
@@ -83,6 +83,28 @@ Page {
...
 
+			ListRadioButtonGroup {
+				//% "Position of AC loads"
+				text: qsTrId("settings_system_ac_position")
+				dataItem.uid: Global.systemSettings.serviceUid + "/Settings/SystemSetup/HasAcInLoads"
+				optionModel: [
+					{
+						//% "AC input & output"
+						display: qsTrId("settings_system_ac_input_and_output"),
+						//% "Use this option when AC-loads are present on the input of the Inverter/Charger. Use this option if unsure."
```
blammit, line 94: The gui-v1 commit shows this text in a notification after the option is clicked, but I think if we can, it's nicer to show it as a caption so that you can see the explanation before you choose the option.

blammit:
Note that changes to this setting don't affect the UI directly. Instead, it will affect how systemcalc reports the AC in/out loads.

DanielMcInnes (approved):

PR #1530, opened 2024-10-01
Make 'Tank sensor' value read-only in tank setup menu

### LINKED ISSUE #1519: Tank input: Sensor value writable
The "Setup" menu of a tank input (GX Tank 140 or built-in) displays the current sensor value:

image

The sensor value field should be read-only, but its writable as well:
image

The field will be set to the new value and changes back to the actual sensor value after a second. This bug occurs for all sensor types
This bug was reported on [community](https://community.victronenergy.com/t/bug-sensor-value-a-writeable-field-in-v3-50-24/5179)

This bug can be reproduced on a GX device without any sensor or resistors attached. The status of the sensor will be "open circuit" but the sensor value is still writable (Cerbo GX built-in Tank input): 
image

DanielMcInnes:
Fixes #1519

### DIFF
--- a/pages/settings/devicelist/tank/PageTankSetup.qml
+++ b/pages/settings/devicelist/tank/PageTankSetup.qml
@@ -118,13 +118,12 @@ Page {
 				suffix: "s"
 			}
 
-			ListSpinBox {
+			ListTextItem {
 				//% "Sensor value"
 				text: qsTrId("devicelist_tanksetup_sensor_value")
 				dataItem.uid: root.bindPrefix + "/RawValue"
 				allowed: defaultAllowed && dataItem.isValid
-				suffix: rawUnit.value || ""
-				decimals: 1
+				secondaryText: dataItem.isValid ? Units.formatNumber(dataItem.value, 1) + (rawUnit.value || "") : "--"
 			}
 
 			ListNavigationItem {


chriadam (approved):

DanielMcInnes:
image

pages/settings/devicelist/tank/PageTankSetup.qml
```diff
@@ -118,13 +118,12 @@ Page {
...
 
-			ListSpinBox {
+			ListTextItem {
 				//% "Sensor value"
 				text: qsTrId("devicelist_tanksetup_sensor_value")
 				dataItem.uid: root.bindPrefix + "/RawValue"
 				allowed: defaultAllowed && dataItem.isValid
-				suffix: rawUnit.value || ""
-				decimals: 1
+				secondaryText: dataItem.value === undefined ? "--" : Units.formatNumber(dataItem.value, 1) + (rawUnit.value || "")
```
DanielMcInnes, line 126: The unit for this item comes from a string, so we can't use ListQuantityItem/Field

blammit: Yeah I see, we don't support adding a freeform unit string at the moment in QuantityLabel/quantityinfo. I suppose we can add that later on if this comes up more regularly.

Could you use `dataItem.isValid` instead of `dataItem.value === undefined`? Otherwise LGTM.

blammit (approved):

pages/settings/devicelist/tank/PageTankSetup.qml
DanielMcInnes: Done, thanks Bea

Issue #1462, opened 2024-09-16
add Node-RED submenu with factory reset


DanielMcInnes:
as per gui-v1 commit [c2f0e33d3da7abc0b865f9125f542291a722ded1
](https://github.com/victronenergy/gui/commit/c2f0e33d3da7abc0b865f9125f542291a722ded1)

    Adds option to reset node-red from the user interface.
    User is asked for confirmation before the reset.

mpvader:
To test, set the Image type (in the settings -> firmware menu) to Large, and then update.

and otherwise Rein can help to explain as usual

chriadam:
Fixed

PR #1536, opened 2024-10-02
Add Node-RED submenu with factory reset

chriadam:
Fixes issue #1462

### DIFF
--- a/pages/settings/PageSettingsNodeRed.qml
+++ b/pages/settings/PageSettingsNodeRed.qml
@@ -1,0 +1,61 @@
+/*
+** Copyright (C) 2024 Victron Energy B.V.
+** See LICENSE.txt for license information.
+*/
+
+import QtQuick
+import Victron.VenusOS
+
+Page {
+	id: root
+
+	GradientListView {
+		model: ObjectModel {
+			ListRadioButtonGroup {
+				id: nodered
+
+				text: qsTrId("settings_large_node_red")
+				dataItem.uid: Global.venusPlatform.serviceUid + "/Services/NodeRed/Mode"
+				allowed: dataItem.isValid
+				optionModel: [
+					{ display: CommonWords.disabled, value: VenusOS.NodeRed_Mode_Disabled },
+					{ display: CommonWords.enabled, value: VenusOS.NodeRed_Mode_Enabled },
+					//% "Enabled (safe mode)"
+					{ display: qsTrId("settings_large_enabled_safe_mode"), value: VenusOS.NodeRed_Mode_EnabledWithSafeMode },
+				]
+			}
+
+			ListButton {
+				id: resetButton
+				//% "Node-RED factory reset"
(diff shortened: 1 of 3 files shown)


chriadam:
@ReinvdZee any chance you could help me test this one?  I did: Settings -> Firmware -> Online Updates -> Image Type = Large, then performed a firmware update.  Now the Settings page has a "Venus OS Large Features" entry, and I can see the Node Red option and interact with it.  But I'm not sure if it's doing what it's supposed to do, since there's no feedback after I trigger the reset.

pages/settings/PageSettingsNodeRed.qml
```diff
@@ -0,0 +1,61 @@
...
+			ListButton {
+				id: resetButton
+				//% "Node-RED factory reset"
+				text: qsTrId("page_settings_nodered_factory_reset")
+				secondaryText: CommonWords.press_to_reset
+				allowed: true
+				onClicked: Global.dialogLayer.open(factoryResetConfirmationDialogComponent)
+
+				VeQuickItem {
+					id: nodeRedFactoryResetItem
```
DanielMcInnes, line 37: can this be moved into factoryResetConfirmationDialogComponent?

DanielMcInnes (approved):

ReinvdZee:
@chriadam 

There is currently no feedback from venus-platform, other than node-red being disabled (`/Services/NodeRed/Mode` set to 0), and everything under `/data/home/nodered/.node-red/node_modules` deleted.

Instead of writing `1` to `/Services/NodeRed/FactoryReset`, the GUI should write the last 4 digits of the device's unique ID (`com.victronenergy.platform/Device/UniqueId`) to this field. This is to prevent resetting it easily by hand, from `dbus-spy` for example.

See [my comment](https://github.com/victronenergy/venus/issues/1293#issuecomment-2233237447) on the original issue.

pages/settings/PageSettingsNodeRed.qml
chriadam: the dialog will be destroyed after being accepted, and I'd prefer not to rely on VeQuickItem doing the write synchronously for it to succeed.

chriadam:
Thanks @ReinvdZee , updated PR

ReinvdZee:
@chriadam,

I've just tested this and it does not work. I just found out that although we initially agreed to write the last 4 digits of the unique ID to `/Services/NodeRed/FactoryReset`, the gui needs to write just a `1` to it to initiate the reset. So it probably was correct the first time. So please revert back to writing just a `1` instead of the 4 last digits of the unique ID.

Apologies for the confusion.

chriadam:
No problem at all, will revert

chriadam:
Hopefully I did that correctly.  If someone could sanity check for me, that'd be greatly appreciated.

ReinvdZee:
Tested again, works fine!

PR #1548, opened 2024-10-09
Improve Generator card display of running state and runtime

### LINKED ISSUE #1432: Improve Generator card display of running state and runtime
The current generator card looks like this:

image

There are two issues here:

- The "Running by SOC condition" text is longer than what's desired in the design
- The runtime should be clear about whether it's showing hours-minutes or minutes-seconds. When under a minute, it's clear that it's seconds because the number will change every second, but the UI should be more explicit about this.

Updated design from Serj is like this:

image
image

So the tasks to be done are:

- [x] Update the generator card to show the generator state and condition, separated by a dot if necessary. If it was auto-started (i.e. any start condition other than "Manual start" or "Test run") then show "Auto-started" as the state text.
- [x] Update the runtime format in the generator card: show a "s" suffix when the runtime is less than a minute, otherwise show the "hh:mm" format without a suffix.
- [x] Similarly update the runtime format in the generator start/stop dialog. (This is the dialog that is shown when "manual stop" is pressed.)

DanielMcInnes:
Fixes #1432

### DIFF
--- a/components/Utils.js
+++ b/components/Utils.js
@@ -129,6 +129,12 @@ function formatAsHHMMSS(seconds, showUnits) {
 	return s
 }
 
+function formatGeneratorRuntime(seconds) {
+	return seconds < 60
+			? secondsToString(seconds)
+			: formatAsHHMM(seconds / 60)
+}
+
 function reactToSignalOnce(sig, slot) {
 	var f = function() {
 		slot.apply(this, arguments)
--- a/data/Generators.qml
+++ b/data/Generators.qml
@@ -22,8 +22,28 @@ QtObject {
 		model.removeDevice(generator.serviceUid)
 	}
 
-	function stateToText(state, conditionCode) {
+	function reset() {
+		model.clear()
+	}
+
+	function stateAndCondition(state, conditionCode) {
 		switch (state) {
+		case VenusOS.Generators_State_WarmUp:
+		case VenusOS.Generators_State_CoolDown:
+		case VenusOS.Generators_State_Stopping:
+		case VenusOS.Generators_State_Error:
+			return stateText(state)
+		}
+
+		return runningByText(conditionCode)
+	}
+
+	function stateText(state) {
+		switch (state) {
+		case VenusOS.Generators_State_Stopped:
+			return CommonWords.stopped
+		case VenusOS.Generators_State_Running:
+			return CommonWords.running_status
 		case VenusOS.Generators_State_WarmUp:
 			//% "Warm-up"
 			return qsTrId("page_generator_warm_up")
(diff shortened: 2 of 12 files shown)


data/Generators.qml
```diff
@@ -70,8 +89,62 @@ QtObject {
...
+			return CommonWords.running_status
+		}
+
+		return stateToText(state, conditionCode)
+	}
+
+	function autoStartReason(conditionCode) {
+		switch (conditionCode) {
+		case VenusOS.Generators_RunningBy_LossOfCommunication:
+			//: Translations should not be longer than the longest engineering text in this list.
```
blammit, line 118: It might be difficult to keep to this text length limit in some languages. I think it's fine to have longer text and wrap it -- Serj was more concerned about the text wrapping with the larger font size, but now that you've added the second line of smaller "Auto-started" text, it would be okay to wrap this smaller line of text.

Also could you capitalize the first letter, e.g. "Loss of communication"?

pages/controlcards/GeneratorCard.qml
```diff
@@ -28,11 +28,28 @@ ControlCard {
...
+	Label {
+		id: autoStarted
+
+		anchors {
+			top: root.status.bottom
+			topMargin: Theme.geometry_controlCard_status_topMargin
+			left: parent.left
+			leftMargin: Theme.geometry_controlCard_contentMargins
+		}
+		//% "Auto-started • %1"
```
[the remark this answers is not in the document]
DanielMcInnes: Updated as suggested. 
Can you explain why using \u2022 is better than using "•" ?

data/Generators.qml
DanielMcInnes: Done

pages/controlcards/GeneratorCard.qml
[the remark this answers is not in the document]
chriadam: it's not clear which unicode character is being specified, if you paste it, since there are multiple codepoints which have a similar look.

components/Utils.js
```diff
@@ -129,6 +129,12 @@ function formatAsHHMMSS(seconds, showUnits) {
 	return s
 }
 
+function formatGeneratorRuntime(seconds) {
+	return seconds < 60
+			? secondsToString(seconds)
+			: formatAsHHMMSS(seconds / 60)
```
chriadam, line 135: should this be formatAsHHMM(seconds/60) ?

pages/controlcards/GeneratorCard.qml
```diff
@@ -28,11 +28,32 @@ ControlCard {
...
+			top: root.status.bottom
+			topMargin: Theme.geometry_controlCard_status_topMargin
+			left: parent.left
+			leftMargin: Theme.geometry_controlCard_contentMargins
+			right: parent.right
+			rightMargin: Theme.geometry_controlCard_contentMargins
+		}
+		text: root.generator.isAutoStarted
+			  ? //% "Auto-started \u2022 %1"
+				qsTrId("controlcard_generator_autostarted").arg(root.generator.runningByText)
```
blammit, line 44: Indentation is off?

pages/settings/PageGenerator.qml
```diff
@@ -79,7 +79,7 @@ Page {
 
 			text: CommonWords.state
 			allowed: root.startStopBindPrefix === root.generator0ServiceUid
-			secondaryText: activeCondition.isValid ? Global.generators.stateToText(generatorState.value, activeCondition.value) : '---'
+			secondaryText: activeCondition.isValid ? Global.generators.stateAndCondition(generatorState.value, activeCondition.value) : '---'
```
blammit, line 82: Now the page shows this when the generator is running with a certain condition:

image

which doesn't indicate clearly that it is running, unlike "Running by SOC condition". I think we should either show that full text here, or to avoid duplicate/similar translations, just show the same text as the generator card, e.g. "Auto-started [dot] SOC condition".

components/Utils.js
DanielMcInnes: updated as suggested

pages/controlcards/GeneratorCard.qml
DanielMcInnes: It looks fine in qtcreator. If I auto-indent the code, this is what it gives me. I think it's just how it displays in github

pages/settings/PageGenerator.qml
DanielMcInnes: Updated to show that same text as the generator card

blammit: I'm seeing just the condition number, like this:

image

DanielMcInnes: Whoops, fixed

blammit (approved):

pages/settings/PageGenerator.qml
blammit: LGTM :)

Issue #1554, opened 2024-10-09
Genset total power not displayed on devicelist summary when `/Ac/Power` is not present.


ReinvdZee:
The summary of a connected genset device in the devicelist displays the state and the total power drawn. The model uses the path `/Ac/Power` to fetch the total power. However, some gensets do not support this field, and an invalid power value is displayed:

![image](https://github.com/user-attachments/assets/fea34da0-b049-4fd0-aee3-921952a3661c)

gui-v1 just sums the power values of the individual phases when this field is unavailable. V2 should do the same.

mpvader:
Hey @ReinvdZee when or by what services is this total power missing?

reason to ask it that I prefer to solve this there, rather than have to burden user interface with this.

Either we mandate having /Ac/Power for all things todo with AC; or the opposite: we remove it everywhere.

With everywhere I mean also beyond of the scope of gensets, ie this is also for Multis, chargers, and so forth. At the moment I don’t remember if we are going to the direction of adding it everywhere or the opposite; I’ll ask @izak .

ReinvdZee:
It was decided ([discussion on slack](https://victrondevelopment.slack.com/archives/C7C1ZU6UD/p1729141554596209)) that the drivers need to make sure that the /Ac/Power path is always there. So we won't solve this here.

PR #1566, opened 2024-10-11
Calculate power from phases if  is not available

ReinvdZee:
https://github.com/victronenergy/gui-v2/issues/1554

### DIFF
--- a/pages/settings/devicelist/AcInDeviceModel.qml
+++ b/pages/settings/devicelist/AcInDeviceModel.qml
@@ -17,11 +17,41 @@ BaseDeviceModel {
 		delegate: Device {
 			id: device
 
-			readonly property real power: _power.isValid ? _power.value : NaN
 			readonly property int gensetStatusCode: _gensetStatusCode.isValid ? _gensetStatusCode.value : -1
+
+			readonly property real power: {
+				if (_power.isValid)
+					return _power.value
+
+				var power = 0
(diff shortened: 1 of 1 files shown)


pages/settings/devicelist/AcInDeviceModel.qml
```diff
@@ -17,13 +17,44 @@ BaseDeviceModel {
 		delegate: Device {
 			id: device
 
-			readonly property real power: _power.isValid ? _power.value : NaN
+			readonly property real power: calculatePower()
```
chriadam, line 20: any reason to define a separate function, rather than just doing it as a binding expression?

ReinvdZee: No not in particular. Thought this was more readable. But I can change it.

chriadam: Bindings can be optimised more efficiently by the compiler (at least, it used to be the case, and probably still is).  Given that the function isn't called from any other point, I don't think there's any gain to separating it into a separate function, but there is some (small) benefit to inlining the code into the binding expression.  So that would be my preference - but I don't feel strongly about it either way ;-)

chriadam (approved):

ReinvdZee:
It was decided that the drivers need to make sure that the `/Ac/Power` path is always there. So we won't solve this here.

Issue #1562, opened 2024-10-10
Firmware menu: prevent user from switching to backup firmware if Security profile is set to undetermined


mpvader:
part of https://github.com/victronenergy/venus-private/issues/442.

Jeroen implemented this for gui-v1 and will implement it for gui-v2 as well and make a PR

PR #1568, opened 2024-10-11
a network security profile must be set before downgrading

jhofstee:
If it is not set a user can lock himself out when switching to an older version when switching from VRM.

closes https://github.com/victronenergy/gui-v2/issues/1562

### DIFF
--- a/pages/settings/PageSettingsRootfsSelect.qml
+++ b/pages/settings/PageSettingsRootfsSelect.qml
@@ -18,6 +18,10 @@ Page {
 		uid: Global.systemSettings.serviceUid + "/Settings/System/AutoUpdate"
 	}
 	VeQuickItem {
+		id: securityProfile
+		uid: Global.systemSettings.serviceUid + "/Settings/System/SecurityProfile"
+	}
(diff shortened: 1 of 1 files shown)


pages/settings/PageSettingsRootfsSelect.qml
```diff
@@ -60,6 +64,13 @@ Page {
 				allowed: root._switchingEnabled
 
 				onClicked: {
+					if (securityProfile.value === VenusOS.Security_Profile_Indeterminate) {
```
chriadam, line 67: do we also need to check `securityProfile.isValid`?

blammit: Should be fine without it, if it is already checking the value directly.

blammit (approved):

pages/settings/PageSettingsRootfsSelect.qml
jhofstee: There is no need for that, undefined shouldn't occur in the first place and it won't match either.

Issue #1558, opened 2024-10-10
pvinverter: option to enable limiter if supported


DanielMcInnes:
as per gui-v1 commit c6829d68b3eded79fa3d608e6395d5af53bd9ca1

    pvinverter: option to enable limiter if supported
    
    We're expanding the ability to limit PV-inverters other than
    Fronius and ABB/Fimer. For such inverters, dbus-fronius will
    place a marker value in localsettings that indicate whether
    the setting should be available on this inverter. This allows
    enabling the setting for other inverters, previously not supported.
    
    The setting defaults to enabled.

chriadam:
Fixed.

PR #1569, opened 2024-10-11
pvinverter: option to enable limiter if supported

izak:
https://github.com/victronenergy/gui-v2/issues/1558

### DIFF
--- a/pages/settings/PageSettingsFroniusInverter.qml
+++ b/pages/settings/PageSettingsFroniusInverter.qml
@@ -24,6 +24,12 @@ Page {
 		uid: bindPrefix + "/Phase"
 	}
 
+	VeQuickItem {
+		id: limiterSupportedItem
+
+		uid: bindPrefix + "/LimiterSupported"
+	}
+
 	GradientListView {
 		model: ObjectModel {
 			PvInverterPositionRadioButtonGroup {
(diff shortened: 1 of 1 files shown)


pages/settings/PageSettingsFroniusInverter.qml
```diff
@@ -58,6 +64,17 @@ Page {
 				text: qsTrId("page_settings_fronius_inverter_show")
 				dataItem.uid: bindPrefix + "/IsActive"
 			}
+
+			ListRadioButtonGroup {
+				//% "Power limiting"
+				text: qsTrId("page_settings_fronius_inverter_power_limiting")
+				dataItem.uid: bindPrefix + "/EnableLimiter"
+				allowed: limiterSupportedItem.value
```
blammit, line 72: Maybe should be `allowed: !!limiterSupportedItem.value` to ensure the value is coerced to a bool?

blammit (approved):
Minor comment, LGTM otherwise

pages/settings/PageSettingsFroniusInverter.qml
izak: Let's rather make it `allowed: limiterSupportedItem.value === 1`. I can make the change and rebase/squash it?

blammit: Yes, please go ahead :)

PR #1573, opened 2024-10-14
Add RS system alarms and make VE.Bus alarm fixes

blammit:


pages/vebusdevice/PageVeBusAlarms.qml
```diff
@@ -37,11 +37,21 @@ Page {
...
+
 						text: alarmStatusModel.displayTexts[index]
 						bindPrefix: veBusDevice.serviceUid
-						numOfPhases: multiPhase ? root.numberOfPhases : 1
+						numOfPhases: root.numberOfPhases
+						multiPhase: alarmDelegate.multiPhase
 						alarmSuffix: pathSuffix
-						errorItem: errorItem
+						errorItem: alarmDelegate.errorItem
+						allowed: defaultAllowed && (!showIfMulti || root.isMulti)
```
DanielMcInnes, line 54: is this logic correct? If `isMulti` is true, and `showIfMulti` is false, should we show the alarm?

DanielMcInnes (approved):
lgtm, 1 question

pages/vebusdevice/PageVeBusAlarms.qml
blammit: Yeah it is a bit unclear like this. Perhaps the property should be named `showOnlyIfMulti` instead of `showIfMulti`?

This is supposed to port the behaviour from gui-v1 where there are `show: isMulti` bindings, but these are only set for some items. So we want:

- If `show: isMulti` is set, then `allowed` should depend on `isMulti`
- If this is not set, then we don't care, and can just use `defaultAllowed`.

blammit: Updated to rename to `showOnlyIfMulti`.

pages/settings/devicelist/rs/PageRsSystemAlarms.qml
```diff
@@ -0,0 +1,67 @@
...
+	}
+
+	ObjectModel {
+		id: validAlarmsModel
+
+		VeBusAlarm {
+			id: phaseRotationAlarm
+
+			//% "Phase rotation"
+			text: qsTrId("rs_phase_rotation")
```
chriadam, line 27: nitpick, but I'd prefer ids like: `rssystemalarm_phase_rotation`

pages/settings/devicelist/rs/PageRsSystemAlarms.qml
```diff
@@ -0,0 +1,67 @@
+/*
+** Copyright (C) 2023 Victron Energy B.V.
```
chriadam, line 2: nit: 2024

chriadam (approved):
nits, but LGTM

pages/settings/devicelist/rs/PageRsSystemAlarms.qml
blammit: Done

pages/settings/devicelist/rs/PageRsSystemAlarms.qml
blammit: Done

Issue #743, opened 2024-02-12
Add support for Thai and Chinese for webassembly builds


dirkjanfaber:
The current webassembly builds lack support for Thai and Chinese because of font rendering problems. 

The needed fonts are `NotoSansThai.ttf` and `DroidSansFallback.ttf`. Both do reside in `/usr/lib/fonts` on a GX device, but might need to be made accessible for web retrieval (can be done with creating a symbolic link from place on the webserver).

See https://github.com/victronenergy/gui-v2/pull/673#issuecomment-1931533857 for further info.

mpvader:
hey @blammit , and @chriadam and @DanielMcInnes : this issue was found by our Chinese distributor, he reported it yesterday.

And to be honest I had forgotten about this issue a bit. Its not a blocker for official release of Venus OS with gui-v2, but for sure its a nice to have soon after.

Or at a minimum, we can change the text of this popup to say that Thai and Chinese are not supported yet when accessing this UI, and therefore its available in English only; and expected 2025 Q1. Or something like that.

I have no idea how much work there is in this one; and if it needs discussion with people. For example: if we'd statically host those files on a url, same domain for example as where gui-v2 is downloaded from, would that work and be sufficient for you?

Then we'd arrange hosting on both VRM and locally on the GX and problem easily solved?

blammit:
> I have no idea how much work there is in this one; and if it needs discussion with people. For example: if we'd statically host those files on a url, same domain for example as where gui-v2 is downloaded from, would that work and be sufficient for you?

> Then we'd arrange hosting on both VRM and locally on the GX and problem easily solved?

Yes if it's on the same domain then that should be sufficient. For local wasm the font file can be put in `/var/www/venus/gui-beta` and then for VRM access it would have to be some pre-defined location that can be loaded by gui-v2.

mpvader:
thanks, then thats what we'll do.

@jhofstee can you make those two fonts available on the webserver? for people accessing http://[IP]/gui-v2. 

I'll arrange the fonts to be on VRM.

jhofstee:
I would prefer to have them in http://[ip/fonts instead, so it can be shared. Or does that cause any issue?

chriadam:
Any path is fine, so long as it's served by the nginx webserver we can load it

jhofstee:
@blammit there is nothing blocking this issue, it is trivial to copy / symlink fonts.

jhofstee:
[...]

PR #1581, opened 2024-10-16
WASM: load font files from url

chriadam:
This allows us to load fonts required to display languages like Chinese and Thai.

Also improve handling of language selection so that if changing language fails the radio button list view isn't left in a state which doesn't match the actual currently selected language.

Fixes issue #743

### DIFF
--- a/components/listitems/ListRadioButtonGroup.qml
+++ b/components/listitems/ListRadioButtonGroup.qml
@@ -36,6 +36,7 @@ ListNavigationItem {
 			? optionModel[currentIndex].value
 			: undefined
 
+	property bool updateCurrentIndexOnClick: true
 	property bool updateDataOnClick: true
 	property var popDestination: null   // if undefined, will not automatically pop page when value is selected
 	property var validatePassword
@@ -218,7 +219,11 @@ ListNavigationItem {
 					}
 
 					onClicked: {
-						optionsListView.currentIndex = model.index
+						if (root.updateCurrentIndexOnClick) {
+							optionsListView.currentIndex = model.index
+						} else {
+							optionsListView.selectionChanged = true
+						}
 						if (bottomContentLoader.sourceComponent === passwordComponent) {
 							bottomContentLoader.item.focusPasswordInput()
 						} else {
--- a/data/SystemSettings.qml
+++ b/data/SystemSettings.qml
@@ -197,18 +197,19 @@ QtObject {
 	}
 
 	property VeQuickItem language: VeQuickItem {
+		id: languageChangeListener
+		property bool changingLanguage: false
 		uid: root.serviceUid + "/Settings/Gui/Language"
 		onValueChanged: {
-			if (value !== undefined && !Global.changingLanguage
+			if (!languageChangeListener.changingLanguage
+					&& value !== undefined
 					&& value != Language.toCode(Language.current)) {
-				Global.changingLanguage = true
+				languageChangeListener.changingLanguage = true
 				if (!Language.setCurrentLanguageCode(value)) {
 					// failed.  set the settings value back to the previous one.
 					setValue(Language.toCode(Language.current))
-					Qt.callLater(function() { Global.changingLanguage = false })
-				} else {
-					Qt.callLater(Global.main.retranslateUi)
 				}
+				Qt.callLater(function() { languageChangeListener.changingLanguage = false })
 			}
 		}
 	}
--- a/src/language.cpp
+++ b/src/language.cpp
@@ -323,19 +323,3 @@ bool Language::installTranslatorForLanguage(QLocale::Language language)
 
 	return true;
 }
-
-void Language::retranslate()
-{
-	QQmlEngine* engine = nullptr;
-	QQmlContext* context = QQmlEngine::contextForObject(this);
-
-	if (context) {
-		engine = QQmlEngine::contextForObject(this)->engine();
-	}
-
-	if (engine) {
-		engine->retranslate();
-	} else {
-		qCWarning(venusGui) << "Unable to retranslate";
-	}
-}
--- a/src/language.cpp
+++ b/src/language.cpp
@@ -88,6 +88,8 @@ LanguageModel::LanguageModel(QObject *parent)
 	addLanguage("Українська", "uk", QLocale::Ukrainian);
 	addLanguage("中文", "zh_CN", QLocale::Chinese);
 	addLanguage("العربية", "ar", QLocale::Arabic);
+
+	setCurrentLanguage(Language::create()->getCurrentLanguage());
 }
 
 LanguageModel::~LanguageModel()
--- a/src/language.cpp
+++ b/src/language.cpp
@@ -17,7 +17,7 @@ using namespace Victron::VenusOS;
 
 namespace {
 
-QUrl fontUrlForLanguage(QLocale::Language language)
+QUrl fontUrlForLanguage(QLocale::Language language, const QString &urlPrefix = QString())
 {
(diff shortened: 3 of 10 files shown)


src/main.cpp
```diff
@@ -221,6 +221,14 @@ void initBackend(bool *enableFpsCounter, bool *skipSplashScreen)
 
 #if defined(VENUS_WEBASSEMBLY_BUILD)
 
+EM_JS(const char *, getLocationHrefUtf8, (), {
```
blammit, line 224: I think this `EM_JS` block is indented with four spaces instead of tabs, my pre-commit hooks are expanding them

blammit (commented):
So nice to have this working now :) Just have some minor comments.

src/language.cpp
```diff
@@ -203,19 +201,16 @@ QVariant LanguageModel::data(const QModelIndex &index, int role) const
...
 #if defined(VENUS_WEBASSEMBLY_BUILD)
-	// For languages that cannot be rendered by the default font in wasm, show the language name
+	// For languages that cannot be rendered by a supported font in wasm, show the language name
 	// in English, so that the name can be rendered.
 	switch (language) {
-	case QLocale::Chinese:
-		return QStringLiteral("Chinese %1").arg(m_unsupportedLanguageMessage);
-	case QLocale::Thai:
-		return QStringLiteral("Thai %1").arg(m_unsupportedLanguageMessage);
+		//case QLocale::<Unsupported>: return QStringLiteral("Unsupported %1").arg(m_unsupportedLanguageMessage);
```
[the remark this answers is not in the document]
chriadam: No, I left all of the plumbing in for unsupported languages, in case we have some in future.  But, now that I think about it, that seems unlikely.  So, I'll just remove all of the unsupported language plumbing altogether.

src/main.cpp
chriadam: thanks, will fix

components/listitems/ListRadioButtonGroup.qml
```diff
@@ -36,6 +36,7 @@ ListNavigationItem {
 			? optionModel[currentIndex].value
 			: undefined
 
+	property bool updateCurrentIndexOnClick: true
```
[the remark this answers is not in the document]
chriadam: I will look into this

DanielMcInnes:
Re. common_words_ess - I notice that this translates correctly for desktop builds, but not for wasm builds.

DanielMcInnes:
At startup, we call `Language::Language()` which installs the translator for English, but don't call QQmlEngine::retranslate().
Then, when the VeQuickItem with uid: "/Settings/Gui/Language" gets created, we detect that the language is already English, so we don't call `retranslate()`. 
If we change languages to German and then back to English, we finally call `retranslate()`.

A possible fix is to call `retranslate()` after installing a translator, this is what is recommended here: https://doc.qt.io/qt-6/qqmlengine.html#retranslate

dmcinnes@dmcinnes-System-Product-Name:~/git/victron/gui-v2$ git diff abf484b569cd01f402650ec9b37234e07f405066
diff --git a/src/language.cpp b/src/language.cpp
index 2698c594..cdbd75ea 100644
--- a/src/language.cpp
+++ b/src/language.cpp
@@ -355,6 +355,7 @@ bool Language::installTranslatorForLanguage(QLocale::Language language)
        m_currentLanguage = language;
        m_fontFileUrl = fontUrlForLanguage(language, fontUrlPrefix());
 
+       retranslate();
        return true;
 }
 
dmcinnes@dmcinnes-System-Product-Name:~/git/victron/gui-v2$

components/listitems/ListRadioButtonGroup.qml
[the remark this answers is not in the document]
chriadam: I would prefer that the UI reflects the backend state directly.  Updating the UI on click to be something is not reflective of the backend state, and then updating the UI back to be the old value if the backend state doesn't change, seems confusing to me.

src/language.cpp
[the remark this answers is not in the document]
chriadam: done

src/main.cpp
chriadam: done

chriadam:
@DanielMcInnes we previously didn't want to do that directly, since in the past the QML engine would re-evaluate ALL bindings when the language changed (not just the bindings which involved translated strings), and thus the UI would be left in an indeterminate state (since all delegates will be destroyed due to model property reevaluation etc etc).  But, you are right: we can do it directly, now.

components/listitems/ListRadioButtonGroup.qml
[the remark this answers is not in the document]
blammit: We already have `updateDataOnClick` to control whether the backend state is affected by a click, so also adding `updateCurrentIndexOnClick` seems confusing to me. It would be better if they were at least consolidated somehow.

But I'm happy to leave this for now and rework it later.

blammit (approved):
Nice to get rid of the rebuild-on-language-change part too.

data/SystemSettings.qml
```diff
@@ -197,18 +197,19 @@ QtObject {
 	}
 
 	property VeQuickItem language: VeQuickItem {
+		id: languageChangeListener
 		uid: root.serviceUid + "/Settings/Gui/Language"
+		property bool changingLanguage: false
```
DanielMcInnes, line 202: properties up the top please

data/SystemSettings.qml
```diff
@@ -197,18 +197,19 @@ QtObject {
...
-				Global.changingLanguage = true
+				languageChangeListener.changingLanguage = true
 				if (!Language.setCurrentLanguageCode(value)) {
 					// failed.  set the settings value back to the previous one.
 					setValue(Language.toCode(Language.current))
-					Qt.callLater(function() { Global.changingLanguage = false })
-				} else {
-					Qt.callLater(Global.main.retranslateUi)
 				}
+				Qt.callLater(function() { languageChangeListener.changingLanguage = false })
```
DanielMcInnes, line 212: does this need to be wrapped in `callLater` ?

src/language.cpp
```diff
@@ -28,13 +28,14 @@ QUrl fontUrlForLanguage(QLocale::Language language)
...
-	Q_UNUSED(language);
-	// On wasm, the custom font files are not available. Just use the default font.
-	return defaultFontUrl;
-
+	// On wasm, load the custom font files from relative path /fonts/<fontfile>
+	const QString fileName = fontFileNames.value(language);
+	if (!fileName.isEmpty()) {
+		return QUrl(QStringLiteral("%1%2").arg(urlPrefix, fileName));
+	}
 #elif not defined(VENUS_DESKTOP_BUILD)
```
DanielMcInnes, line 36: `#elif`  -> `#else`

DanielMcInnes (approved):
lgtm, nits

data/SystemSettings.qml
chriadam: done

data/SystemSettings.qml
chriadam: yes, as if the change fails, that can cause a side effect e.g. setting the value BACK to the old value, and we don't want to spin infinitely.

src/language.cpp
chriadam: no, we don't want to look in /usr/lib/fonts on Windows Desktop

data/SystemSettings.qml
DanielMcInnes: fair enough

src/language.cpp
DanielMcInnes: all good, I can't read

PR #1582, opened 2024-10-16
Fix path for Lynx parallel BMS system voltage

### LINKED ISSUE #1522: Lynx parallel BMS on WASM not displaying system voltage when BMS is pending
If one of the paralleled BMSes is 'pending' (waiting until its voltage level is in range with the system voltage), we show the system voltage in the menu of that BMS. However, on WASM, this field is not set properly:

image

On the local UI, the field is showing the value correctly.
See [original issue](https://github.com/victronenergy/gui-v2/issues/1269)

DanielMcInnes:
Fixes #1522

### DIFF
--- a/pages/settings/devicelist/battery/PageBattery.qml
+++ b/pages/settings/devicelist/battery/PageBattery.qml
@@ -137,7 +137,7 @@ Page {
 
 				//% "System voltage"
 				text: qsTrId("devicelist_battery_system_voltage")
-				dataItem.uid: BackendConnection.serviceUidForType("battery.lynxparallel") + _n2kDeviceInstance.value + "/Dc/0/Voltage"
+				dataItem.uid: BackendConnection.serviceUidFromName("com.victronenergy.battery.lynxparallel" + _n2kDeviceInstance.value, _n2kDeviceInstance.value) + "/Dc/0/Voltage"
 				allowed: defaultAllowed && _n2kDeviceInstance.isValid && root.battery.state === VenusOS.Battery_State_Pending
 				unit: VenusOS.Units_Volt_DC
 			}
--- a/pages/settings/devicelist/battery/PageBattery.qml
+++ b/pages/settings/devicelist/battery/PageBattery.qml
@@ -12,7 +12,7 @@ Page {
 	property var battery
 
 	readonly property bool isFiamm48TL: productId.value === ProductInfo.ProductId_Battery_Fiamm48TL
-	readonly property bool isParallelBms: productId.value === ProductInfo.ProductId_Battery_ParallelBms
+	readonly property bool isParallelBms: nrOfBmses.dataItem.isValid
 
 	title: battery.name
 
@@ -126,7 +126,7 @@ Page {
 				//% "Total Capacity"
 				text: qsTrId("devicelist_battery_total_capacity")
 				dataItem.uid: root.battery.serviceUid + "/Capacity"
-				allowed: defaultAllowed && numberOfBms.allowed
+				allowed: defaultAllowed && root.isParallelBms
 				unit: VenusOS.Units_AmpHour
 			}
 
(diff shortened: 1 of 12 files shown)


pages/settings/devicelist/battery/PageBattery.qml
```diff
@@ -137,7 +137,9 @@ Page {
 
 				//% "System voltage"
 				text: qsTrId("devicelist_battery_system_voltage")
-				dataItem.uid: BackendConnection.serviceUidForType("battery.lynxparallel") + _n2kDeviceInstance.value + "/Dc/0/Voltage"
+				dataItem.uid: BackendConnection.type === BackendConnection.MqttSource
+							  ? "mqtt/battery/" + _n2kDeviceInstance.value  + "/Dc/0/Voltage"
+							  : BackendConnection.serviceUidForType("battery.lynxparallel") + _n2kDeviceInstance.value + "/Dc/0/Voltage"
```
DanielMcInnes, line 142: @blammit this is one of those cases where the mqtt path is quite different to the dbus path.

dbus path:
dbus/com.victronenergy.battery.lynxparallel1/Dc/0/Voltage

mqtt path:
mqtt/battery/1/Dc/0/Voltage

I'm not sure how best to handle this. Who defines the mqtt paths? Maybe the real fix here is to change the mqtt path.

src/backendconnection.cpp
```diff
@@ -581,8 +581,8 @@ QString BackendConnection::serviceUidForType(const QString &serviceType) const
 	//  - Mock: mock/com.victronenergy.system
 
 	return m_type == MqttSource
-			? QStringLiteral("mqtt/%1/0").arg(serviceType)
-			: QStringLiteral("%1/com.victronenergy.%2").arg(uidPrefix()).arg(serviceType);
+			? QStringLiteral("mqtt/%2/%1").arg(deviceInstance).arg(serviceType.split(".")[0])
```
DanielMcInnes, line 584: This fixes the uid for serviceTypes with the form "foo.bar". Currently, "battery.lynxparallel" is the only serviceType that calls this function with the form of "foo.bar", everywhere else that calls BackendConnection::serviceUidForType() does it with a simpler serviceType, that is not dot-separated.

The intent is for the behavior to remain unchanged for every other caller of BackendConnection::serviceUidForType()

DanielMcInnes:
It occurs to me that dbus-flashmq (https://github.com/victronenergy/dbus-flashmq.git) is the canonical source for converting dbus paths to mqtt paths. I'm not sure it's worth changing to use this at this stage of the project, but it's useful to know anyway. See the following code snippet:
```
// shortservicename.cpp
std::string ShortServiceName::make_short(std::string service)
{
    if (service.find("com.victronenergy.") != std::string::npos)
    {
        const std::vector<std::string> parts = splitToVector(service, '.');
        service = parts.at(2);
    }
    else if (service.find(".") != std::string::npos || service.find("/") != std::string::npos)
    {
        throw ValueError("String doesn't look like com.victronenergy.something or something without dots or slashes.");
    }

    return service;
}
```

pages/settings/devicelist/battery/PageBattery.qml
```diff
@@ -137,7 +137,7 @@ Page {
 
 				//% "System voltage"
 				text: qsTrId("devicelist_battery_system_voltage")
-				dataItem.uid: BackendConnection.serviceUidForType("battery.lynxparallel") + _n2kDeviceInstance.value + "/Dc/0/Voltage"
+				dataItem.uid: BackendConnection.serviceUidForType("battery.lynxparallel", _n2kDeviceInstance.value) + "/Dc/0/Voltage"
```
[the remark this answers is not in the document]
DanielMcInnes: updated as suggested. Added unit tests for BackendConnection.serviceUidForType().

src/backendconnection.cpp
```diff
@@ -658,5 +659,17 @@ QVariant BackendConnection::mockValue(const QString &uid) const
 	return QVariant();
 }
 
+BackendConnectionTester::BackendConnectionTester() {
```
blammit, line 662: `{` on the next line

pages/settings/devicelist/battery/PageBattery.qml
```diff
@@ -138,16 +138,16 @@ Page {
...
 
 			ListTextItem {
-				id: numberOfBms
+				id: nrOfBmses
 				//% "Number of BMSes"
 				text: qsTrId("devicelist_battery_number_of_bmses")
 				dataItem.uid: root.battery.serviceUid + "/NumberOfBmses"
-				allowed: defaultAllowed && dataItem.isValid
+				allowed: defaultAllowed && root.isParallelBms
 			}
```
blammit, line 151: The "Alarms" menu also needs a binding to `root.isParallelBms`.

DanielMcInnes: done

src/backendconnection.cpp
DanielMcInnes: Done

tests/backendConnection/tst_backendConnection.cpp
```diff
@@ -0,0 +1,15 @@
+/*
```
blammit, line 1: Could you lower-case these file names to `tst_backendconnection.cpp/qml`?

pages/settings/devicelist/battery/PageBattery.qml
blammit: I don't see the change, was it left out?

DanielMcInnes: are you talking about this thing in PageBattery.qml:
```			ListNavigationItem {
				text: CommonWords.alarms
				allowed: !root.isParallelBms
				onClicked: {
					Global.pageManager.pushPage("/pages/settings/devicelist/battery/PageBatteryAlarms.qml",
							{ "title": text, "bindPrefix": root.battery.serviceUid })
				}
			}
```
?

blammit: Oh sorry, completely missed that.

blammit (approved):

PR #1583, opened 2024-10-16
Fix dbus path in Fronius inverter pages, small cleanup

izak:
The dbus path for Fronius inverters was wrong (it had the dbus/ prefix twice). Also, `phaseItem` was unused.


blammit (approved):
LGTM, thanks!

PR #1594, opened 2024-10-18
Add onboarding/welcome page

### LINKED ISSUE #1441: Add onboarding page
when opening gui-v2 for the first time, welcome user, give option to go to a yet to be written blogpost (on vrm/wasm it can be a link, native it can be a QR code (bit-ly link @jepefe !) , and explain where to switch back to previous gui.

for switching back, see this issue: https://github.com/victronenergy/venus/issues/1342

Next step: @jepefe discusses this with Serj.

blammit:
This is shown before the splash screen, if /Settings/Gui2/OnBoarding is 0.

Fixes #1441


blammit:
To view the onboarding in mock mode, change SystemSettingsImpl.qml to set `Gui2/OnBoarding = 0`.

pages/welcome/WelcomePage.qml
```diff
@@ -0,0 +1,111 @@
...
+		}
+
+		Label {
+			Layout.fillWidth: true
+			Layout.fillHeight: true
+			Layout.verticalStretchFactor: root.qrCodeUrl ? -1 : 1
+			wrapMode: Text.Wrap
+			text: root.text
+			onLinkActivated: (linkText) => {
+				BackendConnection.openUrl(linkText)
```
chriadam, line 68: expected that this does nothing on the cerbo?

pages/welcome/WelcomeView.qml
```diff
@@ -0,0 +1,175 @@
...
+			left: parent.left
+			right: parent.right
+			top: header.bottom
+			bottom: parent.bottom
+		}
+	}
+
+	VeQuickItem {
+		id: onboardingDone
+		uid: Global.systemSettings.serviceUid + "/Settings/Gui2/OnBoarding"
```
chriadam, line 173: does this need guarding similarly to the other one: `Global.dataManagerLoaded ? Global.systemSettings.serviceUid + "/Settings/Gui2/OnBoarding" : ""`

pages/welcome/WelcomePage.qml
blammit: Yes, but there are no links in the cerbo version (only in the wasm version) so this won't get activated.

pages/welcome/WelcomeView.qml
blammit: It doesn't need it, since the Loader that loads this view has a binding of `active: Global.dataManagerLoaded`.

chriadam (approved):
LGTM.

Tested on cerbo, local wasm, and VRM wasm.

One point to note: the setting value is just 1 for "done", which means that if an installer accesses from VRM to do something, it will show the Welcome, if they skip it, then the user boots their Cerbo, the Welcome page won't show in that case (since it has already been skipped).  This is probably fine / as per spec, but might be worth discussing (should we have flags for DoneWasm vs DoneCerbo and show on each, or?)

PR #1596, opened 2024-10-18
Set different onboarding 'completed' states for wasm vs device

blammit:
This allows the onboarding/welcome screen to be shown on the device, even if it was previously shown on wasm, and vice-versa.

Contributes to #1441


chriadam (approved):
LGTM, but let's wait until we hear from Matthijs before merging.

blammit:
Merging as per slack discussion.

PR #1602, opened 2024-10-22
Fix the tank merge threshold for 7 inch screens

### LINKED ISSUE #1599: Tanks are not grouped as expected in 7"/wasm screen size
When there are more tanks than what can fit into the view, tanks of the same type should be grouped to allow all tanks to be visible.

This is reproducible in mock mode in the "8 tanks" configuration. The tanks are grouped as expected in the 5" screen size:

image

But not in the 7"/wasm screen size. You can see the third black water tank is partially off-screen:

image

DanielMcInnes:
Fixes #1599

5" screens can fit 5 gauges, 7" screens can fit 6 gauges


blammit (approved):
We should calculate the threshold instead of hardcoding it in a value, but that's for another day :)

PR #1603, opened 2024-10-22
Ensure onboarding/welcome screen cannot get stuck

### LINKED ISSUE #1601: Onboarding is stuck?
On this system, https://vrm.victronenergy.com/installation/151755/diagnostics, 50ft yacht demo system in our lab, while running Venus OS v3.50~34, three people, including me, had the issue that they couldn't in the end close the onboarding pages - while operating in WASM.

Both the SKIP and the CLOSE button didn't work.

I didn't see any relevant messages in the dev. console of the browser.

Also haven't heard anyone else complain about the same, which is strange since by now there must have been 1000s of people that have seen the onboarding page.

Just in case its relevant: the onboarding setting does exist in localsettings, and setting that to 0 immediately closed the onboard for me. Which was nice, since this is a demo system which marketing dept. is using to make screenshots required for the launch of gui-v2.

So; at least do a review of the code to see if there is some condition or similar in which case the onboarding page refuses to close?

blammit:
In case a write to /Settings/Gui2/OnBoarding fails, set a bool property that will cause the onboarding/welcome screen to be closed.

Fixes #1601

### DIFF
--- a/data/SystemSettings.qml
+++ b/data/SystemSettings.qml
@@ -10,6 +10,7 @@ QtObject {
 	id: root
 
 	readonly property string serviceUid: BackendConnection.serviceUidForType("settings")
+	readonly property bool needsOnboarding: !_onboardingState.done
 
 	property int electricalQuantity: VenusOS.Units_None
 	property int temperatureUnit: VenusOS.Units_None
@@ -105,6 +106,10 @@ QtObject {
 		default:
 			return ""
 		}
+	}
+
+	function setOnboardingDone() {
+		_onboardingState.setDoneFlag()
 	}
 
 	property VeQuickItem accessLevel: VeQuickItem {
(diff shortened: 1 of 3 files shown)


data/SystemSettings.qml
```diff
@@ -300,6 +305,26 @@ QtObject {
...
+		function setDoneFlag() {
+			_forceOnboardingDone = true
+			if (Qt.platform.os === "wasm") {
+				setValue(value | VenusOS.OnboardingState_DoneWasm)
+			} else {
+				setValue(value | VenusOS.OnboardingState_DoneNative)
+			}
+		}
+
+		uid: Global.systemSettings.serviceUid + "/Settings/Gui2/OnBoarding"
```
chriadam, line 325: this can probably just be root.serviceUid

chriadam (approved):
one nitpick, but LGTM

data/SystemSettings.qml
blammit: Done

PR #1604, opened 2024-10-22
Add light-theme image for QR code

blammit:
Do the qrCodeUrl binding in the delegate. Otherwise if the qrCodeUrl model data changes based on the light/dark theme (which can happen during startup when the theme color value is read) the model and instantiator objects will be rebuilt.


DanielMcInnes (approved):

mr-manuel:
Is there a reason, thy the QR code is not generated in Qt?

blammit:
Mainly that it follows a specific design, and it would be preferable to reduce external dependencies for the onboarding view.

PR #1605, opened 2024-10-22
Show ESS status numbers in the battery widget

### LINKED ISSUE #424: ESS status numbers are not shown in gui-v2
In gui-v1 the bottom of the screen shows ESS status numbers, like this:

image

This “ESS#1” at the bottom of the screen is showing the state of values under com.victronenergy.system/SystemState, e.g. SystemState/LowSoc, SystemState/BatteryLife. 

@jepefe Do we need to add this info somewhere? (If so, maybe it could be added to the ESS control cards?)

DanielMcInnes:
Fixes #424

### DIFF
--- a/components/SystemReason.qml
+++ b/components/SystemReason.qml
@@ -1,0 +1,58 @@
+/*
+** Copyright (C) 2024 Victron Energy B.V.
+** See LICENSE.txt for license information.
+*/
+
+import QtQuick
+import Victron.VenusOS
+
+QtObject {
+	id: root
+
+	readonly property bool chargeOrDischargeDisabled: chargeDisabled.value || dischargeDisabled.value || false
+	readonly property string serviceUid: BackendConnection.serviceUidForType("system")
+	//% "ESS %1"
+	readonly property string text: flags.length > 0 ? qsTrId("ess_flags").arg(flags.join(" ")) : ""
+
+	readonly property var flags: {
+		var r = []
+		for (var i=0; i<flagItems.length; i++) {
+			if (flagItems[i].value) r.push('#' + (i+1).toString(16))
+		}
+		return r
+	}
+
+	readonly property alias lowSoc: _lowSoc
+	readonly property alias batteryLife: _batteryLife
+	readonly property alias chargeDisabled: _chargeDisabled
+	readonly property alias dischargeDisabled: _dischargeDisabled
+	readonly property alias slowCharge: _slowCharge
+	readonly property alias userChargeLimited: _userChargeLimited
+	readonly property alias userDischargeLimited: _userDischargeLimited
+
+	// Flags to monitor
+	readonly property list<VeQuickItem> flagItems: [
+		VeQuickItem { id: _lowSoc;					uid: serviceUid + "/SystemState/LowSoc" },
+		VeQuickItem { id: _batteryLife;				uid: serviceUid + "/SystemState/BatteryLife" },
+		VeQuickItem { id: _chargeDisabled;			uid: serviceUid + "/SystemState/ChargeDisabled" },
+		VeQuickItem { id: _dischargeDisabled;		uid: serviceUid + "/SystemState/DischargeDisabled" },
+		VeQuickItem { id: _slowCharge;				uid: serviceUid + "/SystemState/SlowCharge" },
+		VeQuickItem { id: _userChargeLimited;		uid: serviceUid + "/SystemState/UserChargeLimited" },
+		VeQuickItem { id: _userDischargeLimited;	uid: serviceUid + "/SystemState/UserDischargeLimited" }
+	]
+
+	function descriptionText() { // not used yet, will be needed for the new battery charging/discharge design
+		if (chargeDisabled.value && dischargeDisabled.value) {
+			//% "ESS %1 Charge/Discharge Disabled"
+			return qsTrId("systemreason_charge_discharge_disabled").arg(flags.join(" "))
+		} else if (chargeDisabled.value) {
+			//% "ESS %1 Charge Disabled"
+			return qsTrId("systemreason_charge_disabled").arg(flags.join(" "))
+		} else if (dischargeDisabled.value) {
+			//% "ESS %1 Discharge Disabled"
+			return qsTrId("systemreason_discharge_disabled").arg(flags.join(" "))
+		}
+
+		return ""
+	}
+}
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -266,6 +266,7 @@ set (VENUS_QML_MODULE_SOURCES
     components/StatusBar.qml
     components/SwipePageModel.qml
     components/SwipeViewPage.qml
+    components/SystemReason.qml
     components/TabBar.qml
     components/TankItem.qml
     components/TankGauge.qml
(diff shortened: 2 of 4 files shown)


DanielMcInnes:
image

DanielMcInnes:
Converting to draft, I need to clarify requirements.
Gui-v1 shows "ESS #..." if any of the following are true:
LowSoc, BatteryLife, ChargeDisabled, DischargeDisabled, SlowCharge, UserChargeLimited, UserDischargeLimited

I assume (but am not certain) that gui-v2 should only show "ESS #..." if ChargeDisabled and / or DischargeDisabled is true. I.e. , if "LowSoc" is true, I don't think we should show "ESS #1", as then we wouldn't have room to show volts / amps / watts

tests/screenblanker/tst_screenblanker.qml
```diff
@@ -38,7 +38,7 @@ TestCase {
 		let endTime = new Date();
 
 		// Check the timeout roughly follows the display off time
-		fuzzyCompare(blanker.displayOffTime, endTime - startTime, 50)
+		fuzzyCompare(blanker.displayOffTime, endTime - startTime, 100)
```
DanielMcInnes, line 41: this was failing for me

components/widgets/BatteryWidget.qml
```diff
@@ -174,6 +174,21 @@ OverviewWidget {
...
+			Label {
+				id: essStatus
+
+				text: systemReason.text
+				font.pixelSize: Theme.font_size_body1
+				width: parent.width
+				wrapMode: Text.WordWrap
+				color: Theme.color_font_secondary
+				// There is not enough room to display the system reason text when the nav bar is visible
+				visible: Global.pageManager.interactivity === VenusOS.PageManager_InteractionMode_Idle && systemReason.flags.length > 0
```
DanielMcInnes, line 186: we need to hide this when the nav bar is visible, otherwise we get this:
Screenshot from 2024-10-23 11-17-31

chriadam: As per meeting discussion: reduce the font size of the inverter status label, and put the ESS code numbers into the inverter (I guess we might need `maxLineCount: 2; wrapMode: Text.Wrap` for those, since the phase bar graphs take up some horizontal space and there are 8 possible codes).

tests/screenblanker/tst_screenblanker.qml
DanielMcInnes: reverted, I think this failed because I was running multiple instances of gui-v2

components/widgets/BatteryWidget.qml
DanielMcInnes: Updated as per meeting discussion. I ended up leaving the inverter status label font size the same, everything fits, and it looks better this size somehow, looks a bit funny when I reduce it.
Screenshot from 2024-10-25 12-14-17
Screenshot from 2024-10-25 12-14-15
Screenshot from 2024-10-25 12-14-03
Screenshot from 2024-10-25 12-13-57

components/SystemReason.qml
```diff
@@ -0,0 +1,58 @@
...
+		VeQuickItem { id: _dischargeDisabled;		uid: serviceUid + "/SystemState/DischargeDisabled" },
+		VeQuickItem { id: _slowCharge;				uid: serviceUid + "/SystemState/SlowCharge" },
+		VeQuickItem { id: _userChargeLimited;		uid: serviceUid + "/SystemState/UserChargeLimited" },
+		VeQuickItem { id: _userDischargeLimited;	uid: serviceUid + "/SystemState/UserDischargeLimited" }
+	]
+
+	function descriptionText() { // not used yet, will be needed for the new battery charging/discharge design
+		if (chargeDisabled.value && dischargeDisabled.value) {
+			//% "ESS %1 Charge/Discharge Disabled"
+			return qsTrId("charge_discharge_disabled").arg(flags.join(" "))
```
chriadam, line 47: The translation id should be more context-specific e.g. "ess_charge_discharge_disabled" or location-specific e.g. "systemreason_charge_discharge_disabled", because there are a variety of contexts in which "charge_discharge_disabled" might be sensible, and we want to avoid id clashes.

components/SystemReason.qml
```diff
@@ -0,0 +1,58 @@
...
+	readonly property alias lowSoc: _lowSoc
+	readonly property alias batteryLife: _batteryLife
+	readonly property alias chargeDisabled: _chargeDisabled
+	readonly property alias dischargeDisabled: _dischargeDisabled
+	readonly property alias slowCharge: _slowCharge
+	readonly property alias userChargeLimited: _userChargeLimited
+	readonly property alias userDischargeLimited: _userDischargeLimited
+
+	// Flags to monitor
+	readonly property list<VeQuickItem> flagItems: [
```
chriadam, line 34: from the meeting, I thought there were 8 codes, but there are only 7 here - expected?

components/widgets/BatteryWidget.qml
chriadam: Looks good.  Does the string "ESS" ever get translated into other languages?  Wondering if it's larger/longer in German or something, in which case it might cause problems.  But, at least in my testing, I tried to switch between a variety of languages and that string "ESS" was never translated...

chriadam: Hmm, "ESS" entry in Settings page doesn't seem to get translated either.  Ok, never mind :-)

components/SystemReason.qml
```diff
@@ -0,0 +1,58 @@
...
+import QtQuick
+import Victron.VenusOS
+
+QtObject {
+	id: root
+
+	readonly property bool chargeOrDischargeDisabled: chargeDisabled.value || dischargeDisabled.value
+	readonly property string serviceUid: BackendConnection.serviceUidForType("system")
+	//% "ESS %1"
+	readonly property string text: qsTrId("ess_flags").arg(flags.join(" "))
```
chriadam, line 15: should this be: `flags.length > 0 ? qsTrId("ess_flags").arg(flags.join(" ")) : ""` ?

chriadam: Or, should it only show if the system has some non-null ESS mode?  (is such a thing possible?)

components/SystemReason.qml
```diff
@@ -0,0 +1,58 @@
...
+** See LICENSE.txt for license information.
+*/
+
+import QtQuick
+import Victron.VenusOS
+
+QtObject {
+	id: root
+
+	readonly property bool chargeOrDischargeDisabled: chargeDisabled.value || dischargeDisabled.value
```
chriadam, line 12: maybe this only affects mock mode, but when I commented out `systemReason.dischargeDisabled.setValue(1)` from mock/SystemSettingsImpl.qml I get: `SystemReason.qml:12:2: Unable to assign [undefined] to bool`

blammit:
With "Boat/Motorhome Demo 1" I'm seeing just "ESS" without any text -- should it be just omitted altogether instead?

image

components/SystemReason.qml
DanielMcInnes: I half-remember 8 codes from the meeting as well, but gui-v1 only has 7

components/SystemReason.qml
DanielMcInnes: done

components/SystemReason.qml
DanielMcInnes: yes, updated

components/SystemReason.qml
DanielMcInnes: it's easy enough to fix anyway, updated. 

`readonly property bool chargeOrDischargeDisabled: chargeDisabled.value || dischargeDisabled.value || false`

DanielMcInnes:
> With "Boat/Motorhome Demo 1" I'm seeing just "ESS" without any text -- should it be just omitted altogether instead?
> image

yes it should be omitted, fixed now

chriadam (approved):

PR #1608, opened 2024-10-22
update the UI selection text

jhofstee:
The UI selection now also switches the default index page; remote console or the gui-v2 wasm. What exactly is being switched depends on the device, since not all of them can run gui-v2.

If the device can run gui-v2, the setting is now called 'User interface', otherwise 'User interface (Remote Console)' to make it clear that only remote console will be switched. Remove the Onscreen UI references, since it no longer affects only the Onscreen UI.

### DIFF
--- a/pages/settings/PageSettingsDisplay.qml
+++ b/pages/settings/PageSettingsDisplay.qml
@@ -172,8 +172,11 @@ Page {
 
 				property ModalWarningDialog _restartDialog
 
-				//% "Onscreen UI (GX Touch & Ekrano)"
-				text: qsTrId("settings_display_onscreen_ui")
+				text: onScreenGuiv2Possible.value
+					  //% "User interface"
+					? qsTrId("settings_display_onscreen_ui")
+					  //% "User interface (Remote Console)"
+					: qsTrId("settings_display_remote_console_ui")
 				dataItem.uid: Global.systemSettings.serviceUid + "/Settings/Gui/RunningVersion"
 				writeAccessLevel: VenusOS.User_AccessType_User
 				updateDataOnClick: false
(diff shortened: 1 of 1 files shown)


pages/settings/PageSettingsDisplay.qml
```diff
@@ -171,9 +171,13 @@ Page {
 				id: runningVersion
 
 				property ModalWarningDialog _restartDialog
+				property VeQuickItem onScreenGuiv2Possible: VeQuickItem { uid: Global.venusPlatform.serviceUid + "/Gui/OnScreenGuiv2Supported" }
 
-				//% "Onscreen UI (GX Touch & Ekrano)"
-				text: qsTrId("settings_display_onscreen_ui")
+				text: onScreenGuiv2Possible.value === true ?
```
DanielMcInnes, line 176: you can get rid of the `=== true`, eg.
`onScreenGuiv2Possible.value ?`

DanielMcInnes (approved):
lgtm, nits

pages/settings/PageSettingsDisplay.qml
jhofstee: Yes, that can also be done. I am perfectly fine if you just make such small changes while applying. That saves a round trip.

pages/settings/PageSettingsDisplay.qml
```diff
@@ -172,8 +172,11 @@ Page {
 
 				property ModalWarningDialog _restartDialog
 
-				//% "Onscreen UI (GX Touch & Ekrano)"
-				text: qsTrId("settings_display_onscreen_ui")
+				text: onScreenGuiv2Possible.value
+					  //% "User interface"
+					? qsTrId("settings_display_onscreen_ui")
```
blammit, line 177: @jhofstee I also made a change to put the `?` and `:` at the beginning of the lines here.

Also made `onScreenGuiv2Possible` a child item rather than a property, since it doesn't need to be a property.

blammit (approved):

pages/settings/PageSettingsDisplay.qml
```diff
@@ -198,28 +201,37 @@ Page {
...
 							? qsTrId("settings_restarting_app")
-							  //% "Application restarted"
-							: qsTrId("settings_app_restarted")
+							  //: %1 = The name of the setting being updated
+							  //% "%1 updated"
+							: qsTrId("settings_app_restarted").arg(settingText)
 						description: BackendConnection.type === BackendConnection.DBusSource
 							  //: %1 = the UI version that the system is switching to
-							  //% "Onscreen UI will switch to %1."
+							  //% "User interface will switch to %1."
```
blammit, line 229: Also fixed `inferface` -> `interface`

PR #1609, opened 2024-10-23
Battery list: show values directly from from system /Batteries

### LINKED ISSUE #1572: Battery detail page data issues
Most genset controllers provide a starter battery voltage reading. If it is available via `com.victronenergy.genset.*/StarterVoltage`, Venus OS seems to interpret the genset service as a battery. This is really helpful, as a low starter battery (voltage) could cause a genset start failure.

However, gui-v2 does not yet seem to be able to interpret the service correctly. In the overview of all batteries in the system (via tap on the SystemBattery in the Overview), no helpful values are given for the genset starter battery.

192 168 254 90_gui-v2_(Ekrano GX) (5)
192 168 254 90_gui-v2_(Ekrano GX) (6)

For comparison: On VRM, the service is always called “Generator Starter Battery” and shows the only available value, the voltage. I would expect something similar on gui-v2.

Bildschirmfoto 2024-10-11 um 18 29 56

blammit:
The /Batteries list provides values like "power", "voltage", and "soc", so these can be used to show the values in the battery list. This is better than assuming each /Batteries entry comes from a service that provides paths like /Power, /Voltage, /Soc, etc. and fetching the data values from those paths for each entry.

For example, the /Batteries list may contain a starter battery provided by a genset service, and in this case the voltage is provided by com.victronenergy.genset.*/StarterVoltage instead of /Voltage. The backend uses the /StarterVoltage to set the 'voltage' value for the entry in the /Batteries list, so the UI should show that value, instead of trying to find a /Voltage path that does not exist.

Also rework the UI layout for each delegate in the battery list, since the SOC should be vertically centered if the mode/discharging text is not available.

Fixes #1572

### DIFF
--- a/pages/battery/BatteryListPage.qml
+++ b/pages/battery/BatteryListPage.qml
@@ -76,6 +76,8 @@ Page {
 					QuantityLabel {
 						id: socLabel
 
+						readonly property int statusLevel: Theme.getValueStatus(value, VenusOS.Gauges_ValueType_FallingPercentage)
+
 						width: parent.width
 						height: nameLabel.height
 						alignment: Text.AlignRight
@@ -83,6 +85,14 @@ Page {
 						unit: VenusOS.Units_Percentage
 						font.pixelSize: Theme.font_size_body2
 						visible: !isNaN(batteryDelegate.device.soc)
+						valueColor: batteryDelegate.device.mode === VenusOS.Battery_Mode_Idle ? Theme.color_font_primary
+								: statusLevel === Theme.Critical ? Theme.color_red
+								: statusLevel === Theme.Warning ? Theme.color_orange
+								: Theme.color_green
+						unitColor: batteryDelegate.device.mode === VenusOS.Battery_Mode_Idle ? Theme.color_font_secondary
+								: statusLevel === Theme.Critical ? Theme.color_red
+								: statusLevel === Theme.Warning ? Theme.color_orange
+								: Theme.color_green
 					}
 
 					Label {
--- a/pages/battery/BatteryListPage.qml
+++ b/pages/battery/BatteryListPage.qml
@@ -4,102 +4,114 @@
 */
 
 import QtQuick
+import QtQuick.Layouts
 import QtQuick.Controls.impl as CP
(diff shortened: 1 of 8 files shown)


DanielMcInnes:
Should we be seeing the leading separator?

Screenshot from 2024-10-23 12-15-48

blammit:
@DanielMcInnes It's not in the design, but without it, it's difficult to make the "My starter battery" and the "0.03V" line up. We can fudge it but it won't really line up exactly anyway. The problem is that the design of having those multiple numbers in a row does not normally need to align horizontally with anything else.

DanielMcInnes:
> @DanielMcInnes It's not in the design, but without it, it's difficult to make the "My starter battery" and the "0.03V" line up. We can fudge it but it won't really line up exactly anyway. The problem is that the design of having those multiple numbers in a row does not normally need to align horizontally with anything else.
> 
> So we can either fudge it or add the separators 🤔

fair enough

pages/battery/BatteryListPage.qml
```diff
@@ -132,34 +173,65 @@ Page {
...
+		property real soc: NaN
+		property real temperature: NaN
+		property int timetogo: 0
+		property real voltage: NaN
+		readonly property int mode: Global.batteries.batteryMode(power)
+	}
+
 	Component {
-		id: batteryComponent
+	   id: batteryComponent
```
DanielMcInnes, line 231: formatting

pages/battery/BatteryListPage.qml
```diff
@@ -132,34 +173,65 @@ Page {
...
 
-		Battery {
-			id: battery
+		BatteryListDevice {
+		   id: battery
```
DanielMcInnes, line 234: formatting

DanielMcInnes (approved):
lgtm, formatting nits

pages/battery/BatteryListPage.qml
blammit: Done

Serjle:
@DanielMcInnes 
divider
Separator just between numbers

PR #1611, opened 2024-10-23
Show confirmation dialog on reboot

### LINKED ISSUE #1577: Show confirmation dialog on reboot
In Settings > General, when "Reboot now" is clicked, show a confirmation dialog with OK/Cancel options so that the user can cancel this action if it was accidentally clicked.

DanielMcInnes:
Fixes #1577

### DIFF
--- a/pages/settings/PageSettingsGeneral.qml
+++ b/pages/settings/PageSettingsGeneral.qml
@@ -5,6 +5,7 @@
 
 import QtQuick
 import Victron.VenusOS
+import QtQuick.Templates as T
 
 Page {
 	id: root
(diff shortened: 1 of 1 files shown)


pages/settings/PageSettingsGeneral.qml
```diff
@@ -298,13 +298,31 @@ Page {
...
+
+				Component {
+					id: confirmRebootDialogComponent
+
+					ModalWarningDialog {
+						//% "Press 'OK' to reboot"
+						title: qsTrId("press_ok_to_reboot")
+						dialogDoneOptions: VenusOS.ModalDialog_DoneOptions_OkAndCancel
+						onAccepted: {
+							timer.start()
```
DanielMcInnes, line 311: DialogLayer doesn't handle opening a dialog immediately after another dialog has closed, use a timer to do it instead

chriadam (approved):

pages/settings/PageSettingsGeneral.qml
blammit: @DanielMcInnes does it need to be 500ms? Will a 0-interval time do? Or is it related to the duration of the dialog fade-out animation or such?

DanielMcInnes: I think there's a bug in DialogLayer.open(). If you open a new dialog before the old one is completely gone (eg. in the 'onAccepted' handler rather than the 'onClosed' handler, the new dialog gets closed when the old one gets destroyed.

I found a nicer way to do it, via 'onClosed' instead of 'onAccepted', so I can get rid of the hacky Timer

chriadam (approved):

PR #1614, opened 2024-10-23
add info for user to troubleshoot MQTT

jhofstee:
This adds the MQTT connection errors to VRM online portal menu as well. The MQTT connection errors are not shown if a bridge is disabled. If an error is present all error status are shown unless explicitly reported as disabled.

### DIFF
--- a/pages/settings/PageSettingsLogger.qml
+++ b/pages/settings/PageSettingsLogger.qml
@@ -116,6 +116,8 @@ Page {
 			}
 
 			ListSwitch {
+				id: useHttps
+
 				//% "Use secure connection (HTTPS)"
 				text: qsTrId("settings_https_enabled")
 				dataItem.uid: Global.systemSettings.serviceUid + "/Settings/Vrmlogger/HttpsEnabled"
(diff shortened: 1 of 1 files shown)


pages/settings/PageSettingsLogger.qml
```diff
@@ -142,6 +144,42 @@ Page {
 				}
 			}
 
+			ListTextItem {
+				id: connectionStatus
+
+				// Status of vrmlogger
+				property VeQuickItem httpsConnectionError: VeQuickItem {
```
blammit, line 151: Doesn't need to be a property, since the parent is an `Item`. Same for the VeQuickItems below.

pages/settings/PageSettingsLogger.qml
```diff
@@ -142,6 +144,42 @@ Page {
...
+					uid: Global.venusPlatform.serviceUid + "/Mqtt/Bridges/GXdbus/ConnectionStatus"
+				}
+				property VeQuickItem mqttRpcConnected: VeQuickItem {
+					uid: Global.venusPlatform.serviceUid + "/Mqtt/Bridges/GXrpc/Connected"
+				}
+				property VeQuickItem mqttRpcStatus: VeQuickItem {
+					uid: Global.venusPlatform.serviceUid + "/Mqtt/Bridges/GXrpc/ConnectionStatus"
+				}
+
+				property bool anyError:
```
blammit, line 172: Can be readonly

pages/settings/PageSettingsLogger.qml
blammit: Oh never mind, I see that you want to access these as properties.

pages/settings/PageSettingsLogger.qml
blammit: I made this change, and force-pushed.

blammit:
Haven't tested this against a specific installation with all these configurations, but page loads fine.

blammit (approved):

PR #1616, opened 2024-10-24
Only show onboarding when setting is available

blammit:
Show the onboarding if the setting indicates it is needed, instead of when it is "not done". This way, the onboarding will not show if the setting is not actually available.


blammit:
Adding this so that the onboarding doesn't show up when testing on a VRM installation that isn't actually updated to the latest version.

PR #1620, opened 2024-10-24
Adjust conditions for showing arrow icon for positive/negative AC power on Brief/Overview

blammit:


components/AcInputDirectionIcon.qml
```diff
@@ -7,9 +7,12 @@ import QtQuick
 import QtQuick.Controls.impl as CP
 import Victron.VenusOS
 
-// When ESS feedback to grid is enabled, show an arrow indicating the flow direction.
+// When the power is negative, show the "<" green arrow, and don't show a minus symbol. Do this
+// regardless of whether ESS feedback is enabled.
```
chriadam, line 11: Really?  "ESS feedback" is whether we allow feeding power back into the grid from the battery, right?  So, we should only show the < green arrow if ESS feedback is enabled, surely?

components/ElectricalQuantityLabel.qml
```diff
@@ -13,15 +13,13 @@ QuantityLabel {
...
-	readonly property bool _feedbackEnabled: Global.systemSettings.essFeedbackToGridEnabled && acInputMode
 	readonly property real _value: !_dataObjectValid ? NaN
 		: _unitAmps ? dataObject.current
 		: dataObject.power
 
-	// For AC inputs, if an AcInputDirectionIcon is present to indicate when power is negative,
-	// then the minus sign is not necessary. In this case, if feed-in to grid is enabled, omit
-	// the minus sign and just show the absolute value.
-	value: _feedbackEnabled ? Math.abs(_value) // will return NaN if _value is NaN.
+	// Foo AC inputs, the AcInputDirectionIcon should be present to indicate when power is negative,
```
chriadam, line 20: typo: Foo -> For

components/AcInputDirectionIcon.qml
chriadam: I just read the slack thread.  Very confusing...

chriadam:
I still see a negative symbol on the brief page side panel when the AC Input is a 3-phase Generator, and the feed-in exceeds draw.

components/ElectricalQuantityLabel.qml
blammit: Fixed

blammit:
> I still see a negative symbol on the brief page side panel when the AC Input is a 3-phase Generator, and the feed-in exceeds draw.

Turns out that was never implemented for the generator in the Brief side panel. Have added that now (last commit).

chriadam:
I guess it's not needed since a generator cannot consume power, I guess.  If the generator is feeding power into the grid, the feed-in indicator should show on the grid and not on the generator.

blammit:
> As per discussion, seems we don't need it

Never mind, we do need it. Added :)

chriadam (approved):
LGTM.

PR #1625, opened 2024-10-24
PV inverter: make meaning of limiting option clear

izak:
Add some text to indicate what the limiter option is for. It is so a user can disable it where it interferes with normal operation.

Also, hide the limiter option if the PV-inverter is not to be shown to the user, ie if it belongs to another installation and is not used here.

### DIFF
--- a/pages/settings/PageSettingsFroniusInverter.qml
+++ b/pages/settings/PageSettingsFroniusInverter.qml
@@ -54,6 +54,7 @@ Page {
 			}
 
 			ListRadioButtonGroupNoYes {
+				id: isActive
 				//% "Show"
 				text: qsTrId("page_settings_fronius_inverter_show")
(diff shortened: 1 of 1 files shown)


pages/settings/PageSettingsFroniusInverter.qml
```diff
@@ -63,11 +64,17 @@ Page {
 				//% "Power limiting"
 				text: qsTrId("page_settings_fronius_inverter_power_limiting")
 				dataItem.uid: bindPrefix + "/EnableLimiter"
-				allowed: limiterSupportedItem.value === 1
+				allowed: isActive.dataItem.value === 1 && limiterSupportedItem.value === 1
```
blammit, line 67: You can do this without adding the `isActive` id, as it is in the same scope. But LGTM either way.

blammit (approved):

PR #1635, opened 2024-10-28
Actually ensure heatpump services appear in Device List

### LINKED ISSUE #1634: heatpump services not showing in Device List
65d9b36d13df52111e6dba47894ad322d3ac57ce is not sufficient as heatpump services are not in AllDevicesModel.

blammit:
65d9b36 is not sufficient; add heatpump services to AllDevicesModel so that these services appear as delegates in the Device List.

It is not necessary to add dbus/HeatPumpsImpl.qml and mqtt/HeatPumpsImpl.qml since the list of heatpump services is only used in AllDevicesModel, and there is no need for an always-present model like Global.heatPumps.

Also update AcInDeviceModel so that it can read mock services as well.

Fixes #1634

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -836,6 +836,7 @@ set(Mock_QML_MODULE_SOURCES
     data/mock/EssImpl.qml
     data/mock/EvChargersImpl.qml
     data/mock/GeneratorsImpl.qml
+    data/mock/HeatPumpsImpl.qml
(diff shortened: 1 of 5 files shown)


blammit:
Should be able to verify via VRM as well once access to https://vrm.victronenergy.com/installation/473156/dashboard is added.

chriadam (approved):

PR #1638, opened 2024-10-28
Fix list of locales used in the vkb

### LINKED ISSUE #1632: Chinese and other fonts rendering issue
https://community.victronenergy.com/t/gui-v2-3-50-37-on-screen-keyboard-input-issue-chinese/8925/1

DanielMcInnes:
Fixes #1632

### DIFF
--- a/components/InputPanel.qml
+++ b/components/InputPanel.qml
@@ -5,6 +5,7 @@
 
 import QtQuick
 import QtQuick.VirtualKeyboard as QtVirtualKeyboard
+import QtQuick.VirtualKeyboard.Settings
 import Victron.VenusOS
 
 // *** This file can be edited directly on the cerbo filesystem,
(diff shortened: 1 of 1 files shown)


components/InputPanel.qml
```diff
@@ -108,4 +109,27 @@ QtVirtualKeyboard.InputPanel {
...
+				if (status === FontLoader.Ready) {
+					languageModel.setFontFamily(source, name)
+				}
+			}
+		}
+	}
+
+
+	Component.onCompleted: {
+		VirtualKeyboardSettings.activeLocales = ["en_US", "cs_CZ", "da_DK", "de_DE", "es_ES", "fr_FR", "it_IT", "nl_NL", "pl_PL", "ru_RU", "ro_RO", "sv_SE", "th_TH", "tr_TR", "uk_UA", "zh_CN", "ar_AR"]
```
chriadam, line 133: Do we also need to set VirtualKeyboardSettings.locale to match the locale set by gui-v2?  Or, does it pick it up automatically already?

chriadam: e.g. `property string localeName: Language.localeName; onLocaleNameChanged: VirtualKeyboardSettings.locale = localeName` (then might need to add Q_PROPERTY(QString localeName ... ) returning QLocale::name(), to language.h/cpp)

chriadam: Well, we can do this in a different PR, it shouldn't block this one.

chriadam (approved):

PR #1642, opened 2024-10-28
BatteryListPage: don't confuse uids of virtual and real batteries

### LINKED ISSUE #1572: Battery detail page data issues

blammit:
The /Batteries list may have virtual battery entries, with an id like 'com.victronenergy.battery.ttyUSB1:1' and no 'instance' value. For these entries, use the id as the device uid in the device model, instead of assigning a default DeviceInstance=0. Otherwise, it may conflict with the uid of a real battery.

Fixes #1572

### DIFF
--- a/data/mock/BatteriesImpl.qml
+++ b/data/mock/BatteriesImpl.qml
@@ -106,12 +106,13 @@ QtObject {
 	}
 
 	function updateBatteriesList() {
+		const dummyBatteryServiceName = dummyBattery.serviceUid.substring(BackendConnection.uidPrefix().length + 1)
 		const batteryList = [
 			{
 				// System battery
 				active_battery_service: true,
 				current: dummyBattery.current,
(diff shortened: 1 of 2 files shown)


chriadam (approved):
LGTM

Issue #1637, opened 2024-10-28
VE.Bus AC Input 1 and 2 issues


mpvader:
@vejpasop I encountered this by accident, while looking into an unrelated issue where the current limit of EVCS can't be set via gui-v2. Please take this over from me, test more completely and propose how it should be work + list of test conditions.

(Possible) issues that I've seen: 

### 1. Both the input current limit entries shown in the list have the same name.

Whats expected instead is that they have different names; or we show just one.

Twice the same name in the control card:
![image](https://github.com/user-attachments/assets/09998fc9-46df-4e8d-a5f6-c023665f46eb)

Twice the same name shown in details page:
![image](https://github.com/user-attachments/assets/c95ddd47-9e3b-4a3b-be8d-be0cfe89bec7)

### 2. the 50A is shown as editable but its not

Better to show it as a non-editable version right from the get go and a toast explaining why when user clicks that.

![image](https://github.com/user-attachments/assets/5d022c04-fbad-431d-beb9-be3379ba2bb3)

### 3. Notification text needs amending. 

It now says "This setting is disabled. Possible reasons are "Overruled by remote" is not enabled or an assistant is preventing the adjustment. Please, check the inverter configuration with VEConfigure". In 99% of cases, its disabled by design by the installer, lets not tell end-users to go into VEConfigure.

New text: "This current limit is fixed in the system configuration, it is not possible to adjust it".

blammit:
PR https://github.com/victronenergy/gui-v2/pull/1643 fixes this.

Note that this only works if the vebus service is the primary one on the system, according to `com.victronenergy.system/VebusService`. Otherwise, it is assumed the `/SystemSetup/AcInput<X>` cannot be applied for this vebus service, and it will still just show "Input current limit" instead of something more specific.

vejpasop:
Successfully tested on Venus V3.50~42.  
![image](https://github.com/user-attachments/assets/75172fff-ce73-46c2-a39c-3193337ff7c7)
![image](https://github.com/user-attachments/assets/5f6fce3b-d90d-4e9e-a2cf-f134db2e6d66)
![image](https://github.com/user-attachments/assets/f12f40bc-92ce-4361-808e-56a1e8abc154)
![image](https://github.com/user-attachments/assets/6ae8ef18-7804-4671-bfbc-aa146ddd137c)
![image](https://github.com/user-attachments/assets/66b85489-1d76-410b-8c2d-35b74b7be399)
[...]

PR #1643, opened 2024-10-29
vebus input fixes

blammit:
Fixes https://github.com/victronenergy/gui-v2/issues/1637

### DIFF
--- a/components/InverterChargerModeButton.qml
+++ b/components/InverterChargerModeButton.qml
@@ -14,14 +14,20 @@ ListItemButton {
 	readonly property string serviceType: BackendConnection.serviceTypeFromUid(serviceUid)
 	readonly property int writeAccessLevel: VenusOS.User_AccessType_User
 	readonly property bool userHasWriteAccess: Global.systemSettings.canAccess(writeAccessLevel)
+	readonly property bool modeAdjustable: modeIsAdjustable.isValid && !!modeIsAdjustable.value
 
 	text: serviceType !== "inverter" || isInverterChargerItem.value === 1
 			? Global.inverterChargers.inverterChargerModeToText(modeItem.value)
 			: Global.inverterChargers.inverterModeToText(modeItem.value)
+
+	// TODO need to show a different indicator (like in settings pages) when a control is disabled
+	// due to reduced user access level. This is different from when the control is disabled due to
+	// the mode not being adjustable.
 	enabled: userHasWriteAccess
+	showEnabled: modeAdjustable
 
 	onClicked: {
-		if (modeIsAdjustable.isValid && !modeIsAdjustable.value) {
+		if (!modeAdjustable) {
 			if (dmc.isValid) {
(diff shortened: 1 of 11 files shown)


components/InverterChargerModeButton.qml
```diff
@@ -14,22 +14,29 @@ ListItemButton {
...
+
+	// TODO need to show a different indicator (like in settings pages) when a control is disabled
+	// due to reduced user access level. This is different from when the control is disabled due to
+	// the mode not being adjustable.
 	enabled: userHasWriteAccess
+	showEnabled: modeAdjustable
 
 	onClicked: {
-		if (modeIsAdjustable.isValid && !modeIsAdjustable.value) {
+		if (!modeAdjustable) {
```
blammit, line 30: There's a behavioural change here. Previously, it would only show the notification if /ModeIsAdjustable was valid (and set to 0). Now it will show the notification even if /ModeIsAdjustable is not valid. But looking at gui-v1, I think this change aligns with the original logic more closely.

chriadam (approved):
One question: in my test on device using a demo, the text shown on the button was "Grid current limit" and when pressing it, the dialog shows: "Input current limit\nAC input 2".  I assume that this was because the "first device" happened to have the name "AC input 2"?  If so, fine, although slightly confusing (I mean, why was "AC input 1" not considered the first device?).  Also, strange that there is a text mismatch (Grid current limit vs Input current limit) but that is probably expected as we reuse the same modal dialog for different ac input types?

But, LGTM.

components/controls/Button.qml
```diff
@@ -12,17 +12,18 @@ import Victron.VenusOS
 CT.Button {
 	id: root
 
-	property color color: !enabled ? Theme.color_font_disabled
+	property color color: !showEnabled ? Theme.color_font_disabled
 			: down ? Theme.color_button_down_text
 			: Theme.color_font_primary
-	property color backgroundColor: !enabled ? Theme.color_background_disabled
+	property color backgroundColor: !showEnabled ? Theme.color_background_disabled
```
DanielMcInnes, line 18: the background color changes look good tom - are they in the design somewhere? I had a quick look and couldn't see what a disabled button should look like
image

DanielMcInnes (approved):

components/controls/Button.qml
blammit: Not in the design, they just do the job for now -- might have to update them later on.

blammit:
> One question: in my test on device using a demo, the text shown on the button was "Grid current limit" and when pressing it, the dialog shows: "Input current limit\nAC input 2". I assume that this was because the "first device" happened to have the name "AC input 2"? If so, fine, although slightly confusing (I mean, why was "AC input 1" not considered the first device?). Also, strange that there is a text mismatch (Grid current limit vs Input current limit) but that is probably expected as we reuse the same modal dialog for different ac input types?

Added a patch to show e.g. "Grid current limit" if possible.

chriadam (approved):
LGTM.

PR #1644, opened 2024-10-29
Ensure the VKB locale matches the chosen language

chriadam:
Contributes to issue #1632

### DIFF
--- a/components/InputPanel.qml
+++ b/components/InputPanel.qml
@@ -21,6 +21,8 @@ QtVirtualKeyboard.InputPanel {
 	property real toContentY
 	property real toHeight
 
+	readonly property string localeName: Language.currentLocaleName
+
 	function acceptMouseEvent(item, itemMouseX, itemMouseY) {
 		if (!Qt.inputMethod.visible || !item || !focusedItem) {
 			return false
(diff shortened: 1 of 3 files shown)


components/InputPanel.qml
```diff
@@ -128,8 +130,9 @@ QtVirtualKeyboard.InputPanel {
 		}
 	}
 
-
+	onLocaleNameChanged: VirtualKeyboardSettings.locale = localeName
 	Component.onCompleted: {
 		VirtualKeyboardSettings.activeLocales = ["en_US", "cs_CZ", "da_DK", "de_DE", "es_ES", "fr_FR", "it_IT", "nl_NL", "pl_PL", "ru_RU", "ro_RO", "sv_SE", "th_TH", "tr_TR", "uk_UA", "zh_CN", "ar_AR"]
+		VirtualKeyboardSettings.locale = localeName
```
DanielMcInnes, line 136: If you change the gui language to Arabic, we change Language.currentLocaleName to 'ar_EG' (Arabic - Egypt), which is not supported by the vkb, so the changes don't get applied. The only Arabic that the vkb knows about is 'ar_AR' The vkb will stay on the previously selected language. Apart from that, lgtm.

DanielMcInnes: FYI, vkb supports these locales: 
[ar_AR,bg_BG,cs_CZ,da_DK,de_DE,el_GR,en_GB,en_US,es_ES,es_MX,et_EE,fa_FA,fi_FI,fr_CA,fr_FR,he_IL,hi_IN,hr_HR,hu_HU,id_ID,it_IT,ja_JP,ko_KR,ms_MY,nb_NO,nl_NL,pl_PL,pt_BR,pt_PT,ro_RO,ru_RU,sk_SK,sl_SI,sq_AL,sr_SP,sv_SE,th_TH,tr_TR,uk_UA,vi_VN,zh_CN,zh_TW]

chriadam: Thanks!  I will add a hardcoded fixup for this case.

DanielMcInnes (approved):
lgtm

PR #1645, opened 2024-10-29
DeviceList: avoid duplicate tanks after changing fluid type

### LINKED ISSUE #1631: Entry in the device list doubles after changing settings?
first step is to see if this is reproducible.

https://community.victronenergy.com/t/small-bugs-3-50-37/8943/4

Note that Item two in that post, a suggestion, has been listed listed elsewhere already to look into later (gui-v2 doc). No need for action on that. 

for me this  bug is not a blocker for v3.50 release. Totally ok to leave it, to prevent breaking other things if fix is non trivial.

IMG_3335

blammit:
If a tank moves from one tank model to another, due to changing its fluid type, then AggregateDeviceModel must remove the tank from the old model.

Normally AggregateDeviceModel does not remove device entries because they are typically only removed from their source models when they are disconnected, and in that case, the AggregateDeviceModel just shows them as "Not connected" instead of removing them. However, in this case, the devices should be removed from AggregateDeviceModel, as they are still valid and not disconnected.

Fixes #1631

### DIFF
--- a/src/aggregatedevicemodel.cpp
+++ b/src/aggregatedevicemodel.cpp
@@ -88,6 +88,7 @@ void AggregateDeviceModel::setSourceModels(const QVariantList &models)
 			connect(device, &BaseDevice::validChanged, this, &AggregateDeviceModel::deviceValidChanged);
 		}
 		connect(model, &BaseDeviceModel::rowsInserted, this, &AggregateDeviceModel::sourceModelRowsInserted);
+		connect(model, &BaseDeviceModel::rowsAboutToBeRemoved, this, &AggregateDeviceModel::sourceModelRowsAboutToBeRemoved);
 	}
 	m_sourceModels = models;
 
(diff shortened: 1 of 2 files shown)


src/aggregatedevicemodel.cpp
```diff
@@ -230,6 +231,42 @@ void AggregateDeviceModel::cleanUp()
...
+	// model. (If the device is invalid, do not remove it; in that case, it should remain in this
+	// model but be marked as "Not connected", as per deviceValidChanged().)
+	for (int i = last; i >= first; --i) {
+		BaseDevice *device = sourceModel->deviceAt(i);
+		if (!device) {
+			qmlInfo(this) << "cannot find device to remove at index " << i << " from model "
+						  << sourceModel->objectName() << " with count:" << sourceModel->count();
+			continue;
+		}
+		if (device->isValid()) {
```
chriadam, line 254: seems simple enough, I just have one question: is it ever possible that a disconnected device will have its validity change asynchronously?  Or is that impossible, i.e. the only other way the source model knows to remove the device (in the disconnection case) is if its isValid() goes to false?

DanielMcInnes (approved):
lgtm

src/aggregatedevicemodel.cpp
blammit: Yes, a disconnected device could be reconnected, and in that case it should be re-added to the source model, and then in sourceModelRowsInserted() we will make change it from "Not connected" back into a normal connected device.

Typically we add a device to a source model when valid=true, and remove it when valid=false. But it's not guaranteed that it will always happen in this way.

src/aggregatedevicemodel.cpp
```diff
@@ -230,6 +231,42 @@ void AggregateDeviceModel::cleanUp()
...
+			const int removalIndex = indexOf(DeviceInfo::infoId(device, sourceModel));
+			if (removalIndex >= 0) {
```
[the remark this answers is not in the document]
blammit: Done

src/aggregatedevicemodel.cpp
chriadam: I think I explained myself poorly.  My question was meant to be: is there a case where the source model removes the row, and valid=true, but really the device is just disconnected (and so we should NOT remove it from m_devicesInfos (I assume in this case, valid would change to false asynchronously)?

Or, will valid always already equal false here if it has been disconnected (and that is the reason that it was removed from the source model)?

chriadam (approved):

src/aggregatedevicemodel.cpp
blammit: We shouldn't run into this, as we should be removing devices from device models when the device valid=false. But, this is not actually enforced anywhere.

If this case happens, then the device will be removed from the aggregate model and then the device would become invalid later on. Which might be problematic for the source DeviceModel if it doesn't get removed there, but should be okay for the aggregate model, I think, since it has disconnected from the signals of the Device.

PR #1656, opened 2024-10-30
Set 'layer.enabled' to false wherever text is displayed

### LINKED ISSUE #1639: Inverter/charger power values render strange
all other numbers look nice and smooth, but these two do not. Is that just how it is (then fine for me) or small mistake somewhere?

this is seen on WASM; and also if I enlarge or decrease the size of my webbrowser they remain a pixelated, while all other numbers show nicely, even when enlarged a lot.

image

DanielMcInnes:
On wasm builds, text looks rough wherever layer.enabled is true Fixes #1639

### DIFF
--- a/components/ThreePhaseQuantityTable.qml
+++ b/components/ThreePhaseQuantityTable.qml
@@ -46,6 +46,7 @@ QuantityTable {
 	]
 	labelHorizontalAlignment: Qt.AlignRight
 	headerComponent: AsymmetricRoundedRectangle {
+		layer.enabled: false // if 'layer.enabled' is true, any child text looks rough on wasm builds
 		width: root.width
 		height: Theme.geometry_vebusDeviceListPage_quantityTable_header_height
 		roundedSide: VenusOS.AsymmetricRoundedRectangle_RoundedSide_Top
--- a/components/SegmentedButtonRow.qml
+++ b/components/SegmentedButtonRow.qml
@@ -6,6 +6,7 @@
 import QtQuick
 import QtQuick.Controls as C
 import QtQuick.Controls.impl as CP
+import QtQuick.Templates as T
 import Victron.VenusOS
 
 Item {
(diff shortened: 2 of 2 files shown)


components/ThreePhaseQuantityTable.qml
```diff
@@ -46,6 +46,7 @@ QuantityTable {
 	]
 	labelHorizontalAlignment: Qt.AlignRight
 	headerComponent: AsymmetricRoundedRectangle {
+		layer.enabled: false // if 'layer.enabled' is true, any child text looks rough on wasm builds
```
chriadam, line 49: was a layer enabled deliberately for AsymmetricRoundedRectangle previously (e.g. to ensure that transparency-stacking doesn't occur, or something), or is the layer involved here just the supersample layer for antialiasing?

DanielMcInnes: AsymmetricRoundedRectangle has the following comment:

```
	// ensure that the entire texture is faded equally during animations
	// otherwise the clip messes with this on WASM platform.
	layer.enabled: true
```
so it sounds like something to do with animations does wrong is wasm builds

chriadam: RIght.  So we can't set `layer.enabled: false` in this case.

I wonder if we can just only use the AsymmetricRoundedRectangle in the BACKGROUND of the control, so that the Text element isn't a child of the AsymmetricRoundedRectangle (and thus the text won't be pre-rendered to the layer and then blitted with downscaling to the output buffer).

DanielMcInnes: I updated the SegmentedButtonRow as suggested. I've left AsymmetricRoundedRectangle in ThreePhaseQuantityTable as is, as there aren't any animations to worry about there

chriadam (approved):

PR #1657, opened 2024-10-30
Overview solar: fix multi-phase display for PV inverters

### LINKED ISSUE #1654: Overview page: Solar yield widget: undefined 0 W
I have multiple tests systems. On the test systems with more than one phase I'm seeing "undefined: 0 W" in the solar yield tile on the overview page. There is only an AC coupled inverter with the same phase amount as the rest of the system. Shouldn't in this case the phases be displayed as in the other tiles?

This requirements are met, so that [ThreePhaseDisplay.qml](https://github.com/victronenergy/gui-v2/blob/main/components/ThreePhaseDisplay.qml) is displayed.

https://github.com/victronenergy/gui-v2/blob/a64c1f1e4399ce6c6ced6f031af8514e1ea3d158/components/widgets/SolarYieldWidget.qml#L56

Top left: 1 phase
Top right: 1 phase
Bottom left: 2 phase
Bottom right: 3 phase

grafik

blammit:
There are two use cases for showing PV inverter multi-phase data:

1) Overview SolarYieldWidget, which shows a ThreePhaseDisplay for the PV inverter. ThreePhaseDisplay needs a QAbstractItemModel-based model to create the display delegates using the model roles.

2) PvInverterPage, which needs a QObject for each phase to create a property binding to each phase measurement property (power, current, etc.) so that valueForModelIndex() will be invoked when a measurement value changes.

Change PvInverter to use PhaseModel provide the QAbstractItemModel- based model for case 1, and also provide a getPhase() function for case 2.

QuantityTable could be reworked in future to avoid the need for a valueForModelIndex() function that requires QObject-bindings, thereby removing the need for case 2.

Fixes the regression from a9038b76bfddc53ebc3c3db55061ad4bd671556a. EvCharger does not need a similar fix, since the QAbstractItemModel- based model is not required for current EVCS features in the UI.

Fixes #1654

### DIFF
--- a/data/common/PvInverter.qml
+++ b/data/common/PvInverter.qml
@@ -17,14 +17,14 @@ Device {
 	readonly property real power: _power.isValid ? _power.value : NaN
 	readonly property real voltage: _voltage.isValid ? _voltage.value : NaN
 
-	readonly property QtObject phases: QtObject {
-		property int count
-
+	readonly property PhaseModel phases: PhaseModel {
 		function updateCount(maxPhaseCount) {
-			count = Math.max(count, maxPhaseCount)
+			// pvinverter services do not have /NumberOfPhases, so manually update the phase model
+			// count when phase measurements are detected.
+			phaseCount = Math.max(count, maxPhaseCount)
 		}
 
-		function get(index) {
+		function getPhase(index) {
 			return _phases.objectAt(index)
 		}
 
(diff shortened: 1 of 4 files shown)


blammit:
Can be tested with:

- venus-docker, e.g. with `./simulate.sh --with-pvinverter w`. Be sure to use a simulation that does not have a solar charger; if there are both solar chargers and PV inverters, the Overview solar widget will not show the 3-phase display.
- mock mode, with e.g. configuration number 2, which has a 3-phase PV inverter.

data/common/PvInverter.qml
```diff
@@ -17,14 +17,14 @@ Device {
...
+	readonly property PhaseModel phases: PhaseModel {
 		function updateCount(maxPhaseCount) {
-			count = Math.max(count, maxPhaseCount)
+			// pvinverter services do not have /NumberOfPhases, so manually update the phase model
+			// count when phase measurements are detected.
+			phaseCount = Math.max(count, maxPhaseCount)
 		}
 
-		function get(index) {
+		function getPhase(index) {
```
DanielMcInnes, line 27: is this different to `PhaseModel::get(int index)`?

DanielMcInnes (approved):

chriadam (approved):

data/common/PvInverter.qml
blammit: This returns a QtObject, whereas PhaseMode::get() returns a QVariantMap; I renamed this function so it wouldn't conflict with PhaseModel::get(), and to try to make it clear that this returns a phase object.

PR #1658, opened 2024-10-30
Overview: don't connect EVCS to Inverter/Charger

### LINKED ISSUE #1624: Overview: Inconsistent "wiring" of the EVCS tile
EVCS tile should always be connected to either "AC Loads" or "Essential Loads", but never "Inverter/Charger". Currently, that's not the case if there is only a "AC Loads" and no "Essential Loads" tile.

192 168 254 95_gui-v2_(Ekrano GX)

blammit:
When AC Loads are combined (rather than split into separate AC Loads and Essential Loads), connect the EVCS box to the AC Loads box, instead of the Inverter/Charger box.

Fixes #1624

### DIFF
--- a/pages/OverviewPage.qml
+++ b/pages/OverviewPage.qml
@@ -840,23 +840,22 @@ SwipeViewPage {
 		EvcsWidget {
 			id: evcsWidget
 
-			// The EVCS widget may have connectors to the Inverter/Charger, AC Loads or Essential
-			// Loads, depending on whether AC loads are split, and the "Position" configuration
-			// of the EV chargers on the system.
+			// The EVCS widget may have connectors to the AC Loads or Essential Loads, depending on
+			// whether AC loads are split into AC Loads + Essential Loads, and also depending on the
+			// "Position" configuration of the EV chargers on the system.
 			//
 			// If showing combined AC loads:
-			//  - connect to Inverter/Charger
+			//  - connect to AC Loads
 			// If splitting loads into (input) AC Loads and (output) Essential Loads:
 			//  - connect to AC Loads, if there are any EV chargers with /Position=1 (AC-In)
 			//  - connect to Essential Loads, if there are any EV chargers with /Position=0 (AC-Out)
-			readonly property bool connectToInverterCharger: visible && !Global.system.showInputLoads
 			readonly property bool connectToAcLoads: visible
-					&& Global.system.showInputLoads
-					&& Global.evChargers.acInputPositionCount > 0
+					&& (!Global.system.showInputLoads  // AC loads are combined
+						|| Global.evChargers.acInputPositionCount > 0) // AC loads are split, and AC-in position is in use
 			readonly property bool connectToEssentialLoads: visible
-					&& Global.system.showInputLoads
-					&& Global.system.hasAcOutSystem
-					&& Global.evChargers.acOutputPositionCount > 0
+					&& Global.system.showInputLoads     // AC loads are split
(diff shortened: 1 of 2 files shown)


pages/OverviewPage.qml
```diff
@@ -840,23 +840,22 @@ SwipeViewPage {
...
+			//  - connect to AC Loads
 			// If splitting loads into (input) AC Loads and (output) Essential Loads:
 			//  - connect to AC Loads, if there are any EV chargers with /Position=1 (AC-In)
 			//  - connect to Essential Loads, if there are any EV chargers with /Position=0 (AC-Out)
-			readonly property bool connectToInverterCharger: visible && !Global.system.showInputLoads
 			readonly property bool connectToAcLoads: visible
-					&& Global.system.showInputLoads
-					&& Global.evChargers.acInputPositionCount > 0
+					&& ( !Global.system.showInputLoads  // AC loads are combined
+						|| (Global.system.showInputLoads && Global.evChargers.acInputPositionCount > 0) ) // AC loads are split, and AC-in position is in use
```
DanielMcInnes, line 854: this expression can be simplified a bit
```
( !Global.system.showInputLoads  // AC loads are combined
    || Global.evChargers.acInputPositionCount > 0 )
```

pages/OverviewPage.qml
```diff
@@ -840,23 +840,22 @@ SwipeViewPage {
...
 			readonly property bool connectToEssentialLoads: visible
-					&& Global.system.showInputLoads
-					&& Global.system.hasAcOutSystem
-					&& Global.evChargers.acOutputPositionCount > 0
+					&& Global.system.showInputLoads     // AC loads are split
+					&& Global.system.hasAcOutSystem     // Essential Loads should be visible
+					&& Global.evChargers.acOutputPositionCount > 0  // AC-in position is in use
```
DanielMcInnes, line 858: should the comment be: `// AC-out position is in use`

pages/OverviewPage.qml
```diff
@@ -840,23 +840,22 @@ SwipeViewPage {
...
-					&& Global.system.showInputLoads
-					&& Global.evChargers.acInputPositionCount > 0
+					&& ( !Global.system.showInputLoads  // AC loads are combined
+						|| (Global.system.showInputLoads && Global.evChargers.acInputPositionCount > 0) ) // AC loads are split, and AC-in position is in use
 			readonly property bool connectToEssentialLoads: visible
-					&& Global.system.showInputLoads
-					&& Global.system.hasAcOutSystem
-					&& Global.evChargers.acOutputPositionCount > 0
+					&& Global.system.showInputLoads     // AC loads are split
+					&& Global.system.hasAcOutSystem     // Essential Loads should be visible
```
DanielMcInnes, line 857: Is it worth adding a property `Global.system.showEssentialLoads` ?

pages/OverviewPage.qml
blammit: Updated, thanks

pages/OverviewPage.qml
blammit: Oops, fixed

pages/OverviewPage.qml
blammit: I'm not sure, it might be confusing as `showInputLoads` means "Split into AC loads + Essential Loads" but `hasAcOutSystem` still has to be `true` to get the Essential Loads to be shown. These property names map to the ones used in gui-v1 to use consistent terminology.

DanielMcInnes (approved):

chriadam (approved):

PR #1663, opened 2024-10-31
Overview: allow drilldown on non-active AC input

### LINKED ISSUE #1405: Generator drilldown on WASM doesn't work
On Venus OS 3.50~14. 
The WASM correctly shows the blue outline on the generator widget:
image

But the drilldown does not work. It opens an empty page. Locally, this works properly. 
This is on a system with a vebus system connected, if that matters. This can be tested with the DSE simulation on venus-docker.

@blammit
I suppose we need to publish the instance number of the main generator? 
As indicated by @izak's comment [here](https://github.com/victronenergy/vrm-portal/issues/314#issuecomment-1980630339) (issue needs to be reopened when this is fixed) this may also be useful for VRM.

blammit:
Any AC input widgets on the Overview page should be clickable and take the user to the drilldown/settings page, even if it is not the currently active AC input.

Also allow PageAcIn.qml and PageRsSystem.qml to provide the device name internally, instead of relying on the launching page to provide the name via the 'title' property.

Fixes #1405

### DIFF
--- a/components/widgets/AcInputWidget.qml
+++ b/components/widgets/AcInputWidget.qml
@@ -23,7 +23,7 @@ AcWidget {
 	quantityLabel.leftPadding: acInputDirectionIcon.visible ? (acInputDirectionIcon.width + Theme.geometry_acInputDirectionIcon_rightMargin) : 0
 	quantityLabel.acInputMode: true
 	phaseCount: connected ? input.phases.count : 0
-	enabled: !!input
+	enabled: !!inputInfo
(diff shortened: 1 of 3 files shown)


blammit:
Can test with:

- https://vrm.victronenergy.com/installation/473156/dashboard
- Any mock configuration with multiple AC inputs, e.g. number 25 (Grid disconnected)

chriadam (approved):

PR #1664, opened 2024-10-31
Disable onboarding if touch is disabled

### LINKED ISSUE #1660: Don't show onboarding if touch is disabled
It's difficult to get past the onboarding if touch is disabled and the cursor is not visible. So, don't show onboarding if touch is disabled. 

This is provided by the `/Settings/Gui/TouchEnabled` setting.

blammit:
It's hard to skip onboarding without touch, so disable onboarding if touch is disabled.

Fixes #1660

### DIFF
--- a/data/SystemSettings.qml
+++ b/data/SystemSettings.qml
@@ -11,6 +11,8 @@ QtObject {
 
 	readonly property string serviceUid: BackendConnection.serviceUidForType("settings")
 	readonly property bool needsOnboarding: _onboardingState.needsOnboarding
+			// It's hard to skip onboarding without touch, so disable onboarding if touch is disabled.
+			&& _touchEnabled.isValid && _touchEnabled.value !== 0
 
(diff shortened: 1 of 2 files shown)


data/SystemSettings.qml
```diff
@@ -11,6 +11,8 @@ QtObject {
 
 	readonly property string serviceUid: BackendConnection.serviceUidForType("settings")
 	readonly property bool needsOnboarding: _onboardingState.needsOnboarding
+			// It's hard to skip onboarding without touch, so disable onboarding if touch is disabled.
+			&& _touchEnabled.value !== 0
```
chriadam, line 15: if the value is null or undefined, I guess needsOnboarding would still be true in this case?

blammit: Hm. Yes if we don't know whether touch is enabled, then probably better to skip the onboarding. Fixed.

chriadam (approved):

Issue #1633, opened 2024-10-27
VRM Portal setting is not set back to full when clicking cancel


mpvader:
This is about the Are you sure popup thats shown when in WASM as a user changing the VRM Portal setting from Full to something else.

When I click cancel, it says Read-only. I expect it to then have gone back to Full.

![image](https://github.com/user-attachments/assets/4b3fc722-e65a-4a0d-bffb-7be277ea9932)

chriadam:
I will take this one.  I hope the fix is basically just to set the "don't update currentIndex on click" flag.

chriadam:
Fixed.

PR #1672, opened 2024-11-01
Don't prematurely update the current index in VRM portal options

chriadam:
The default behaviour of ListRadioButtonGroup is to set the current index to the clicked option immediately upon click.

In the VRM portal options case, we don't want to do this. Instead, the current index should always track the backend value, and we show the confirmation dialog on click, before attempting to set the backend value.

Thus we need to set the updateCurrentIndexOnClick flag to false.

Fixes issue #1633

### DIFF
--- a/pages/settings/PageSettingsLogger.qml
+++ b/pages/settings/PageSettingsLogger.qml
@@ -45,6 +45,7 @@ Page {
 				text: qsTrId("settings_logging_vrm_portal")
 				dataItem.uid: Global.systemSettings.serviceUid + "/Settings/Network/VrmPortal"
 				updateDataOnClick: false
+				updateCurrentIndexOnClick: false // track backend value, set only when confirmation dialog is accepted.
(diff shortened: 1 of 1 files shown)


pages/settings/PageSettingsLogger.qml
```diff
@@ -45,6 +45,7 @@ Page {
 				text: qsTrId("settings_logging_vrm_portal")
 				dataItem.uid: Global.systemSettings.serviceUid + "/Settings/Network/VrmPortal"
 				updateDataOnClick: false
+				updateCurrentIndexOnClick: false
```
blammit, line 48: Can you add a comment to say this is because you only want the selected option to change after the dialog box has been confirmed?

blammit (approved):

pages/settings/PageSettingsLogger.qml
chriadam: done

Issue #1661, opened 2024-10-30
Ukrainian translation of battery and settings menu header


dirkjanfaber:
In Ukrainian the word "Battery" no longer gets translated into  "Батарея". See https://github.com/victronenergy/gui-v2/blob/main/i18n/venus-gui-v2_uk.ts#L4425

Strangely enough this is not problem for other languages, which correctly translate the term. 
![image](https://github.com/user-attachments/assets/89a4a390-47f5-4dfd-af12-04bd9520b427)

And before v3.50~37 it was working correctly:
![image](https://github.com/user-attachments/assets/56414c42-759e-4760-b3ff-748db2857909)

Feel free to test on https://vrm.victronenergy.com/installation/102195/dashboard

And also noticing that, when you switch translations, the menu header doesn't get updated until you re-enter the settings menu.
![image](https://github.com/user-attachments/assets/b7affaca-45aa-4e0a-af66-a9998efba610)

podarok:
Tnx

imval:
There is a similar issue for the French translation with the word "Automatic" that is not correctly translated in the UI in a few places, even though the key is present in Poeditor (the translation platform).
![image](https://github.com/user-attachments/assets/06ff772f-3eaf-4cd4-a9df-1be5e129ad09)
![image](https://github.com/user-attachments/assets/cfe23ab7-7ed1-4016-943e-7813069911b7)

Dirk-Jan asked me to add this in this issue.

podarok:
title seems like not only Ukrainian related @dirkjanfaber 
I have no rights to change it, on you

chriadam:
The id for the "Battery" string is "common_words_battery", but this is missing from the locale-specific .ts files in the repository, although it IS present in the locale-specific .ts file in the build artifacts directory, from which the .qm files are generated... Interestingly, the translation is marked as "unfinished" there!

```
    <message id="common_words_battery">
        <location filename="../../gui-v2/components/CommonWords.qml" line="83"/>
        <source>Battery</source>
        <translation type="unfinished">Батарея</translation>
    </message>
```

chriadam:
I think we're running lrelease "too soon" i.e. on a version of the .ts which hasn't been properly updated by lupdate.
We have some dependency logic in the CMakeLists.txt file to try to enforce proper ordering, but it seems like this might be failing.

dirkjanfaber:
[...]

PR #1678, opened 2024-11-01
Update in-source translation files to build baseline

chriadam:
We previously disabled automatic-update of in-source .ts files. This means that we rely on lupdate to update the build-dir copies before lrelease runs on those.

In some cases, this is causing problems.

This commit updates the in-source versions with the latest build-dir copies, so that at least for now, it must be in sync.

TODO: fix this properly in future.

Contributes to issue #1661


dirkjanfaber:
This needs to be done just after downloading the new translations from poeditor. So don't think this PR is needed.
I'll adjust the scripting for the downloading to also build the project and copy over the generated files.

Issue #1685, opened 2024-11-03
Relay UI: Celsius/Fahrenheit bug


mpvader:
https://community.victronenergy.com/t/cerbo-venus-3-50-fahrenheit-bug-in-relay-ui/10211

mpvader:
fixed already right @chriadam ?

chriadam:
Yes, fixed, closing.

PR #1691, opened 2024-11-04
Convert relay temperatures to display unit

chriadam:
Contributes to issue #1685


chriadam:
I'm having difficulty testing this one.  The only installation I found with a temperature mode relay was 68572 but navigating to Settings -> Relay -> Temperature Control Rules shows just an empty page...

blammit (approved):

PR #1694, opened 2024-11-04
Overview: animate AC Loads -> EVCS connection when loads are combined

### LINKED ISSUE #1693: Overview: AC Loads -> EVCS connection doesn't animate when loads are combined
See slack discussion: https://victrondevelopment.slack.com/archives/C020QG3JVEW/p1730695967989319

blammit:
a9208f086c73faf36bf37ffb029e2a216afeb292 removed the Inverter/Charger -> EVCS connector, so that the AC Loads -> EVCS connector is used instead when AC loads are combined. However, the AC Loads -> EVCS connector also needs to be updated to animate the connection when needed, i.e. when there is EVCS power for either AC-in or AC-out positions.

The fix restores the connectToInverterCharger that was removed in a9208f086c73faf36bf37ffb029e2a216afeb292 and renames it to connectToCombinedAcLoads, to make the new bindings more clear.

Fixes #1693

### DIFF
--- a/data/mock/EvChargersImpl.qml
+++ b/data/mock/EvChargersImpl.qml
@@ -72,9 +72,9 @@ QtObject {
 				onTriggered: {
 					const zeroPower = Math.random() < 0.2
 					let totalPower = 0
-					for (let i = 0; i < evCharger.phases.count; ++i) {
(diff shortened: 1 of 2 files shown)


chriadam (approved):
LGTM

PR #1698, opened 2024-11-05
1552 implement connectivity icons

DanielMcInnes:


DanielMcInnes:
screenshots:

Screenshot from 2024-11-05 13-23-35
Screenshot from 2024-11-05 13-22-17
Screenshot from 2024-11-05 13-21-46

components/WifiModel.qml
```diff
@@ -0,0 +1,94 @@
+/*
+** Copyright (C) 2024 Victron Energy B.V.
```
blammit, line 2: Can you add more detail to this commit fb48198d02140e8af7b9a2f8a55d78f297930445 ? Has the code changed aside from moving it into a component with a ListModel base type?

blammit (commented):
Looks great. I noticed that if the "Silence alarm" or "Acknowledge alert" button is a bit longer, then it will run over the top of the wifi/gsm icons. Wonder if those button texts should be elided, or if the icons can just move to the other side of the clock time.

mr-manuel:
@blammit in future there will also be an additional icon for Tailscale, just to help you making decisions for the space

components/WifiModel.qml
```diff
@@ -0,0 +1,94 @@
...
+							  })
+					break
+				}
+			}
+
+			// Insert newly discovered networks
+			if (!found) {
+				// Services are sorted by favorite and signal strength, try to maintain order
+				model.insert(i, {
+								 "network": network,
```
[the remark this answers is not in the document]
DanielMcInnes: done

components/WifiModel.qml
DanielMcInnes: done

DanielMcInnes:
> Looks great. I noticed that if the "Silence alarm" or "Acknowledge alert" button is a bit longer, then it will run over the top of the wifi/gsm icons. Wonder if those button texts should be elided, or if the icons can just move to the other side of the clock time.

I tried moving the icons to the other side of the clock time, and it just looks wrong to me somewhow.
image

Eliding the button text is not easy to do. Our Button implementation uses QtQuick.Controls.impl.IconLabel for the contentItem, which doesn't expose an elide property. I'm not sure it is a problem if the alarm button does cover the connectivity buttons, if there is an alarm, it is more important than the wifi strength. What do you think? Ask Serj, or leave it as is?

blammit:
> Eliding the button text is not easy to do. Our Button implementation uses QtQuick.Controls.impl.IconLabel for the contentItem, which doesn't expose an elide property. I'm not sure it is a problem if the alarm button does cover the connectivity buttons, if there is an alarm, it is more important than the wifi strength. What do you think? Ask Serj, or leave it as is?

Yeah I think you're right, it's not a big deal, as you say the buttons have higher priority. LGTM.

blammit (approved):
LGTM but just need to check when to merge this exactly.

PR #1700, opened 2024-11-05
integrate HasAcOutSystem into system settings

izak:
HasAcOutSystem used to be an ESS specific setting. With the RS including ESS support internally, make this a system setting.

This also relies on systemcalc v2.201 to detect whether there is any sort of ESS functionality. Loads on the input side (only) makes no sense in a non-ESS system, and therefore that option is hidden.

The text that explains what the settings do are also updated to be more sensible.

https://github.com/victronenergy/venus-private/issues/450

### DIFF
--- a/pages/settings/PageSettingsHub4.qml
+++ b/pages/settings/PageSettingsHub4.qml
@@ -66,19 +66,11 @@ Page {
 			]
 		}
 
-		ListSwitch {
-			id: acOutInUse
-			//% "Inverter AC output in use"
-			text: qsTrId("settings_ess_inverter_ac_output_in_use")
-			dataItem.uid: Global.systemSettings.serviceUid + "/Settings/SystemSetup/HasAcOutSystem"
-			allowed: defaultAllowed && withoutGridMeter.currentIndex === 0
-		}
-
 		ListRadioButtonGroup {
 			//% "Self-consumption from battery"
 			text: qsTrId("settings_ess_self_consumption_battery")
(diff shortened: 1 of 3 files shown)


pages/settings/PageSettingsHub4.qml
```diff
@@ -279,6 +271,11 @@ Page {
 				uid: Global.system.serviceUid + "/Control/ScheduledSoc"
 			}
 
+			VeQuickItem {
+				id: acOutInUseItem
```
DanielMcInnes, line 275: The id `acOutInUseItem` is confusing for me. Could you add a comment explaining what this VeQuickItem means? From the uid, it sounds like it means that this system has an ac out system, in which case `hasAcOutSystem` would be a better name. `acOutInUseItem` sounds like it means that the ac out system is currently providing power to something.

pages/settings/PageSettingsSystem.qml
```diff
@@ -86,23 +98,36 @@ Page {
 			ListRadioButtonGroup {
 				//% "Position of AC loads"
 				text: qsTrId("settings_system_ac_position")
-				dataItem.uid: Global.systemSettings.serviceUid + "/Settings/SystemSetup/HasAcInLoads"
+				allowed: defaultAllowed
```
blammit, line 101: The `allowed: defaultAllowed` seems unnecessary, unless it is working around some bug in ListRadioButtonGroup?

DanielMcInnes (approved):
1 nit, then lgtm

blammit (approved):

Issue #1701, opened 2024-11-05
VE.Bus restart does not work after a fresh boot


izak:
The /SystemReset path comes up as invalid after a reboot, causing the option to be disabled:

![Selection_192](https://github.com/user-attachments/assets/28ce7085-4c0a-4f96-bbba-797ecc373ecb)

I looked at how this was done in the old GUI, and I'm unsure what the idea was. The button is "editable" if the path is valid, and once you pressed it once it gets that little triangle, but in all cases you can reset the VE.Bus regardless of whether the path is valid or not. So this is a regression in GUI-v2.

blammit:
Fixed by https://github.com/victronenergy/gui-v2/pull/1702

PR #1702, opened 2024-11-05
Fix VE.Bus system reset being disabled

izak:
https://github.com/victronenergy/gui-v2/issues/1701

### DIFF
--- a/pages/vebusdevice/PageVeBusAdvanced.qml
+++ b/pages/vebusdevice/PageVeBusAdvanced.qml
@@ -197,7 +197,6 @@ Page {
 								//% "Press to redetect"
 							 : qsTrId("vebus_device_press_to_redetect")
 				writeAccessLevel: VenusOS.User_AccessType_User
-				enabled: redetectSystem.isValid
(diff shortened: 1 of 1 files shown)


pages/vebusdevice/PageVeBusAdvanced.qml
```diff
@@ -197,7 +197,7 @@ Page {
 								//% "Press to redetect"
 							 : qsTrId("vebus_device_press_to_redetect")
 				writeAccessLevel: VenusOS.User_AccessType_User
-				enabled: redetectSystem.isValid
+				enabled: userHasWriteAccess
```
blammit, line 200: The default binding in `ListButton` is `enabled: userHasWriteAccess`, so you could just remove the `enabled` binding here and below.

blammit (approved):
Minor comment, LGTM

pages/vebusdevice/PageVeBusAdvanced.qml
blammit: I made this small change, and will merge.

PR #1703, opened 2024-11-06
Show both AC inputs in the Overview when necessary, and add a "Preferred" input.

blammit:
Do not assume that only the system's active AC input (according to system /Ac/ActiveIn) can produce measurement data. This is true for inputs connected to a Multi/Quattro or a Multi RS system (i.e. on vebus or acsystem services), but it is not true for inputs connected on dedicated energy meters (e.g. Grid and Generator inputs connected on grid and genset services).

Replace ActiveAcInput.qml with AcInput.qml, and always provide Global.acInputs.input1 and Global.acInputs.input2, instead only only Global.acInputs.activeInput.

AcInput has an 'active' property that is true for dedicated energy meters, or for the 'active' input on a Multi/Quattro or a Multi RS system.

Show both AC input 1 and AC input 2 in the Overview page, and remove Global.acInputs.activeInput. The Brief page can use the "preferred" input instead of the system active input; the preferred input is the AcInput of a Grid/Shore type that is 'active', or otherwise, any other 'active' input.


DanielMcInnes:
I don't see the electrons animating from either of the ac inputs. The grid input is small, but I would expect to see something coming from the generator

image

blammit:
> I don't see the electrons animating from either of the ac inputs. The grid input is small, but I would expect to see something coming from the generator. Also, should I be seeing direction icons on these widgets?

All of the electrons should only animate when the power is over the 30W threshold.

The direction icons are only shown if the power is negative, or if it is positive and "Feedback to grid" is enabled.

DanielMcInnes:
> > I don't see the electrons animating from either of the ac inputs. The grid input is small, but I would expect to see something coming from the generator. Also, should I be seeing direction icons on these widgets?
> 
> All of the lines should only animate their electrons when the power is over the +/-30W threshold.
> 
> The direction icons are only shown if the power is negative, or if it is positive and "Feedback to grid" is enabled. (Also see this [slack thread](https://victrondevelopment.slack.com/archives/C020QG3JVEW/p1729233878491899?thread_ts=1729160215.114419&cid=C020QG3JVEW).)

The generator is producing 12kW, should it show electrons?

blammit:
> The generator is producing 12kW, should it show electrons?

No, it has to be over 30W.

components/AcInputDirectionIcon.qml
```diff
@@ -11,12 +11,13 @@ import Victron.VenusOS
...
-			? (Global.acInputs.activeInput.power < 0 ? "qrc:/images/icon_to_grid.svg" : "qrc:/images/icon_from_grid.svg")
+	required property AcInput input
+
+	visible: input && (input.power < 0 || Global.system.feedbackEnabled)
+	source: !!input
+			? (input.power < 0 ? "qrc:/images/icon_to_grid.svg" : "qrc:/images/icon_from_grid.svg")
 			: ""
-	color: !Global.acInputs.activeInput || (Global.acInputs.activeInput.power || 0) === 0 ? Theme.color_background_disabled
-			: Global.acInputs.activeInput.power < 0 ? Theme.color_green
+	color: !input || (input.power || 0) === 0 ? Theme.color_background_disabled
```
DanielMcInnes, line 20: When feedback is enabled, but the power === 0, it doesn't seem right to show the icon. I can't see any mention of it in the slack thread.

blammit: Changed to set opacity=0 for icon in this case.

DanielMcInnes (approved):
lgtm

blammit:
Updated with renaming, as discussed with Chris:

active -> operational
preferred -> highlighted

Issue #1665, opened 2024-10-31
Fallback or translations of startpage -> timeout don’t always work?


mpvader:
this is german,

english does work for me. And german too.. so can’t reproduce. And I expect that soon this will be translated - if you can’t reproduce / don’t see something special, then pls close it as non-reproducible. If there are further reports we can always re open it.

![IMG_3352](https://github.com/user-attachments/assets/311942c4-a77b-4a81-97a1-5455f7ed5c63)

mpvader:
someone else having it all the time:

![image](https://github.com/user-attachments/assets/89d8d152-3926-49ec-8244-eaad4be1e12d)

mpvader:
updated: 
<img width="629" alt="image" src="https://github.com/user-attachments/assets/0555faaf-d3f5-4c20-92d7-8d5bed77c95f">

chriadam:
I believe this is (effectively) a duplicate of #1661

chriadam:
@dirkjanfaber do you know, was this one also verified to be fixed in v3.51~2?  If so, will close.

dirkjanfaber:
Just tested with Romanian and this page. Looks good! Closing. 

![image](https://github.com/user-attachments/assets/7a64b2e8-12b9-4118-a2d9-3b31a3442861)

dirkjanfaber:
I think we need to re-open this, @chriadam 

<img width="965" alt="image" src="https://github.com/user-attachments/assets/c4599ff4-e005-4fa3-a4eb-0e8bbcc63989">

Of course it needs to be translated properly, but as long as the translation isn't there yet, it should show the English variant and not the key. 

https://github.com/victronenergy/gui-v2/blob/main/i18n/venus-gui-v2_sv.ts#L8320

philipptrenz:
Another find for the Spanish language

See translation here: https://github.com/victronenergy/gui-v2/blob/main/i18n/venus-gui-v2_es.ts#L8437

![Bildschirmfoto 2024-11-06 um 10 03 23](https://github.com/user-attachments/assets/5417ade5-7c45-4ff2-9e08-5c17be514e49)

chriadam:
I should be able to simply install the english catalogue as a fallback catalogue, in case some translation is missing from a non-english translation catalogue in some version.

mpvader:
Is there more remaining work for this one @chriadam ?

chriadam:
No, closing as fixed.

PR #1706, opened 2024-11-08
Always install English as the fallback catalogue

chriadam:
This ensures that even if another language is chosen as the active language, if some translated string is missing from that language's catalogue, the English translation will be used as a fallback.

Contributes to issue #1665

### DIFF
--- a/src/language.cpp
+++ b/src/language.cpp
@@ -205,10 +205,13 @@ Language::Language(QQmlEngine*) : QObject(nullptr)
 
 void Language::init() // call this after setting the fontUrlPrefix if required.
 {
+	// Install English as the default fallback catalogue.
+	installTranslatorForLanguage(QLocale::English);
+
 	// Load appropriate translations for current locale, e.g. :/i18n/venus-gui-v2_fr.qm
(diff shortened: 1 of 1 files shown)


dirkjanfaber (approved):
Looks good. Tested it for Swedish and indeed seeing the text in English when there is no translation available. Nice!

image

src/language.cpp
```diff
@@ -205,10 +205,13 @@ Language::Language(QQmlEngine*) : QObject(nullptr)
 
 void Language::init() // call this after setting the fontUrlPrefix if required.
 {
+	// Install English as the default fallback catalogue.
+	installTranslatorForLanguage(QLocale::English);
```
DanielMcInnes, line 209: maybe log a warning if this fails

DanielMcInnes (approved):

src/language.cpp
dirkjanfaber: That is a nice idea, but would only do that in combination with a `--verbose` / `-v` QCommandLineOption. Else it might flood the logfiles.

PR #1711, opened 2024-11-08
Improvements for Multi-RS

izak:
Some items were duplicated on the product page. Those have been removed.

The ESS menu is moved one up, from the product page to the Inverter/Charger page. Easier for users to reach.

An item was added for the Active SOC limit. This is the battery SOC Limit the RS will maintain when ESS mode is set to optimised with BatteryLife. Therefore it is shown only when in that ESS mode, so that the user will know which MinSoc is being used.

### DIFF
--- a/pages/settings/devicelist/rs/PageRsSystem.qml
+++ b/pages/settings/devicelist/rs/PageRsSystem.qml
@@ -94,14 +94,6 @@ Page {
 			}
 
 			ListNavigationItem {
-				text: CommonWords.ess
-				onClicked: {
-					Global.pageManager.pushPage("/pages/settings/devicelist/rs/PageRsSystemEss.qml",
-							{ "title": text, "bindPrefix": root.bindPrefix })
-				}
-			}
-
-			ListNavigationItem {
 				//% "RS devices"
 				text: qsTrId("settings_rs_devices")
 				onClicked: {
--- a/pages/invertercharger/OverviewInverterChargerPage.qml
+++ b/pages/invertercharger/OverviewInverterChargerPage.qml
@@ -152,6 +152,15 @@ Page {
 			}
 
 			ListNavigationItem {
+				text: CommonWords.ess
+				allowed: root.serviceType === "acsystem"
+				onClicked: {
+					Global.pageManager.pushPage("/pages/settings/devicelist/rs/PageRsSystemEss.qml",
+							{ "title": text, "bindPrefix": root.serviceUid })
+				}
+			}
+
+			ListNavigationItem {
 				text: CommonWords.product_page
 				onClicked: {
 					if (root.serviceType === "inverter") {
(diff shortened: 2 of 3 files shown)


components/dialogs/ESSMinimumSOCDialog.qml
```diff
@@ -42,6 +42,7 @@ ModalDialog {
 				width: parent.width - (2 * Theme.geometry_modalDialog_content_horizontalMargin)
 				value: root.minimumStateOfCharge
 				from: 0
+				stepSize: 5
```
DanielMcInnes, line 45: Just checking that this is intentional. Did somebody ask for it? Not sure why we would want coarser control of minimum soc

pages/settings/devicelist/rs/PageRsSystem.qml
```diff
@@ -26,56 +26,12 @@ Page {
 
 	GradientListView {
 		model: ObjectModel {
-			ListItem {
```
DanielMcInnes, line 29: Was this mode button is removed, because it is available onOverviewInverterChargerPage? 
You can also get to this page by clicking on the grid widget. In this case, the user won't see a mode button, or the current limit buttons etc. Is that the desired behavior?

blammit: And this page is also opened when clicking on an RS system in the Device List, so I assume you'd want to see the mode and current limit buttons there as well.

izak: > Was this mode button is removed, because it is available onOverviewInverterChargerPage?

Indeed. I didn't realise there is another path to get to the same place. I will have to think about this some more.

components/dialogs/ESSMinimumSOCDialog.qml
izak: Yes, this was intentional. Users rarely need to be able to set the SOC to a precise number, and for many years the minsoc has always been in steps of 5%. I don't know if it is just myself, and maybe being a tad OCD, but I like to set it to a round number, and right now it is also just a little bit too squirely. I find the notchy 5% steps to be easier to use.

The other reason I did this was because the existing ESS controls in the old ported Hub4 pages still use the steps (and plus and minus buttons). So now we have two parts that behave differently, and I thought it would be nice to make them behave the same.

izak:
OK, to make this a bit less controversial, I made these changes.

1. The 5% step is removed. Instead I changed the one place where the step was still used (ESS Submenu of RS) to use the same common slider instead of the spinbuttons. So this is now consistent.
2. I restored the controls I removed because I thought they were duplicated.
3. I kept the change where ESS moves to the Inverter/Charger menu. I think that is fine, and not having it when you click on the grid block (on systems without a grid meter) is not only alright, it is better than having it there, which can be confusing.

This is hopefully the last modification to make gui-v2 really nice for the RS.

DanielMcInnes (approved):
lgtm

blammit (approved):

PR #1715, opened 2024-11-11
Fix path for modbus client tcp

### LINKED ISSUE #1707: Modbus TCP/UDP devices menu controls not working properly on local GUI
Some parts of the "Modbus TCP/UDP devices" settings menu are not working correctly on the onscreen GUI.

The "Press to scan" button does not work at all, I can press it but nothing happens.
Starting a scan from the WASM works. On the WASM, the button changes to displaying the scan progress, the button on the local GUI does not change at all.

If there are discovered modbus devices available, the "Discovered devices" list shows them on the WASM, but not on the local GUI.

On my [test system](https://vrm.victronenergy.com/installation/323213/dashboard)(3.52~2), all the required data seems to be available on dbus (full contents of `com.victronenergy.modbusclient.tcp`):

```
com.victronenergy.modbusclient.tcp
Devices/ve_HQ23492WWJA/Enabled                                           1
Devices/ve_HQ23492WWJA/Name                                      VM-3P75CT
Devices/ve_HQ23492WWJA/Serial                                  HQ23492WWJA
Scan                                                                     0
ScanProgress                                                             -
```
Maybe there is something missing for the GUI to consider the service valid? 
On GUI-V1 everything works as expected.

DanielMcInnes:
Fixes #1707

### DIFF
--- a/pages/settings/PageSettingsModbus.qml
+++ b/pages/settings/PageSettingsModbus.qml
@@ -9,7 +9,7 @@ import Victron.VenusOS
 Page {
 	id: root
 
-	property string service: BackendConnection.serviceUidForType("modbusclient")
+	property string service: BackendConnection.serviceUidFromName("com.victronenergy.modbusclient.tcp", 0)
 	property string settings: Global.systemSettings.serviceUid + "/Settings/ModbusClient/tcp"
 
 	VeQuickItem {
--- a/data/mock/SystemSettingsImpl.qml
+++ b/data/mock/SystemSettingsImpl.qml
@@ -64,7 +64,7 @@ QtObject {
 	}
 
 	function setMockModbusClientValue(key, value) {
-		Global.mockDataSimulator.setMockValue(BackendConnection.serviceUidForType("modbusclient") + key, value)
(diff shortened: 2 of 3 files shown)


pages/settings/PageSettingsModbus.qml
```diff
@@ -9,7 +9,7 @@ import Victron.VenusOS
 Page {
 	id: root
 
-	property string service: BackendConnection.serviceUidForType("modbusclient")
+	property string service: BackendConnection.serviceUidForType("modbusclient.tcp")
```
blammit, line 12: When calling `BackendConnection.serviceUidForType()` there's an implicit assumption that the device instance is zero. I think that's okay for `system`, `settings` and core services like that, but for other services I think it'd be good to be explicit that we're using zero as the device instance. What if instead of calling `serviceUidForType()`, these were changed to:

```
property string service: BackendConnection.serviceUidFromName("com.victronenergy.modbusclient.tcp", 0)
```

blammit (approved):

pages/settings/PageSettingsModbus.qml
DanielMcInnes: updated as suggested

PR #1717, opened 2024-11-11
Show Grid/Shore first in Overview AC inputs

### LINKED ISSUE #1716: Show Grid at the top in Overview page
If there are both Grid/Shore and Generator inputs, then prefer to show the Grid/Shore first in the Overview page.

blammit:
When there is a Grid/Shore input as well as a Generator input, prefer to show the Grid/Shore input first.

Fixes #1716

### DIFF
--- a/pages/OverviewPage.qml
+++ b/pages/OverviewPage.qml
@@ -16,8 +16,8 @@ SwipeViewPage {
 	// Preferred order for the input widgets on the left hand side. When placing widgets, avoid / minimize connectors crossing each other.
 	readonly property var _leftWidgetOrder: [
 		// Top widgets: these widgets have to be up the top, as they connect to the 'inverter/charger widget', which is at the top of the center column.
-		VenusOS.OverviewWidget_Type_AcInput1,
-		VenusOS.OverviewWidget_Type_AcInput2,
+		VenusOS.OverviewWidget_Type_AcInputPriority,
+		VenusOS.OverviewWidget_Type_AcInputOther,
 		// End top widgets
 
 		// Middle widgets: these widgets can connect to both the inverter/charger widget and the battery widget in the center column.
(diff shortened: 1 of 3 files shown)


pages/OverviewPage.qml
```diff
@@ -16,8 +16,8 @@ SwipeViewPage {
 	// Preferred order for the input widgets on the left hand side. When placing widgets, avoid / minimize connectors crossing each other.
 	readonly property var _leftWidgetOrder: [
 		// Top widgets: these widgets have to be up the top, as they connect to the 'inverter/charger widget', which is at the top of the center column.
-		VenusOS.OverviewWidget_Type_AcInput1,
-		VenusOS.OverviewWidget_Type_AcInput2,
+		VenusOS.OverviewWidget_Type_AcInputPriority,
```
chriadam, line 19: did we discuss using "Highlighted" instead of "Preferred" or "Priority"?  Or was that for something else?

blammit: The "highlighted" input also refers to whether the input is connected/active, whereas this is purely about which one is shown first (which only looks at whether it is Grid/Shore).

How about `AcInputPrioritySource` and `AcInputOtherSource`?

pages/OverviewPage.qml
```diff
@@ -257,9 +257,15 @@ SwipeViewPage {
 
 		// Add AC-in widgets.
 		const acInputConfigs = [
-			{ input: Global.acInputs.input1, widgetType: VenusOS.OverviewWidget_Type_AcInput1 },
-			{ input: Global.acInputs.input2, widgetType: VenusOS.OverviewWidget_Type_AcInput2 },
+			{ input: Global.acInputs.input1, widgetType: VenusOS.OverviewWidget_Type_AcInputPriority },
+			{ input: Global.acInputs.input2, widgetType: VenusOS.OverviewWidget_Type_AcInputOther },
 		]
+		if (Global.acInputs.isGridOrShore(Global.acInputs.input2)
```
blammit, line 263: Cleaned this up to use the function from `Global.acInputs`.

pages/OverviewPage.qml
DanielMcInnes: I vote for `OverviewWidget_Type_AcInputGridOrShore` and `OverviewWidget_Type_AcInputGenerator`

DanielMcInnes (approved):

pages/OverviewPage.qml
chriadam: Well, in the case where there is ONLY a generator input, it will be the first one, right?

chriadam (approved):

pages/OverviewPage.qml
blammit: Yes, we don't care so much about the type in the layout, just which one is higher up in the view.

Issue #1626, opened 2024-10-24
overview issue when multi is off


mpvader:
See https://community.victronenergy.com/t/some-failure-with-em24-and-gui-v2-on-v3-50-37-and-34-also-now-vrm-data-is-missing/8779/1

@izak pls triage and if indeed gui-v2 issue then prepare a recording, or other instructions for Qinetic to reproduce, fix and test it.

no need to fix or report this on vrm yet.

izak:
Looks like a gui-v1 vs v2 parity issue.

The feature to assume the grid is present based on the presence of a grid meter (even if the Multi is off) is new in Venus 3.50. In the past gui-v1 would have behaved exactly the same.

These two changes:

https://github.com/victronenergy/gui/commit/8794e5aed69dbeeb3b3dae246bbac4b39d7a5fdb
https://github.com/victronenergy/gui/commit/dc217efd5f1337a83f232b87ae61f6ac67fc883f

So it looks like this is something that did not quite make it to gui-v2 yet.

To reproduce, you can use the attached simulation files, which emulates a Multi that was switched off at the front panel, and a grid meter.

[grid.csv](https://github.com/user-attachments/files/17519074/grid.csv)
[vebus.csv](https://github.com/user-attachments/files/17519076/vebus.csv)

Play with the `play.py` from dbus-recorder, eg:

```
python3 play.py vebus.csv grid.csv
```

Result on GUI-v1 vs v2:

![Selection_182](https://github.com/user-attachments/assets/2eb2b4db-b112-4ee3-86f2-f5974cffdbce)
![Selection_183](https://github.com/user-attachments/assets/5e1e5b42-44c5-4ffb-8ae8-4dfb4881f483)


![Selection_181](https://github.com/user-attachments/assets/1b523ea7-5431-4e1c-a4ef-9d98aed7d48a)

blammit:
Thanks for the `csv` files. I played them and the problem I'm seeing is that `com.victronenergy.system/Ac/ActiveIn/Source` is not set, and gui-v2 relies on that to find the currently-active AC input, to show it on the Brief/Overview pages. 

However, `com.victronenergy.system/Ac/ActiveIn/ServiceType` is set to `grid` as expected, so I've made a patch that checks this value in the case that the `/Source` is not available: https://github.com/victronenergy/gui-v2/pull/1674

Could it be that the `/Source` does not get set in the case that the multi is off?

Also, note that:
[...]

PR #1718, opened 2024-11-11
Use system /Ac/ActiveIn as primary determinant of highlighted input 

blammit:
If system /Ac/ActiveIn/Source or /Ac/ActiveIn/ServiceType is set, then choose the first AC input with a matching source or service type (regardless of the power value, or whether it is a Grid/Shore/Genset).

If neither /Ac/ActiveIn/Source nor /Ac/ActiveIn/ServiceType are set, then choose the first input that is Grid/Shore, or otherwise, any operational input.

Issue #1626

### DIFF
--- a/data/AcInputs.qml
+++ b/data/AcInputs.qml
@@ -11,14 +11,24 @@ QtObject {
 
 	property AcInput input1
 	property AcInput input2
+	readonly property int activeInSource: _activeInSource.value ?? VenusOS.AcInputs_InputSource_NotAvailable
+	readonly property string activeInServiceType: _activeInServiceType.value || ""
 
-	// When the UI has to choose to show a single AC input (e.g. for Brief AC-in gauge), prefer the
-	// first Grid/Shore input that is operational, or otherwise, any operational input.
-	readonly property AcInput highlightedInput: _highlightInput(input1) ? input1
-			: _highlightInput(input2) ? input2
-			: input1?.operational ? input1
-			: input2?.operational ? input2
-			: null
+	readonly property AcInput highlightedInput: {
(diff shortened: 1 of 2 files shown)


data/AcInputs.qml
```diff
@@ -11,14 +11,24 @@ QtObject {
...
 	property AcInput input2
-
-	// When the UI has to choose to show a single AC input (e.g. for Brief AC-in gauge), prefer the
-	// first Grid/Shore input that is operational, or otherwise, any operational input.
-	readonly property AcInput highlightedInput: _highlightInput(input1) ? input1
-			: _highlightInput(input2) ? input2
-			: input1?.operational ? input1
-			: input2?.operational ? input2
-			: null
+	readonly property int activeInSource: _activeInSource.value ?? VenusOS.AcInputs_InputSource_NotAvailable
```
DanielMcInnes, line 14: does this do the same as: 
`readonly property int activeInSource: _activeInSource.value || VenusOS.AcInputs_InputSource_NotAvailable`

DanielMcInnes (approved):

data/AcInputs.qml
blammit: The outcome is the same, but calculated slightly differently:

- when using `??`, if `value` is undefined or null, then `VenusOS.AcInputs_InputSource_NotAvailable` will be returned
- when using `||`, if `value` is undefined or null or zero, then `VenusOS.AcInputs_InputSource_NotAvailable` will be returned

I figure using `??` might be more clean because it's only using the fallback value if the backend value is undefined/null.

chriadam (approved):

PR #1721, opened 2024-11-12
Load demo image if available

blammit:


ApplicationContent.qml
```diff
@@ -32,6 +32,13 @@ Item {
 		Component.onCompleted: Global.mainView = mainView
 	}
 
+	Image {
+		y: Theme.geometry_statusBar_height
+		width: parent.width
```
chriadam, line 37: maybe set this to zero if the demo image file name is empty?

DanielMcInnes (approved):
Maybe add a debug message if we are displaying the demo image. I can just imagine leaving a 'real' looking demo image lying around, and spending ages trying to figure out why none of the nav buttons work

ApplicationContent.qml
blammit: Done

blammit:
> Maybe add a debug message if we are displaying the demo image. I can just imagine leaving a 'real' looking demo image lying around, and spending ages trying to figure out why none of the nav buttons work

Done

DanielMcInnes:
currently the demo image displays on every page, not just the brief page. I assume this has to change?

PR #1726, opened 2024-11-12
Fix run-time null error on Brief page for DC Inputs Icon

MikeTrahearn-Qinetic:
Fix that Global.dcInputs.model.firstObject can become null (e.g. if simulator stops) causing the inputType to become inaccessible.
Use optional chaining to return a null into the Global.dcInputs.inputTypeIcon() function which returns the default value "qrc:/images/icon_dc_24.svg" in this case.


blammit (approved):
Please wrap the commit message to 80 characters. LGTM otherwise.

MikeTrahearn-Qinetic:
Commit message amended. FYI I have now turned on "fixed width font for commit messages" and show vertical rule at 80 chars. So now I don't have to count ;-)
Rebased on main.

PR #1727, opened 2024-11-12
Remove ListRadioButtonGroupNoYes component

### LINKED ISSUE #865: Get rid of ListRadioButtonGroupNoYes component

MikeTrahearn-Qinetic:
Update PageSettingsFroniusInverter to use ListSwitch.

Fixes #865


pages/settings/PageSettingsFroniusInverter.qml
```diff
@@ -47,7 +47,7 @@ Page {
 				]
 			}
 
-			ListRadioButtonGroupNoYes {
+			ListSwitch {
 				id: isActive
```
blammit, line 51: Looks like this id is not used anywhere, can you remove it as well?

MikeTrahearn-Qinetic:
Rebased on main
removed the unreferenced id
Modified comment (80 char format)

DanielMcInnes (approved):

PR #1729, opened 2024-11-12
Fix PageSettingsFroniusInverters uids

MikeTrahearn-Qinetic:
Use [bindPrefix + "/Inverters" in all cases.

### DIFF
--- a/pages/settings/PageSettingsFroniusInverters.qml
+++ b/pages/settings/PageSettingsFroniusInverters.qml
@@ -13,11 +13,7 @@ Page {
 
 	GradientListView {
 		model: VeQItemTableModel {
-			uids: BackendConnection.type === BackendConnection.DBusSource
-				  ? [bindPrefix + "/Inverters"]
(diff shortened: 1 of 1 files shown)


blammit:
@MikeTrahearn-Qinetic There are many uids related to fronius inverters, so "Fix PageSettingsFroniusInverters uids" is not very useful as a commit message - can you explain what issue this is fixing? In this case, it is fixing the list of uids for showing the list of Fronius inverters on the system, and specifically, the change ensures that the list appears as expected in mock mode.

MikeTrahearn-Qinetic:
Commit message updated as requested. Force Pushed changes successfully after rebase on main.

DanielMcInnes:
@MikeTrahearn-Qinetic is there an installation where I can test this?

MikeTrahearn-Qinetic:
The only one we have presently is Izak Cerbo VRM but it isn't responding on my machine here.

DanielMcInnes (approved):

PR #1734, opened 2024-11-13
Don't invalidate CustomName for Device

### LINKED ISSUE #1705: Cached device name switches to `/ProductName`, even though `/CustomName` was set before the device disconnected
On Venus OS 3.52~1.
Gui-v2 caches the name of devices to show them when the device disconnects. Also, the `/CustomName` is preferred over the `/ProductName`, as per the dbus specification. However, when a Lynx BMS with a `/CustomName` set is disconnected, it switches over to displaying the `/ProductName`: 

[Screencast from 07-11-24 13:04:33.webm](https://github.com/user-attachments/assets/336c111a-9f8f-48db-a15d-515b1a9c72f5)

This does not happen on gui-v1, but also not with devices other than the Lynx Smart BMS. Maybe the `/CustomName` is deleted before the `/ProductName` is in this case. Gui-v1 [checks](https://github.com/victronenergy/gui/blob/master/src/dbus_service.cpp#L44) if the dbus service is still synchronized before using the `/ProductName` in case the `/CustomName` disappeared. As far as I can see, gui-v2 does not do this.

DanielMcInnes:
Fixes #1705

### DIFF
--- a/components/Device.qml
+++ b/components/Device.qml
@@ -15,6 +15,10 @@ BaseDevice {
 
 	readonly property VeQuickItem _customName: VeQuickItem {
 		uid: root.serviceUid ? root.serviceUid + "/CustomName" : ""
+		// When some devices (eg. BMSes), are turned off, the custom name value changes to 'undefined'
+		// before they become invalid. See https://github.com/victronenergy/gui-v2/issues/1705.
+		// Setting 'invalidate' to false retains the last valid value
+		invalidate: false
 	}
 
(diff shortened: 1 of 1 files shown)


blammit (approved):

blammit:
Could you add something to the commit to explain what this fixes? E.g. that the Device List will always show the last custom name, rather than the product name.

DanielMcInnes:
> Could you add something to the commit to explain what this fixes? E.g. that the Device List will always show the last valid custom name, rather than the product name.

Done

Issue #1730, opened 2024-11-12
Spinboxes for VE.Bus ESS Max Charge and Inverter Power does not work correctly.


izak:
Settings -> ESS

The Spin boxes for maximum charge- and inverter power has a minimum of -1 (which means disabled), and goes in steps of 50. This causes the spinbox to go 49, 99, 149, etc which is not what we want.

The steps should be from 0 in 50W steps. "-1" should not be an option.

![Selection_198](https://github.com/user-attachments/assets/33c7d9a5-25b0-4f7d-9cad-3e0102a7dc27)

![Selection_199](https://github.com/user-attachments/assets/e9e16bcc-e202-43fd-9a35-2b03c47ea68d)

PR #1736, opened 2024-11-13
fix VE.Bus ESS spinbox behaviour

izak:
These values use -1 to indicate that it is not set, but the RangeModel prevents the value from actually every going negative, thereby making the switch unresponsive.

By using the value of the dataItem instead, this works as in the past.

Closes https://github.com/victronenergy/gui-v2/issues/1730

### DIFF
--- a/pages/settings/PageSettingsHub4Peakshaving.qml
+++ b/pages/settings/PageSettingsHub4Peakshaving.qml
@@ -84,6 +84,7 @@ Page {
 
 				//% "Limit system AC import current"
 				text: qsTrId("settings_ess_limit_system_ac_import_current")
+				checkable: true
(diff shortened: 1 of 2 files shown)


pages/settings/PageSettingsHub4Peakshaving.qml
```diff
@@ -84,6 +84,7 @@ Page {
 
 				//% "Limit system AC import current"
 				text: qsTrId("settings_ess_limit_system_ac_import_current")
+				checkable: true
```
blammit, line 87: I see, this was something leftover that should have been updated in https://github.com/victronenergy/gui-v2/pull/980 when we updated the checkable behaviour for various switches.

PR #1741, opened 2024-11-13
Rename 'VRM online portal' -> 'VRM Portal mode'

blammit:
Align with the terminology used in other UIs.

### DIFF
--- a/pages/SettingsPage.qml
+++ b/pages/SettingsPage.qml
@@ -60,8 +60,8 @@ SwipeViewPage {
 				page: "/pages/settings/PageSettingsDisplay.qml"
 			},
 			{
-				//% "VRM online portal"
-				text: qsTrId("settings_vrm_online_portal"),
+				//% "VRM Portal mode"
+				text: qsTrId("settings_vrm_portal_mode"),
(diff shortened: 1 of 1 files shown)


blammit:
**Note:** Do not merge before 1.1.0, to avoid breaking translations.

pages/SettingsPage.qml
```diff
@@ -60,8 +60,8 @@ SwipeViewPage {
 				page: "/pages/settings/PageSettingsDisplay.qml"
 			},
 			{
-				//% "VRM online portal"
-				text: qsTrId("settings_vrm_online_portal"),
+				//% "VRM Portal mode"
```
DanielMcInnes, line 63: I think this is still called "VRM online portal" in gui-v1. Why are we changing this?

blammit: Ah, I guess it hasn't been fixed in gui-v1. Basically Matthijs asked for the naming to be aligned as it's called just "VRM Portal" in other products, e.g. if you go to vrm.victronenergy.com.

DanielMcInnes (approved):

Issue #1301, opened 2024-07-05
Grid drilldown widget shows incorrect firmware version


chriadam:
Using ESS demo, clicking the Grid widget from Overview page leads to a drilldown page which shows "Firmware version: v0.02" but gui-v1 shows "Firmware version: 2" and dbus-spy also shows value: "2".

chriadam:
I looked into this, and it seems that what is happening is as follows:

1) the value of the setting `com.victronenergy.grid.cgwacs_ttyUSB0_di30_mb1/FirmwareVersion` is `2` (a number).
2) in gui-v1 this is just displayed directly (see qml/PageDeviceInfo.qml)
3) in gui-v2 we pass this value to the versionText() function within FirmwareVersion.qml, which has some rules for converting version numbers into strings (which are then displayed in the UI).  It documents some of the patterns which it uses during conversion, and the case we are hitting for this bug is:

a) the value is a number (2), and not a string ("2")
b) so it assumes that the version number is a hex code, so converts to hex
c) this hex value is of the form 0x00000A
d) so it returns the string: "v0.0A"

--

I guess the question is: is the logic described above only intended for some specific type of version, and not in this (grid) case?

In gui-v1 there is e.g. `return firmwareVersion.value >= 0x109` in `qml/PageSolarCharger.qml` which suggests that at least some devices assume that the version is given as a number which is meaningful when converted to hex.

Are there some specific rules for when the number/hex/string conversion rules should be applied?  @jepefe perhaps knows?

jepefe:
Gui-v1 uses the GetText which provides the correct version format while gui-v2 GetText is not available so the raw value from DBus is used. 

@izak Is it possible to change dbus-cgwacs to report `/FirmwareVersion` with the expected value? Or will it break something else?

izak:
I assigned this to myself. For CG meters the firmware version, which is really just a 16-bit integer, is published as an integer on dbus, but really for third party meters where we don't control the numeric->text translation, it is better that the meter publishes its firmware version as text.

jhofstee:
The GetText version is included in Venus candidate, version v3.52~6.

blammit:
Fixed by https://github.com/victronenergy/gui-v2/pull/1743

PR #1743, opened 2024-11-15
better energy meter firmware version formatting

izak:
If the firmware version is from an energy meter (this will also include some gensets), and is already published as a string, don't reformat it according to Victron formatting. Use the string as is. This moves the responsibility of firmware formatting to the driver.

I would have preferred to make this more generic, simply assuming that all string values can be displayed as is, but I'm concerned this might be too sweeping a change.

https://github.com/victronenergy/gui-v2/issues/1301


blammit:
> I would have preferred to make this more generic, simply assuming that all string values can be displayed as is, but I'm concerned this might be too sweeping a change.

If you think that should be the case then I'm happy to go with it. The problem is that we don't have an extensive set of test cases for the various types of firmware strings. The `tst_firmwareversion.qml` test has a non-exhaustive list.

blammit (approved):

izak:
Ho @blammit , I made the change to use the text version if it exists for all devices. I looked back at how it was done in Gui-v1, and that always used the text representation (from the dbus GetText call) for firmware versions. So it seems correct enough that if the value is already text (and therefore GetText == GetValue in that case), we can use the value as is.

This brought up an interesting question: If all strings are shown as is, what else is there? Well, it turns out I know one driver in Venus (for the IMT insolation sensor) that publishes firmware as a float. So now there is a check for integer in place of the old check for hexadecimal, which is the same thing really.

blammit:
@izak Looks great with the simplified change.  Can you add a test for the float case to tst_firmwareversion.qml as well? If you configure with `-DRUN_UNIT_TESTS=ON` it will run the unit tests when compiled.

blammit (approved):

Issue #1593, opened 2024-10-17
Add EVCS control card


blammit:
Add a control card for each EVCS service. It should show similar data to the control card in VRM:

![image](https://github.com/user-attachments/assets/b0f19e9b-1d12-4863-a66a-306c20c1671d)

MikeTrahearn-Qinetic:
Should there be one control card for each EVCS?

blammit:
> Should there be one control card for each EVCS?

Yes.

MikeTrahearn-Qinetic:
@blammit  There is quite a lot that can be done here and I have gotten a new card with a few test control working on it already. However we should decide on actual UI design for this based on existing UI precedents, for example the Inverter/Charger Control card and its various controls and the existing EVCS Settings controls.

MikeTrahearn-Qinetic:
Screen shots of work in progress - EVCS Cards are now working showing the name, status, mode, current and charging; current is only enabled when the mode is manual as per the EvChargerPage.

<img width="256" alt="EVCS Cards - Light" src="https://github.com/user-attachments/assets/6785c8fb-9dbb-4885-9185-c9051e28b4dd">
<img width="256" alt="EVCS Card Mode - Light" src="https://github.com/user-attachments/assets/39303502-5b6e-4638-9ade-eaec92c6b6f7">
<img width="256" alt="EVCS Card Current - Light" src="https://github.com/user-attachments/assets/0bb9f6fb-ff76-49d0-ad0d-e2fad2194d1e">
<img width="256" alt="EVCS Cards - Dark" src="https://github.com/user-attachments/assets/976fce91-6748-4308-b0cb-a3272e08544d">
<img width="256" alt="EVCS Card Mode - Dark" src="https://github.com/user-attachments/assets/135f6387-9f4c-4fc1-bfd1-737998fe7a05">
<img width="256" alt="EVCS Card - Current - Dark" src="https://github.com/user-attachments/assets/2ec02905-1968-49bc-a9ef-e64b78548f08">

PR #1744, opened 2024-11-18
Miketrahearn/1593 add evcs control card

MikeTrahearn-Qinetic:
Adding PR for review.

### DIFF
--- a/components/dialogs/EvcsChargerModeDialog.qml
+++ b/components/dialogs/EvcsChargerModeDialog.qml
@@ -1,0 +1,44 @@
+/*
+** Copyright (C) 2023 Victron Energy B.V.
+** See LICENSE.txt for license information.
+*/
+
+import QtQuick
+import Victron.VenusOS
+
+// Dialog for changing the charging mode for EVCS chargers
+
+ModalDialog {
+	id: root
+
+	property int mode
+
+	title: CommonWords.mode
+
+	contentItem: Column {
+
+		anchors {
+			top: parent.top
+			left: parent.left
+			right: parent.right
+			margins: Theme.geometry_modalDialog_content_horizontalMargin
+		}
+
+		Repeater {
+			id: repeater
+			model: Global.evChargers.modeOptionModel
+			delegate: Column {
+				width: parent.width
+
+				ListRadioButton {
+					flat: true
+					checked: modelData.value === root.mode
+					text: modelData.display
+					onClicked: root.mode = modelData.value
+				}
+
+				SeparatorBar { visible: model.index !== repeater.count - 1 }
+			}
+		}
+	}
+}
--- a/pages/controlcards/EVCSCard.qml
+++ b/pages/controlcards/EVCSCard.qml
@@ -1,0 +1,94 @@
+/*
+** Copyright (C) 2023 Victron Energy B.V.
+** See LICENSE.txt for license information.
+*/
+
+import QtQuick
+import QtQuick.Controls.impl as CP
+import Victron.VenusOS
+
+ControlCard {
+	id: root
+
+	// can only assume this is a Device because there could be Energy Meters in the Global.evChargers.model
+	// as well as EvCharger
+	property Device evCharger
+
+	readonly property int writeAccessLevel: VenusOS.User_AccessType_Installer
+	readonly property bool userHasWriteAccess: Global.systemSettings.canAccess(writeAccessLevel)
+
+	icon.source: "qrc:/images/icon_charging_station_24.svg"
+
+	//: %1 = the EVCS name
+	//% "EVCS (%1)"
+	title.text: qsTrId("controlcard_evcs_title").arg(evCharger.name)
+	status.text: Global.evChargers.chargerStatusToText(evCharger.status)
+
+	// do not show this card if not an EvCharger
+	visible: modeItem.isValid
+
+	VeQuickItem {
+		id: modeItem
+		uid: root.evCharger.serviceUid + "/Mode"
+	}
+
+	Column {
+		anchors {
+			top: root.status.bottom
+			topMargin: Theme.geometry_controlCard_status_bottomMargin
+			left: parent.left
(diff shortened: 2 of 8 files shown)


components/widgets/EvcsWidget.qml
```diff
@@ -48,7 +48,7 @@ OverviewWidget {
 		id: singleEvChargerComponent
 
 		Column {
-			readonly property var evCharger: Global.evChargers.model.deviceAt(0)
+			readonly property EvCharger evCharger: Global.evChargers.model.deviceAt(0)
```
blammit, line 51: These `var` -> `EvCharger` changes have to be removed, since it won't work in mock mode otherwise. Or it can just be changed to a `Device` type.

components/dialogs/EvcsChargerModeDialog.qml
```diff
@@ -0,0 +1,47 @@
...
+// Dialog for changing the charging mode for EVCS chargers
+
+ModalDialog {
+	id: root
+
+	property int mode
+
+	title: CommonWords.mode
+
+	height: header.height + contentHeight + footer.height
```
blammit, line 18: Is this height binding necessary?

blammit (commented):
Looks great. I have some comments about some minor things but it mostly looks good, nice work.

pages/controlcards/EVCSCard.qml
```diff
@@ -0,0 +1,89 @@
...
+			topMargin: Theme.geometry_controlCard_status_bottomMargin
+			left: parent.left
+			right: parent.right
+		}
+
+		ListItem {
+			//% "Mode"
+			text: CommonWords.mode
+			flat: true
+			enabled: root.userHasWriteAccess
```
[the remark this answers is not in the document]
MikeTrahearn-Qinetic: Done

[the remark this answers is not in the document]
MikeTrahearn-Qinetic: The EvChargerPage's Mode ListRadioButtonGroup uses a uid and a dataItem and its allowed property is defaultAllowed && dataItem.isValid, therefore the correct fix in this case should be:

enabled: root.userHasWriteAccess && modeItem.isValid

in order to correctly match the behaviour of the EvChargerPage. This is because the ListItem in the EVCSCard case does not use a uid on its dataItem and so defaultAllowed would not reflect the modeItem's isValid state.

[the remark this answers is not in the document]
MikeTrahearn-Qinetic: Done

pages/controlcards/EVCSCard.qml
```diff
@@ -0,0 +1,89 @@
...
+			enabled: modeItem.value === VenusOS.Evcs_Mode_Manual
+		}
+
+		FlatListItemSeparator {}
+
+		ListSwitch {
+			//% "Charging"
+			text: CommonWords.charging
+			flat: true
+			// the EvCharger doesn't have a startStop uid property so we provide one to the ListSwitch
```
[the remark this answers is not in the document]
MikeTrahearn-Qinetic: DONE

components/widgets/EvcsWidget.qml
MikeTrahearn-Qinetic: DONE

components/dialogs/EvcsChargerModeDialog.qml
MikeTrahearn-Qinetic: Removed height binding in EvcsChargerModeDialog

The EvcsChargerModeDialog should use the default sizing behaviour.
This was originally added due to copying the design of the InverterChargerMode dialog (code and Figma) but there was clearly a reason for that dialog having that binding whereas no other Dialogs appear to do so. Therefore we keep the EvcsChargerModeDialog acting with a default behaviour to be consistent with everything else.

MikeTrahearn-Qinetic:
@blammit - all changes complete as discussed.

pages/controlcards/EVCSCard.qml
```diff
@@ -0,0 +1,86 @@
...
+
+ControlCard {
+	id: root
+
+	// can only assume this is a Device because there could be Energy Meters in the Global.evChargers.model
+	// as well as EvCharger
+	property Device evCharger
+
+	// do not show this card if not an EvCharger (check if the /Mode is valid
+	visible: modeItem.isValid
```
blammit, line 18: Can you move this to below the property declarations?

pages/controlcards/EVCSCard.qml
```diff
@@ -0,0 +1,86 @@
...
+			enabled: root.userHasWriteAccess && modeItem.isValid
+
+			content.children: ListItemButton {
+				width: Math.min(implicitWidth, Theme.geometry_veBusDeviceCard_modeButton_maximumWidth)
+				text: Global.evChargers.chargerModeToText(modeItem.value)
+				onClicked: Global.dialogLayer.open(modeDialogComponent, { mode: modeItem.value })
+			}
+		}
+
+		FlatListItemSeparator {}
```
blammit, line 55: These separators should be hidden if the item above is hidden.

blammit (commented):
Can you squash some of the commits together? Looks like there are really only these three distinct changes:

- Add Mock simulation data for EVChargers' /SetCurrent value
- Add EVCS Control Cards
- Remove redundant InverterChargerEssModeDialog Component delegate

Also, the "Add EVCS Control Cards" commit needs "Fixes #1593".

This can be moved out of WIP status. We'll just mark it for 1.1.0 so it's not merged prematurely.

pages/controlcards/EVCSCard.qml
MikeTrahearn-Qinetic: Done

pages/controlcards/EVCSCard.qml
MikeTrahearn-Qinetic: Done

PR #1749, opened 2024-11-19
Add support for the SolarSense 750 irradiance sensor

dirkjanfaber:
Add support for the bluetooth SolarSense 750 irradiance sensor.

Tested on https://vrm.victronenergy.com/installation/348124/dashboard and working as intended. One of the things to notice is that the device is currently only visible on the dbus when there is irradiance, so during our daytime.

### DIFF
--- a/pages/settings/devicelist/PageMeteo.qml
+++ b/pages/settings/devicelist/PageMeteo.qml
@@ -74,6 +74,14 @@ Page {
 			}
 
 			ListQuantityItem {
+				dataItem.uid: bindPrefix + "/WindDirection"
+				//% "Wind direction"
+				text: qsTrId("page_meteo_wind_direction")
+				allowed: dataItem.isValid
+				unit: VenusOS.Units_CardinalDirection
+			}
+
+			ListQuantityItem {
 				dataItem.uid: bindPrefix + "/InstallationPower"
 				//% "Installation Power"
 				text: qsTrId("page_meteo_installation_power")
--- a/pages/settings/devicelist/PageMeteo.qml
+++ b/pages/settings/devicelist/PageMeteo.qml
@@ -14,8 +14,12 @@ Page {
 
 	VeQuickItem {
 		id: deviceInstance
-
 		uid: bindPrefix + "/DeviceInstance"
+	}
+
+	VeQuickItem {
+		id: productId
+		uid: root.bindPrefix + "/ProductId"
 	}
 
 	GradientListView {
@@ -34,16 +38,18 @@ Page {
 			ListTemperatureItem {
 				//% "Cell temperature"
 				text: qsTrId("page_meteo_cell_temperature")
+				allowed: dataItem.isValid
 				dataItem.uid: bindPrefix + "/CellTemperature"
 				precision: 1
 			}
 
 			ListTemperatureItem {
-				text: sensor2.dataItem.isValid ?
-						  //% "External temperature (1)"
-						  qsTrId("page_meteo_external_temperature_1") :
-						  //% "External temperature"
-						  qsTrId("page_meteo_external_temperature")
+				text: sensor2.dataItem.isValid
+					//% "External temperature (1)"
+					? qsTrId("page_meteo_external_temperature_1")
+					//% "External temperature"
+					: qsTrId("page_meteo_external_temperature")
+				allowed: dataItem.isValid
 				dataItem.uid: bindPrefix + "/ExternalTemperature"
 				precision: 1
 			}
@@ -67,6 +73,67 @@ Page {
 				precision: 1
 			}
 
+			ListQuantityItem {
+				dataItem.uid: bindPrefix + "/InstallationPower"
+				//% "Installation Power"
+				text: qsTrId("page_meteo_installation_power")
+				allowed: dataItem.isValid
+				unit: VenusOS.Units_Watt
+				precision: 1
+			}
+
+			ListQuantityItem {
+				dataItem.uid: bindPrefix + "/TodaysYield"
+				//% "Today's yield"
+				text: qsTrId("page_meteo_daily_yield")
+				allowed: dataItem.isValid
+				unit: VenusOS.Units_Energy_KiloWattHour
+				precision: 1
+			}
+
+			ListItem {
+				id: sensorBattery
+
+				//% "Sensor battery"
+				text: qsTrId("page_meteo_battery_voltage")
+				allowed: defaultAllowed && batteryVoltage.isValid
+
+				content.children: [
+					QuantityLabel {
+						id: batteryVoltageLabel
+						anchors.verticalCenter: parent.verticalCenter
+						font.pixelSize: Theme.font_size_body2
+						value: batteryVoltage.value === undefined ? NaN : batteryVoltage.value
+						unit: VenusOS.Units_Volt_DC
+						VeQuickItem {
+							id: batteryVoltage
+							uid: bindPrefix + "/BatteryVoltage"
+						}
+					},
+					Label {
+						anchors.verticalCenter: parent.verticalCenter
+						text: {
+							if (lowBattery.isValid) {
+								const low = lowBattery.value === 1
+								//% "Low"
+								return low ? qsTrId("meteo_sensor_battery_status_low")
+													: CommonWords.ok
+							} else {
+								return ""
+							}
+						}
+						color: lowBattery.value === 1 ? Theme.color_red : Theme.color_green
+						font.pixelSize: Theme.font_size_body2
+						verticalAlignment: Text.AlignVCenter
+
+						VeQuickItem {
+							id: lowBattery
+							uid:  bindPrefix + "/Alarms/LowBattery"
+						}
+					}
+				]
+			}
+
 			ListNavigationItem {
 				id: settingsMenu
 
@@ -75,6 +142,15 @@ Page {
 														   "title": CommonWords.settings,
 														   meteoSettingsPrefix: root.settingsPrefix
 													   })
+				allowed: productId.value === ProductInfo.ProductId_MeteoSensor_Imt
+			}
+
+			ListNavigationItem {
+				text: CommonWords.device_info_title
+				onClicked: {
+					Global.pageManager.pushPage("/pages/settings/PageDeviceInfo.qml",
+									{ "title": text, "bindPrefix": root.bindPrefix })
+				}
 			}
 		}
 	}
--- a/src/units.cpp
+++ b/src/units.cpp
@@ -94,7 +94,9 @@ int Units::defaultUnitPrecision(VenusOS::Enums::Units_Type unit) const
 	case VenusOS::Enums::Units_Temperature_Celsius:    // fall through
 	case VenusOS::Enums::Units_Temperature_Fahrenheit: // fall through
 	case VenusOS::Enums::Units_Temperature_Kelvin:     // fall through
-	case VenusOS::Enums::Units_RevolutionsPerMinute: return 0;
+	case VenusOS::Enums::Units_RevolutionsPerMinute:   // fall through
+	case VenusOS::Enums::Units_CardinalDirection:      // fall through
+		return 0;
 	default:
 		// VoltAmpere
 		// Amp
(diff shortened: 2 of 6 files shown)


dirkjanfaber:
Added some more changes, showing the wind direction as well. Obviously not too much at home in the qml world, so there are probably some better ways of achieving the same.

pages/settings/devicelist/PageMeteo.qml
```diff
@@ -73,13 +73,12 @@ Page {
 				precision: 1
 			}
 
-			ListQuantityItem {
+			ListWindDirectionItem {
 				dataItem.uid: bindPrefix + "/WindDirection"
 				//% "Wind direction"
 				text: qsTrId("page_meteo_wind_direction")
 				allowed: dataItem.isValid
-				unit: VenusOS.Units_Degrees
```
blammit: How about adding `Units_CardinalDirection` or such, and then updating `Units::getDisplayTextWithHysteresis()` to format the `quantity.unit` depending on the degrees value? That way we can keep this as a `ListQuantityItem` and use the precision handling in `units.cpp`, and we won't need a separate `Units:: formatWindDirection()`.

Also this is the only code that uses `Units_Degrees`, so it can be removed from `enums.h` if it's removed here.

blammit (commented):
Is there an installation that can be used for testing these configurations?

dirkjanfaber:
> Is there an installation that can be used for testing these configurations?

Yes, it is available on https://vrm.victronenergy.com/installation/102195/dashboard.

pages/settings/devicelist/PageMeteo.qml
```diff
@@ -67,6 +73,75 @@ Page {
...
+				dataItem.uid: bindPrefix + "/WindDirection"
+				//% "Wind direction"
+				text: qsTrId("page_meteo_wind_direction")
+				allowed: dataItem.isValid
+				precision: 0
+			}
+
+			ListQuantityItem {
+				dataItem.uid: bindPrefix + "/InstallationPower"
+				//% "Installation Power""
```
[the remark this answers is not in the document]
dirkjanfaber: Thanks. Updated that. ✅

pages/settings/devicelist/PageMeteo.qml
```diff
@@ -67,6 +73,75 @@ Page {
...
+				text: qsTrId("page_meteo_installation_power")
+				allowed: dataItem.isValid
+				unit: VenusOS.Units_Watt
+				precision: 1
+			}
+
+			ListQuantityItem {
+				dataItem.uid: bindPrefix + "/TodaysYield"
+				//% "Todays yield""
```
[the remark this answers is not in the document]
dirkjanfaber: Thanks. Updated that. ✅

pages/settings/devicelist/PageMeteo.qml
dirkjanfaber: Adjusted that. Still not completely certain if I did everything the correct way. So leaving this open for another review.

pages/settings/devicelist/PageMeteo.qml
```diff
@@ -67,6 +73,76 @@ Page {
...
+				//% "Installation Power"
+				text: qsTrId("page_meteo_installation_power")
+				allowed: dataItem.isValid
+				unit: VenusOS.Units_Watt
+				precision: 1
+			}
+
+			ListQuantityItem {
+				dataItem.uid: bindPrefix + "/TodaysYield"
+				//% "Today's yield""
```
blammit, line 96: Extra quotation mark on the end?

pages/settings/devicelist/PageMeteo.qml
```diff
@@ -14,10 +14,14 @@ Page {
 
 	VeQuickItem {
 		id: deviceInstance
-
 		uid: bindPrefix + "/DeviceInstance"
 	}
 
+        VeQuickItem {
```
blammit, line 20: Can you check the spaces vs tabs that are being added? Currently the repo is using tabs instead of spaces (though it'd be nice to change that), and I see three blocks of code that are using spaces instead of tabs:

- this block here
- line 103-145 in this file
- line 156-163 in this file

src/units.cpp
```diff
@@ -429,5 +435,31 @@ int Units::unitToVeUnit(VenusOS::Enums::Units_Type unit) const
...
+		qtTrId("direction_south"),
+		//% "SW"
+		qtTrId("direction_southwest"),
+		//% "W"
+		qtTrId("direction_west"),
+		//% "NW"
+		qtTrId("direction_northwest")
+	};
+	int index = static_cast<int>((degrees + 22.5) / 45.0) % 8;
+	return QString(directions[index]);
```
blammit, line 461: Can this be just `return directions[index]`?

Also `index` can be const.

src/units.cpp
```diff
@@ -257,6 +259,10 @@ quantityInfo Units::getDisplayTextWithHysteresis(VenusOS::Enums::Units_Type unit
 		return quantity;
 	}
 
+	if (unit == VenusOS::Enums::Units_CardinalDirection) {
+		quantity.unit += " - " + formatWindDirection(static_cast<int>(value));
```
blammit, line 263: I'm wondering, is the " - " necessary? Can we just show "35° NW" instead of  "35° - NW"?

pages/settings/devicelist/PageMeteo.qml
blammit: Looks great :) Can you add a few tests to `tst_units.qml` for the new unit? The test run when compiled if configured with `-DRUN_UNIT_TESTS=ON`.

CMakeLists.txt
```diff
@@ -347,6 +347,7 @@ set (VENUS_QML_MODULE_SOURCES
     components/listitems/ListTextItem.qml
     components/listitems/ListTextItemSecondaryLabel.qml
     components/listitems/ListTimeSelector.qml
+    components/listitems/ListWindDirectionItem.qml
```
blammit, line 350: Looks like this can be removed now

pages/settings/devicelist/PageMeteo.qml
```diff
@@ -75,7 +151,16 @@ Page {
 														   "title": CommonWords.settings,
 														   meteoSettingsPrefix: root.settingsPrefix
 													   })
+				allowed: productId.value == ProductInfo.ProductId_MeteoSensor_Imt
```
DanielMcInnes, line 154: `===`

src/units.cpp
```diff
@@ -429,5 +435,31 @@ int Units::unitToVeUnit(VenusOS::Enums::Units_Type unit) const
 	return static_cast<int>(::unitToVeUnit(unit));
 }
 
+QString Units::formatWindDirection(int degrees) const {
+	// Normalize degrees to 0-360 range
+	degrees = (degrees + 360) % 360;
+
+	const QString directions[] = {
+		//% "N"
+		qtTrId("direction_north"),
```
DanielMcInnes, line 444: I've never used `qtTrId` before. Can you test whether it updates when you change languages? Eg. Add a dummy translation for the language of your choice for each of these strings (eg. "dummy N", "dummy NE", ...).

src/units.cpp
dirkjanfaber: Yes, that is a good suggestion.

DanielMcInnes (approved):
lgtm

src/units.cpp
```diff
@@ -257,6 +261,12 @@ quantityInfo Units::getDisplayTextWithHysteresis(VenusOS::Enums::Units_Type unit
 		return quantity;
 	}
 
+	if (unit == VenusOS::Enums::Units_CardinalDirection) {
+		value = fmod(value + 360, 360);
+		quantity.number = formattingLocale()->toString(value, 'f', 0);
```
blammit, line 266: Is there a reason for forcing the precision to be always 0? If the default precision for CardinalDirection is 0, then can we just use the default precision processing further down in this method?

That way you could specify a non-zero precision for some custom case if necessary.

tests/units/tst_units.qml
```diff
@@ -45,6 +45,22 @@ TestCase {
 		expect(VenusOS.Units_Percentage, 100, "100", "%")
 	}
 
+	function test_winddirection() {
```
blammit, line 48: Could you camel case test_winddirection -> test_windDirection?

src/units.h
```diff
@@ -98,6 +98,8 @@ class Units : public QObject
 		qreal normalized = qMax(fromMin, qMin(fromMax, n));
 		return qFuzzyIsNull(fromRange) ? 0.0 : ((((normalized - fromMin) / fromRange) * toRange) + toMin);
 	}
+
+	Q_INVOKABLE QString formatWindDirection(int degrees) const;
```
blammit, line 102: This no longer needs Q_INVOKABLE, and could be a private function.

src/units.cpp
```diff
@@ -257,6 +261,12 @@ quantityInfo Units::getDisplayTextWithHysteresis(VenusOS::Enums::Units_Type unit
...
+		quantity.unit += " " + formatWindDirection(static_cast<int>(value));
```
blammit, line 267: I'm wondering, is the extra space strictly necessary? It's not added for any other units.

pages/settings/devicelist/PageMeteo.qml
```diff
@@ -75,7 +150,16 @@ Page {
 														   "title": CommonWords.settings,
 														   meteoSettingsPrefix: root.settingsPrefix
 													   })
+				allowed: productId.value === ProductInfo.ProductId_MeteoSensor_Imt
 			}
+
+                        ListNavigationItem {
```
blammit, line 156: Note, this block (line 155-162) is still using tabs instead of spaces.

components/listitems/ListWindDirectionItem.qml
```diff
@@ -0,0 +1,34 @@
+/*
+** Copyright (C) 2023 Victron Energy B.V.
+** See LICENSE.txt for license information.
+*/
+
+import QtQuick
+import Victron.VenusOS
+
+ListItem {
+   id: root
```
blammit, line 10: ListWindDirectionItem.qml is no longer used and can be removed

blammit (commented):
Looks great with the tests, just have some more comments about minor things.

src/units.cpp
dirkjanfaber: Thought about it for a while and came to the conclusion that the degrees is the unit and the cardinal direction is the rough direction on what those degrees are in. 
To me it makes sense to have a space in between, as those are two separate things. 
I don't think we currently have other units that have a comparable thing.

blammit: Okay, sounds good, maybe we can revise this if we have other units later on with a similar requirement.

components/listitems/ListWindDirectionItem.qml
blammit: @dirkjanfaber This file is still in the PR?

src/units.cpp
dirkjanfaber: Agreed. Also not 100% sure if all of the languages will confirm to the current format. But we can always change when needed.

src/units.h
```diff
@@ -39,6 +39,8 @@ class Units : public QObject
 	QML_SINGLETON
 	Q_PROPERTY(QString numberFormattingLocaleName READ numberFormattingLocaleName CONSTANT FINAL)
 
+	QString formatWindDirection(int degrees) const;
```
blammit, line 42: Can you move this below the last `scaleNumber()` function:

```
private:
     QString formatWindDirection(int degrees) const;
```

Also the progress commits should be squashed -- can you squash this PR into one or two commits?

blammit: To make it easier, I've moved the function and squashed the commits into two; I hope you don't mind. This will make it easier to fold the changes into https://github.com/victronenergy/gui-v2/pull/1762.

I've also wrapped the commit messages to 72 characters.

blammit (approved):

blammit:
Merging now so we can have them as part of the API renaming in https://github.com/victronenergy/gui-v2/pull/1762.

PR #1751, opened 2024-11-19
Overview: show multi-phase solar total in watts, not amps

### LINKED ISSUE #1750: Overview: multi-phase solar shows total in Amps instead of Watts
Given this setup:

- system has only one PV inverter, which has multiple phases
- system has no PV chargers
- user has selected to show in Amps instead of Watts

The Overview solar box shows its total in Amps, but it should be in Watts, since it does not make sense to add Amp values together.

Reproducible with https://vrm.victronenergy.com/installation/473156/dashboard: 

image

If gui-v2 is patched to only show one AC input, you can see the Overview shows 3 phases for the PV inverter, with Amps for the phases and also for the total, when the total should be in Watts:

image

blammit:
PvMonitor maxPhaseCount was incorrect as Instantiator::objectAt() returns an invalid object when the /NumberOfPhases value is initialized on startup. Wait until the end of the event loop to ensure objectAt() returns a valid value.

This fixes the bug where multi-phase solar totals were shown in amps, as the maxPhaseCount is used in System.qml to determine whether there are multi-phase PV inverters on the system.

Fixes #1750

### DIFF
--- a/data/System.qml
+++ b/data/System.qml
@@ -62,6 +62,8 @@ QtObject {
 					// cannot be summed across multiple phases.
 					return NaN
 				}
+				// There are one or more PV inverters, which are all single-phase, so it's safe to
+				// return a total current as they should all have the same PV output voltage.
 				return _pvMonitor.totalCurrent
 			} else if (Global.solarChargers.model.count > 0) {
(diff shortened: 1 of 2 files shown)


chriadam (approved):
LGTM

Issue #1338, opened 2024-07-16
Avoid rebuilding models for ListQuantityGroup and ListTextGroup 


blammit:
ListQuantityGroup and ListTextGroup models are assembled using models like this:

```
ListQuantityGroup {
    text: "Multiple quantities"
    textModel: [
        { value: 5.1, unit: VenusOS.Units_Volt_DC },
        { value: 48, unit: VenusOS.Units_Watt }
    ]
}
```

This is conveniently straightforward to specify, but is problematic because when the quantity values change, the array is rebuilt, which in turn causes the delegates to be rebuilt.

So instead, the model should be a model or list object that does not rebuild its list entries when a value is modified. 

**Note:** ListRadioButtonGroup has a similar issue with its `optionModel`, but the list values in those models rarely change (if at all) so it's probably fine to keep those as-is.

## Solution 1

A simple solution is to use a list property of QObjects instead:

```
ListQuantityGroup {
    text: "Multiple quantities"
    textModel: [
        QuantityObject { value: 5.1; unit: VenusOS.Units_Volt_DC }
        QuantityObject { value: 48; unit: VenusOS.Units_Watt }
    ]
}


// ListQuantityGroup.qml
ListItem {
    property list<QuantityInfo> textModel
}

```

But there is still a problem where non-visible items need to be hidden in the group. (See https://github.com/victronenergy/gui-v2/issues/1087#issuecomment-2179943534) For example, currently we do this:

```
ListQuantityGroup {
    text: "Multiple quantities"
    textModel: [
        { value: someVeItem.value, unit: VenusOS.Units_Watt; visible: someVeItem.isValid },  // hide this conditionally
        { value: 5.1, unit: VenusOS.Units_Volt_DC },
    ]
}
```

This is problematic as the separator between the two items is not hidden as expected.

## Solution 2

Perhaps a more general-purpose solution would be better, where we use ObjectModel with QuantityLabel:

```
ListQuantityGroup {
    model: ObjectModel {
        QuantityLabel { value: someVeItem.value, unit: VenusOS.Units_Watt; visible: someVeItem.isValid }
        QuantityLabel { value: 5.1, unit: VenusOS.Units_Volt_DC },
    }
}
```

This would still cause problems with separators if some of the labels should be non-visible. But if a VisibleItemModel is added then we can use that instead of ObjectModel, and omit the `visible` condition:

```
ListQuantityGroup {
    model: VisibleItemModel {
[...]

PR #1757, opened 2024-11-25
Device List: rework to load per-service delegate files 

blammit:
The _deviceDisplayInfo() in DeviceListPage is problematic as it causes binding updates to all delegates whenever any of the displayed delegate values change. 

Instead of using this single gigantic function to configure delegates, load per-service QML files that define their own delegate UIs. This improvement also means that custom list items can be provided for individual services, instead of requiring all services to use a similar delegate UI.

All delegates should have the same appearance as before, except for the Inverter delegate, which has been fixed to only show the power for the currently active phase.

Issue #1338

(This only reworks the UI delegates; the second part of the task is to rework the model loading, which will be done separately.)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -214,6 +214,7 @@ set (VENUS_QML_MODULE_SOURCES
     components/CurrentLimitButton.qml
     components/DateSelector.qml
     components/Device.qml
+    components/DeviceListDelegate.qml
     components/DynamicValueRange.qml
     components/ElectricalQuantityLabel.qml
     components/EnvironmentGauge.qml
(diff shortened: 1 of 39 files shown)


CMakeLists.txt
```diff
@@ -588,6 +589,36 @@ set (VENUS_QML_MODULE_SOURCES
...
     pages/settings/devicelist/dc-in/PageDcMeterHistory.qml
+    pages/settings/devicelist/delegates/AcInDeviceListDelegate.qml
+    pages/settings/devicelist/delegates/DcMeterDeviceListDelegate.qml
+    pages/settings/devicelist/delegates/DeviceListDelegate_acload.qml
+    pages/settings/devicelist/delegates/DeviceListDelegate_acsystem.qml
+    pages/settings/devicelist/delegates/DeviceListDelegate_alternator.qml
+    pages/settings/devicelist/delegates/DeviceListDelegate_battery.qml
+    pages/settings/devicelist/delegates/DeviceListDelegate_charger.qml
+    pages/settings/devicelist/delegates/DeviceListDelegate_dcdc.qml
+    pages/settings/devicelist/delegates/DeviceListDelegate_dcgenset.qml
```
DanielMcInnes, line 600: I can't find this file

pages/settings/devicelist/delegates/DeviceListDelegate_dcdc.qml
```diff
@@ -0,0 +1,12 @@
+/*
+** Copyright (C) 2024 Victron Energy B.V.
+** See LICENSE.txt for license information.
+*/
+
+import QtQuick
+import Victron.VenusOS
+
+DcMeterDeviceListDelegate {
+	id: root
```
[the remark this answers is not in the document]
blammit: They looked a bit empty otherwise, but it's really not necessary. Have updated to remove them.

components/DeviceListDelegate.qml
```diff
@@ -0,0 +1,17 @@
+/*
+** Copyright (C) 2023 Victron Energy B.V.
+** See LICENSE.txt for license information.
+*/
+
+import QtQuick
+import Victron.VenusOS
+import QtQuick.Controls.impl as CP
```
chriadam, line 8: unnecessary import?

pages/settings/devicelist/DeviceListPage.qml
```diff
@@ -3,230 +3,56 @@
...
+
+			on_LoadCustomDelegateChanged: {
+				let delegateUri
+				if (_loadCustomDelegate) {
+					const serviceType = BackendConnection.serviceTypeFromUid(device.serviceUid)
+					if (!serviceType) {
+						console.warn("DeviceList: cannot load delegate, cannot read service type from serviceUid:", device.serviceUid)
+						return
+					}
+					setSource("delegates/DeviceListDelegate_%1.qml".arg(serviceType), {
```
chriadam, line 37: Hrm.  I would probably prefer a strongly typed component instead of a lookup of a qml file by string.  Something like:

```
QtObject {
    property Component abcComponent: Component {
        DeviceListDelegate_abc { }
    }
    property Component xyzComponent: Component {
        DeviceListDelegate_xyz { }
    }
    
    function delegateForService(service) {
        switch (service) {
        case 'abc': return abcComponent;
        case 'xyz': return xyzComponent;
        default: break;
        }
        return null;
    }
}
```

blammit: Any particular reason? If we make it strongly typed then it's harder to load them dynamically for third parties.

components/DeviceListDelegate.qml
blammit: Removed

pages/settings/devicelist/DeviceListPage.qml
chriadam: True, good point.

chriadam (approved):

PR #1765, opened 2024-11-27
Respect write access in spinbox, date/time selectors and misc buttons

### LINKED ISSUE #1764: Spinboxes do not respect access roles
Spinboxes are editable in the "User" role, where editing is prevented in gui-v1. For example:

- Settings > ESS > Grid set point
- Settings > DVCC > Maximum charge current (when "Limit charge current" is activated)

blammit:
Fixes #1764

### DIFF
--- a/components/listitems/ListClearHistoryButton.qml
+++ b/components/listitems/ListClearHistoryButton.qml
@@ -37,7 +37,7 @@ ListButton {
 		id: timer
 		interval: 2000
 	}
-	enabled: !timer.running
+	enabled: userHasWriteAccess && !timer.running
 
 	onClicked: {
 		/*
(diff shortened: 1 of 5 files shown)


chriadam:
FirmwareCheckListButton overrides the enabled binding from ListButton, so might need fixing - maybe set button.enabled instead?
ListClearHistoryButton seems to do the same, but also has an allowed property, not sure if it takes access into account or not.
Question: what about SettingsSlider.qml?  I guess it's not a List* item, so probably not related.

blammit:
> FirmwareCheckListButton overrides the enabled binding from ListButton, so might need fixing - maybe set button.enabled instead?
ListClearHistoryButton seems to do the same, but also has an allowed property, not sure if it takes access into account or not.
Question: what about SettingsSlider.qml? I guess it's not a List* item, so probably not related.

Thanks, updated to fix this. Yep SettingsSlider can be ignored, since ListSlider will check the permissions.

chriadam (approved):
LGTM

PR #1767, opened 2024-11-27
Add a breadcrumb header when navigating the settings pages

### LINKED ISSUE #1231: Provide breadcrumbs header for the settings pages
GX devices have a lot of settings, sometimes in deep hierarchy. Replace the static settings page titles with a breadcrumbs pattern that visualises the user’s current location in the settings hierarchy, and makes it easy to jump to specific level in the parent hierarchy. See the Figma design for more details.

DanielMcInnes:
Fixes #1231


pages/settings/devicelist/delegates/DcMeterDeviceListDelegate.qml
```diff
@@ -17,7 +17,7 @@ DeviceListDelegate {
 
 	onClicked: {
 		Global.pageManager.pushPage("/pages/settings/devicelist/dc-in/PageDcMeter.qml",
-				{ bindPrefix : root.device.serviceUid })
+				{ "title": text, bindPrefix : root.device.serviceUid })
```
chriadam, line 20: I assume this won't update if we change translation.  Can we do `"title": Qt.binding(function() { return root.text })` or does that not work here? (e.g. maybe root will go out of scope in some case?)

components/StatusBar.qml
```diff
@@ -98,11 +100,53 @@ Rectangle {
 		onClicked: root.leftButtonClicked()
 	}
 
+	Breadcrumbs {
+		id: breadcrumbs
```
chriadam, line 104: I'm not sure whether this was already defined somewhere, if so I apologise for the noise, but when using Overview drilldowns, I noticed the Breadcrumbs will effectively "hide" the controls button and clock.  Well, I guess that being able to navigate the drilldown hierarchy quickly is more important?  But, thought I'd raise it for discussion just in case.

components/Breadcrumbs.qml
```diff
@@ -0,0 +1,63 @@
...
+
+		readonly property bool isTopBreadcrumb: index === root.count - 1
+		readonly property bool isBottomBreadcrumb: index === 0
+		readonly property color iconColor: isTopBreadcrumb ? primaryBreadcrumbColor : secondaryBreadcrumbColor
+
+		height: root.height
+		width: row.width
+
+		Row {
+			id: row
```
blammit, line 31: Use a less generic name than `row`?

components/StatusBar.qml
```diff
@@ -98,11 +100,53 @@ Rectangle {
...
+			left: leftButton.right
+			leftMargin: Theme.geometry_settings_breadcrumb_horizontalMargin
+			right: rightButtonRow.left
+			verticalCenter: leftButton.verticalCenter
+		}
+		height: Theme.geometry_settings_breadcrumb_height
+		model: root.pageStack.depth + 1 // '+ 1' because we insert a dummy breadcrumb with the text "Settings"
+		visible: count >= 2
+		primaryBreadcrumbColor: Theme.color_settings_breadcrumb_background_top_page
+		secondaryBreadcrumbColor: Theme.color_settings_breadcrumb_background
```
blammit, line 140: Looks like `primaryBreadcrumbColor` and `secondaryBreadcrumbColor` are only set from here. Could these be assigned as initial values in `Breadcrumbs.qml` instead?

components/StatusBar.qml
```diff
@@ -98,11 +100,53 @@ Rectangle {
...
+	Breadcrumbs {
+		id: breadcrumbs
+
+		function getText(index) {
+			return index === 0
+					? Global.pageManager.navBar.activeButtonText // eg: "Settings"
+					: pageStack.get(index - 1).title // eg: "Device list"
+		}
+
+		onClicked: function(index) {
```
blammit, line 112: Can you put this `onClicked` handler below the anchors and other property initializations?

components/Breadcrumbs.qml
```diff
@@ -0,0 +1,63 @@
...
+				text: getText(index)
+			}
+
+			CP.ColorImage {
+				color: iconColor
+				source: "qrc:/images/breadcrumb_rhs.svg"
+			}
+		}
+
+		MouseArea {
```
blammit, line 58: Use a `PressArea` instead to get the nice press effect?

components/Breadcrumbs.qml
```diff
@@ -0,0 +1,63 @@
...
+			}
+
+			Label {
+				horizontalAlignment: Text.AlignHCenter
+				verticalAlignment: Text.AlignVCenter
+				height: parent.height
+				width: implicitWidth + Theme.geometry_settings_breadcrumb_horizontalMargin * 2
+				color: isTopBreadcrumb ? Theme.color_settings_breadcrumb_primaryText : Theme.color_settings_breadcrumb_secondaryText
+				background: Rectangle {
+					color: iconColor
```
blammit, line 46: Would prefer to qualify these references with `delegate.iconColor` so it's clear where they come from.

Also would be nicer to have a less generic id than `delegate`.

components/Breadcrumbs.qml
```diff
@@ -0,0 +1,63 @@
...
+
+	signal clicked(index : int)
+
+	function getText(index) {
+		return ""
+	}
+
+	orientation: ListView.Horizontal
+	currentIndex: count - 1
+	clip: true
```
blammit, line 19: A faded gradient on the left/right edge might be nicer than a hard edge, like the gradient at the bottom of the settings list views.

blammit (commented):
Very nice! I really like this feature.

mr-manuel:
Looks awesome! I tested it and found a little bug, that was not present before. If I navigate to a 4th or 5th level menu entry and then go back, it jumps from the 4th or 5th level directly to the 2nd level and not to the 3rd or 4th level like it should.

Here a comparison. First the build of this PR and then the current main:

venus-os-breadcrumb

components/StatusBar.qml
DanielMcInnes: This is worth bringing up. Yes, this will hide the controls buttons and the clock

components/Breadcrumbs.qml
DanielMcInnes: Updated as suggested

components/StatusBar.qml
DanielMcInnes: Done

components/StatusBar.qml
DanielMcInnes: Done

components/Breadcrumbs.qml
DanielMcInnes: Done

components/Breadcrumbs.qml
DanielMcInnes: Done x 2

components/Breadcrumbs.qml
DanielMcInnes: You're right, it does look nicer with the gradient. Updated as suggested

DanielMcInnes:
> Looks awesome! I tested it and found a little bug, that was not present before. If I navigate to a 4th or 5th level menu entry and then go back, it jumps from the 4th or 5th level directly to the 2nd level and not to the 3rd or 4th level like it should.
> 
> Clicking on a breadcrumb to navigate back works as expected.
> 
> Here a comparison. First the build of this PR and then the current main:
> 
> venus-os-breadcrumb [ venus-os-breadcrumb ](https://private-user-images.githubusercontent.com/12592765/391033045-efba87dd-d58f-45f0-b5c2-18155046ac4e.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MzMxOTk5OTksIm5iZiI6MTczMzE5OTY5OSwicGF0aCI6Ii8xMjU5Mjc2NS8zOTEwMzMwNDUtZWZiYTg3ZGQtZDU4Zi00NWYwLWI1YzItMTgxNTUwNDZhYzRlLmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDEyMDMlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQxMjAzVDA0MjEzOVomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPWJkN2I5ZWFhYmY2MzAwNzE2OWVhYWFmYjYyNzQ5NDk3NWIwMzQ2ZTFhOTRmMjcwMzQ1NDU4OWU2M2YxZmQ5M2ImWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0In0.W9WGrzhmCmJqyK4gA6KZ0Y72XX4DP-NScCw5EDVsa70) [ ](https://private-user-images.githubusercontent.com/12592765/391033045-efba87dd-d58f-45f0-b5c2-18155046ac4e.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MzMxOTk5OTksIm5iZiI6MTczMzE5OTY5OSwicGF0aCI6Ii8xMjU5Mjc2NS8zOTEwMzMwNDUtZWZiYTg3ZGQtZDU4Zi00NWYwLWI1YzItMTgxNTUwNDZhYzRlLmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDEyMDMlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQxMjAzVDA0MjEzOVomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPWJkN2I5ZWFhYmY2MzAwNzE2OWVhYWFmYjYyNzQ5NDk3NWIwMzQ2ZTFhOTRmMjcwMzQ1NDU4OWU2M2YxZmQ5M2ImWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0In0.W9WGrzhmCmJqyK4gA6KZ0Y72XX4DP-NScCw5EDVsa70)

Thanks @mr-manuel , this is an excellent bug report. I have fixed in the latest version.

blammit (approved):
Looks great 👍

components/Breadcrumbs.qml
```diff
@@ -0,0 +1,108 @@
...
+
+ListView {
+	id: root
+
+	property color primaryBreadcrumbColor: Theme.color_settings_breadcrumb_background_top_page
+	property color secondaryBreadcrumbColor: Theme.color_settings_breadcrumb_background
+
+	signal clicked(index : int)
+
+	function getText(index) {
```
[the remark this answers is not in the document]
DanielMcInnes: updated as suggested

PR #1772, opened 2024-12-02
Load settings pages with uids instead of objects

blammit:


data/dbus/SolarChargersImpl.qml
```diff
@@ -18,7 +18,18 @@ QtObject {
 		}
 
 		delegate: SolarCharger {
+			id: solarCharger
 			serviceUid: model.uid
+
+			onValidChanged: {
+				if (!!Global.solarChargers) {
```
DanielMcInnes, line 25: Just curious why this change was made. To my eyes, it looked better before - is it needed for some reason?

pages/vebusdevice/PageVeBus.qml
```diff
@@ -9,46 +9,43 @@ import Victron.VenusOS
...
+	required property string bindPrefix
+	readonly property bool isMulti: numberOfAcInputs.isValid && numberOfAcInputs.value > 0
 	readonly property bool chargeInProcess: preferRenewableEnergy.value === 0
 
 	title: veBusDevice.name
 
-	VeQuickItem {
-		id: _numberOfPhases
+	Device {
+		id: veBusDevice
```
DanielMcInnes, line 19: how does this get its serviceUid?

DanielMcInnes:
There are still some settings pages that get loaded with objects rather than uids, eg:

PageSettingsModbusDevices.qml: Global.pageManager.pushPage("/pages/settings/PageSettingsModbusAddDevice.qml", { devices: _devices } )

PageBattery.qml: Global.pageManager.pushPage("/pages/settings/devicelist/battery/PageBatteryDetails.qml",
							{ "title": text, "bindPrefix": root.bindPrefix, "details": batteryDetails })

pages/vebusdevice/PageVeBus.qml
blammit: Oops - forgot to add it. Fixed now.

data/dbus/SolarChargersImpl.qml
blammit: It removes the dependency between the `SolarCharger` type and the `Global.solarChargers.model`, so that it is possible to create a `SolarCharger` object that is just used for fetching the data within that objects, without adding it to the global model.

E.g. in `DeviceListDelegate_inverter.qml`, this is convenient for creating an `Inverter` object to get the relevant AC data, without necessarily adding that inverter object to the global inverters model.

Going ahead, I'd like to do this across the board where possible, so that device types like `SolarCharger` can be created independently of the global models. This will make it more flexible to use those types on sub-pages without creating global objects that are only used on some pages.

blammit:
> There are still some settings pages that get loaded with objects rather than uids, eg:

This is okay -- the main thing is to be able to create the Device pages without passing in objects. It might be useful to make these changes for those other sub-pages as well, not sure, but for now they are not essential.

DanielMcInnes (approved):
lgtm

PR #1781, opened 2024-12-04
Scheduled charging menu for Multi-RS

izak:
This simply repeats the Scheduled charging menu inside the RS ESS section.

It includes an additional fix to hide the minsoc setting when the ESS mode is Keep Batteries Charged.


pages/settings/devicelist/rs/PageRsSystemEss.qml
```diff
@@ -35,6 +39,7 @@ Page {
 				//% "Minimum SOC (unless grid fails)"
 				text: qsTrId("settings_rs_ess_min_soc")
 				button.text: Units.getCombinedDisplayText(VenusOS.Units_Percentage, essMinSocItem.value)
+				allowed: defaultAllowed && isModeOptimized
```
blammit, line 42: Could use `root.isModeOptimized` instead of just `isModeOptimized` to make the scope clear (and similarly below), otherwise LGTM.

blammit (approved):

Issue #1104, opened 2024-05-20
Make spinbox number editable


blammit:
Make the number spinbox editable - e.g. replace it with a text input when clicked - so that the user has an alternative to just pressing the -/+ buttons, which may take a long time if the desired number is not close to the original value.

Some issues for investigation/discussion:

- Should the text input have a frame/border to indicate it is editable?
- On the Cerbo/Ekrano, we need to show the VKB for number entry. How can we show the VKB above the NumberSelectorDialog? (Preliminary investigation suggest we may need to use a Popup instead of a Dialog. In that case to consider whether NumberSelectorDialog and other dialogs need to be changed to extend Popup instead, and consider any implications of such a change.)
- The VKB in the current position would cover the text input. Should the dialog/popup move upwards when it receives focus? Or what else can be done to make the input fully visible?

jpetrell:
Also reported by the [customer](https://community.victronenergy.com/questions/279997/venus-os-v3408-available-for-testing.html?childToView=286325#answer-286325) independently "AC input current limit shows fractions, but increase/decrease works only +-1A, can't activate keyboard, can't set 16.2 or 16.0".

blammit:
The AC input current limit issue is fixed by https://github.com/victronenergy/gui-v2/pull/1199

MikeTrahearn-Qinetic:
The VKB overarching issue probably needs to be discussed first before any changes are made to the spin box with respect its primaryLabel and secondaryLabel in its contentItem.

MikeTrahearn-Qinetic:
This is also requested here:
https://victrondevelopment.slack.com/archives/C020QG3JVEW/p1731734054739599

MikeTrahearn-Qinetic:
For reference: https://bugreports.qt.io/browse/QTBUG-128406

PR #1784, opened 2024-12-05
miketrahearn/1104 make spinbox number editable

MikeTrahearn-Qinetic:
Made the SpinBox editable, including enhancing ModalDialog so that it moves to ensure that any text inputs are always shown above the virtual keyboard, which is now shown above all Dialogs where required.

- NumberSelectionDialog is specifically modified to meet the requirements of this ticket.
- Colors used for additional text selection colors are Theme.color_blue and Theme.color_white where non-dark/light theme awareness is required.
- Fixed SpinBox's indicators enabled states on limits.
- Commits comments explain each step of change.

Input Current Limit editable Dark
Input Current Limit editable Keys Dark

Input Current Limit editable Light
Input Current Limit editable Keys Light

Charge Current editable Dark
Charge Current editable Keys Dark

Charge Current editable Light
Charge Current editable Keys Light

### DIFF
--- a/components/controls/SpinBox.qml
+++ b/components/controls/SpinBox.qml
@@ -7,6 +7,7 @@ import QtQuick
 import QtQuick.Controls as C
 import QtQuick.Templates as CT
 import QtQuick.Controls.impl as CP
+import QtQuick.Layouts
 import Victron.VenusOS
 
 // SpinBox uses a binding to increase 'stepSize' when the user holds a button down for a while. This allows the spin box to quickly change arbitrarily large values.
@@ -17,11 +18,12 @@ CT.SpinBox {
 	id: root
 
 	property alias label: primaryLabel
-	property string secondaryText
+	property alias secondaryText: secondaryLabel.text
 	property int indicatorImplicitWidth: Theme.geometry_spinBox_indicator_minimumWidth
 	property int orientation: Qt.Horizontal
 	property int _scalingFactor: 1
 	property int _originalStepSize
+	property alias suffix: suffixLabel.text
 
 	signal maxValueReached()
 	signal minValueReached()
@@ -45,29 +47,104 @@ CT.SpinBox {
 	}
 
 	contentItem: Item {
+
+		// needed for QQuickSpinBoxPrivate to read the "text" property of the contentItem
+		// so that it can call the valueFromText() function
+		readonly property alias text: primaryLabel.text
+
 		Column {
 			id: valueColumn
 
-			width: Math.max(primaryLabel.implicitWidth, secondaryLabel.implicitWidth)
+			width: Math.max(primaryTextInput.implicitWidth, secondaryLabel.implicitWidth)
 			anchors.centerIn: parent
 
-			Label {
-				id: primaryLabel
+			Item  {
+				id: primaryTextInput
 
-				width: parent.width
-				text: root.textFromValue(root.value, root.locale)
-				color: root.enabled ? Theme.color_font_primary : Theme.color_background_disabled
-				font.pixelSize: root.secondaryText.length > 0 ? Theme.font_size_h2 : Theme.font_size_h3
-				horizontalAlignment: Qt.AlignHCenter
-				verticalAlignment: Qt.AlignVCenter
+				width: primaryRowLayout.implicitWidth + Theme.geometry_textField_horizontalMargin * 2
+				height: primaryRowLayout.height
+				anchors.horizontalCenter: parent.horizontalCenter
+
+				MouseArea {
+					anchors.fill: parent
+					enabled: root.editable
+					onClicked: primaryLabel.forceActiveFocus()
+				}
+
+				Rectangle {
+					anchors.fill: parent
+					visible: root.editable
+					color: "transparent"
+					border.color: Theme.color_blue
+					border.width: Theme.geometry_button_border_width
+					radius: Theme.geometry_button_radius
+				}
+
+				RowLayout {
+					id: primaryRowLayout
+
+					anchors.centerIn: parent
+
+					TextInput {
+						id: primaryLabel
+
+						color: root.enabled ? Theme.color_font_primary : Theme.color_background_disabled
+						font.family: Global.fontFamily
+						font.pixelSize: root.secondaryText.length > 0 ? Theme.font_size_h2 : Theme.font_size_h3
+						horizontalAlignment: Qt.AlignHCenter
+						verticalAlignment: Qt.AlignVCenter
+						selectedTextColor: Theme.color_white
+						selectionColor : Theme.color_blue
+						readOnly: !root.editable
+						selectByMouse: !readOnly
+						validator: root.validator
+						inputMethodHints: root.inputMethodHints
+
+						onAccepted: {
+							// Note that the text may at this time represent a value out of SpinBox
+							// to/from range, so clamp it here.
+							let v = root.valueFromText(text, root.locale)
+							if (v < root.from) {
+								v = root.from
+							} else if (v > root.to) {
+								v = root.to
+							}
+
+							// Force-update the displayed text, to guarantee the text is in sync
+							// with the numeric value, even if the value has not changed due to the
+							// user entering an out-of-range value on consecutive attempts.
+							text = root.textFromValue(v, root.locale)
+							root.value = v
+							primaryLabel.focus = false
+						}
+
+						Connections {
+							target: root
+							function onValueChanged() {
+								// Update the displayed text when the initial value is set or when
+								// the up/down buttons are pressed.
+								primaryLabel.text = root.textFromValue(root.value, root.locale)
+							}
+						}
+					}
+
+					Label {
+						id: suffixLabel
+
+						visible: text.length
+						color: primaryLabel.color
+						font: primaryLabel.font
+						horizontalAlignment: primaryLabel.horizontalAlignment
+						verticalAlignment: primaryLabel.verticalAlignment
+					}
+
+				}
 			}
 
 			Label {
 				id: secondaryLabel
 
-				width: primaryLabel.width
 				height: text.length ? implicitHeight : 0
-				text: root.secondaryText
 				color: Theme.color_font_secondary
 				font.pixelSize: Theme.font_size_caption
 				horizontalAlignment: Qt.AlignHCenter
@@ -114,6 +191,18 @@ CT.SpinBox {
 			source: 'qrc:/images/icon_minus.svg'
 			opacity: root.enabled ? 1.0 : 0.4 // TODO add Theme opacity constants
 		}
+	}
+
+	textFromValue: function(value, locale) {
+		return Units.formatNumber(value)
+	}
+	valueFromText: function(text, locale) {
+		let value = Units.formattedNumberToReal(text)
+		if (isNaN(value)) {
+			// don't change the current value
+			value = root.value
+		}
+		return value
 	}
 
 	Timer {
--- a/components/dialogs/ModalDialog.qml
+++ b/components/dialogs/ModalDialog.qml
@@ -13,6 +13,7 @@ T.Dialog {
 	property string secondaryTitle
 	property int dialogDoneOptions: VenusOS.ModalDialog_DoneOptions_SetAndCancel
 	property alias canAccept: doneButton.enabled
+	readonly property real centeredY: (parent.height - height) / 2
 
 	// Optional functions: called when accept/reject is attempted.
 	// These should return true if the accept/reject can be executed, and false otherwise.
@@ -32,7 +33,9 @@ T.Dialog {
 
 	readonly property string rejectTextCancel: CommonWords.cancel
 
-	anchors.centerIn: parent
+	// Use x/y positioning instead of anchors, so that the dialog can be moved upwards when needed.
+	x: (parent.width - width) / 2
+	y: centeredY
 
 	/*
 	If you specify implicitWidth & implicitHeight here, and shrink the browser (or desktop app) window from 100% -> 0%,
@@ -169,6 +172,82 @@ T.Dialog {
 				root.accept()
 			}
 		}
+	}
+
+	property QtObject _stateManager: QtObject {
+		id: stateManager
+
+		readonly property Item inputItem: root.visible && Qt.inputMethod.visible ?
+											  (root.contentItem.Window.activeFocusItem as TextField ??
+											   root.contentItem.Window.activeFocusItem as TextInput ??
+											   // root.contentItem.Window.activeFocusItem as TextArea ?? // not used
+											   root.contentItem.Window.activeFocusItem as TextEdit) : null
+
+		property real targetDialogY: 0
+
+		onInputItemChanged: {
+			if (!inputItem) {
+				dialogStateGroup.state = "default"
+				return
+			}
+
+			const currentDialogOffset = root.y - root.centeredY // 0 or negative
+			const inputItemBottomPos = inputItem.mapToItem(Global.mainView, 0, inputItem.implicitHeight).y - currentDialogOffset
+
+			targetDialogY = root.centeredY
+
+			const vkbTopPos = Global.mainView.height - Qt.inputMethod.keyboardRectangle.height
+
+			if (inputItemBottomPos > vkbTopPos) {
+				// Note: moving the Dialog while in "focused" state will change to
+				// the new location immediately without any animation.
+				targetDialogY += (vkbTopPos - inputItemBottomPos)
+			}
+
+			dialogStateGroup.state = "focused"
+		}
+		property StateGroup dialogStateGroup: StateGroup {
+
+			state: "default"
+
+			states: [
+				State {
+					name: "default"
+					PropertyChanges {
+						// reset to the "default" binding explicitly
+						// so we can get the transition
+						root.y: root.centeredY
+					}
+				},
+				State {
+					name: "focused"
+					PropertyChanges {
+						root.y: stateManager.targetDialogY
+					}
+				}
+			]
+
+			transitions: [
+				Transition {
+					to: "*"
+					NumberAnimation {
+						target: root
+						property: "y"
+						duration: Theme.animation_inputPanel_slide_duration
+						easing.type: Easing.InOutQuad
+					}
+				}
+			]
+		}
+	}
+
+	MouseArea {
+		// placed behind the background, contentItem, header and footer
+		parent: contentItem.parent
+		anchors.fill: parent
+		z: -1
+		enabled: !!stateManager.inputItem
+		onClicked: focus = true
 	}
 }
 
--- a/components/dialogs/NumberSelectorDialog.qml
+++ b/components/dialogs/NumberSelectorDialog.qml
@@ -56,15 +56,25 @@ ModalDialog {
 				anchors.horizontalCenter: parent.horizontalCenter
 				width: parent.width - 2*Theme.geometry_modalDialog_content_horizontalMargin
 				height: Theme.geometry_timeSelector_spinBox_height
+				editable: true
 				indicatorImplicitWidth: root.decimals > 0
 						? Theme.geometry_spinBox_indicator_minimumWidth
 						: Theme.geometry_spinBox_indicator_maximumWidth
 				textFromValue: function(value, locale) {
-					return Units.formatNumber(value / root._multiplier(), root.decimals) + root.suffix
+					return Units.formatNumber(value / root._multiplier(), root.decimals)
+				}
+				valueFromText: function(text, locale) {
+					let value = Units.formattedNumberToReal(text) * root._multiplier()
+					if (isNaN(value)) {
+						// don't change the current value
+						value = spinBox.value
+					}
+					return value
 				}
 				from: Math.max(Global.int32Min, root.from * root._multiplier())
 				to: Math.min(Global.int32Max, root.to * root._multiplier())
 				stepSize: root.stepSize * root._multiplier()
+				suffix: root.suffix
 
 				onValueChanged: {
 					if (_initialized) {
--- a/ApplicationContent.qml
+++ b/ApplicationContent.qml
@@ -4,6 +4,7 @@
 */
 
 import QtQuick
+import QtQuick.Controls as QtQuickControls
 import Victron.VenusOS
 
 Item {
@@ -155,6 +156,8 @@ Item {
 		source: Global.isGxDevice
 				? "qrc:/qt/qml/Victron/VenusOS/components/InputPanel.qml"
 				: "qrc:/qt/qml/Victron/VenusOS/components/WasmVirtualKeyboardHandler.qml"
+		parent: QtQuickControls.Overlay.overlay
+		z: 1
 	}
 
 	// Sometimes, the wasm code may crash. Use a watchdog to detect this and reload the page when necessary.
--- a/components/InputPanel.qml
+++ b/components/InputPanel.qml
@@ -52,7 +52,7 @@ QtVirtualKeyboard.InputPanel {
 
 	states: State {
 		name: "visible"
-		when: Qt.inputMethod.visible
+		when: Qt.inputMethod.visible && !!root.focusedFlickable
 
 		PropertyChanges {
 			target: root.focusedFlickable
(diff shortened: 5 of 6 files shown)


MikeTrahearn-Qinetic:
Please note that [PR ](https://github.com/victronenergy/gui-v2/pull/1785) contains a separate fix for the enabled status of the SpinBox indicators such that the screen shots shown here do not show the indicators disabled when the limits are reached.

MikeTrahearn-Qinetic:
I will be testing on cerbo before requiring approvals.

components/dialogs/NumberSelectorDialog.qml
```diff
@@ -78,6 +87,16 @@ ModalDialog {
 					spinBox.value = Math.round(root.value * root._multiplier())
 					_initialized = true
 				}
+
+				validator: DoubleValidator {
+					// text editing needs validating in the non-multiplied range
+					// this stops you entering a value outside this range
+					locale: root.locale.name
```
blammit, line 94: The other day I said that the spinbox should use the number formatting for the app locale, but this is incorrect. A while ago we changed the behaviour so that it would always use the C locale for number formatting, so e.g. the decimal separator will always be "." and never a "," regardless of the app locale. This was because there were various difficulties in consistently supporting alternative formats across the board, and also users may prefer the "." formatting regardless of the selected locale.

components/controls/SpinBox.qml
```diff
@@ -42,32 +44,82 @@ CT.SpinBox {
...
+
+				}
 			}
 
 			Label {
 				id: secondaryLabel
 
-				width: primaryLabel.width
 				height: text.length ? implicitHeight : 0
-				text: root.secondaryText
```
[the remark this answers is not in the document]
MikeTrahearn-Qinetic: The text for the suffix label is a property alias at the root SpinBox.qml now so this specific line is removed.

components/controls/SpinBox.qml
```diff
@@ -42,32 +44,82 @@ CT.SpinBox {
...
+						readOnly: !root.editable
+						selectByMouse: !readOnly
+						validator: root.validator
+						inputMethodHints: root.inputMethodHints
+					}
+
+					Label {
+						id: suffixLabel
+
+						visible: text.length
```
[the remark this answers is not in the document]
MikeTrahearn-Qinetic: I agree - I will see what I can do to make this happen.

components/dialogs/ModalDialog.qml
```diff
@@ -170,5 +173,95 @@ T.Dialog {
...
+		}
+	}
+
+	MouseArea {
+		// placed behind the contentItem
+		parent: contentItem.parent
+		anchors.fill: parent
+		z: -1
+		enabled: !!stateManager.inputItem
+		onClicked: focusScope.focus = true
```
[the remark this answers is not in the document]
MikeTrahearn-Qinetic: The FocusScope was added initially by you in your test branch. I will take a look to see if we can remove it.

components/dialogs/NumberSelectorDialog.qml
MikeTrahearn-Qinetic: Done - in SpinBox.qml and NumberSelectorDialog.qml

components/dialogs/ModalDialog.qml
```diff
@@ -170,5 +173,95 @@ T.Dialog {
...
+
+		property real targetDialogY: 0
+
+		onInputItemChanged: {
+			if(!inputItem) {
+				dialogStateGroup.state = "default"
+				return
+			}
+
+			dialogStateGroup.state = "interrupted"
```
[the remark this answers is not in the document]
MikeTrahearn-Qinetic: The "interrupted" state ensures that any Transition currently in progress is halted before measurements are taken to ensure we aren't trying to do this on a still moving target. I have proven that the animations can continue between text fields changing focus to another. If we can categorically prove that we can get away with not interrupting the transition (try slowing it down to 2000 for example and it still works ok - which I don't think it will, as I tested the state changes with such a slow transition before speeding it back up again) then I would be ok to remove this state. Note that it works in concert with the deliberate choice of the behaviour of the PropertyChanges restoreEntryValues false in the "focused" state. I had seen in your branch that you had thought that explicit: true would work around that problem, but in a separate test application (attached) I proved that this isn't adequate to correctly prevent the animation from causing the measured location of the new text input to be slightly wrong.
I hope that explains the rationale here.
[AnimationTest.zip](https://github.com/user-attachments/files/18087726/AnimationTest.zip)

components/controls/SpinBox.qml
```diff
@@ -116,6 +168,24 @@ CT.SpinBox {
...
+	}
+
+	textFromValue: function(value, locale) {
+		return Units.formatNumber(value)
+	}
+	valueFromText: function(text, locale) {
+		// close the Virtual Keyboard on Return/Enter pressed
+		// even if the value hasn't actually changed e.g. in the case
+		// the value is about to be set out of range.
+		root.focus = false
```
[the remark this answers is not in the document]
MikeTrahearn-Qinetic: Sadly no. Well, normally yes... and I understand the question. But the scenario here was solving the issue when a text was entered that hit a limit, and then you would not be able to lose the focus. When we look at the docs for:
[editingFinished](https://doc.qt.io/qt-6/qml-qtquick-textinput.html#editingFinished-signal) or
[accepted](https://doc.qt.io/qt-6/qml-qtquick-textinput.html#accepted-signal)
we can see that when there is a validator in the mix, neither of these are going to fire if the value is out of the range - OR - and specifically, if the text change would not actually change the value.
So we needed to hunt for the one place where we could inject the focus = false after the text was entered but was still being processed by the SpinBox, and before the validator kicked in. This is the solution unfortunately.

components/dialogs/NumberSelectorDialog.qml
```diff
@@ -56,15 +56,24 @@ ModalDialog {
...
 						: Theme.geometry_spinBox_indicator_maximumWidth
 				textFromValue: function(value, locale) {
-					return Units.formatNumber(value / root._multiplier(), root.decimals) + root.suffix
+					return Units.formatNumber(value / root._multiplier(), root.decimals)
+				}
+				valueFromText: function(text, locale) {
+					// close the Virtual Keyboard on Return/Enter pressed
+					// even if the value hasn't actually changed e.g. in the case
+					// the value is about to be set out of range.
+					spinBox.focus = false
```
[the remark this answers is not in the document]
MikeTrahearn-Qinetic: See explanation above.

components/controls/SpinBox.qml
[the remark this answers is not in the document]
MikeTrahearn-Qinetic: I am actually looking into what event sequences unfold if I actually get rid of the TEXT validator. In this case, we get the onEditingFinished and onAccepted signals, even if the text input represents an out-of-range value. Then there might be better injection points to FIX the displayText - which only happens the first time automatically since the value starts at 1.20 and we change it to something over-limit and the value changes to the MAX (1.50). Since the value changes, the text is updated with the max value. Subsequent text entries over the limit don't change the value - and the the displayTest stays as is - WRONG. So this is what I have to fix and am looking into now.

components/dialogs/ModalDialog.qml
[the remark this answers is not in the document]
blammit: Great, yes the FocusScope in the prototype was used to remove the focus from the text fields, but in this current setup it shouldn't be necessary.

MikeTrahearn-Qinetic:
We should also turn off the VKB Text Selection Handles after discussing and agreeing with @blammit as part of this PR and not requiring any other ticket, and quoting the reason as [QTBUG-114551](https://bugreports.qt.io/browse/QTBUG-114551).

components/controls/SpinBox.qml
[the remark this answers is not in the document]
MikeTrahearn-Qinetic: DONE

components/dialogs/ModalDialog.qml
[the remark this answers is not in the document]
MikeTrahearn-Qinetic: Yes - it can be removed. DONE.

components/controls/SpinBox.qml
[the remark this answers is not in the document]
MikeTrahearn-Qinetic: The solution now is to remove the TEXT validators:
- the TextInput's onAccepted fires and sets its focus to false regardless of the validity of the text.
- there is no need to set focus = false anywhere else and these have all been removed.
- this closes the VBK in all cases, whatever the entered value is - even if it is BLANK.

Now: how to fix 
a) when you enter an invalid value (which is now allowed) AND 
b) how to ensure that if the value entered is invalid or out of bounds and it only self-corrects the first time only (as observed)?

This is awkward, but see the valueFromText() function that now has to stimulate a change to the spinBox.value whenever the value is ABOUT to change to something valid but different. The clumsy part is that it has to ensure that even when entering limits (from or to) we must still make a change else the textFromValue() won't trigger and the displayText() won't self-correct.
e.g. a partial entry of "1" with no decimals will be corrected to "1.0.0". 
e.g. an out of range "-1" will correct to "1.0.0" or "5" > "1.5.0".
and finally "" (or any other NaN input) > last known good value.
(noting in the test case the properties from: 1.0.0 and to: 1.5.0)

So we now have a solution to solving both a) and b) above.

MikeTrahearn-Qinetic:
@blammit I have rebased on main - and cleaned up the commit history as there was a lot of noise in there.
This means some PR comments are now "outdated" but the fixes are still present. It is now therefore easier to do a cleaner review on the new commit stack.

components/dialogs/NumberSelectorDialog.qml
```diff
@@ -56,15 +56,34 @@ ModalDialog {
...
+					let value = Number.fromLocaleString(locale, text) * root._multiplier()
+					if(isNaN(value)) {
+						// don't change the current value
+						value = spinBox.value
+					}
+					// need to set spinBox.value to a different temporary valid value
+					// to illicit a value change so it always updates the displayText properly
+					// in all cases, to something that is valid
+					if(value <= spinBox.from) {
+						spinBox.value = spinBox.to
```
blammit, line 76: I would expect `valueFromText()` to just return a value, without modifying the stored `spinBox.value` in the process. What if this clamping was done in the TextInput `onAccepted` instead? From there you can call `valueFromText(text)` and update `root.value` as needed.

components/controls/SpinBox.qml
```diff
@@ -45,29 +47,88 @@ CT.SpinBox {
...
+
+				MouseArea {
+					anchors.fill: parent
+					enabled: root.editable
+					onClicked: primaryLabel.forceActiveFocus()
+				}
+
+				Rectangle {
+					anchors.fill: parent
+
```
blammit, line 77: Unnecessary newline? Similarly below `visible`

components/controls/SpinBox.qml
```diff
@@ -45,29 +47,88 @@ CT.SpinBox {
...
 
-				width: parent.width
-				text: root.textFromValue(root.value, root.locale)
-				color: root.enabled ? Theme.color_font_primary : Theme.color_background_disabled
-				font.pixelSize: root.secondaryText.length > 0 ? Theme.font_size_h2 : Theme.font_size_h3
-				horizontalAlignment: Qt.AlignHCenter
-				verticalAlignment: Qt.AlignVCenter
+				width: primaryRowLayout.implicitWidth + Theme.geometry_textField_horizontalMargin * 2
+				height: primaryRowLayout.height
+
```
blammit, line 66: Unnecessary newline?

components/dialogs/ModalDialog.qml
```diff
@@ -170,5 +173,91 @@ T.Dialog {
...
+
```
blammit, line 199: I can see the values being affected in your example app by the interrupted state, and I can see how an interrupted state could be useful when restoreEntryValues is being used, but I'm not understanding how it relates to this onInputItemChanged handler where it sets the state to "interrupted" and then "focused" immediately afterwards. I tried slowing down the animation duration as well, but the dialog y still moved to the expected position each time. Let's look over this part together and see what's happening.

components/dialogs/ModalDialog.qml
```diff
@@ -170,5 +173,91 @@ T.Dialog {
...
+		readonly property Item inputItem: root.visible && Qt.inputMethod.visible ?
+											  (root.contentItem.Window.activeFocusItem as TextField ??
+											   root.contentItem.Window.activeFocusItem as TextInput ??
+											   // root.contentItem.Window.activeFocusItem as TextArea ?? // not used
+											   root.contentItem.Window.activeFocusItem as TextEdit) : null
+
+		// vkbTopPos is const - it will always be the same value when the keyboard is open;
+		// - we only need this value when the keyboard is open (inputItem is not null)
+		readonly property real vkbTopPos: Global.mainView.height - Qt.inputMethod.keyboardRectangle.height
```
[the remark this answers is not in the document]
MikeTrahearn-Qinetic: Done. However I will always prefer to use bindings rather than assignments because they are type safe. I also appreciate the potential performance gain if that is an issue. A compromise would be to have a type safe property that was assigned. Style over function on this one.

components/dialogs/NumberSelectorDialog.qml
```diff
@@ -56,15 +56,34 @@ ModalDialog {
...
+				editable: true
 				indicatorImplicitWidth: root.decimals > 0
 						? Theme.geometry_spinBox_indicator_minimumWidth
 						: Theme.geometry_spinBox_indicator_maximumWidth
 				textFromValue: function(value, locale) {
-					return Units.formatNumber(value / root._multiplier(), root.decimals) + root.suffix
+					return Units.formatNumber(value / root._multiplier(), root.decimals)
+				}
+				valueFromText: function(text, locale) {
+					let value = Number.fromLocaleString(locale, text) * root._multiplier()
```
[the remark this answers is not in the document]
MikeTrahearn-Qinetic: DONE

components/dialogs/ModalDialog.qml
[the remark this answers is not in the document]
blammit: It's not really a performance issue here, but I don't see any gain from having a separate property. The lack of type safety is just built-in from using JavaScript statements, and I don't think we need to try to avoid that.

components/dialogs/NumberSelectorDialog.qml
MikeTrahearn-Qinetic: I should explain further what is going on here. Architecturally I am fighting with the present incomplete implementation of a decimal-supporting SpinBox. This should be made a feature of SpinBox for all cases. At the moment the NumberSelectorDialog attempts to finish off the implementation. That being the case, I have to select with some care, the injection point for the case where the text is not valid and as you can see, duplicate some of the effort across SpinBox and NumberSelectorDialog which isn't ideal. I agree that valueFromText should return a value, but you see, the SpinBox is clamping internally - there's nothing I can do about that - and so it should. But this leads to some of the onward mechanics failing to trigger in certain use cases. If I move all the logic to the onAcceptedHandler of the SpinBox text I then have two problems to solve: the overriding of the SpinBox's normal valueFromText machinery, and the duplication of it, and, referring to my first point, there would be no nice way to do this in the NumberSelectorDialog at all in that case as we are not outside the SpinBox and don't (shouldn't) have access to the internal TextInput. So the net effect is this (I agree) not so nice solution until such time as the SpinBox itself has decimal capabilities.

components/controls/SpinBox.qml
MikeTrahearn-Qinetic: DONE

components/controls/SpinBox.qml
MikeTrahearn-Qinetic: DONE

components/dialogs/ModalDialog.qml
MikeTrahearn-Qinetic: The change of the inputItem can happen mid transition.
I can try to make this "interrupted" state disappear, but in doing so I know there are going to be edge cases that break. I guess I will have to spend more time analysing this over again.

MikeTrahearn-Qinetic: I have now pushed a change which removes the "interrupted" state.
This has the effect that moving the Dialog while in "focused" state will change to the new location immediately without any animation. No other calculation side-effects were seen. I concede that the likelihood of this actually happening is very low given the lack of a TAB key in most cases.

components/dialogs/NumberSelectorDialog.qml
MikeTrahearn-Qinetic: Additionally since the SpinBox's valueFromText function call is overridden in the NumberSelectorDialog, we can't simply call valueFromText() in the TextInput's onAccepted handler because it would call the wrong valueFromText function. This would not be the case if the SpinBox would have its decimals and multipliers inside its implementation.

blammit: Yeah it would be nicer if SpinBox had the decimal support, instead of having it only in NumberSelectorDialog. If that makes things easier for this implementation, and if it seems straightforward to do, then it would be fine to do it as part of this PR.

components/dialogs/ModalDialog.qml
```diff
@@ -170,5 +173,82 @@ T.Dialog {
...
+				dialogStateGroup.state = "default"
+				return
+			}
+
+			const currentDialogOffset = root.y - root.centeredY // 0 or negative
+			const inputItemBottomPos = inputItem.mapToItem(Global.mainView, 0, inputItem.implicitHeight).y - currentDialogOffset
+
+			targetDialogY = root.centeredY
+
+			let vkbTopPos = Global.mainView.height - Qt.inputMethod.keyboardRectangle.height
```
blammit, line 199: Can be `const`

components/dialogs/ModalDialog.qml
```diff
@@ -170,5 +173,82 @@ T.Dialog {
...
+				// Note: moving the Dialog while in "focused" state will change to
+				// the new location immediately without any animation.
+				targetDialogY += (vkbTopPos - inputItemBottomPos)
+			}
+
+			dialogStateGroup.state = "focused"
+		}
+
+		property StateGroup dialogStateGroup: StateGroup {
+
```
blammit, line 211: Extra newline?

components/controls/SpinBox.qml
```diff
@@ -116,6 +172,27 @@ CT.SpinBox {
 		}
 	}
 
+	textFromValue: function(value, locale) {
+		return Units.formatNumber(value)
+	}
+	valueFromText: function(text, locale) {
+		let value = Units.formattedNumberToReal(text)
+		if(isNaN(value)) {
```
blammit, line 180: Add space between `if` and `(`

blammit (commented):
The editable input works very nicely and plays well with the VKB. I'm about to look into the `valueFromText()` issue further; will let you know how it it goes.

Can you squash some of the commits? At least these commits could be squashed into previous ones (and maybe others too?) as they are amendments/fixes for the original commit that added the feature:
30e8f256e9141f4c2c1968a8d4630b74018b891d
a8a785231f1e60189ed1102639b83deccee08389
c06b14f5c1e91e3e6f1a74da943b4f3f2bfd7096
b70a15a929bf876d0fd556c74a63f9881ebb8a96
c83a9b63bc7e3b9efa294883d3b60c2d25645f5b
56e4284226f53655bc2b05a5bd400fa31c0aad37

And link to the task in the commits that are relevant to it, e.g. put "Part of #1104" in the last line of the commit message.

Also, the commit message for 788deba5853a52e76e1820b2e66ce21fb2abf693 still refers to a DoubleValidator that has been removed.

components/dialogs/ModalDialog.qml
MikeTrahearn-Qinetic: DONE

components/dialogs/ModalDialog.qml
MikeTrahearn-Qinetic: I would say not, but I've removed it as requested.

components/controls/SpinBox.qml
MikeTrahearn-Qinetic: DONE

components/dialogs/NumberSelectorDialog.qml
MikeTrahearn-Qinetic: using one of my favourite logging categories **qt.qml.binding.removal.info=true** I can verify that the line `text: root.textFromValue(root.value, root.locale)` does not break the binding when the text is edited. The binding is still active and is not overwritten; the text will be set inside the C++ using a setText method which doesn't break the binding.

MikeTrahearn-Qinetic: Making all the changes above: I observe:
- open the test spin box dialog, it should show 1.2 >> GOOD
- edit it to be 6 > clamps to 1.5 >> GOOD
- now edit it to be 1.2 >> 1.5 BAD.
The assumption that the text binding breaks has led you down a certain path that seems to not work in all cases whereas as it was, it did. There's no absolutely ideal way to handle this - so we need to compromise otherwise against the lack of DoubleSpinBox in a way that works in all the cases we have which I believe I have done. I don't like it, but it works in all the edge cases, text changes (valid or otherwise), backend value changes and even locale.

MikeTrahearn-Qinetic:
I re-evaluated all the commits and created a set of just four salient commits for the whole feature.

components/dialogs/NumberSelectorDialog.qml
blammit: Nice, thanks for finding the bug in the patch. I didn't know about `qt.qml.binding.removal.info`, that seems very useful. That's good to know the text is set from `setText()` without breaking the text binding. 

We should still aim to avoid modifying `root.value` from `textFromValue()`, since `textFromValue()` isn't expected to modify the `value` itself, and I'm concerned that doing this may produce other unexpected behaviour in the future. Also, doing the clamping in `onAccepted` allows the clamping to be done in a single place, instead of from both `NumberSelectorDialog` and `SpinBox`.

Here is an updated patch that fixes the issue you mentioned:

```
diff --git a/components/controls/SpinBox.qml b/components/controls/SpinBox.qml
index 789ace22..33751f24 100644
--- a/components/controls/SpinBox.qml
+++ b/components/controls/SpinBox.qml
@@ -88,7 +88,6 @@ CT.SpinBox {
 					TextInput {
 						id: primaryLabel
 
-						text: root.textFromValue(root.value, root.locale)
 						color: root.enabled ? Theme.color_font_primary : Theme.color_background_disabled
 						font.family: Global.fontFamily
 						font.pixelSize: root.secondaryText.length > 0 ? Theme.font_size_h2 : Theme.font_size_h3
@@ -102,9 +101,31 @@ CT.SpinBox {
 						inputMethodHints: root.inputMethodHints
 
 						onAccepted: {
-							// Note that the text may at this time represent a value out of SpinBox to/from range.
+							// Note that the text may at this time represent a value out of SpinBox
+							// to/from range, so clamp it here.
+							let v = root.valueFromText(text, root.locale)
+							if (v < root.from) {
+								v = root.from
+							} else if (v > root.to) {
+								v = root.to
+							}
+
+							// Force-update the displayed text, to guarantee the text is in sync
+							// with the numeric value, even if the value has not changed due to the
+							// user entering an out-of-range value on consecutive attempts.
+							text = root.textFromValue(v, root.locale)
+							root.value = v
 							primaryLabel.focus = false
 						}
+
+						Connections {
+							target: root
+							function onValueChanged() {
+								// Update the displayed text when the initial value is set or when
+								// the up/down buttons are pressed.
+								primaryLabel.text = root.textFromValue(root.value, root.locale)
+							}
+						}
 					}
 
 					Label {
@@ -181,15 +202,6 @@ CT.SpinBox {
 			// don't change the current value
 			value = root.value
 		}
-		// need to set spinBox.value to a different temporary valid value
-		// to illicit a value change so it always updates the displayText properly
-		// in all cases, to something that is valid
-		if (value <= root.from) {
-			root.value = root.to
-		} else {
-			// if the value is >= to or any other value above from, set it to from
-			root.value = root.from
-		}
 		return value
 	}
 
diff --git a/components/dialogs/NumberSelectorDialog.qml b/components/dialogs/NumberSelectorDialog.qml
index b551c619..da7078db 100644
--- a/components/dialogs/NumberSelectorDialog.qml
+++ b/components/dialogs/NumberSelectorDialog.qml
@@ -69,15 +69,6 @@ ModalDialog {
 						// don't change the current value
 						value = spinBox.value
 					}
-					// need to set spinBox.value to a different temporary valid value
-					// to illicit a value change so it always updates the displayText properly
-					// in all cases, to something that is valid
-					if (value <= spinBox.from) {
-						spinBox.value = spinBox.to
-					} else {
-						// if the value is >= to or any other value above from, set it to from
-						spinBox.value = spinBox.from
-					}
 					return value
 				}
 				from: Math.max(Global.int32Min, root.from * root._multiplier())

```

MikeTrahearn-Qinetic: I've accepted your patch and rebased.

blammit (approved):
Verified on wasm and cerbo, LGTM

PR #1785, opened 2024-12-05
Fix SpinBox's up and down indicators to disable when limits are reached

MikeTrahearn-Qinetic:
The overall enabled state of the SpinBox will cascade properly into the up and down indicators, and will take into account the limits being reached automatically due to the SpinBox's internal logic setting the explicit enabled property on the indicators. In so adjusting the color binding from using root.enabled to just the indicator's [effective] enabled we get the correct disabled coloring when the limits are reached.


blammit (approved):

blammit:
Note, the commit title says "Fix SpinBox's up and down indicators to disable", but the fix is about whether it looks enabled, rather than whether it is enabled.

MikeTrahearn-Qinetic:
Updated the commit comment to reflect the actual change.

PR #1787, opened 2024-12-06
Reduce global device models

blammit:


components/AllDevicesModel.qml
```diff
@@ -13,25 +13,25 @@ AggregateDeviceModel {
...
-		meteoDeviceModel,
-		motorDriveDeviceModel,
-		pulseMeterDeviceModel,
+		meteoModel,
+		motorDriveModel,
+		pulseMeterModel,
 		Global.pvInverters.model,
 		Global.solarChargers.model,
-		unsupportedDeviceModel,
+		unsupportedModel,
```
chriadam, line 27: I prefer unsupportedDeviceModel actually :-/

blammit: Changed to `unsupportedDevices` as discussed.

Issue #1491, opened 2024-09-19
Always show battery SOC percentage in center of brief page


mpvader:
As its now, its only shown if the brief page only has one item. And hidden when there are two or more circle gauges. Further input here:

https://community.victronenergy.com/t/gui-v2-on-v3-50-7-can-we-have-soc-in-the-circle-on-the-brief-page/124

Next step: @jepefe will discuss this with Serj

chriadam:
I have created  a prototype (see the above PR).  It supports showing a center-value in 1 or 2 or 3 gauge case, but NOT in 4 gauge case (there's not enough space in the middle).  The 1-gauge case is unchanged from the current implementation (except that you can now choose to show a temperature value there instead of the battery state of charge value, as is currently supported).  There are settings to control: 1) which (battery or temperature sensor) device whose value should be shown in the center of the brief page, 2) whether to show the label above the value or not, 3) when the center value should be shown (e.g. only 1 gauge case, vs 1 or 2, vs 1 or 2 or 3 gauge case).  Please see the attached screenshots for the various 2 and 3 gauge cases, both with and without labels.  @Serj do you have any thoughts on this prototype?  Are there any specific changes you would make, or ways this could be improved?  Matthijs asked me to speak to you to get some proper design input so that we can iterate the prototype into a proper solution.

<img width="602" alt="2-gauge-battery-soc-with-label" src="https://github.com/user-attachments/assets/3a629b37-47e6-4ba0-a507-7f82ad3100c4" />
<img width="602" alt="2-gauge-battery-soc-without-label" src="https://github.com/user-attachments/assets/268f9b6a-a88a-4089-a027-063a4512a308" />
<img width="602" alt="2-gauge-temperature-with-label" src="https://github.com/user-attachments/assets/3ca4b4a9-4e0f-4b45-82b0-3db2caec5986" />
<img width="602" alt="2-gauge-temperature-without-label" src="https://github.com/user-attachments/assets/c611e4eb-de53-4784-b346-832a78924356" />
<img width="602" alt="3-gauge-battery-soc-with-label" src="https://github.com/user-attachments/assets/b9797028-dfa8-4560-a377-5b5c01304b61" />
<img width="602" alt="3-gauge-battery-soc-without-label" src="https://github.com/user-attachments/assets/20111a13-da84-4fd4-8fe1-b3e6e72cfb49" />
[...]

PR #1788, opened 2024-12-06
Show battery SoC inside circular multi gauge on Brief Page

chriadam:
Contributes to issue #1491

### DIFF
--- a/components/CenterValueDisplay.qml
+++ b/components/CenterValueDisplay.qml
@@ -1,0 +1,83 @@
+/*
+** Copyright (C) 2024 Victron Energy B.V.
+** See LICENSE.txt for license information.
+*/
+
+import QtQuick
+import QtQuick.Controls.impl as CP
+import Victron.VenusOS
+
+Item {
+	id: root
+
+	required property int gaugeCount
+	required property real maximumWidth
+	readonly property bool isTemperature: _valueDevice.isValid && BackendConnection.serviceTypeFromUid(_valueDevice.value) === "temperature"
+
+	property alias icon: icon
+	property alias label: label
+	property alias quantity: quantity
+
+	implicitWidth: label.visible ? Math.max(icon.width + label.width, quantity.width) : quantity.width
+	implicitHeight: label.visible ? icon.height + quantity.height : quantity.height
+	visible: (!root._valueDisplay.isValid || (root.gaugeCount <= root._valueDisplay.value)) && (_device.isValid && !isNaN(_device.value))
+
+	property VeQuickItem _valueDisplay: VeQuickItem {
+		uid: Global.systemSettings.serviceUid + "/Settings/Gui/BriefView/CenterValueDisplay"
+	}
+
+	property VeQuickItem _labelDisplay: VeQuickItem {
+		uid: Global.systemSettings.serviceUid + "/Settings/Gui/BriefView/CenterValueLabelDisplay"
+	}
+
+	property VeQuickItem _valueDevice: VeQuickItem {
+		uid: Global.systemSettings.serviceUid + "/Settings/Gui/BriefView/CenterValueDevice"
+	}
+
+	property VeQuickItem _device: VeQuickItem {
+		uid: _valueDevice.value + "/" + (root.isTemperature ? "Temperature" : "Soc")
+	}
+
+	CP.ColorImage {
+		id: icon
+		anchors.right: label.left
+		width: Theme.geometry_icon_size_small
+		height: Theme.geometry_icon_size_small
+		visible: label.visible
+		color: Theme.color_font_primary
+		source: root.isTemperature ? "qrc:/images/icon_temp_32.svg" : "qrc:/images/icon_battery_24.svg"
+	}
+	Label {
+		id: label
+		anchors {
+			// usually, center the label over the quantity,
+			// except when there is only one gauge or the quantity only has 1 digit
+			// in which case take icon width into account (i.e. properly centered).
+			horizontalCenterOffset: (root.gaugeCount == 1 || quantity.valueText.length <= 1) ? icon.width/2 : 0
+			horizontalCenter: quantity.horizontalCenter
+			verticalCenter: icon.verticalCenter
+		}
+		visible: !root._labelDisplay.isValid || (root.gaugeCount <= root._labelDisplay.value)
+		font.pixelSize: (root.isTemperature && gaugeCount > 2) ? Theme.font_size_body1 : Theme.font_size_body2
+		color: Theme.color_font_primary
+		text: root.isTemperature ? CommonWords.temperature : CommonWords.battery
+		width: Math.min(implicitWidth, root.maximumWidth)
+		elide: Text.ElideRight
+	}
+	QuantityLabel {
+		id: quantity
+		anchors {
+			top: icon.visible ? icon.bottom : parent.top
+			horizontalCenter: parent.horizontalCenter
+		}
+		font.pixelSize: root.isTemperature && gaugeCount > 2 ? Theme.font_briefPage_battery_percentage_pixelSize - 12
+			: root.isTemperature && gaugeCount > 1 ? Theme.font_briefPage_battery_percentage_pixelSize - 8 // temperature units takes lots of space.
+			: gaugeCount <= 2 ? Theme.font_briefPage_battery_percentage_pixelSize + 4 // larger font size if we have more space.
+			: valueText.length < 3 ? Theme.font_briefPage_battery_percentage_pixelSize // default font size.
+			: Theme.font_briefPage_battery_percentage_pixelSize - 8 // 5-inch "100%" fits with this font size with 3 gauges.
+		unit: root.isTemperature ? Global.systemSettings.temperatureUnit : VenusOS.Units_Percentage
+		value: !root._device.isValid ? NaN
+			: root.isTemperature ? Global.systemSettings.convertFromCelsius(root._device.value)
+			: root._device.value
+	}
+}
--- a/components/CircularSingleGauge.qml
+++ b/components/CircularSingleGauge.qml
@@ -12,8 +12,6 @@ import Victron.Gauges
 Item {
 	id: gauges
 
-	property alias icon: icon
-	property alias name: nameLabel.text
 	property alias voltage: voltageLabel.value
 	property alias current: currentLabel.value
 	property alias value: arc.value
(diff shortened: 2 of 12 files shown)


chriadam:
Still TODO: add settings to control when/whether to show, and also whether to show the icon+label above.

DanielMcInnes:
> Still TODO: add settings to control when/whether to show, and also whether to show the icon+label above.

Will you be adding the settings as part of this PR, or in a different one?

chriadam:
I did add them already, that comment is now out of date.  Will remove it.

pages/settings/PageSettingsDisplayBrief.qml
```diff
@@ -4,12 +4,153 @@
...
+				{ "display": qsTrId("settings_briefview_centervalue_display_lessthanfourgauges"), "value": VenusOS.BriefView_CenterValue_SingleDoubleOrTripleGauge }
+			]
+			currentIndex: {
+				let i
+				for (i = 0; i < optionModel.length; ++i) {
+					if (optionModel[i].value == dataItem.value) {
+						return i
+					}
+				}
+				return VenusOS.BriefView_CenterValue_SingleDoubleOrTripleGauge
```
DanielMcInnes, line 38: Can you do away with this logic, and just use ListRadioButtonGroup's `defaultIndex` property? Likewise for centerValueDeviceGroup

pages/settings/PageSettingsDisplayBrief.qml
```diff
@@ -4,12 +4,153 @@
...
 
 import QtQuick
+import QtQml.Models
 import Victron.VenusOS
 import Victron.Gauges
 
 Page {
 	id: root
 
+	property ObjectModel settingsModel: ObjectModel {
```
DanielMcInnes, line 14: does this need to be a property?

pages/settings/PageSettingsDisplayBrief.qml
```diff
@@ -4,12 +4,153 @@
...
+			id: centerValueDeviceGroup
+			//: Which device value to show in the center of the circular gauges
+			//% "Center value device"
+			text: qsTrId("settings_briefview_centervalue_device")
+			dataItem.uid: Global.systemSettings.serviceUid + "/Settings/Gui/BriefView/CenterValueDevice"
+			updateDataOnClick: false // don't update data automatically.
+
+			property var devices: []
+
+			property Instantiator deviceInstantiator: Instantiator {
```
DanielMcInnes, line 87: do 'deviceInstantiator', 'batteryDevicesModel' and 'systemBatteryService' need to be properties?

DanielMcInnes (commented):
is there a design for this? To my eyes, the center text is a bit too low and a bit too far to the left. More cowbell!
Screenshot from 2024-12-11 13-07-55

components/CenterValueDisplay.qml
```diff
@@ -0,0 +1,83 @@
...
+*/
+
+import QtQuick
+import QtQuick.Controls.impl as CP
+import Victron.VenusOS
+
+Item {
+	id: root
+
+	implicitWidth: label.visible ? Math.max(icon.width + label.width, quantity.width) : quantity.width
```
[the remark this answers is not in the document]
chriadam: done

components/CenterValueDisplay.qml
```diff
@@ -0,0 +1,83 @@
...
+		elide: Text.ElideRight
+	}
+	QuantityLabel {
+		id: quantity
+		anchors {
+			top: icon.visible ? icon.bottom : parent.top
+			horizontalCenter: parent.horizontalCenter
+		}
+		font.pixelSize: root.isTemperature && gaugeCount > 2 ? Theme.font_briefPage_battery_percentage_pixelSize - 12
+			: root.isTemperature && gaugeCount > 1 ? Theme.font_briefPage_battery_percentage_pixelSize - 8 // temperature units takes lots of space.
```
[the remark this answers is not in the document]
chriadam: Currently, we don't have Theme values.  Should we add them?  I personally don't know that there's much need to.  They all look correct in both geometries, yes.

components/CenterValueDisplay.qml
```diff
@@ -0,0 +1,83 @@
...
+	implicitHeight: label.visible ? icon.height + quantity.height : quantity.height
+	visible: (!root._valueDisplay.isValid || (root.gaugeCount <= root._valueDisplay.value)) && (_device.isValid && !isNaN(_device.value))
+
+	property alias icon: icon
+	property alias label: label
+	property alias quantity: quantity
+
+	required property int gaugeCount
+	required property real maximumWidth
+	readonly property bool isTemperature: _valueDevice.isValid && BackendConnection.serviceTypeFromUid(_valueDevice.value) == "temperature"
```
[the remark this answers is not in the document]
chriadam: done

components/CenterValueDisplay.qml
```diff
@@ -0,0 +1,83 @@
...
+	}
+
+	property VeQuickItem _device: VeQuickItem {
+		uid: _valueDevice.value + "/" + (root.isTemperature ? "Temperature" : "Soc")
+	}
+
+	CP.ColorImage {
+		id: icon
+		anchors.right: label.left
+		width: 24
```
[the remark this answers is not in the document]
chriadam: done

pages/settings/PageSettingsDisplayBrief.qml
chriadam: thanks, done

pages/settings/PageSettingsDisplayBrief.qml
chriadam: no, fixed

pages/settings/PageSettingsDisplayBrief.qml
chriadam: batteryDevicesModel does, the others don't - fixed.

chriadam:
There is no design, currently - I was asked to prototype something up, then show it to Serj for feedback and improvement.  I will attach screenshots to the task and contact him.

DanielMcInnes (approved):

chriadam:
We can't merge this until we get proper feedback from Serj, and iterate it.  Also, we need to get whatever properties are required, added into the base system.

pages/settings/PageSettingsDisplayBrief.qml
```diff
@@ -4,12 +4,139 @@
...
 import Victron.Gauges
 
 Page {
 	id: root
 
+	ObjectModel {
+		id: settingsModel
+		ListRadioButtonGroup {
+			//: Whether to show a value in the center of the circular gauges
+			//% "Center value display"
```
blammit, line 18: Would be nice to just have a single "Center display" list item, that takes you to another page with the various center display options.

pages/settings/PageSettingsDisplayBrief.qml
```diff
@@ -4,12 +4,139 @@
...
+			defaultIndex: 3 // VenusOS.BriefView_CenterValue_SingleDoubleOrTripleGauge
+
+			onOptionClicked: (index) => {
+				dataItem.setValue(optionModel[index].value)
+			}
+		}
+
+		ListRadioButtonGroup {
+			//: Whether to show a descriptive label above the value in the center of the circular gauges
+			//% "Center value label"
```
blammit, line 41: It's not clear what the "display" vs "label" means in the settings page; would be nice to have bottomChildren captions for both of them to briefly expand on what these configurations do.

components/CenterValueDisplay.qml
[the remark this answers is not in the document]
blammit: Or just use `fitMode` to fit to the available width?

pages/settings/PageSettingsDisplayBrief.qml
```diff
@@ -4,12 +4,139 @@
...
+	ObjectModel {
+		id: settingsModel
+		ListRadioButtonGroup {
+			//: Whether to show a value in the center of the circular gauges
+			//% "Center value display"
+			text: qsTrId("settings_briefview_centervalue_display")
+			dataItem.uid: Global.systemSettings.serviceUid + "/Settings/Gui/BriefView/CenterValueDisplay"
+			updateDataOnClick: false // don't update data automatically.
+			optionModel: [
+				//% "No center value displayed"
```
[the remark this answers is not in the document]
blammit: Do we really need separate settings for 1 gauge, 2 gauge etc? Could we just have a switch with on/off for whether the display and label should be shown? If the CenterValueDisplay uses a horizontal size-to-fit for the text, then it could be shown regardless of the number of gauges present.

components/CenterValueDisplay.qml
```diff
@@ -0,0 +1,83 @@
...
+
+	property alias icon: icon
+	property alias label: label
+	property alias quantity: quantity
+
+	implicitWidth: label.visible ? Math.max(icon.width + label.width, quantity.width) : quantity.width
+	implicitHeight: label.visible ? icon.height + quantity.height : quantity.height
+	visible: (!root._valueDisplay.isValid || (root.gaugeCount <= root._valueDisplay.value)) && (_device.isValid && !isNaN(_device.value))
+
+	property VeQuickItem _valueDisplay: VeQuickItem {
```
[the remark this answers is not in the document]
MikeTrahearn-Qinetic: The number of times we have property blah inside a QtObject, it would be far more readable if we invented:
QMLObject.qml:

```
import QtQml

QtObject {
     default property alias _children
     property list<QtObject> _children
}
```

Then you could do for example:

```
QMLObject {
    VeQuickItem {
         id: valueDisplay
         uid: Global.systemSettings.serviceUid + "/Settings/Gui/BriefView/CenterValueDisplay"
   }
}
```
 
Which then brings some nice advantages:
- Much more readable
- Private object ids no longer needs an underscore (we only have one for _children we never directly use)
- Private objects is not [directly] accessible outside the component (unless you are misbehaving and use _children)
- The ambiguity of using the property name rather than the id of the object assigned to it is removed
- The potential for overwriting or shadowing the property is removed.
- No need for an inner QtObject to keep things private - and then, no need for a property to assign it

Even better: if the QMLObject was defined in C++ (not hard) then the property could be declared FINAL to avoid property shadowing of the _children property.

components/CenterValueDisplay.qml
```diff
@@ -0,0 +1,83 @@
...
+		id: icon
+		anchors.right: label.left
+		width: Theme.geometry_icon_size_small
+		height: Theme.geometry_icon_size_small
+		visible: label.visible
+		color: Theme.color_font_primary
+		source: root.isTemperature ? "qrc:/images/icon_temp_32.svg" : "qrc:/images/icon_battery_24.svg"
+	}
+	Label {
+		id: label
```
MikeTrahearn-Qinetic, line 51: It's been mentioned to be about removing extra lines used just for readability so I'm just checking here; what's the QML coding style? New line after id only or not?

MikeTrahearn-Qinetic: Also do we keep spacing around operators for example ? * / etc?

components/CenterValueDisplay.qml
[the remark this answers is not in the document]
MikeTrahearn-Qinetic: Is there room enough for 100% (four characters)?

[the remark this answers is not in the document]
MikeTrahearn-Qinetic: Re theme values, it would be better if the Figma design had fewer of them overall and that they were nice intervals of 4 or 8. Figma Variables can allow you do define numerical constants and then bind them to other things like border.large, font.small, spacing.cosy etc. That way, the constants can be "modified" for smaller screens and everything follows - and this can be checked in Figma very easily FIRST. It reduces the number of uncontrolled numerical values all over the theme and brings a focus on the design system's consistency over having lots of Theme.spacingInTheCaseOfThisSpecificThingy type properties.

components/CircularSingleGauge.qml
MikeTrahearn-Qinetic, line 1: Consider changing
`progressColor: Theme.color_darkOk,Theme.statusColorValue(gauges.status)`
to
`progressColor: {Theme.color_darkOk; return Theme.statusColorValue(gauges.status);}`
to get rid of the M30 "Don't use comma expressions" warning - or - consider why this is necessary and do it a different way.

data/mock/config/LevelsPageConfig.qml
```diff
@@ -164,6 +164,11 @@ QtObject {
 				{ type: VenusOS.Tank_Type_FreshWater, level: 75, capacity: 20 },
 			]
 		},
+		{
+			name: "No tanks",
```
MikeTrahearn-Qinetic, line 168: "yes peas" ;-)

pages/settings/PageSettingsDisplayBrief.qml
```diff
@@ -4,12 +4,139 @@
...
+			}
+
+			onOptionClicked: (index) => {
+				dataItem.setValue(devices[index].serviceUid)
+			}
+
+			currentIndex: {
+				let systemBatteryIndex = -1
+				let i
+				for (i = 0; i < devices.length; ++i) {
```
MikeTrahearn-Qinetic, line 106: consider moving the scope of access of "i" inside the for loop:
'for (let i = 0; i < devices.length; ++i) {'
else "i" can be used outside the loop which doesn't look like the intention.

pages/settings/PageSettingsDisplayBrief.qml
```diff
@@ -4,12 +4,139 @@
...
+					if (devices[i].serviceUid == activeBatteryService.value) {
+						systemBatteryIndex = i
+					}
+				}
+				return systemBatteryIndex
+			}
+
+			Component.onCompleted: {
+				var tempDevices = devices
+				devices = tempDevices // force update signal.
```
MikeTrahearn-Qinetic, line 119: Be careful! [QTBUG-131901](https://bugreports.qt.io/browse/QTBUG-131901) has been fixed so this code is a ticking time bomb. Please consider a different means of forcing the update signal if required.

chriadam:
Obsoleted by https://github.com/victronenergy/gui-v2/pull/1948, closing this one now.

PR #1799, opened 2024-12-12
Fix SettingsPage's GradientListView to add clipping

### LINKED ISSUE #1714: Settings: clicking on the top row causes the Generator Start/Stop menu option to be triggered
See thread and video here: https://victrondevelopment.slack.com/archives/C04RYU0UMDE/p1731142689206319

MikeTrahearn-Qinetic:
Clip the GradientListView to prevent delegates that extend beyond the ListView's top and bottom edges from being clickable outside of those boundaries.

This was reproduced and the root cause discovered.

Tested the fix on Desktop, WASM and Cerbo as "can no longer reproduce".

Fixes #1714

### DIFF
--- a/pages/SettingsPage.qml
+++ b/pages/SettingsPage.qml
@@ -23,6 +23,8 @@ SwipeViewPage {
 	GradientListView {
 		id: settingsListView
 
+		clip: true
+
 		model: [
 			{
 				//% "Device list"


chriadam (approved):

chriadam:
Nice catch!

PR #1800, opened 2024-12-12
Add power factor to energy meter page

martinbosma:
image

### DIFF
--- a/src/units.cpp
+++ b/src/units.cpp
@@ -82,6 +82,7 @@ int Units::defaultUnitPrecision(VenusOS::Enums::Units_Type unit) const
 {
 	switch (unit) {
 	case VenusOS::Enums::Units_Energy_KiloWattHour:  return 3;
+	case VenusOS::Enums::Units_PowerFactor:          return 3;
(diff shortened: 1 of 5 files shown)


src/units.cpp
```diff
@@ -154,6 +155,8 @@ QString Units::defaultUnitString(VenusOS::Enums::Units_Type unit, int formatHint
 		return QStringLiteral("kPa");
 	case VenusOS::Enums::Units_CardinalDirection:
 		return DegreesSymbol;
+	case VenusOS::Enums::Units_PowerFactor:
+		return QStringLiteral();
```
blammit, line 159: An empty QStringLiteral works fine, though the codebase normally uses `QString()` for an empty string.

Could you add some tests to `tests/units/tst_units.qml` as well? The tests will run when the app is compiled, if configured with `-DRUN_UNIT_TESTS=ON`.

martinbosma: @blammit Updated to `QString()` and added unit tests for `VenusOS::Enums::Units_PowerFactor`

blammit (approved):

blammit:
Looks okay to merge for today's release.

PR #1802, opened 2024-12-13
dvcc: add setting for control of all multis

izak:
Originally DVCC parameters were used only to control one Multi or Quattro. A new feature in systemcalc makes it possible to control all of them, in systems where there may be more than one. To avoid sudden changes in existing systems, it is an option that defaults to off.

### DIFF
--- a/pages/settings/PageSettingsDvcc.qml
+++ b/pages/settings/PageSettingsDvcc.qml
@@ -211,6 +211,31 @@ Page {
 					uid: bmsName.serviceUid ? "%1/CustomName".arg(bmsName.serviceUid) : ""
 				}
 			}
+
+			ListSwitch {
+				//% "Managed battery controls all Multis and Quattros"
+				text: qsTrId("settings_dvcc_control_all_vebus_devices")
+				dataItem.uid: Global.systemSettings.serviceUid + "/Settings/SystemSetup/DvccControlAllMultis"
+				allowed: defaultAllowed && commonSettings.dvccActive &&
+						 nrVebusDevices.isValid && nrVebusDevices.value > 1
+
+				bottomContentChildren: [
+					PrimaryListLabel {
+						topPadding: 0
+						bottomPadding: 0
+						color: Theme.color_font_secondary
+						//% "When enabled, the managed battery controls all Multis and Quattros. "
+						//% "When disabled, only the device on the built-in port is controlled. "
+						//% "See documentation for further information."
+						text: qsTrId("settings_dvcc_control_all_vebus_devices_label")
+					}
+				]
+
+				VeQuickItem {
+					id: nrVebusDevices
+					uid: Global.system.serviceUid + "/Devices/NumberOfVebusDevices"
+				}
+			}
 		}
 	}
 }


pages/settings/PageSettingsDvcc.qml
```diff
@@ -211,6 +211,32 @@ Page {
...
 				}
 			}
+
+			ListSwitch {
+				id: dvccControlAllVebusSwitch
+				//% "Managed battery controls all Multis and Quattros"
+				text: qsTrId("settings_dvcc_control_all_vebus_devices")
+				dataItem.uid: Global.systemSettings.serviceUid + "/Settings/SystemSetup/DvccControlAllMultis"
+				allowed: defaultAllowed && commonSettings.dvccActive &&
+				         nrVebusDevices.isValid && nrVebusDevices.value > 1
```
DanielMcInnes, line 221: It looks like you have your editor set up for 'tabs as spaces'? 
If you use QtCreator, you can set it up like this under 'Edit -> Preferences':
image

pages/settings/PageSettingsDvcc.qml
```diff
@@ -211,6 +211,32 @@ Page {
 					uid: bmsName.serviceUid ? "%1/CustomName".arg(bmsName.serviceUid) : ""
 				}
 			}
+
+			ListSwitch {
+				id: dvccControlAllVebusSwitch
```
DanielMcInnes, line 216: I think this id can be removed, it's not referenced anywhere

pages/settings/PageSettingsDvcc.qml
```diff
@@ -211,6 +211,32 @@ Page {
...
+					PrimaryListLabel {
+						topPadding: 0
+						bottomPadding: 0
+						color: Theme.color_font_secondary
+						//% "When enabled, the managed battery controls all Multis and Quattros. "
+						//% "When disabled, only the device on the built-in port is controlled. "
+						//% "See documentation for further information."
+						text: qsTrId("settings_dvcc_control_all_vebus_devices_label")
+					}
+				]
```
DanielMcInnes, line 233: @blammit when we have a PrimaryListLabel with a lot of text like this, the text extends a long way to the right. Do you think this should have a bigger rightMargin? I don't think this is covered in the design. It looks ok to me, but not great.
Screenshot from 2024-12-17 12-15-45

blammit: Yeah good question, I'm not sure. Depending on how big the margin is, it might look odd if it wraps early. Might be nice to try out a prototype and see how it looks?

pages/settings/PageSettingsDvcc.qml
izak: I prefer tabs, although I will adapt to what is used in the project. Here the spaces are deliberate, to make it line up with the `allowed:` above it. Also see line 189 where a mix of tabs and spaces is used. I did not want to assume that a tab is always 4 spaces for everyone, even if I have it set up that way. What is the preference here?

pages/settings/PageSettingsDvcc.qml
izak: Right. It was used to hide something else, but then I dropped it. I will remove this id.

blammit (approved):

pages/settings/PageSettingsDvcc.qml
DanielMcInnes: I guess if it's already done like that elsewhere, it's ok.

pages/settings/PageSettingsDvcc.qml
DanielMcInnes: Looks good enough for now

DanielMcInnes (commented):
lgtm

PR #1804, opened 2024-12-17
Add a watchdog to detect crashed wasm builds and reload the page

### LINKED ISSUE #1803: Add a watchdog to restart wasm builds

DanielMcInnes:
Fixes #1803

### DIFF
--- a/.github/patches/index.html.patch
+++ b/.github/patches/index.html.patch
@@ -1,6 +1,5 @@
-diff -u index.html venus-gui-v2.html
---- index.html	2024-05-16 22:12:44.106129598 +0300
-+++ venus-gui-v2.html	2024-07-05 12:44:44.396192697 +0200
+--- venus-gui-v2.html	2024-12-19 11:52:03.677046276 +1000
++++ index.html	2024-12-19 13:54:58.232268241 +1000
 @@ -12,15 +12,18 @@
      <title>venus-gui-v2</title>
      <style>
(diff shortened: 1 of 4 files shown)


src/backendconnection.cpp
```diff
@@ -193,6 +193,16 @@ void BackendConnection::onReloadPageTimerExpired()
...
 }
 
+EM_JS(bool, hitWatchdog, (), {
+	watchdogHit = true // 'watchdogHit' is defined in venus-gui-v2.html, which checks it periodically and reloads the page if not hit regularly.
+});
+
+// Sometimes, the wasm code may crash. Use a watchdog to detect this and reload the page when necessary.
+void BackendConnection::onWatchdogTimerExpired()
+{
+	hitWatchdog();
```
chriadam, line 203: just checking, is it possible to just do: `emscripten_run_script("watchdogHit = true")` instead of defining an EM_JS function?

chriadam:
There's an extra commit which needs to be squashed, by the looks

src/backendconnection.cpp
DanielMcInnes: yes it is, updated as suggested

DanielMcInnes:
> There's an extra commit which needs to be squashed, by the looks

done

.github/patches/index.html.patch
```diff
@@ -63,3 +62,19 @@ diff -u index.html venus-gui-v2.html
...
+     <script type="text/javascript" src="qtloader.js"></script>
++    <script type="text/javascript">
++        var watchdogHit = false // this gets set to 'true' by a timer in BackendConnection::onWatchdogTimerExpired()
++        setInterval(function() {
++            if (!watchdogHit) {
++                console.error("Watchdog timer expired - reloading page")
++                location.reload()
++            }
++            watchdogHit = false
++        }, 20000);
```
DanielMcInnes, line 77: What do you think of making the first timeout 40 seconds (or more), and then changing it to 5 seconds? That would allow for very slow internet connections to start a wasm session, and also be a bit more responsive once loaded.

.github/patches/index.html.patch
```diff
@@ -63,3 +62,26 @@ diff -u index.html venus-gui-v2.html
...
+     </script>
+     <script src="venus-gui-v2.js"></script>
+     <script type="text/javascript" src="qtloader.js"></script>
++    <script type="text/javascript">
++        var watchdogHit = false // this gets set to 'true' by a timer via BackendConnection::hitWatchdog()
++        console.log("starting watchdog timer")
++        setTimeout(function(){ // wait 2 minutes for the page to load, then set a 10 second watchdog timer
++            checkWatchdog()
++            setInterval(checkWatchdog, 10000)
++        }, 120000)
```
chriadam, line 75: code LGTM, but I don't know about these intervals.  We should ask Jesús I think

chriadam (approved):

PR #1806, opened 2024-12-18
Improve VE.Bus inverter support

martinbosma:
Some items were visible for inverters while not applicable, such as active input and AC phase data. I copied the behavior mainly from gui-v1.


blammit (approved):
Okay, so a `vebus` service could be an inverter/charger, or just a plain inverter, is that right? I had thought that plain inverters would use the `inverter` service, but maybe that is only for older products?

martinbosma:
Correct, and the vebus inverters are actually the older products but still being produced.

PR #1808, opened 2024-12-19
Miketrahearn/1492 alarm indicator prototype

MikeTrahearn-Qinetic:
This is a prototype of the new Alarms UI and API


MikeTrahearn-Qinetic:
Updated the filtering logic and visual appearance according to Victron requirements.

MikeTrahearn-Qinetic:
TODO: review all cases where we might be using the /Alarm or /Alert values (0 or 1) as these are proposed to be depreciated for Gui-v2 in favour of additional paths describing the counters in more detail.*

MikeTrahearn-Qinetic:
TODO: replace "acknowledgeAll" with "silenceAll" (Notifications.qml /AcknowledgeAll) and consider if we need to change the wording elsewhere for example in the StatusBar "Acknowledge alerts" and "Silence Alarm" buttons

MikeTrahearn-Qinetic:
@ReinvdZee this branch has been fully updated to adhere to the proposed UI and backend API changes for Notifications and may be tested with matching prototype firmware.

Couple of outstanding questions regarding the text on some controls now we are moving away from the the term "alert"...

- StatusBar "Silence alarm" button: replace translation with "Silence Alarms" (pluralise)
- StatusBar "Acknowledge alerts" button: replace translation with "Acknowledge notifications"?
- NotificationsPage: "No current alerts": replace translation with "No current notifications"?

What are your thoughts on these suggestions or just leave them as they are (the logic behind them is updated to the new API), as we are just talking about visual text updates.

components/NotificationCounter.qml
```diff
@@ -0,0 +1,40 @@
...
+			centerIn: parent
+			// just to apply a tiny fix for the font's baseline
+			// without having to use TextMetrics
+			verticalCenterOffset: 0.5
+		}
+		// always white over a red background
+		color: Theme.color_white
+		text: root.count
+		verticalAlignment: Text.AlignVCenter
+		font.pixelSize: 14
```
chriadam, line 38: I suspect this should come from a theme token

components/NotificationCounter.qml
```diff
@@ -0,0 +1,40 @@
+/*
+** Copyright (C) 2023 Victron Energy B.V.
```
chriadam, line 2: 2025

components/NotificationCounter.qml
```diff
@@ -0,0 +1,40 @@
...
+Rectangle {
+	id: root
+
+	property int count: 0
+	property color borderColor: "transparent"
+
+	// fixed height, minimum width (height as circle)
+	// but can extend its width to accomodate the text
+	height: 20
+	width: Math.max(height, countText.width + 12)
```
chriadam, line 18: Is the +12 basically the sum of left+right paddings?  Should those paddings be defined as theme tokens?

components/NotificationCounter.qml
```diff
@@ -0,0 +1,40 @@
...
+
+Rectangle {
+	id: root
+
+	property int count: 0
+	property color borderColor: "transparent"
+
+	// fixed height, minimum width (height as circle)
+	// but can extend its width to accomodate the text
+	height: 20
```
chriadam, line 17: Theme token

data/Notifications.qml
```diff
@@ -11,33 +11,184 @@ QtObject {
...
+			toastedNotif.toast?.close(true)
+			toastedNotif.toast = null
+			toastedNotif.notification = null
+		}
+
+		function checkAndRemoveExistingToast(notif: BaseNotification) {
+			if(toast && (notif.type === VenusOS.Notification_Alarm ||
+						 (notif.type === VenusOS.Notification_Warning && toast.category !== VenusOS.Notification_Alarm) ||
+						 toast.category === VenusOS.Notification_Info)) {
+				close()
```
chriadam, line 54: just checking: hopefully this doesn't "acknowledge" the underlying notification, but instead just merely closes the toast popup?

src/notificationsmodel.cpp
```diff
@@ -200,3 +155,64 @@ QHash<int, QByteArray> NotificationsModel::roleNames() const
 {
 	return m_roleNames;
 }
+
+void NotificationsModel::roleChangedHandler(BaseNotification *notification, NotificationRoles role)
+{
+	if(notification == nullptr) {
```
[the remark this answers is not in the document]
MikeTrahearn-Qinetic: DONE - and all other cases in changed files.

src/notificationsortfilterproxymodel.cpp
```diff
@@ -0,0 +1,155 @@
+/*
+** Copyright (C) 2023 Victron Energy B.V.
```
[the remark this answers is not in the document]
MikeTrahearn-Qinetic: DONE - also in all other changed files for this PR.

src/notificationsortfilterproxymodel.cpp
```diff
@@ -0,0 +1,155 @@
...
+		return d->m_sortFunction.call(args).toBool();
+	}
+	return QSortFilterProxyModel::lessThan(sourceLeft, sourceRight);
+}
+
+void NotificationSortFilterProxyModel::updateCount()
+{
+	const int count = rowCount();
+
+	if(m_count == count) {
```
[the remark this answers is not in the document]
MikeTrahearn-Qinetic: DONE

data/Notifications.qml
MikeTrahearn-Qinetic: Yes, it just closes the Toast. The notification delegate remains intact as no changes to the underlying model are made when closing a Toast.

src/notificationsmodel.h
```diff
@@ -7,117 +7,70 @@
...
+	Q_ENUM(NotificationRoles);
 
 	explicit NotificationsModel(QObject *parent = nullptr);
 
 	int count(const QModelIndex& parent = QModelIndex()) const;
 	int rowCount(const QModelIndex &parent) const override;
 	QVariant data(const QModelIndex& index, int role) const override;
 
-	Q_INVOKABLE void insertByDate(Victron::VenusOS::BaseNotification *notification);
+	//Q_INVOKABLE void insertByDate(Victron::VenusOS::BaseNotification *notification);
```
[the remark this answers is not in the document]
MikeTrahearn-Qinetic: Removed.

components/NotificationCounter.qml
MikeTrahearn-Qinetic: DONE

components/NotificationCounter.qml
MikeTrahearn-Qinetic: I would have to define another one - but also need to check the figma design first - and then, I suspect it isn't appropriate to the confirmed max 2 digits - this is why I want the source of truth to be the source of truth; when it isn't then we have to invent new ones which gets messy over time.

src/notificationsortfilterproxymodel.cpp
```diff
@@ -0,0 +1,155 @@
...
+		if (!d->m_engine)
+			qWarning() << "NotificationSortFilterProxyModel can't sort without a JavaScript engine";
+	}
+	if (d->m_engine && d->m_sortFunction.isCallable()) {
+		const QVariantMap left = get(sourceLeft.row());
+		const QVariantMap right = get(sourceRight.row());
+		QJSValueList args = { d->m_engine->toScriptValue(left), d->m_engine->toScriptValue(right) };
+		return d->m_sortFunction.call(args).toBool();
+	}
+	return QSortFilterProxyModel::lessThan(sourceLeft, sourceRight);
```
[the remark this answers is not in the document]
MikeTrahearn-Qinetic: Since the m_sortFunction is defined in JavaScript, the return value here (assuming such a function is defined) will be entirely determine by what JavaScript defines. In the case of Notifications, the sorting is only done on the type (int) and the dateTime (QDateTime) so I don't think any locale awareness is required at this time. Should it need to be, the sort function should be revisited.

```// Notifications.qml
function _notificationSortFunction(leftNotification: var, rightNotification: var) : bool {
        return leftNotification.type < rightNotification.type &&
                leftNotification.dateTime > rightNotification.dateTime
}

File: components/NotificationCounter.qml (hunk 1 above)
Comment (MikeTrahearn-Qinetic) [reply components/NotificationCounter.qml]: DONE

File: components/NotificationCounter.qml (hunk 4 above)
Comment (MikeTrahearn-Qinetic) [reply components/NotificationCounter.qml]: DONE

File: components/NotificationCounter.qml (hunk 3 above)
Comment (MikeTrahearn-Qinetic) [reply components/NotificationCounter.qml:18]: This one is weird because it is a bit of a manual gap between the left and the right of the red circle - which is actually a horizontally expanding "lozenge" shape. The 12 here represents a left 6 and right 6 pseudo-padding I supposed to make sure that the rounded ends of the "lozenge" are far enough away from the text that the text fits inside it. Thinking about it, there might be a better way to do this.

Comment (MikeTrahearn-Qinetic) [reply components/NotificationCounter.qml:18]: I also need to update the graphic for the bell - there's a "cut-out" version of it in Figma which I need to use when the red dot is displayed.

**Comment by MikeTrahearn-Qinetic** [comment]
@chriadam @blammit : The code in the PR functions now in accordance to testing by @ReinvdZee and he is now approved the new functionality against his new backend and has set "go" for merge (both ends together are necessary for this change). Please could you review the Guiv2 side for any final tweaks or approval. Thanks!

File: components/NavBar.qml [hunk 11]
```diff
@@ -79,20 +81,19 @@ FocusScope {
...
+				Loader {
+					z: 1 // to get it on top of the IconLabel
 					anchors {
+						left: parent.horizontalCenter
+						leftMargin: 1
 						top: parent.top
-						topMargin: Theme.geometry_navigationBar_notifications_redDot_topMargin
-						horizontalCenter: parent.horizontalCenter
-						horizontalCenterOffset: Theme.geometry_navigationBar_notifications_redDot_horizontalCenterOffset
+						topMargin: 6
```
chriadam, line 90: the constants here (1 and 6) should probably be theme tokens, I think?

data/Notifications.qml
```diff
@@ -32,12 +145,74 @@ QtObject {
 		uid: root.serviceUid + "/AcknowledgeAll"
 	}
 
-	readonly property VeQuickItem _alarm: VeQuickItem {
-		uid: root.serviceUid + "/Alarm"
+	readonly property bool statusBarNotifcationIconVisible: activeOrUnAcknowledgedModel.count > 0
```
chriadam, line 148: nitpick: typo: should be statusBarNotificationIconVisible

data/Notifications.qml
```diff
@@ -32,12 +145,74 @@ QtObject {
...
+	readonly property int statusBarNotifcationIconPriority: alarms.hasActive || !alarms.hasActive && alarms.hasUnAcknowledged ? VenusOS.Notification_Alarm :
```
chriadam, line 149: same typo here, missing i between f and c

data/mock/NotificationsImpl.qml
```diff
@@ -96,92 +87,133 @@ QtObject {
...
-	}
 
-	function removeLastNotification() {
-		const notif = notifications[notifications.length - 1]
-		const notificationId = notif.notificationId
-		if (notif.active) {
-			_numberOfActiveNotifications.setValue(Math.max(0, _numberOfActiveNotifications.value - 1))
+		// Whether we created a new Notification or recycled an existing one,
+		// we increment (or wrap) the nextNotificationId.
+		nextNotificationId++
```
chriadam, line 156: does anything bind to this, and if so, will this cause problems in the maxNotificationCount case?  I would prefer something like:
```
nextNotificationId = ((nextNotificationId + 1) === maxNotificationCount) ? 0 : (nextNotificationId + 1);
```
to avoid this intermediate value issue.

data/Notifications.qml
MikeTrahearn-Qinetic: As in other places, fixed.

data/mock/NotificationsImpl.qml
MikeTrahearn-Qinetic: Accepted suggestion after testing the maximumNotificationCount scenario remains identical.

chriadam (approved):
A couple of nitpicks, but aside from that LGTM.  Nice work!

components/NavBar.qml
MikeTrahearn-Qinetic: One of them was still in existence but needed updating and the other added. Done.

MikeTrahearn-Qinetic:
Resolved all comments - going for a merge when the tests have passed.

PR #1809, opened 2024-12-19
Update device-type pages to provide their own titles

blammit:
When pushing a page that uses a device name for the page title, have each page provide this title internally (using the device name) instead of passing the title in as a fixed name.

This way, the page title is updated if the device name changes (e.g. if the custom name changes) and also we don't need to duplicate the code that passes in the page title as a parameter when pushing the page.

### DIFF
--- a/components/widgets/DcInputWidget.qml
+++ b/components/widgets/DcInputWidget.qml
@@ -36,7 +36,6 @@ OverviewWidget {
 	onClicked: {
 		if (root.inputs.count === 1) {
 			Global.pageManager.pushPage(root.detailUrl, {
-				"title": root.inputs.firstObject.name,
 				"bindPrefix": root.inputs.firstObject.serviceUid
 			})
 		} else {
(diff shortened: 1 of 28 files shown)


DanielMcInnes (approved):
lgtm. There are 11 device-type pages that have:
```
	title: device.name

	Device {
		id: device
		serviceUid: root.serviceUid
	}
```
Do you think it's worth adding a 'DevicePage' type to encapsulate this?

blammit:
> Do you think it's worth adding a 'DevicePage' type to encapsulate this?

I'm reluctant to if that's the only benefit. If you have a device-type page and you want to override the title, then there's no point in using DevicePage, and then do you still extend DevicePage just for the semantics? Also, once we have a base-type component, it creates restrictions, e.g. what if you have a device-type page but it needs to extend something else? I would prefer to follow the "composition over inheritance" idea here, unless we get more requirements for device-type pages that make it more necessary to have a base DevicePage.

PR #1810, opened 2024-12-19
Move startstop1 uid finding feature into a separate component

blammit:
This is a general cleanup to make it clear how to reuse this functionality from elsewhere if that is needed.

### DIFF
--- a/components/GensetStartStop1Finder.qml
+++ b/components/GensetStartStop1Finder.qml
@@ -1,0 +1,62 @@
+/*
+** Copyright (C) 2024 Victron Energy B.V.
+** See LICENSE.txt for license information.
+*/
+
+import QtQuick
+import Victron.VenusOS
+
+/*
+	This component provides a way to find the startstop1 uid for a genset service.
+	On D-Bus this is always "com.victronenergy.generator.startstop1", but on a MQTT backend this
+	is less easily located.
+
+	For some background context, there are two start/stop generator types:
+
+	- startstop0
+		The original start/stop generator. This turns the Cerbo's built-in relay on/off.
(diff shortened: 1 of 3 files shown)


components/GensetStartStop1Finder.qml
```diff
@@ -0,0 +1,60 @@
...
+	- startstop1
+		Added for Fischer Panda generators. This works with an actual genset, by sending a command
+		to it instead of closing the relay.
+		On D-Bus, the uid is "com.victronenergy.generator.startstop1".
+		On MQTT, the service uid is found by locating the generator service with a /GensetService
+			value that matches that of a genset or dcgenset service. It might turn out to be
+			"mqtt/generator/1", but that is not guaranteed.
+
+	So, unlike for many other service types, these services these do not refer to different generator
+	instances; instead, they are used to distinguish between two different types of generators.
```
blammit, line 30: @izak could you please check these comments - is this correct or have I misunderstood the startstop0/startstop1 differentiation?

DanielMcInnes (approved):

components/GensetStartStop1Finder.qml
izak: This is correct. It is not ideal of course: we prefer that the 4th component in the service path not be used for identification, but right now it is what it is.

The only correction I need to make is that startstop1 is for more than just FischerPanda. It will control any genset service with start-stop capability, which also includes ComAp and DSE controllers and maybe more in future. The concept is the same though: Genset services that have their own built-in start/stop functionality that can be remotely commanded.

blammit: Thanks! Updated as suggested.

PR #1812, opened 2024-12-23
Miketrahearn/1239 align genset settings and control card

MikeTrahearn-Qinetic:


MikeTrahearn-Qinetic:
@blammit should I be considering using your new work in #1810 for this ticket?

components/PageGensetModel.qml
```diff
@@ -54,10 +54,37 @@ ObjectModel {
...
+		Component {
+			id: confirmationDialogComponent
+
+			ModalWarningDialog {
+				dialogDoneOptions: VenusOS.ModalDialog_DoneOptions_OkAndCancel
+
+				//% "Disable autostart?"
+				title: qsTrId("controlcard_generator_disableautostartdialog_title")
+
+				//% "Autostart will be disabled and the generator won't automatically start based on the configured conditions.\nIf the generator is currently running due to a autostart condition, disabling autostart will also stop it immediately.
```
DanielMcInnes, line 80: missing trailing "
Also, you can move both of these strings to CommonWords if you want

pages/controlcards/GeneratorCard.qml
```diff
@@ -90,7 +91,7 @@ ControlCard {
 				//% "Disable autostart?"
 				title: qsTrId("controlcard_generator_disableautostartdialog_title")
 
-				//% "Autostart will be disabled and the generator won't automatically start based on the configured conditions."
+				//% "Autostart will be disabled and the generator won't automatically start based on the configured conditions.\nIf the generator is currently running due to a autostart condition, disabling autostart will also stop it immediately.
```
DanielMcInnes, line 94: missing trailing "

MikeTrahearn-Qinetic: Noted.

components/PageGensetModel.qml
MikeTrahearn-Qinetic: Noted.

components/PageGensetModel.qml
```diff
@@ -54,10 +54,37 @@ ObjectModel {
...
+				   }
+
+		Component {
+			id: confirmationDialogComponent
+
+			ModalWarningDialog {
+				dialogDoneOptions: VenusOS.ModalDialog_DoneOptions_OkAndCancel
+
+				//% "Disable autostart?"
+				title: qsTrId("controlcard_generator_disableautostartdialog_title")
```
DanielMcInnes, line 78: if you want to reuse the translation from GeneratorCard.qml, you don't need the `//% "Disable autostart?"` above it. We don't usually reuse translations from other files like this, except for CommonWords.qml. It should work though.

components/PageGensetModel.qml
MikeTrahearn-Qinetic: Fixed trailing "

pages/controlcards/GeneratorCard.qml
MikeTrahearn-Qinetic: Fixed trailing "

components/PageGensetModel.qml
MikeTrahearn-Qinetic: Can I check with @blammit about adding new translations to the global common words because I have a vague recollection of a conversation about limiting these cases given other complex discussions around its use.

MikeTrahearn-Qinetic: I think the consensus is that we don't want to put such long and contextual phrases in common words.

DanielMcInnes: Are you intentionally using the translation id "controlcard_generator_disableautostartdialog_title" in 2 different places? If so, I'm not sure which engineering text would be used - could be ambiguous?

components/PageGensetModel.qml
```diff
@@ -54,10 +54,37 @@ ObjectModel {
...
 
 	ListSwitch {
+		id: autostartSwitch
 		//% "Auto start functionality"
 		text: qsTrId("ac-in-genset_auto_start_functionality")
 		allowed: root.gensetEnabled.value === 1
 		dataItem.uid: root.startStopBindPrefix ? root.startStopBindPrefix + "/AutoStartEnabled" : ""
+		updateDataOnClick: false
+
+		onClicked: if (!checked) {
```
blammit, line 64: Enclose the `onClicked` code inside braces since it has multiple lines. If it's a single-line statement then it's more down to personal preference, though we should probably standardize on this.

data/common/Generator.qml
```diff
@@ -56,6 +57,10 @@ Device {
 		uid: serviceUid + "/AutoStartEnabled"
 	}
 
+	readonly property VeQuickItem _startstop1Enabled: VeQuickItem {
+		uid: serviceUid + "/startstop1/Enabled"
```
blammit, line 61: The path should be just "/Enabled", no need for the "/startstop1". 

On D-Bus this will produce a VeQuickItem uid like `dbus/com.victronenergy.generator.startstop1/Enabled`. On MQTT, the uid will be like  `mqtt/generator/1/Enabled`.

data/common/Generator.qml
```diff
@@ -14,6 +14,7 @@ Device {
 	readonly property int manualStartTimer: _manualStartTimer.isValid ? _manualStartTimer.value : 0
 	readonly property int runtime: _runtime.value || 0
 	readonly property int runningBy: _runningBy.isValid ? _runningBy.value : 0
+	readonly property bool startstop1Enabled: _startstop1Enabled.value === 1
```
blammit, line 17: The name can be just `enabled`.

blammit (commented):
Can you add a reference to the issue number in each commit?

blammit:
> @blammit should I be considering using your new work in https://github.com/victronenergy/gui-v2/pull/1810 for this ticket?

No, that's not necessary; that PR is about finding the generator for a genset service, which you don't need to do here.

components/PageGensetModel.qml
MikeTrahearn-Qinetic: DONE

DanielMcInnes (approved):

blammit (approved):
Could you also patch data/mock/GeneratorsImpl.qml to set Enabled=1 for the mock generators? Like this:

```
diff --git a/data/mock/GeneratorsImpl.qml b/data/mock/GeneratorsImpl.qml
index 642df01d..a5c715bf 100644
--- a/data/mock/GeneratorsImpl.qml
+++ b/data/mock/GeneratorsImpl.qml
@@ -33,6 +33,7 @@ QtObject {
                                _customName.setValue("Start/Stop generator")
                                _state.setValue(VenusOS.Generators_State_Running)
                                _runningBy.setValue(VenusOS.Generators_RunningBy_Soc)
+                               Global.mockDataSimulator.setMockValue(generator.serviceUid + "/Enabled", 1)
                                setAutoStart(true)
                        }
```

Otherwise they won't show up in the generator cards.

~Aside from that, LGTM.~ Could you add the common dialog component as well?

components/PageGensetModel.qml
```diff
@@ -54,10 +54,39 @@ ObjectModel {
...
+			} else {
+				// check if they really want to disable
+				Global.dialogLayer.open(confirmationDialogComponent)
+			}
+		}
+
+		Component {
+			id: confirmationDialogComponent
+
+			ModalWarningDialog {
```
blammit, line 76: Could you move this into a common component to avoid the code duplication with GeneratorCard?

MikeTrahearn-Qinetic: Ah I missed that one...

MikeTrahearn-Qinetic: See https://github.com/victronenergy/gui-v2/pull/1825 for the change that implements this.

PR #1819, opened 2025-01-07
Rework ListItem layout and add ListLink component

blammit:


blammit:
@mr-manuel can you try this out and see if it is suitable?

components/listitems/ListLink.qml
```diff
@@ -0,0 +1,79 @@
...
+import QtQuick.Controls.impl as CP
+import Victron.VenusOS
+
+ListItem {
+	id: root
+
+	property string url
+	property string caption: root.mode === VenusOS.ListLink_Mode_LinkButton ? ""
+		  //: %1 = url text
+		  //% "Scan the QR code with your portable device.<br />Or insert the link: %1"
```
MikeTrahearn-Qinetic, line 16: Just checking on the use of <br /> here - why not use \n instead? Does the Text element that will contain this caption have explicit or automatic support for rich text?

components/listitems/ListLink.qml
```diff
@@ -0,0 +1,79 @@
...
+			//% "Open link"
+			text: qsTrId("listlink_open_link")
+		},
+
+		CP.ColorImage {
+			id: icon
+
+			visible: root.mode === VenusOS.ListLink_Mode_LinkButton
+			anchors.verticalCenter: parent.verticalCenter
+			source: "qrc:/images/icon_open_link_32.svg"
```
MikeTrahearn-Qinetic, line 38: should we be using Qt.resolvedUrl() to make sure that QML files and their associated image resources are portable?

components/listitems/ListLink.qml
```diff
@@ -0,0 +1,79 @@
...
+
+		Item {
+			visible: root.mode === VenusOS.ListLink_Mode_QRCode
+			width: Theme.geometry_listLink_qrCodeSize
+			height: Theme.geometry_listLink_qrCodeSize + (2 * Theme.geometry_listItem_content_verticalMargin)
+
+			Image {
+				anchors.centerIn: parent
+				source: root.mode === VenusOS.ListLink_Mode_QRCode
+						? "image://QZXing/encode/" + root.url + "?correctionLevel=M" + "&format=qrcode"
```
MikeTrahearn-Qinetic, line 51: Consider using string template? image://QZXing/encode/${root.url}?correctionLevel=M&format=qrcode

components/listitems/core/ListItem.qml
```diff
@@ -69,39 +67,46 @@ Item {
...
+			font.pixelSize: flat ? Theme.font_size_body1 : Theme.font_size_body2
+			wrapMode: Text.Wrap
+			width: root.availableWidth - content.width - Theme.geometry_listItem_content_spacing
+		}
 
-		anchors {
-			right: parent.right
-			rightMargin: root.rightPadding
-			verticalCenter: primaryLabel.verticalCenter
+		Row {
```
MikeTrahearn-Qinetic, line 90: Mixing Row (positioner) with GridLayout (layout) - suggest using RowLayout for cleaner sizing in this case

pages/settings/debug/PageSettingsDemo.qml
```diff
@@ -300,6 +327,11 @@ Page {
 					}
 				]
 			}
+
+			ListLink {
+				text: "Victron Energy"
+				url: "http://www.victronenergy.com"
```
MikeTrahearn-Qinetic, line 333: Use secure https://www.victronenergy.com

themes/geometry/FiveInch.json
```diff
@@ -59,6 +58,8 @@
     "geometry_listItem_textField_minimumWidth": 200,
     "geometry_listItem_textField_maximumWidth": 376,
 
+    "geometry_listLink_qrCodeSize": 96,
```
MikeTrahearn-Qinetic, line 61: Checking if this should be a different size to the 7 inch

MikeTrahearn-Qinetic (commented):
General note: the layout strategy is very complex using anchors, spacing properties from numerous places. Perhaps this is just legacy, but I'm left wondering as to the general state of the design if it this complex.

components/listitems/ListLink.qml
blammit: Do you mean, to allow the default images to be overridden? Currently we don't have a case for loading images from anywhere other than the root images directory.

components/listitems/ListLink.qml
blammit: Done

pages/settings/debug/PageSettingsDemo.qml
blammit: Done

themes/geometry/FiveInch.json
blammit: This should be okay for now.

components/listitems/ListLink.qml
blammit: The formattedUrl contains a <font> tag, so then this also needs the <br> to render the newline.

components/listitems/core/ListItem.qml
blammit: Yes true, that might be nicer for consistency as well, though there is a lot of code that assumes the `content` is a Row and therefore allows the setting of `anchors.verticalCenter` within the child, so unless we really need to use a RowLayout, probably simpler to keep it as a Row.

blammit:
> General note: the layout strategy is very complex using anchors, spacing properties from numerous places. Perhaps this is just legacy, but I'm left wondering as to the general state of the design if it this complex.

The layouts of list items can certainly be improved now that we have ported all of the settings from gui-v1 and have a better idea of the different layouts that are required. Ideally the spacing/margins would be consolidated to have a common set across the board, instead of having highly customized values for different items.

components/listitems/ListLink.qml
MikeTrahearn-Qinetic: I didn't mean to replace the images, but if the images are for example located in the same QML Module as the QML referencing it, then if that module ever moves, then the image URLs would break. Using Qt.resolvedUrl() removes this risk - if indeed it is a risk. If the strategy is that all image resources are locked away elsewhere and will never move relative to the QML references (because they are always absolute), then this is not a thing to worry about - only that if ever the images move, ALL your QML would have to be updated. Just food for thought.

components/listitems/core/ListItem.qml
MikeTrahearn-Qinetic: Understood.

components/listitems/ListLink.qml
blammit: Yes, that's the current strategy at the moment, but yes it does mean we will have to update all of the QML if the images resources are moved.

mr-manuel:
@blammit I was now able to test it. Looks good, thanks!

PR #1834, opened 2025-01-13
Fix screen unblanking

### LINKED ISSUE #1831: Screen doesn't turn back on after it's turned off
On the device, click the screen-off button in the top-right corner. The screen is now blank. Tap the screen, and observe it doesn't turn back on as expected.

blammit:
Set the window property instead of relying on the windowChanged signal to provide the value. Otherwise, ScreenBlanker::eventFilter() does not receive the touch events as expected.

Fixes #1831

### DIFF
--- a/ApplicationContent.qml
+++ b/ApplicationContent.qml
@@ -4,6 +4,7 @@
 */
 
 import QtQuick
+import QtQuick.Window
 import QtQuick.Controls as QtQuickControls
 import Victron.VenusOS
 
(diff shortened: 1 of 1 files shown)


ApplicationContent.qml
```diff
@@ -43,12 +45,11 @@ Item {
 		Component.onCompleted: Global.firmwareUpdate = firmwareUpdate
 	}
 
-	onWindowChanged: function (window) { screenBlanker.window = window }
-
 	ScreenBlanker {
 		id: screenBlanker
 		enabled: !Global.splashScreenVisible && !(!!Global.pageManager && Global.pageManager.statusBar.notificationButtonVisible)
 		displayOffTime: displayOffItem.isValid ? 1000*displayOffItem.value : 0.0
+		window: root.window
```
MikeTrahearn-Qinetic, line 52: try window: root.Window.window here and then line 15 can go, as well as the binding in Main.qml::107

DanielMcInnes (approved):

ApplicationContent.qml
DanielMcInnes: That's a new syntax to me. Would `root.Window.window` refer to the Window from Main.qml? A quick google doesn't tell me much

chriadam: https://doc.qt.io/qt-6/qml-qtquick-window.html#window-attached-prop

MikeTrahearn-Qinetic: Window.window refers to the window that the item is rendered in, so yes it does. I asked Bea to check in this use case if Window.window updates at the correct time to make it all work as expected.

blammit: Updated as suggested.

PR #1837, opened 2025-01-13
Ensure solar device name does not overlap with quantity labels

blammit:


blammit:
Note, this bug is in the Overview solar drilldown list:

image

MikeTrahearn-Qinetic (approved):

MikeTrahearn-Qinetic:
Does it need to elide the text?

blammit:
The existing behaviour is to wrap the text, so I'll keep that for now.

PR #1841, opened 2025-01-14
Improve GPS support

dirkjanfaber:
- Show the cardinal direction for the course.
- Add a proper unit for the altitude (meter / feet)

Note that there is also a [patch](https://github.com/victronenergy/veutil/tree/faberd/altitude-support) on veutil for altitude support. 

image

### DIFF
--- a/data/SystemSettings.qml
+++ b/data/SystemSettings.qml
@@ -18,6 +18,7 @@ QtObject {
 	property int temperatureUnit: VenusOS.Units_None
 	property string temperatureUnitSuffix
 	property int volumeUnit: VenusOS.Units_None
+	property int altitudeUnit: VenusOS.Units_None
 	readonly property StartPageConfiguration startPageConfiguration: StartPageConfiguration {
 		systemSettingsUid: root.serviceUid
 	}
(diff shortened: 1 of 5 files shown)


DanielMcInnes:
Hi @dirkjanfaber , is there an installation where I can test this?

dirkjanfaber:
Send info on the installation via Slack. 

Another thing to note is that the Venus settings currently doesn't have a general setting for Altitude (via venus-platform). To me it would seem logical to add it as `/System/Units/Altitude`, comparable with the Kelvin / Celsius TemperatureUnit here: https://github.com/victronenergy/venus-platform/blob/master/src/application.cpp#L237

Alternative is to add it as `/Gps/AltitudeUnit` in line with the existing SpeedUnit, here: https://github.com/victronenergy/venus-platform/blob/master/src/application.cpp#L192. 

I don't have a strong opinion for either option. A bit far fetched, but I can image that altitude might also be used to indicate on which altitude the solar panels, irradiance sensor or some wind generator has been installed. 

It is also a bit awkward that the gps'es don't appear under the device list, as I would have expected initially.

data/SystemSettings.qml
```diff
@@ -266,6 +286,30 @@ QtObject {
 		}
 	}
 
+	property VeQuickItem _altitudeUnit: VeQuickItem {
```
DanielMcInnes, line 289: Is this currently supported by venus-platform?

dirkjanfaber: That line 293 you mean? No, it isn't supported there yet. It could be that `/Settings/System/Gps/AltitudeUnit` is the better option. 

See https://github.com/victronenergy/gui-v2/pull/1841#issuecomment-2593792599.

DanielMcInnes: I expect we will wait for platform support before merging these changes

dirkjanfaber: Initiated [a change](https://github.com/victronenergy/venus-platform/compare/master...faberd/units-altitude) there. Lets indeed wait until that has been merged before continuing this PR.

philipptrenz:
Hi all, can we push this? With the upcoming boat UI also the GPS integration will be boosted

blammit:
@dirkjanfaber do you know if the relevant venus-platform changes have been merged yet?

dirkjanfaber:
> @dirkjanfaber do you know if the relevant venus-platform changes have been merged yet?

No, not yet. Perhaps @ReinvdZee can take a look at [this patch](https://github.com/victronenergy/venus-platform/commit/e8b6e684cf2f19eb646365110e909817355a5ded) and merge it.

ReinvdZee:
Changes to venus-platform are merged.

data/SystemSettings.qml
```diff
@@ -75,10 +76,29 @@ QtObject {
...
+		}
+	}
+
 	function convertFromCelsius(celsius_value) {
 		return Units.convert(celsius_value, VenusOS.Units_Temperature_Celsius, temperatureUnit)
 	}
 
+	function convertFromMeter(meter_value) {
+		return Units.convert(meter_value, VenusOS.Units_Altitude_Meter, altitudeUnit)
+	}
```
blammit, line 100: Looks like `setAltitudeUnit()` and `convertFromMeter()` are not used anymore?

blammit: @dirkjanfaber Could you remind me, what is a good installation for testing this?

Also could you rebase against the latest `main`?

blammit (approved):
Rebased against the latest `main`, as discussed on Slack. Tested on the VRM installation and LGTM.

PR #1842, opened 2025-01-15
Add AllowedItemModel and use it where possible

blammit:


blammit:
This needs some fixes. Closing for now.

PR #1845, opened 2025-01-15
Add VisibleItemModel

blammit:


src/visibleitemmodel.h
```diff
@@ -0,0 +1,73 @@
...
+			ListItem {} // this item will be loaded by the view
+		}
+	}
+*/
+class VisibleItemModel : public QQmlInstanceModel
+{
+	Q_OBJECT
+	Q_DECLARE_PRIVATE(VisibleItemModel)
+
+	Q_PROPERTY(QQmlListProperty<QObject> sourceModel READ sourceModel NOTIFY sourceModelChanged DESIGNABLE false)
```
MikeTrahearn-Qinetic, line 40: Should this property be FINAL?

src/visibleitemmodel.h
```diff
@@ -0,0 +1,73 @@
...
+
+	ListView {
+		// Make a VisibleItemModel with two source items
+		model: VisibleItemModel {
+			ListItem { preferredVisible: false } // this item will not be loaded by the view
+			ListItem {} // this item will be loaded by the view
+		}
+	}
+*/
+class VisibleItemModel : public QQmlInstanceModel
```
MikeTrahearn-Qinetic, line 35: Just checking - what's the general policy of using private APIs? e.g. when we bump from Qt 6.6.3 > Qt 6.8 what source compatibility are we relying on?

src/visibleitemmodel.cpp
```diff
@@ -0,0 +1,242 @@
...
+	}
+
+	void append(QObject *object) {
+		Q_Q(VisibleItemModel);
+		allObjects.append(object);
+
+		if (QQuickItem *item = qobject_cast<QQuickItem *>(object)) {
+			const QVariant preferredVisible = item->property("preferredVisible");
+			if (preferredVisible.isValid()) {
+				QObject::connect(item, SIGNAL(preferredVisibleChanged()), q, SLOT(preferredVisibleChanged()));
```
MikeTrahearn-Qinetic, line 51: _Me just realising why you can't use the new function pointer syntax for connect because item doesn't have the preferredVisible property._

src/visibleitemmodel.cpp
```diff
@@ -0,0 +1,242 @@
...
+	// Always return Referenced flag. Otherwise, when a view sees the item is no longer referenced,
+	// it will unparent the item.
+	return QQmlInstanceModel::Referenced;
+}
+
+QVariant VisibleItemModel::variantValue(int index, const QString &role)
+{
+	Q_D(VisibleItemModel);
+	if (index < 0 || index >= d->visibleItems.count()) {
+		return QString();
```
MikeTrahearn-Qinetic, line 214: Should this be return QVariant()? Or are you supposing the role being passed here assumes a QString type in delegate?

tests/visibleitemmodel/tst_visibleitemmodel.cpp
```diff
@@ -0,0 +1,16 @@
...
+** See LICENSE.txt for license information.
+*/
+
+#include <QtQuickTest/quicktest.h>
+#include <QtQml/QQmlEngine>
+#include "visibleitemmodel.h"
+
+int main(int argc, char **argv) \
+{
+    qmlRegisterType<Victron::VenusOS::VisibleItemModel>("Victron.VenusOS", 2, 0, "VisibleItemModel");
```
MikeTrahearn-Qinetic, line 12: I've done QML unit tests before without having to do qmlRegisterType for exposing C++ types to QML for the QML test harness. Especially as VisibleItemModel is declared as QML_ELEMENT. Is there a different reason for this? Perhaps the QML Test structure needs some assistance to help it out a bit (I haven't looked into that yet).

tests/visibleitemmodel/tst_visibleitemmodel.qml
```diff
@@ -0,0 +1,237 @@
...
+		visible: true
+
+		Rectangle {
+			anchors.fill: parent
+			color: "yellow"
+
+			ListView {
+				id: viewA
+				anchors.fill: parent
+				model: VisibleItemModel {
```
MikeTrahearn-Qinetic, line 38: I'm trying to remember where I saw this (Ulf) it isn't really recommended to assign object declarations to properties but to use an id. Would love to provide a bomb proof reason for this... I just can't remember what it was.

tests/visibleitemmodel/tst_visibleitemmodel.cpp
MikeTrahearn-Qinetic: Ah - it is because the test project doesn't bother linking to guiv2 and manually SOURCEs ../../src/visibleitemmodel.h / cpp.

src/visibleitemmodel.cpp
blammit: Fixed, should be returning QVariant instead.

src/visibleitemmodel.h
blammit: Generally we want to use the more established kind of private APIs, so that the instability is kept to a minimum, and only using them where absolutely necessary.

src/visibleitemmodel.h
blammit: Yes, added.

src/visibleitemmodel.h
DanielMcInnes: Is there a reason why we can't use a public qt base model?

src/visibleitemmodel.h
```diff
@@ -0,0 +1,73 @@
...
+{
+	Q_OBJECT
+	Q_DECLARE_PRIVATE(VisibleItemModel)
+
+	Q_PROPERTY(QQmlListProperty<QObject> sourceModel READ sourceModel NOTIFY sourceModelChanged DESIGNABLE false FINAL)
```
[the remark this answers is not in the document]
blammit: The `DESIGNABLE` probably isn't necessary, it came from a copy-paste.

blammit:
> Is there a reason why we can't use a public qt base model? eg. QAbstractItemModel

It needs to be a QQmlInstanceModel so that it can provide instances directly, unlike QAbstractItemModel. The QtQuick views have built-in behaviour to treat instance models differently; it can fetch delegate instances directly from the model, whereas when it uses a QAbstractItemModel, it will instantiate delegate instances instead.

src/visibleitemmodel.h
blammit: It needs to be a QQmlInstanceModel so that it can provide instances directly, unlike QAbstractItemModel. The QtQuick views have built-in behaviour to treat instance models differently; it can fetch delegate instances directly from the model, whereas when it uses a QAbstractItemModel, it will instantiate delegate instances instead.

src/visibleitemmodel.cpp
```diff
@@ -0,0 +1,242 @@
...
+	return d->visibleItems.at(index)->property(role.toUtf8().constData());
+}
+
+QQmlIncubator::Status VisibleItemModel::incubationStatus(int)
+{
+	// The model does not internally create objects, so any referenced objects are always available.
+	return QQmlIncubator::Ready;
+}
+
+int VisibleItemModel::indexOf(QObject *item, QObject *context) const
```
DanielMcInnes, line 225: unused parameter 'context'

DanielMcInnes (approved):
lgtm. Would be good if @chriadam could review as well

src/visibleitemmodel.cpp
```diff
@@ -0,0 +1,242 @@
...
+
+		if (QQuickItem *item = qobject_cast<QQuickItem *>(object)) {
+			const QVariant preferredVisible = item->property("preferredVisible");
+			if (preferredVisible.isValid()) {
+				QObject::connect(item, SIGNAL(preferredVisibleChanged()), q, SLOT(preferredVisibleChanged()));
+			}
+			// Add the item to the visible item list, if preferredVisible=true, or if it has no
+			// preferredVisible property and thus should not be filtered out.
+			preferredVisibleChanged(item, !preferredVisible.isValid() || preferredVisible.toBool());
+		}
```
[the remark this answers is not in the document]
MikeTrahearn-Qinetic: QQmlListProperty<QObject> should be QQmlListProperty<QQuickItem> (and all subsequent append/remove etc functions and that wouldn't be allowed full stop - unless that's a property type limitation of QmlInstanceModel.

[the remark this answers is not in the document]
blammit: I think it's fine to have a QObject instance in here - the instance can still exist in the source model, but just won't show up in the model that's presented to the view.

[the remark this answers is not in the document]
MikeTrahearn-Qinetic: What if a carefully constructed QObject with a preferredVisible property were added? Contrived, I know, but not enough to understand why QObjects would be allowed here.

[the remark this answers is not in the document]
blammit: It wouldn't be allowed into the model that is presented to the view; it would just be allowed into the source model. I thought this would be simpler because VisibleItemModel will replace all uses of `ObjectModel`, and I don't know if we really need to actively exclude QObjects from this, because there are probably cases where we have some helper QObjects in the model. 

Maybe `FilteredObjectModel` is a better term in some respects. It's really taking an object list, and filtering out everything that has preferredVisible=false from the resulting model, but it also filters out QObjects because it's assumed they would never be "visible". 

Maybe it shouldn't filter out QObjects?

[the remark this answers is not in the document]
MikeTrahearn-Qinetic: The confusion brews because ObjectModel has been used for both objects and items.
In the former case, it is nothing about drawables. In the latter it is. Perhaps the distinction be better made now and not just simply replace one with another. If ObjectModel only ever had QQuickItems in it, then my point remains.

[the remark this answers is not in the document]
blammit: Yeah that's true, maybe it's better to just force VisibleItemModel to only contain QQuickItems within the source model, then it is more clear.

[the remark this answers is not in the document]
blammit: Updated to only allow QQuickItems into the source model, by using `QQmlListProperty<QQuickItem>`.

src/visibleitemmodel.cpp
```diff
@@ -0,0 +1,239 @@
...
+
+QQmlListProperty<QQuickItem> VisibleItemModel::sourceModel()
+{
+	Q_D(VisibleItemModel);
+	return QQmlListProperty<QQuickItem>(this, d,
+									 VisibleItemModelPrivate::sourceModel_append,
+									 VisibleItemModelPrivate::sourceModel_count,
+									 VisibleItemModelPrivate::sourceModel_at,
+									 VisibleItemModelPrivate::sourceModel_clear,
+									 nullptr,
```
MikeTrahearn-Qinetic, line 171: Any reason why the replace function isn't provided?

blammit: Just for convenience really, since we don't have currently have any use cases that replace items in the model.

blammit: Yep, should be okay like this, and if we have use cases for it where we need a more efficient version, we can implement it then.

blammit:
Just seeing if @chriadam has any thoughts on this one before merging.

blammit:
Updated to make VisibleItemModel filter on `effectiveVisible` instead of `preferredVisible` as discussed with @MikeTrahearn-Qinetic, to ensure the filtering is not confused if there are conflicting values for `preferredVisible` vs `showAccessLevel`.

components/listitems/core/ListItem.qml
```diff
@@ -24,6 +24,7 @@ Item {
 	property int writeAccessLevel: VenusOS.User_AccessType_Installer
 	readonly property bool userHasWriteAccess: Global.systemSettings.canAccess(writeAccessLevel)
 	readonly property bool userHasReadAccess: Global.systemSettings.canAccess(showAccessLevel)
+	readonly property bool effectiveVisible: preferredVisible && userHasReadAccess
```
chriadam, line 27: Feel free to ignore this, but it might be a good idea to add a comment here saying something like: "we do NOT want to include `visible` in this check, or the filter model would filter things in and out of the view busily (since `visible` is cascaded from parents) resulting in unwanted delegate constructions etc".

It's true that this information is in the commit message, but a code comment here would increase the locality of this information.

src/visibleitemmodel.cpp
```diff
@@ -0,0 +1,239 @@
...
+
+	static void sourceModel_removeLast(QQmlListProperty<QQuickItem> *prop) {
+		static_cast<VisibleItemModelPrivate *>(prop->data)->removeLast();
+	}
+
+	void append(QQuickItem *item) {
+		Q_Q(VisibleItemModel);
+		allItems.append(item);
+
+		const QVariant effectiveVisible = item->property("effectiveVisible");
```
chriadam, line 48: do we need guards to ensure that item is not null (or, more correctly, to allow appending null item but don't dereference the pointer or attempt to connect any signals etc)

src/visibleitemmodel.cpp
```diff
@@ -0,0 +1,239 @@
...
+		// effectiveVisible property and thus should not be filtered out.
+		effectiveVisibleChanged(item, !effectiveVisible.isValid() || effectiveVisible.toBool());
+
+		emit q->sourceModelChanged();
+	}
+
+	void removeLast() {
+		Q_Q(VisibleItemModel);
+		if (allItems.count() > 0) {
+			QQuickItem *last = allItems.takeLast();
```
chriadam, line 62: would need also to add pointer guards here, if we allow adding null items.

src/visibleitemmodel.cpp
```diff
@@ -0,0 +1,239 @@
...
+			}
+			emit q->sourceModelChanged();
+		}
+	}
+
+	void clear() {
+		Q_Q(VisibleItemModel);
+		if (!allItems.isEmpty()) {
+			for (const QQuickItem *object : allItems) {
+				object->disconnect(q);
```
chriadam, line 80: pointer guards

src/visibleitemmodel.h
```diff
@@ -0,0 +1,76 @@
+/*
+** Copyright (C) 2025 Victron Energy B.V.
+** See LICENSE.txt for license information.
+*/
+
+#ifndef VISIBLEITEMMODEL_H
```
chriadam, line 6: the ship has probably sailed on this one, but I much prefer header guards with project prefixes, like "VICTRON_GUIV2_VISIBLEITEMMODEL_H" to avoid any possible clash of header guard defines with library/included headers.

src/visibleitemmodel.h
```diff
@@ -0,0 +1,76 @@
...
+class VisibleItemModelPrivate;
+
+namespace Victron {
+namespace VenusOS {
+
+/*
+	Provides an instance model that filters out non-visible items from the specified source model.
+
+	If any item in the source model has effectiveVisible=false, then it is filtered out of the
+	instance model, and thus will not be loaded in any view that uses the model.
```
chriadam, line 21: Maybe add a line saying: "Clients may bind the preferredVisible property appropriately to specify whether the item's visibility conditions are met (modulo access control requirements)" or similar, otherwise it's confusing that you talk about effectiveVisible here, but in the example below you set preferredVisible.

src/visibleitemmodel.cpp
```diff
@@ -0,0 +1,239 @@
...
+		static_cast<VisibleItemModelPrivate *>(prop->data)->removeLast();
+	}
+
+	void append(QQuickItem *item) {
+		Q_Q(VisibleItemModel);
+		allItems.append(item);
+
+		const QVariant effectiveVisible = item->property("effectiveVisible");
+		if (effectiveVisible.isValid()) {
+			QObject::connect(item, SIGNAL(effectiveVisibleChanged()), q, SLOT(effectiveVisibleChanged()));
```
chriadam, line 50: Why the string lookup connection, rather than pointer-to-member connection?  Also, I really don't like the trampoline existing in the public class, but I guess the alternative was to define yet another QObject-derived type and instantiate it for each item, etc, which is pretty ugly...  Connecting to a functor would be best, but those are not easily disconnected.  You could use a plain QObject scope/context object and just delete it after removing the item from the list, but then you again still need to instantate the QObject on append and delete it on remove/clear etc...

src/visibleitemmodel.cpp
```diff
@@ -0,0 +1,239 @@
...
+			if (prevVisibleIndex >= 0) {
+				// Insert the new item after that previous visible item.
+				return prevVisibleIndex + 1;
+			}
+		}
+		// There is no previous visible item, so this must be the first visible item.
+		return 0;
+	}
+
+	QVector<QQuickItem *> allItems;
```
chriadam, line 154: OOI what's the ownership of the items?  Does the model take ownership?  If not, should this be a QVector of QPointer to QQuickItem or similar, to guard against random deletion by potential clients?

src/visibleitemmodel.cpp
```diff
@@ -0,0 +1,239 @@
...
+	return QQmlInstanceModel::Referenced;
+}
+
+QVariant VisibleItemModel::variantValue(int index, const QString &role)
+{
+	Q_D(VisibleItemModel);
+	if (index < 0 || index >= d->visibleItems.count()) {
+		return QVariant();
+	}
+	return d->visibleItems.at(index)->property(role.toUtf8().constData());
```
chriadam, line 213: Originally I was going to say that you don't need to guard this, since no null item in allItems could ever be considered visible, so will never get added to visibleItems.  But then I thought: hmm, depending on the ownership semantics, if you end up making allItems and visibleItems vectors of QPointers, then I guess it's possible that the client will have deleted the item at the specified index, so a null pointer check might be required after all.

chriadam:
Mostly LGTM.  A few questions about ownership semantics / null pointer guards.

src/visibleitemmodel.cpp
blammit: It's because `effectiveVisible` is a QML property, so we don't have a reference to the C++ change signal. We could do it via QMetaObject and QMetaProperty, or via QQmlProperty::connectNotifySignal(), if it makes a difference?

src/visibleitemmodel.cpp
blammit: Hm, it would be unusual for clients to delete the objects, but it's certainly possible. Will update to use QPointer.

src/visibleitemmodel.cpp
blammit: Yes, sounds good to have a guard.

src/visibleitemmodel.h
blammit: Added a note about the preferredVisible condition.

src/visibleitemmodel.h
blammit: Added, probably worth updating the other headers too at some point

src/visibleitemmodel.cpp
blammit: Done

src/visibleitemmodel.cpp
blammit: Added

components/listitems/core/ListItem.qml
blammit: Done

src/visibleitemmodel.cpp
blammit: Will merge this but can make the change separately if needed.

PR #1846, opened 2025-01-15
Fix Maximum ESS Grid Feed-in could be set to -1

### LINKED ISSUE #1828: Settings -> ESS -> Grid feed-in goes to -1, 199, 299 etc Watts
While it should be 0, 100, 200, …

I was able to reproduce this on Venus OS v3.60~15 within a few minutes, on this site:

https://vrm.victronenergy.com/installation/473156/firmware-update

to reproduce it, press the down button till you see -1.
 
IMG_3840

reported here:
> I noticed the grid feed in limit goes to -1 instead of 0 then when you increase the limit again it goes in steps of 99 watts instead of 100 watts.

https://community.victronenergy.com/t/venus-os-v3-60-15-available-for-public-testing/13471

MikeTrahearn-Qinetic:
As per Gui-v1, the maxFeedInPower ListSpinBox 'from' property should be set to 0.

Fixes #1828

### DIFF
--- a/pages/settings/PageSettingsHub4Feedin.qml
+++ b/pages/settings/PageSettingsHub4Feedin.qml
@@ -62,6 +62,7 @@ Page {
 				preferredVisible: restrictFeedIn.visible && restrictFeedIn.checked
 				dataItem.uid: Global.systemSettings.serviceUid + "/Settings/CGwacs/MaxFeedInPower"
 				suffix: Units.defaultUnitString(VenusOS.Units_Watt)
+				from: 0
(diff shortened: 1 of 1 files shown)


blammit (approved):
How about "Set maximum ESS Grid Feed-in to -1" as a more concise commit title?

blammit:
> No because that's not what the fix is. We are not setting this to -1. We are preventing it from being set to -1.

Sorry, I meant to say, "Set maximum ESS Grid Feed-in to 0".

PR #1849, opened 2025-01-16
Fix position of "connected" icon in Wi-Fi networks list

blammit:
This also avoids the warning "Cannot anchor to an item that isn't a parent or sibling" when opening PageSettingsWifi.qml.

### DIFF
--- a/pages/settings/PageSettingsWifi.qml
+++ b/pages/settings/PageSettingsWifi.qml
@@ -84,9 +84,10 @@ Page {
 			primaryLabel.leftPadding: Theme.geometry_icon_size_medium + Theme.geometry_listItem_content_spacing
 
 			CP.ColorImage {
+				parent: delagate.primaryLabel
 				anchors {
-					left: delagate.primaryLabel.left
-					verticalCenter: delagate.primaryLabel.verticalCenter
+					left: parent.left
+					verticalCenter: parent.verticalCenter
 				}
 				source: "qrc:/images/icon_checkmark_32.svg"
 				color: Theme.color_green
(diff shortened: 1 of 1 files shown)


MikeTrahearn-Qinetic (approved):

pages/settings/PageSettingsWifi.qml
```diff
@@ -84,9 +84,10 @@ Page {
 			primaryLabel.leftPadding: Theme.geometry_icon_size_medium + Theme.geometry_listItem_content_spacing
 
 			CP.ColorImage {
+				parent: delagate.primaryLabel
```
DanielMcInnes, line 87: Is `delagate` intentionally misspelled to avoid clashing with the `delegate` keyword? I just tried fixing the spelling, and the result works fine on a site with favorite wifi connections such as 151899

DanielMcInnes (approved):
1 nit

pages/settings/PageSettingsWifi.qml
MikeTrahearn-Qinetic: You can indeed use the id spelled the same as a property but my preference would be to give it a meaningful name instead of creating visual ambiguity (even though the compiler will favour using the id. In fact in stricter/newer versions of Qt to reference only the property delegate will create an unqualified reference error because the "deleg**a**te" is in fact a Component, not an object and to force it to do so would require adding pragma ComponentBehavior: Bound - something not available in Qt 6.6.

blammit: I think the "delagate" was a misspelling when the change was first added, and just "delegate" would be fine. But yes, a more meaningful name would be better too.

blammit: On second thoughts, it's probably better to clean it up now, and the id is only used in a few places. Updated to rename it to `accessPointDelegate`.

PR #1851, opened 2025-01-16
Change the number of decimals used to display tank sensor values when…

### LINKED ISSUE #1836: Redundant decimal places shown in tank sensor settings
See screenshots below, as per slack thread: https://victrondevelopment.slack.com/archives/C04RYU0UMDE/p1736602047609179

In the last production release of gui-v2, the tank settings page looks like this, with only 1 decimal places for `cm`:

image

But in the latest beta there are redundant zeros:

image

DanielMcInnes:
… empty / full

Fixes #1836

### DIFF
--- a/pages/settings/devicelist/tank/PageTankSetup.qml
+++ b/pages/settings/devicelist/tank/PageTankSetup.qml
@@ -61,8 +61,8 @@ Page {
 				allowed: dataItem.seen && (!standard.dataItem.isValid || standard.currentValue === 2)
 				dataItem.uid: root.bindPrefix + "/RawValueEmpty"
 				suffix: rawUnit.value || ""
-				decimals: 3
-				stepSize: 0.005
+				decimals: rawUnit.displayDecimals
+				stepSize: rawUnit.stepSize
 			}
 
 			ListSpinBox {
@@ -71,8 +71,8 @@ Page {
 				allowed: dataItem.seen && (!standard.dataItem.isValid || standard.currentValue === 2)
 				dataItem.uid: root.bindPrefix + "/RawValueFull"
 				suffix: rawUnit.value || ""
-				decimals: 3
-				stepSize: 0.005
+				decimals: rawUnit.displayDecimals
+				stepSize: rawUnit.stepSize
 			}
 
 			ListRadioButtonGroup {
@@ -160,6 +160,27 @@ Page {
 
 	VeQuickItem {
 		id: rawUnit
+
+		// The possible values here are not well defined. The doco says: "can be V, and probably also mA and R or O."
+		// At least one installation uses "cm".
+		readonly property int displayDecimals: {
+			switch (value) {
+			case "cm":
+				return 1
+			default:
+				return 3
+			}
+		}
+
+		readonly property real stepSize: {
+			switch (value) {
+			case "cm":
+				return 0.1
+			default:
+				return 0.005
+			}
+		}
+
 		uid: root.bindPrefix + "/RawUnit"
 	}
 }


pages/settings/devicelist/tank/PageTankSetup.qml
```diff
@@ -61,7 +61,7 @@ Page {
 				allowed: dataItem.seen && (!standard.dataItem.isValid || standard.currentValue === 2)
 				dataItem.uid: root.bindPrefix + "/RawValueEmpty"
 				suffix: rawUnit.value || ""
-				decimals: 3
+				decimals: rawUnit.decimals
```
blammit, line 64: It seems odd to have decimals=1 but stepSize=0.005 in the cm case.

pages/settings/devicelist/tank/PageTankSetup.qml
```diff
@@ -160,6 +160,18 @@ Page {
 
 	VeQuickItem {
 		id: rawUnit
+
+		// The possible values here are not well defined. The doco says: "can be V, and probably also mA and R or O."
+		// At least one installation uses "cm".
+		property int decimals: {
+			switch (value) {
+			case "cm":
+				return 1
```
blammit, line 169: I was going to suggest adding `cm` as a unit to `units.h` and providing some default precision for it, but I suppose the 3 decimals for the volts here is already non-standard, so ... maybe it's okay like this even if it's not ideal.

MikeTrahearn-Qinetic: So long as this is an isolated case this _might_ be ok.
However a more consistent approach would be nice if it were possible to define the types of units available for display - and define a) a default decimals for that unit and b) override the chosen unit's decimal on a per-instance basis. I had to deal with this exact problem in a former life so I know how hard b) in particular is to solve.

chriadam: Maybe I'm misunderstanding something, but we do support that already, don't we?  e.g. https://github.com/victronenergy/gui-v2/blob/6040259f4186c4685cf0be91569db7ce2d8d843e/pages/settings/devicelist/PageMeteo.qml#L50 etc.

DanielMcInnes: I asked about this on slack, we'll wait and see what comes back.

blammit: @MikeTrahearn-Qinetic we do have a list of units available for display, and they have default precisions, as per units.h/cpp. This case is a bit different as this is a "raw unit string" so in theory it could be anything.

DanielMcInnes: MV says 1 decimal is fine for cm, and then leave the rest at 3

pages/settings/devicelist/tank/PageTankSetup.qml
DanielMcInnes: Changes step size for the cm case to 0.1

pages/settings/devicelist/tank/PageTankSetup.qml
DanielMcInnes: This page doesn't use default precision, because it's an odd case where we want higher precision for tuning tank sensors. MV says: "You can change it to be 1 decimal for centimeters. And then we leave this. Its not important: sensor value is a debug value. The value that matters is the tank level in liters or m3"

pages/settings/devicelist/tank/PageTankSetup.qml
```diff
@@ -160,6 +160,27 @@ Page {
...
+		property int displayDecimals: {
+			switch (value) {
+			case "cm":
+				return 1
+			default:
+				return 3
+			}
+		}
+
+		property real stepSize: {
```
blammit, line 175: Not sure if a `switch` statement is worthwhile if there's only one case for it, but no strong preference either way.

Also, both properties can be `readonly`.

Otherwise LGTM.

blammit (approved):

pages/settings/devicelist/tank/PageTankSetup.qml
DanielMcInnes: thanks Bea, changed both properties to readonly

PR #1852, opened 2025-01-16
Use shorter title for history page

blammit:
This avoids redundancy in the breadcrumbs where the device name is repeated.


blammit:
This change avoids this kind of situation in the breadcrumbs: 

image

MikeTrahearn-Qinetic (approved):

PR #1801, opened 2024-12-13
Jeroen/move wasm index to cmake

jhofstee:
Adjusting / using the customized index.html is cumbersome, since it is a patch as part of the ci. This just adds the index file to the repository and the CMake project instead, so it can be simply adjusted and will be used when running from e.g. QtCreator.

PR #1860, opened 2025-01-17
Add build scripts and optimized GitHub Actions

mr-manuel:
Some script were added to make it much easier for everyone to locally build GUIv2 for a GX device and as WebAssembly. Currently it works only on X86 processors, due to currently used Emscripten (not for all versions ARM64 binaries are available) and Qt versions (only since 6.7.0 it appears that binaries are available to the community, else a license is required).

GitHub Actions were also modified to execute the new scripts, which makes maintenance less time consuming. For example the emscripten and QT version are now always fetched from an environment file and do not need to be updated in multiple scripts, GitHub Actions and wiki:

https://github.com/victronenergy/gui-v2/blob/a4b7bc19c36cd07944a573e0f30a73d5151bfbd5/scripts/.env#L1-L3

I tested the scripts on:
- Ubuntu 24.04 VM
- Windows 11 with a fresh Ubuntu 24.10 [WSL2 container](https://learn.microsoft.com/en-us/windows/wsl/install)
- GitHub Actions
- macOS 15 (Intel CPU) with [Lima](https://github.com/lima-vm/lima#getting-started)

After this PR is merged, also the [wiki](https://github.com/victronenergy/gui-v2/wiki/How-to-build-venus-gui-v2) has to be modified and simplified.

I added also a `.gitattributes` file to fix automatically the line endings of `.env` and `*.sh` files, else they were always converted to `CRLF` instead of `LF`.


DanielMcInnes:
Building the wasm version fails for old (20.04) ubuntu installations. Not an issue, but should probably be mentioned somewhere. I didn't capture a log, but it was to do with aqt needing a newer version of python than what you get with that old ubuntu. 

Building the wasm version works fine on a fresh ubuntu 24.04 vm.

`build-gx-install-requirements.sh` fails on a fresh ubuntu 24.04 vm:

```dmcinnes@dmcinnes-System-Product-Name:~/git/victron/gui-v2$ ./build-gx-install-requirements.sh
Downloading https://updates.victronenergy.com/feeds/venus/release/sdk/venus-dunfell-x86_64-arm-cortexa8hf-neon-toolchain-v3.53.sh
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  231M  100  231M    0     0  4487k      0  0:00:52  0:00:52 --:--:-- 4538k
sudo: venus-dunfell-x86_64-arm-cortexa8hf-neon-toolchain-v3.53.sh: command not found
ERROR: SDK installation failed.
dmcinnes@dmcinnes-System-Product-Name:~/git/victron/gui-v2$ ```

**Comment by DanielMcInnes** [comment]
If I manually install the toolchain:
```
dmcinnes@dmcinnes-System-Product-Name:/tmp$ sudo ./venus-dunfell-x86_64-arm-cortexa8hf-neon-toolchain-v3.53.sh -y
Victron Energy SDK installer version v3.53
==========================================
The directory "/opt/venus/dunfell-arm-cortexa8hf-neon" already contains a SDK for this architecture.
If you continue, existing files will be overwritten! Proceed [y/N]? Y
Extracting SDK.....................................................................done
Setting it up...done
SDK has been successfully set up and is ready to be used.
Each time you wish to use the SDK in a new shell session, you need to source the environment setup script e.g.
 $ . /opt/venus/dunfell-arm-cortexa8hf-neon/environment-setup-cortexa8hf-neon-ve-linux-gnueabi
dmcinnes@dmcinnes-System-Product-Name:/tmp$ cd
dmcinnes@dmcinnes-System-Product-Name:~$ cd git/victron/gui-v2/
dmcinnes@dmcinnes-System-Product-Name:~/git/victron/gui-v2$ ls
ApplicationContent.qml  build-gx-install-requirements.sh  build-gx.sh  build-wasm-install-requirements.sh  build-wasm.sh  cmake  CMakeLists.txt  CMakeLists.txt.user  components  data  fonts  FrameRateVisualizer.qml  gitdiff  Global.qml  i18n  images  LICENSE.txt  macdeployhelper.sh  Main.qml  nohup.out  pages  qqmlobjectmodel.cpp  README.md  scripts  src  test.py  tests  themes  tools  victron.astyle  windeployhelper.bat
```
... then source the environment setup stuff:
`dmcinnes@dmcinnes-System-Product-Name:~/git/victron/gui-v2$ . /opt/venus/dunfell-arm-cortexa8hf-neon/environment-setup-cortexa8hf-neon-ve-linux-gnueabi`
... then start the gx build:
```
dmcinnes@dmcinnes-System-Product-Name:~/git/victron/gui-v2$ ./build-gx.sh
ERROR: Venus OS SDK was not found.
Execute "./build-gx-install-requirements.sh" once or visit this link for how to install and use the SDK: https://github.com/victronenergy/venus/wiki/howto-install-and-use-the-sdk
dmcinnes@dmcinnes-System-Product-Name:~/git/victron/gui-v2$ . /opt/venus/dunfell-arm-cortexa8hf-neon/environment-setup-cortexa8hf-neon-ve-linux-gnueabi 
dmcinnes@dmcinnes-System-Product-Name:~/git/victron/gui-v2$ ```
... I get this error

DanielMcInnes:
I only have a linux box, I can't test this for mac or windows. Would you expect it to work on these platforms?

blammit:
I understand these scripts make it easier to build, in terms of combining the package installation + configure + build steps, but should these scripts really be in the repo? It looks like they have various specific requirements/assumptions (e.g. Linux builds, doing clean builds, putting build dir inside the repo) so that it may often be easier to build without the scripts. Or, in other cases it is simple enough that you don't need a separate build script (e.g. build-gx.sh). I'm not sure how much benefit we get from adding these scripts, as opposed to having some improved build documentation, and this also creates additional tools that need to be maintained going forward.

DanielMcInnes:
I like the added convenience this brings to linux users, especially building and installing mqtt. I imagine similar scripts would be required for windows & mac users, resulting in 3 sets of build scripts that need to be maintained, which is not ideal. I think this would be better done via cmake, which would give you the same build behavior on all 3 platforms. See `ExternalProject_Add` at https://cmake.org/cmake/help/latest/module/ExternalProject.html . The downside of doing anything in cmake is the difficulty - it is often hugely frustrating.

blammit:
@mr-manuel That's true it would be nice to build with Qt Creator and downloading the SDK yourself, and much more convenient for community members. I think my main objection is that it looks a bit ad hoc in this current form. Unifying the builds under cmake as @DanielMcInnes suggested sounds like a good approach. What about using cmake presets? We could have presets like `linux`, `wasm`, `windows`, `mac`, and have standardized configure+build commands like:

```
cmake --preset wasm  # download/configure for wasm
cmake --build --preset wasm  # run the build
```

mpvader:
@blammit you wrote:

> it would be nice to build with Qt Creator and downloading the SDK yourself, and much more convenient for community members.

Sorry but I don’t agree. Its the opposite. For people to try this out they need to be able to try this very simply without installing and setting up many things. Ie. change a qml file and run the build.

As long as the script works for our github actions + linux + linux on windows then I’m happy. And mac os is an extra nice to have.

By the way: all the rest of Venus OS is also oriented towards linux only; probably builds on Windows using its linux extensions, and might/or might not run on Mac OS; which is fine.

mpvader:
Wrt cmake: doesn’t that make it all a lot less readable? Lets keep it simple.

blammit:
> Sorry but I don’t agree. Its the opposite. For people to try this out they need to be able to try this very simply without installing and setting up many things. Ie. change a qml file and run the build.

My mistake, I meant to say, "it would be nicer than building with Qt Creator...". I agree it's certainly easier for people to try it out if they don't need to download the tools and Qt Creator, etc. I'm just not sure adding build scripts to the repo is the way to go here. These would need to be regularly verified/maintained/supported, and if they don't work for different users' configurations then they will need to do their own hacking anyway. In that sense they seem more like templates for building. But maybe that will be obvious to community members and not really a problem there.

> Wrt cmake: doesn’t that make it all a lot less readable? Lets keep it simple.

Yes it's definitely more work to use cmake, and not as simple as adding build scripts.

If we really need build scripts inside the repo then I'm happy to go ahead with it. @mr-manuel there are already scripts under `scripts` and `tools`.  Maybe these can go under `scripts/build`? (Then perhaps eventually the deployment scripts could be moved from the root directory as well.)

mr-manuel:
I moved the scripts, added some notes and prepared them for supporting installation of the requirements on ARM64 architecture (like Mac's with M chip).

@dirkjanfaber as we wrote, I moved the POEditor scripts from `scripts` to `.github/scripts`.

mr-manuel:
@blammit to make sure all scripts work correctly I can also add a GitHub Action that tests the scripts once they change or the `.env` changes. This way they will always work. This is also possible to do for ARM64 later, so all build possibilities are tested automatically.

blammit:
> @blammit to make sure all scripts work correctly I can also add a GitHub Action that tests the scripts once they change or the .env changes. This way they will always work. This is also possible to do for ARM64 later, so all build possibilities are tested automatically.

That sounds good, to make sure they are verified regularly.

> Maybe these can go under scripts/build

Initially I was thinking this would be better as it would tidy up the root directory, but it does mean the scripts are harder to find, if we want to encourage community members to use them for building. I'm not entirely sure either way. 

Note, the build instructions in the README should also be updated.

scripts/build-gx.sh
```diff
@@ -0,0 +1,82 @@
...
+# Go to the parent directory of the script
+BASE_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )/.." &> /dev/null && pwd )"
+cd "${BASE_DIR}"
+echo "Changed to parent directory: $(pwd)"
+
+# Source the SDK environment
+. /opt/venus/current/environment-setup-cortexa8hf-neon-ve-linux-gnueabi
+
+# Checkout the branch you want to build, if not already on it
+# git checkout -b main origin/main
```
chriadam, line 34: commented out?

scripts/build-gx.sh
```diff
@@ -0,0 +1,82 @@
...
+fi
+
+# Create build directory
+mkdir "build-gx"
+
+cd "build-gx"
+
+
+# Configure the project with CMake, setting the build type to MinSizeRel (minimum size release)
+cmake -DCMAKE_BUILD_TYPE=MinSizeRaptel ..
```
chriadam, line 58: typo?  should it be MinSizeRel ?

scripts/build-wasm.sh
```diff
@@ -0,0 +1,97 @@
...
+if [ -d "artifacts" ]; then
+    rm -rf "artifacts"
+fi
+
+# Prepare files that are needed for the GUIv2 to run
+mkdir -p artifacts/wasm
+mv venus-gui-v2.{html,js,wasm} qtloader.js artifacts/wasm/
+cp ../images/victronenergy.svg ../LICENSE.txt ../.github/patches/Makefile artifacts/wasm/
+mv artifacts/wasm/venus-gui-v2.html artifacts/wasm/index.html
+patch artifacts/wasm/index.html < ../.github/patches/index.html.patch
```
chriadam, line 80: This is going to change, or go away altogether, very soon.  @DanielMcInnes once you have rebased and merged Jeroen's change, can you advise @mr-manuel what changes are needed here?

chriadam:
Thanks!  Looks great, for the most part.  I think we still need the "manual" instructions in the wiki, to ensure that we support folks who don't use Ubuntu 24.04 or higher, and also because I think there is still value in desktop (Windows / Mac OS X) builds for fast-iteration testing.

mr-manuel:
After this PR I would also add a short description on how to launch the build scripts on Windows and macOS. In your case you mean on how to build natively to start the GUI on Windows or macOS?

scripts/build-gx.sh
mr-manuel: Yes, since if you run the script, you have already done that. I left it if someone want's to change/use it, but I can remove it.

scripts/build-wasm.sh
mr-manuel: For reference: https://github.com/victronenergy/gui-v2/pull/1801

DanielMcInnes (approved):
lgtm

mr-manuel:
Rebase was done to include the changes to the `index.html`. You can now merge.

PR #1861, opened 2025-01-19
Add option to override mqtt param for wasm

martinbosma:
Now you can do something like `http://venus.local/gui-v2?mqtt=ws://123.123.1.2/websocket-mqtt`

### DIFF
--- a/.github/patches/index.html.patch
+++ b/.github/patches/index.html.patch
@@ -54,15 +54,23 @@
          async function init()
          {
              const spinner = document.querySelector('#qtspinner');
-@@ -46,6 +74,7 @@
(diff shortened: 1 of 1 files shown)


blammit:
@chriadam could you check this one please?

chriadam:
LGTM, but we should get Jeroen's PR in first, and then make this change directly to the index.html rather than modifying the patch file.

DanielMcInnes:
@martinbosma Jeroen's PR has been merged. These changes can be re-applied to `wasm/index.html`

martinbosma:
Closed through https://github.com/victronenergy/gui-v2/pull/1937

PR #1864, opened 2025-01-20
Prepare to add more generic solar devices

blammit:


pages/solar/PageSolarCharger.qml
```diff
@@ -0,0 +1,344 @@
...
+	VeQuickItem {
+		id: shortCircuitAlarm
+		uid: root.bindPrefix + "/Alarms/ShortCircuit"
+	}
+
+	GradientListView {
+		model: _isModelSupported() ? supportedProductModel : unsupportedProductModel
+	}
+
+	ObjectModel {
```
MikeTrahearn-Qinetic, line 82: Checking - is this a real ObjectModel or a VisibleItemModel?

pages/solar/PageSolarCharger.qml
```diff
@@ -0,0 +1,344 @@
...
+				if (!pvVoltage.value || !pvTotalPower.isValid) {
+					return NaN
+				}
+				return pvTotalPower.value / pvVoltage.value
+			}
+
+			//: PV power for solar charger
+			//% "PV"
+			text: qsTrId("solarcharger_pv")
+			preferredVisible: root.trackerCount < 2
```
MikeTrahearn-Qinetic, line 130: preferredVisible used within an ObjectModel - see previous comment - my guess is this code isn't on top of the other change yet.

pages/solar/PageSolarCharger.qml
```diff
@@ -0,0 +1,344 @@
...
+				Global.pageManager.pushPage("/pages/settings/PageDeviceInfo.qml",
+						{ "title": text, "bindPrefix": root.bindPrefix })
+			}
+		}
+	}
+
+	Component {
+		id: alarmStatusComponent
+
+		Page {
```
MikeTrahearn-Qinetic, line 324: Assume Page is used here for some kind of background and/or padding?

pages/solar/PageSolarParallelOperation.qml
```diff
@@ -1,5 +1,5 @@
 /*
-** Copyright (C) 2023 Victron Energy B.V.
+** Copyright (C) 2025 Victron Energy B.V.
```
MikeTrahearn-Qinetic, line 2: Did this change en-masse or just here?

pages/solar/PageSolarCharger.qml
blammit: Yes, I'll update this to use VisibleItemModel, when the other PR is merged.

pages/solar/PageSolarCharger.qml
blammit: This is okay, preferredVisible comes from ListItem, which is usable without VisibleItemModel.

pages/solar/PageSolarCharger.qml
blammit: It is a new page, that is opened when you click on "Alarm status".

pages/solar/PageSolarParallelOperation.qml
MikeTrahearn-Qinetic: Just wondering if we should do one honking great update in another commit. Maybe we could use some kind of git replace magic here.

blammit: So far we've just been making this "year of file created" rather than when it was modified, but perhaps it should be done on modification instead. If so it would definitely be nicer via git magic.

pages/solar/PageSolarCharger.qml
```diff
@@ -0,0 +1,344 @@
...
+Page {
+	id: root
+
+	required property string bindPrefix
+	readonly property int trackerCount: nrOfTrackers.isValid ? nrOfTrackers.value : 1
+	readonly property SolarCharger solarCharger: Global.solarChargers.model.deviceAt(Global.solarChargers.model.indexOf(bindPrefix))
+
+	function _isModelSupported() {
+		if (!productId.isValid || !firmwareVersion.isValid) {
+			return true
```
DanielMcInnes, line 18: is this meant to return false?

pages/solar/PageSolarCharger.qml
```diff
@@ -0,0 +1,344 @@
...
+
+	title: device.name
+
+	Device {
+		id: device
+		serviceUid: root.bindPrefix
+	}
+
+	VeQuickItem {
+		id: productId
```
DanielMcInnes, line 44: I think you use `device.productId` instead of adding another VeQuickItem here

pages/solar/PageSolarCharger.qml
```diff
@@ -0,0 +1,344 @@
...
+
+			//: PV power for solar charger
+			//% "PV"
+			text: qsTrId("solarcharger_pv")
+			preferredVisible: root.trackerCount < 2
+
+			// PV voltage and current are not visible in parallel mode
+			textModel: [
+				{ value: pvVoltage.value, unit: VenusOS.Units_Volt_DC, visible: pvVoltage.isValid },
+				{ value: pvCurrent(), unit: VenusOS.Units_Amp, visible: pvVoltage.isValid },
```
DanielMcInnes, line 135: maybe: `visible: pvCurrent() != NaN`, to account for  `!pvTotalPower.isValid` as well?

pages/solar/PageSolarCharger.qml
```diff
@@ -0,0 +1,344 @@
...
+		ListQuantity {
+			id: loadQuantityItem
+
+			//% "Load"
+			text: qsTrId("solarcharger_load")
+			dataItem.uid: root.bindPrefix + "/Load/I"
+			unit: VenusOS.Units_Amp
+			preferredVisible: dataItem.isValid && loadState.dataItem.value === 1
+		}
+		ListText {
```
DanielMcInnes, line 241: missing 1 blank line?

pages/solar/PageSolarCharger.qml
blammit: It returns true in gui-v1; I think the intention is, if you can't tell whether it's supported, then assume it is supported.

pages/solar/PageSolarCharger.qml
blammit: Sounds good, updated

pages/solar/PageSolarCharger.qml
blammit: Fixed

pages/solar/PageSolarCharger.qml
blammit: Ah yes, forgot about that, fixed

DanielMcInnes: fair enough, lgtm

DanielMcInnes (approved):

Issue #1232, opened 2024-06-13
Redesigned top-level structure for the settings


jpetrell:
GX devices have accumulated a lot of settings over time, the structure hasn’t kept up with all the changes. Finding what you are looking for can be difficult. Re-organize top-level settings in logical groups for easier access, e.g. Wi-Fi, Ethernet, GSM and Bluetooth under Connectivity. See the Figma design for more details.

DanielMcInnes:
https://www.figma.com/design/j9qNQjp1rHQSbJZRy0sxvk/GX-Touch---Implementation?node-id=6709-103273&p=f&t=YCK5nk5ZRgtQkTzi-0

DanielMcInnes:
Questions for @philipptrenz :
In the new design, is `System Setup -> Charge Control` the same as the old `Settings -> DVCC` ?

philipptrenz:
@DanielMcInnes it relates to
`System Setup` > `Has DC system`

DanielMcInnes:
Hi @philipptrenz , currently in gui-v2 we have a 'Services' page in Settings. It contains the 'tailscale' page, CANbus settings, and a switch for CANbus over tcp/ip. I assume that should go in the new 'Connectivity' page?

philipptrenz:
Hi @DanielMcInnes,

yes, I'd say Tailscale is best placed within "Connectivity". Also the CAN ports are meant to be placed there, see hierarchy tree in Figma. Regarding the "CANbus over tcp/ip" toggle, I guess this is a per CAN port setting and can be placed under the related CAN interface?

DanielMcInnes:
> Hi @DanielMcInnes,
> 
> yes, I'd say Tailscale is best placed within "Connectivity". Also the CAN ports are meant to be placed there, see hierarchy tree in Figma. Regarding the "CANbus over tcp/ip" toggle, I guess this is a per CAN port setting and can be placed under the related CAN interface?

Thanks @philipptrenz , sounds good

DanielMcInnes:
Hi @philipptrenz , what is meant to go in the 'Display Units' page? it seems like a duplicate of 'Data Units'. Also, in 'Data Units', the new design has 'speed' and 'gps format'. I think we support multiple gps receivers, and allow configuring speed & gps format for each gps receiver, so there's not a single system-wide setting for speed & gps format, you have to navigate to a specific gps receiver and configure it there.

mr-manuel:
[...]

PR #1867, opened 2025-01-20
Move "Display units" to display settings and move GPS units there

blammit:
Part of #1232


blammit:
Moved as per slack discussion https://victrondevelopment.slack.com/archives/C06EVSJDR3K/p1737331377169859

DanielMcInnes (approved):

PR #1871, opened 2025-01-22
Replace Global.solarChargers with Global.solarDevices

blammit:
Use Global.solarDevices instead, as that contains not only solarcharger devices, but also any multi and inverter devices that have solar trackers.


DanielMcInnes (approved):

pages/solar/SolarDevicePage.qml
```diff
@@ -26,12 +30,12 @@ Page {
...
 						{
 							title: CommonWords.state,
-							text: VenusOS.solarCharger_stateToText(root.solarCharger.state),
+							text: VenusOS.solarCharger_stateToText(stateItem.value),
 							unit: VenusOS.Units_None,
 						},
 						{
 							title: CommonWords.yield_today,
-							value: solarCharger.dailyHistory(0)?.yieldKwh ?? NaN,
+							value: solarDevice.dailyHistory(0)?.yieldKwh ?? NaN,
```
MikeTrahearn-Qinetic, line 38: consider the qualification of root on solarDevice - sometimes it is used, sometimes not. If the id is available within this scope and isn't being referenced from within component, then it can be omitted as it is unambiguous for the compiler, unless the code makes that otherwise.

data/mock/SystemImpl.qml
```diff
@@ -167,9 +167,9 @@ QtObject {
 		}
 
 		let dcPower = NaN
-		if (solarChargers.model.count) {
-			for (i = 0; i < solarChargers.model.count; ++i) {
-				const charger = solarChargers.model.deviceAt(i)
+		if (solarDevices.model.count) {
+			for (i = 0; i < solarDevices.model.count; ++i) {
```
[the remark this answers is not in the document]
blammit: Updated, there was no need for the separate declaration as the scopes of the two different loops do not clash here.

pages/solar/SolarDevicePage.qml
blammit: Fixed to add the qualification; I'd forgotten to add it here, generally I prefer to use qualified access unless the access is within the same object.

MikeTrahearn-Qinetic: 'tis a good practice!

PR #1879, opened 2025-01-27
Add caption property to ListItem

### LINKED ISSUE #1822: Add caption property to ListItem
There are various list items with caption text, which are implemented by adding custom Labels to the `bottomContent`, like this:

```
bottomContentChildren: [
    PrimaryListLabel {
        width: Math.min(implicitWidth, listItem.maximumContentWidth)
        topPadding: 0
        bottomPadding: 0
        text: "Custom text label"
    }
]
```

Remove these and add a configurable caption to the base ListItem instead.

DanielMcInnes:
Fixes #1822

### DIFF
--- a/components/PageGensetModel.qml
+++ b/components/PageGensetModel.qml
@@ -361,11 +361,8 @@ VisibleItemModel {
 							secondaryText: CommonWords.yesOrNo(dataItem.value)
 							dataItem.uid: root.bindPrefix + "/Settings/BmsPresent"
 							preferredVisible: dataItem.isValid
-							bottomContentChildren: PrimaryListLabel {
-								//% "BMS control is enabled automatically when a BMS is present. Reset it if the system configuration changed or if there is no BMS present."
-								text: qsTrId("genset_bms_control_enabled_automatically")
-								preferredVisible: bmsControlled.dataItem.value === 1
-							}
+							//% "BMS control is enabled automatically when a BMS is present. Reset it if the system configuration changed or if there is no BMS present."
+							caption: bmsControlled.dataItem.value === 1 ? qsTrId("genset_bms_control_enabled_automatically") : ""
 						}
 
 						ListButton {
--- a/components/listitems/ListLink.qml
+++ b/components/listitems/ListLink.qml
@@ -11,10 +11,6 @@ ListItem {
 	id: root
 
 	property string url
-	property string caption: root.mode === VenusOS.ListLink_Mode_LinkButton ? ""
-		  //: %1 = url text
-		  //% "Scan the QR code with your portable device.<br />Or insert the link: %1"
-		: qsTrId("listlink_scan_qr_code").arg(formattedUrl)
 	readonly property string formattedUrl: "<font color=\"%1\">%2</font>".arg(Theme.color_font_primary).arg(url)
 
 	readonly property int mode: Qt.platform.os == "wasm" ? VenusOS.ListLink_Mode_LinkButton : VenusOS.ListLink_Mode_QRCode
@@ -56,15 +52,10 @@ ListItem {
 		}
 	]
 
-	bottomContentChildren: [
-		PrimaryListLabel {
-			font.pixelSize: Theme.font_size_body1
-			horizontalAlignment: Text.AlignLeft
-			color: Theme.color_listItem_secondaryText
-			text: root.caption
-			visible: root.mode === VenusOS.ListLink_Mode_QRCode
-		}
-	]
+	caption: root.mode === VenusOS.ListLink_Mode_LinkButton ? ""
+		  //: %1 = url text
+		  //% "Scan the QR code with your portable device.<br />Or insert the link: %1"
+		: qsTrId("listlink_scan_qr_code").arg(formattedUrl)
 
 	ListPressArea {
 		id: pressArea
--- a/components/listitems/core/ListItem.qml
+++ b/components/listitems/core/ListItem.qml
@@ -14,6 +14,7 @@ Item {
 	property alias content: content
 	property alias bottomContent: bottomContent
 	property alias bottomContentChildren: bottomContent.children
+	property string caption
 	property bool down
 	property bool flat
 	property alias backgroundRect: backgroundRect
@@ -112,6 +113,18 @@ Item {
 			Layout.columnSpan: root.bottomContentSizeMode === VenusOS.ListItem_BottomContentSizeMode_Stretch ? 2 : 1
 			Layout.topMargin: height > 0 ? Theme.geometry_listItem_content_verticalMargin / 2 : 0
 			Layout.bottomMargin: Theme.geometry_listItem_content_verticalMargin
+
+			Label {
+				width: parent.width
+				visible: text !== ""
+				topPadding: 0
+				bottomPadding: 0
+				leftPadding: Theme.geometry_listItem_content_horizontalMargin
+				rightPadding: Theme.geometry_listItem_content_horizontalMargin
+				wrapMode: Text.Wrap
+				color: Theme.color_font_secondary
+				text: root.caption
+			}
 		}
 	}
 }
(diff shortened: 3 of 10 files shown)


components/listitems/ListLink.qml
```diff
@@ -56,15 +52,13 @@ ListItem {
...
-	bottomContentChildren: [
-		PrimaryListLabel {
-			font.pixelSize: Theme.font_size_body1
-			horizontalAlignment: Text.AlignLeft
-			color: Theme.color_listItem_secondaryText
-			text: root.caption
-			visible: root.mode === VenusOS.ListLink_Mode_QRCode
-		}
-	]
+	caption.horizontalAlignment: Text.AlignLeft
```
blammit, line 55: The Text.AlignLeft is redundant?

components/listitems/core/ListItem.qml
```diff
@@ -104,6 +105,13 @@ Item {
 			Layout.columnSpan: root.bottomContentSizeMode === VenusOS.ListItem_BottomContentSizeMode_Stretch ? 2 : 1
 			Layout.topMargin: height > 0 ? Theme.geometry_listItem_content_verticalMargin / 2 : 0
 			Layout.bottomMargin: Theme.geometry_listItem_content_verticalMargin
+
+			PrimaryListLabel {
+				id: caption
+
+				width: Math.min(caption.implicitWidth, root.width - root.leftPadding - root.rightPadding)
```
blammit, line 112: This can be `width: parent.width`, otherwise the text spills onto the right-hand content:

image

blammit (commented):
Can this be reduced to just having a single `caption` property in ListItem? 

- It looks like Theme.color_listItem_secondaryText is always the preferred color, so it doesn't need customization
- I tried setting topPadding=0 and bottomPadding=0 by default in the caption label, and this still looked fine for ListLink, which seems to be the only case where these values aren't initialized to something else
- Components can just set caption to an empty string when it shouldn't be displayed, e.g. in the PageGensetModel.qml case

Also, maybe ListItem can just use a generic Label and style it accordingly, instead of using PrimaryListLabel and giving it a secondary font color. Since it is placing the label inside the `bottomContent`, it can just set `visible` instead of `preferredVisible` (which is intended to be used by the root item, rather than children).

DanielMcInnes:
> Can this be reduced to just having a single `caption` text property in ListItem?

If we added a `caption` text property to ListItem, we would also have to add `captionVisible`, `captionColor`, and arguably `captionTopPadding` and `captionBottomPadding`. Would you prefer it like that?

> 
>     * It looks like Theme.color_listItem_secondaryText is always the preferred color, so it doesn't need customization

I just changed the default color to `Theme.color_font_secondary`. It is overridden in PageGensetModel, ListLink, PageSettingsFroniusInverter and PageSettingsDemo

> 
>     * I tried setting topPadding=0 and bottomPadding=0 by default in the caption label, and this still looked fine for ListLink, which seems to be the only case where these values aren't initialized to something else

PageGensetModel, PageSettingsCanbus, and PageSettingsFroniusInverter also use `caption` without initializing topPadding and bottomPadding

> 
>     * Components can just set caption to an empty string when it shouldn't be displayed, e.g. in the PageGensetModel.qml case


Updated as suggested


> 
> Also, maybe ListItem can just use a generic Label and style it accordingly, instead of using PrimaryListLabel and giving it a secondary font color. 

Updated as suggested

>
> Since it is placing the label inside the `bottomContent`, it can just set `visible` instead of `preferredVisible` (which is intended to be used by the root item, rather than children).

Updated as suggested

blammit:
> PageGensetModel, PageSettingsCanbus, and PageSettingsFroniusInverter also use caption without initializing topPadding and bottomPadding

They seem to look okay with topPadding=0 and bottomPadding=0. For example, here is the PageGensetModel:

image

I'm thinking, how about we just use topPadding=0 and bottomPadding=0 everywhere for visual consistency?

>  just changed the default color to Theme.color_font_secondary. It is overridden in PageGensetModel, ListLink, PageSettingsFroniusInverter and PageSettingsDemo

Let's drop the color customizations, and just use the default secondary colour everywhere for captions, so that it is visually consistent.

> If we added a caption text property to ListItem, we would also have to add captionVisible, captionColor, and arguably captionTopPadding and captionBottomPadding. Would you prefer it like that?

No, that would be annoying :-(

Ideally the captions look the same everywhere, and ideally just a `caption` string property would be sufficient. If we really need customization then perhaps we could add a `captionLabel` property that is aliased to the ListItem label. Are there any cases where we need customizations, if we align the padding and colouring in the current use cases?

components/PageGensetModel.qml
```diff
@@ -361,11 +361,9 @@ ObjectModel {
 							secondaryText: CommonWords.yesOrNo(dataItem.value)
 							dataItem.uid: root.bindPrefix + "/Settings/BmsPresent"
 							preferredVisible: dataItem.isValid
-							bottomContentChildren: PrimaryListLabel {
-								//% "BMS control is enabled automatically when a BMS is present. Reset it if the system configuration changed or if there is no BMS present."
-								text: qsTrId("genset_bms_control_enabled_automatically")
-								preferredVisible: bmsControlled.dataItem.value === 1
-							}
+							caption.color: Theme.color_font_primary
```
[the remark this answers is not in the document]
DanielMcInnes: dropped

DanielMcInnes:
> > PageGensetModel, PageSettingsCanbus, and PageSettingsFroniusInverter also use caption without initializing topPadding and bottomPadding
> 
> They seem to look okay with topPadding=0 and bottomPadding=0. For example, here is the PageGensetModel:
> 
> image
> 
> I'm thinking, how about we just use topPadding=0 and bottomPadding=0 everywhere for visual consistency?
> 
> > just changed the default color to Theme.color_font_secondary. It is overridden in PageGensetModel, ListLink, PageSettingsFroniusInverter and PageSettingsDemo
> 
> Let's drop the color customizations, and just use the default secondary colour everywhere for captions, so that it is visually consistent.
> 
> > If we added a caption text property to ListItem, we would also have to add captionVisible, captionColor, and arguably captionTopPadding and captionBottomPadding. Would you prefer it like that?
> 
> No, that would be annoying :-(
> 
> Ideally the captions look the same everywhere, and ideally just a `caption` string property would be sufficient. If we really need customization then perhaps we could add a `captionLabel` property that is aliased to the ListItem label. Are there any cases where we need customizations, if we align the padding and colouring in the current use cases?

No, there are no cases where we need customizations, if we align the padding and colouring in the current use cases. Updated as suggested.

components/listitems/core/ListItem.qml
```diff
@@ -112,6 +113,20 @@ Item {
 			Layout.columnSpan: root.bottomContentSizeMode === VenusOS.ListItem_BottomContentSizeMode_Stretch ? 2 : 1
 			Layout.topMargin: height > 0 ? Theme.geometry_listItem_content_verticalMargin / 2 : 0
 			Layout.bottomMargin: Theme.geometry_listItem_content_verticalMargin
+
+			Label {
+				id: caption
```
blammit, line 118: id can be removed

components/listitems/core/ListItem.qml
```diff
@@ -14,6 +14,7 @@ Item {
 	property alias content: content
 	property alias bottomContent: bottomContent
 	property alias bottomContentChildren: bottomContent.children
+	property string caption: ""
```
blammit, line 17: "" initialization is not needed?

blammit (approved):
Looks great with the clean-up :)

components/listitems/core/ListItem.qml
DanielMcInnes: thanks, updated

components/listitems/core/ListItem.qml
DanielMcInnes: removed

PR #1880, opened 2025-01-28
Rework trackers and improve SolarDeviceListPage performance

### LINKED ISSUE #1848: Solar device list page can be very slow to load
See https://victrondevelopment.slack.com/archives/C020QG3JVEW/p1736945953241439

SolarDeviceListPage is using Column+Repeater for trackers, as a convenience to include custom headers for "PV chargers" vs "PV inverters" in the view, but this is likely causing the performance issues reported in the thread (with e.g. 30 devices), because:

1) Non-displayed delegates are rendered
2) Each delegate uses ListQuantityGroup, which has its own inefficiencies, as per https://github.com/victronenergy/gui-v2/issues/1338

Just fixing 1) should make a noticeable difference in a long list, but it would be best to fix 2) as well.

To reproduce this in mock mode on device, set the configuration to show at least 20 trackers.

blammit:
Add SolarInputModel to store the data for each tracker and PV inverter entry in the list. This allows the ListView to just create one list item for each entry in the model, instead of creating complex delegates with Column+Repeater objects when showing multiple trackers for a single solarcharger/multi/inverter service. Also, the model sorts its items to allow the use of ListView sections for creating the column headers.

This greatly improves the list loading performance when there are a large number of trackers or PV inverters.

Also, remove the tracker list from SolarDevice.qml as it is only required in specific pages, and those can load their own SolarTracker objects when needed.

Rename SolarDeviceListPage to SolarInputListPage, as it shows a list of trackers and PV inverters, rather than a list of SolarDevice items.

Fixes #1848


src/solarinputmodel.cpp
```diff
@@ -0,0 +1,179 @@
...
+}
+
+int SolarInputModel::rowCount(const QModelIndex &) const
+{
+	return count();
+}
+
+QHash<int, QByteArray> SolarInputModel::roleNames() const
+{
+	static QHash<int, QByteArray> roles = {
```
MikeTrahearn-Qinetic, line 168: I like the use of the {} initialiser for setting the role names.

src/solarinputmodel.h
```diff
@@ -0,0 +1,81 @@
...
+
+  Solar inputs include:
+	- trackers from solarcharger, multi and inverter services
+	- pvinverter services
+*/
+class SolarInputModel : public QAbstractListModel
+{
+	Q_OBJECT
+	QML_ELEMENT
+	Q_PROPERTY(int count READ count NOTIFY countChanged)
```
MikeTrahearn-Qinetic, line 26: add FINAL

blammit: Done

DanielMcInnes (approved):

PR #1890, opened 2025-02-05
Show a toast notification when changing security profiles

DanielMcInnes:
For wasm builds only. "Page will automatically reload in 5 seconds" Fixes 1881

### DIFF
--- a/pages/settings/PageSettingsAccessAndSecurity.qml
+++ b/pages/settings/PageSettingsAccessAndSecurity.qml
@@ -202,6 +202,12 @@ Page {
 							// This guards the wasm version to trigger a reload even if the reply isn't received.
 							BackendConnection.securityProtocolChanged()
 							Global.pageManager.popPage()
+							if (Qt.platform.os === "wasm" && !BackendConnection.vrm) {
+								Global.showToastNotification(VenusOS.Notification_Info,
+															 //% "Page will automatically reload in 5 seconds"
+															 qsTrId("access_and_security_page_will_reload"),
+															 3000)
+							}
 						}
 						dialogDoneOptions: VenusOS.ModalDialog_DoneOptions_OkAndCancel
 						height: securityProfile.pendingProfile === VenusOS.Security_Profile_Secured
--- a/src/backendconnection.cpp
+++ b/src/backendconnection.cpp
@@ -199,6 +199,9 @@ void BackendConnection::onReloadPageTimerExpired()
 // that fails and still triggers a reload (albeit a bit later).
 void BackendConnection::securityProtocolChanged()
 {
+	if (isVrm()) {
+		return;
+	}
 	QTimer *timer = new QTimer(this);
 	connect(timer, &QTimer::timeout, this, &BackendConnection::onReloadPageTimerExpired);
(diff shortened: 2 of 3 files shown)


DanielMcInnes:
Screenshot from 2025-02-05 13-56-01
Screenshot from 2025-02-05 13-55-56

pages/settings/PageSettingsAccessAndSecurity.qml
```diff
@@ -202,6 +202,12 @@ Page {
 							// This guards the wasm version to trigger a reload even if the reply isn't received.
 							BackendConnection.securityProtocolChanged()
 							Global.pageManager.popPage()
+							if (Qt.platform.os === "wasm") {
+								Global.showToastNotification(VenusOS.Notification_Info,
+															 //% "Page will automatically reload in 5 seconds"
```
chriadam, line 207: can you change this so it does something like:
```
Global.showToastNotification(
    VenusOS.Notification_Info,
    BackendConnection.isVrm
        //% "Closing due to security protocol change, you will need to relaunch remote console via VRM"
        ? qsTrId("access_and_security_closing_remote_console")
        //% "Page will automatically reload in 5 seconds"
        : qsTrId("access_and_security_page_will_reload"),
    3000)
```

chriadam: Actually, now I wonder ... do we actually need to reload the page, for VRM connections?

chriadam: Honestly, I think we can just add a check like: "if (!isVrm()) return;" at the start of BackendConnection::securityProtocolChanged(), as I don't think we need to reload the page if we are connected to VRM (rather than to the cerbo's web server directly).  See for example the similar check in BackendConnection::onNetworkConfigChanged().

Then, I think here in this code, we should just raise the toast notification you originally had, but ONLY if we are NOT connected via VRM.  e.g.: add a condition `&& !BackendConnection.isVrm` to the if statement above.

DanielMcInnes: > Honestly, I think we can just add a check like: "if (!isVrm()) return;" at the start of BackendConnection::securityProtocolChanged(), as I don't think we need to reload the page if we are connected to VRM (rather than to the cerbo's web server directly). See for example the similar check in BackendConnection::onNetworkConfigChanged().
> 
> Then, I think here in this code, we should just raise the toast notification you originally had, but ONLY if we are NOT connected via VRM. e.g.: add a condition `&& !BackendConnection.isVrm` to the if statement above.

done, I think.

Do you actually mean: add a check like: "if (isVrm()) return;" ?

chriadam: oops, yes

src/backendconnection.cpp
```diff
@@ -199,6 +199,10 @@ void BackendConnection::onReloadPageTimerExpired()
 // that fails and still triggers a reload (albeit a bit later).
 void BackendConnection::securityProtocolChanged()
 {
+	if (isVrm())
```
chriadam, line 202: style nit: open brace should be on same line as if condition

chriadam (approved):
One style nit, aside from that LGTM

Issue #1857, opened 2025-01-16
On the Brief page, in the center battery icon, show power instead of amps


DanielMcInnes:

chriadam:
Only when the user-selected power preference (watts vs amps) is at watts.

mpvader:
Hi I prioritised this one, its a simple (?) fix, and big improvement. We could even make it a v3.50-fixes thing, would be nice if its indeed simple enough @DanielMcInnes

chriadam:
PR#1891 has been merged.  Closing this issue as fixed.

PR #1891, opened 2025-02-05
Show brief page single center gauge power with selected units

chriadam:
The user can set a setting which determines whether to display electrical quantities as current (amps) or power (watts).

This commit ensures that we respect the value of that setting also for the quantity label shown in the single circular gauge on the brief page.

Contributes to issue #1857

### DIFF
--- a/components/CircularSingleGauge.qml
+++ b/components/CircularSingleGauge.qml
@@ -15,7 +15,8 @@ Item {
 	property alias icon: icon
 	property alias name: nameLabel.text
 	property alias voltage: voltageLabel.value
-	property alias current: currentLabel.value
+	property real current: NaN
+	property real power: NaN
 	property alias value: arc.value
 	property int status
 	property alias caption: captionLabel.text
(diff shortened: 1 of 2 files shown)


components/CircularSingleGauge.qml
```diff
@@ -15,13 +15,17 @@ Item {
...
-	property alias current: currentLabel.value
+	property real current: NaN
+	property real power: NaN
 	property alias value: arc.value
 	property int status
 	property alias caption: captionLabel.text
 	property alias animationEnabled: arc.animationEnabled
 	property alias shineAnimationEnabled: arc.shineAnimationEnabled
 
+	readonly property bool _unitAmps: (Global.systemSettings.electricalQuantity === VenusOS.Units_Amp && !isNaN(gauges.current))
```
MikeTrahearn-Qinetic, line 26: Consider moving _unitAmps inside the apparent one item in which it is used rather than having a public "private" property.
What if the units is watts and isNan(gauges.power)?

components/CircularSingleGauge.qml
```diff
@@ -101,10 +102,13 @@ Item {
 			QuantityLabel {
 				id: currentLabel
 
+				readonly property bool unitAmps: (Global.systemSettings.electricalQuantity === VenusOS.Units_Amp && !isNaN(gauges.current))
```
DanielMcInnes, line 105: maybe:
```
!isNaN(gauges.current)
&& (Global.systemSettings.electricalQuantity === VenusOS.Units_Amp || isNaN(gauges.power))
```

DanielMcInnes (approved):
lgtm, 1 optional suggestion

Issue #1883, opened 2025-01-30
No Voltage and Current displayed in Overview > Solar Yield


SenH:
### `Overview > Solar Yield`
<img width="1056" alt="Image" src="https://github.com/user-attachments/assets/89089524-db22-4001-88d7-623095db937d" />

### Voltage/Current is displayed in `Settings > Devices > Solar`
<img width="1281" alt="Image" src="https://github.com/user-attachments/assets/f352d65a-b4c0-43af-8273-5fc98096c39c" />

I updated today to Venus v3.54 but this has never been displayed since GUIv2 was introduced.

blammit:
This can be tested on installations with a single-phase PV inverter.

May be related to https://github.com/victronenergy/gui-v2/issues/1713.

chriadam:
According to https://github.com/victronenergy/venus/wiki/dbus#pv-inverters the `/Ac/Current` and `/Ac/Voltage` paths are deprecated, and presumably systemcalc isn't publishing appropriate values there, but we use those in PvInverter.qml.  For single-phase inverters, we can just find and use the appropriate per-phase data e.g. `/Ac/L1/Current` and `/Ac/L1/Voltage` directly, however for multi-phase systems I am not sure what we can do (voltage cannot be summed, and summing separate phases of AC current makes no sense either).  Well, I will fix the single-phase case now, and we can discuss the multiphase case further in issue 1713.

chriadam:
Merged PR#1892 to resolve this issue.  Closing as fixed.

Issue #1713, opened 2024-11-10
Display shows unused phase of PV inverters.


tmonxx:
Hi.

In the PV inverter view I can only see the power (W) (picture 1).
In the first view current and voltage are shown from the unused phase. 
After clicking to the second view on one of the inverters (picture 2).
All inverters are connected to L2.
Using 3 Hoymiles inverters. dbus connection to Victron via openDTU: https://github.com/henne49/dbus-opendtu

With GUI-v1 it was no issue.

![IMG_0916](https://github.com/user-attachments/assets/8cbd8257-6056-4294-a351-69bdbf49061e)
![IMG_0917](https://github.com/user-attachments/assets/39671e89-655c-412e-8662-54439d88bc05)

tmonxx:
Just now installed Beta 3.60-39 and it works fine. Good job!

![Image](https://github.com/user-attachments/assets/43922efc-1c17-41e4-9708-ca406987e5ee)

chriadam:
Thanks very much for checking!  Closing task as fixed.

PR #1892, opened 2025-02-05
Fix display of current + voltage for single-phase pvinverters

chriadam:
The /Ac/Current and /Ac/Voltage paths for pvinverters are deprecated / no longer valid.

Instead, determine whether there is a single valid phase available for a given pvinverter, and if so, export that phase's current and voltage as properties.

(If multiple valid phases exist, we cannot export valid aggregate values for current or voltage, so return NaN.)

Contributes to issue #1883

Contributes to issue #1713 in the case where only a single valid phase exists in the pvinverter device.

### DIFF
--- a/data/common/PvInverter.qml
+++ b/data/common/PvInverter.qml
@@ -13,9 +13,9 @@ Device {
 	readonly property int errorCode: _errorCode.isValid ? _errorCode.value : -1
 
 	readonly property real energy: _energy.isValid ? _energy.value : NaN
-	readonly property real current: _current.isValid ? _current.value : NaN
 	readonly property real power: _power.isValid ? _power.value : NaN
-	readonly property real voltage: _voltage.isValid ? _voltage.value : NaN
+	readonly property real current: pvInverter.phases.singlePhaseCurrent
+	readonly property real voltage: pvInverter.phases.singlePhaseVoltage
 
 	readonly property PhaseModel phases: PhaseModel {
 		function updateCount(maxPhaseCount) {
@@ -31,6 +31,7 @@ Device {
 		readonly property Instantiator _phases: Instantiator {
 			model: 3
 			delegate: QtObject {
+				id: phaseObject
 				required property int index
 				readonly property string phaseUid: pvInverter.serviceUid + "/Ac/L" + (index + 1)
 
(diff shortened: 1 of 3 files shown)


DanielMcInnes (approved):

data/common/PvInverter.qml
```diff
@@ -13,9 +13,9 @@ Device {
 	readonly property int errorCode: _errorCode.isValid ? _errorCode.value : -1
 
 	readonly property real energy: _energy.isValid ? _energy.value : NaN
-	readonly property real current: _current.isValid ? _current.value : NaN
 	readonly property real power: _power.isValid ? _power.value : NaN
-	readonly property real voltage: _voltage.isValid ? _voltage.value : NaN
+	readonly property real current: _validSinglePhase ? _validSinglePhase.current : NaN
+	readonly property real voltage: _validSinglePhase ? _validSinglePhase.voltage : NaN
```
blammit, line 18: I could see this data being useful in other multi-phase cases. What if these values could be provided by `PhaseModel` instead? E.g. if we had PhaseModel properties like `current` and `voltage` or `overallCurrent` and `overallVoltage` and these were only set in the single-phase case (i.e. when there is only one phase with valid voltage/current/power)?

data/common/PvInverter.qml
```diff
@@ -42,22 +42,46 @@ Device {
 
 				readonly property VeQuickItem _phaseEnergy: VeQuickItem {
 					uid: phaseUid + "/Energy/Forward"
-					onIsValidChanged: if (isValid) phases.updateCount(index + 1)
+					onIsValidChanged: {
+						if (isValid) {
+							phases.updateCount(index + 1)
+						} else if (phases.count > index) {
+							phases.setValue(index, PhaseModel.EnergyRole, NaN)
+						}
```
blammit, line 50: Maybe worth adding a function to do this to avoid the duplicated code for the four roles?

blammit (approved):
Minor comment, LGTM

chriadam:
done.  merging.

data/common/PvInverter.qml
chriadam: done

Issue #1830, opened 2025-01-12
Tank settings and Levels page shows incorrect tank capacity when greater than 9999 litres


blammit:
As per screenshots, when showing a 10000 litre tank, the capacity and current value are shown as "10":

![image](https://github.com/user-attachments/assets/bbc171a2-05c9-4017-ac4a-e001dfa9db7c)
![image](https://github.com/user-attachments/assets/8570cbd2-2aca-4ab8-afa1-3024486dd83a)

blammit:
Note: also check the font rendering, as the "l" litres symbol is not rendering as expected (screenshot was from Safari).

chriadam:
I suspect it's trying to use kL as the units rather than L, but perhaps something is going wrong with the unit display?

chriadam:
I tested this (with mock mode) on both desktop (windows) and cerbo gx (arm).  In both cases, the unit renders correctly as kL, see:

<img width="599" alt="Image" src="https://github.com/user-attachments/assets/c8c148c6-eb01-4d87-a887-c3416ae0c6e4" />

chriadam:
But, with wasm, we see that it does NOT render properly:

<img width="1007" alt="Image" src="https://github.com/user-attachments/assets/60a534ec-9312-4801-817c-a631fc6d39d1" />

So, this sounds like a font issue to me - the glyph for "kl" doesn't exist in the font used in the web assembly, perhaps?
The unicode codepoint is QStringLiteral("\u3398").

chriadam:
PR#1900 was merged.  Closing this issue as fixed.

PR #1900, opened 2025-02-10
Don't use specific kilolitres codepoint glyph in WebAssembly

chriadam:
The font we use doesn't support that glyph, so work around it.

Contributes to issue #1830

### DIFF
--- a/src/units.cpp
+++ b/src/units.cpp
@@ -308,8 +308,10 @@ quantityInfo Units::getDisplayTextWithHysteresis(VenusOS::Enums::Units_Type unit
 		// Litre scaling is special, only kilo range scaling is supported
 		if (unit == VenusOS::Enums::Units_Volume_Liter) {
 			if (isOverLimit(scaleMatch, VenusOS::Enums::Units_Scale_Kilo, previousScale)) {
-				// \u2113 = l, \u3398 = kl
-				quantity.unit = QStringLiteral("\u3398");
+				// \u2113 = litres symbol.
+				// we don't use \u3398 (kilolitres symbol)
+				// as it isn't available in the required font.
+				quantity.unit = QStringLiteral("k\u2113");
 				quantity.scale = VenusOS::Enums::Units_Scale_Kilo;
 				scaledValue = scaledValue / 1000.0;
 			}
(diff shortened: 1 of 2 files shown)


DanielMcInnes (commented):
lgtm. I also tested the swirly 'litres' symbol by itself, eg. for 5 litres, and that looked ok

DanielMcInnes (approved):

MikeTrahearn-Qinetic:
Is not the Roboto-Regular.ttf built into the application resources?

src/units.cpp
```diff
@@ -309,7 +309,15 @@ quantityInfo Units::getDisplayTextWithHysteresis(VenusOS::Enums::Units_Type unit
 		if (unit == VenusOS::Enums::Units_Volume_Liter) {
 			if (isOverLimit(scaleMatch, VenusOS::Enums::Units_Scale_Kilo, previousScale)) {
 				// \u2113 = l, \u3398 = kl
+#if defined(VENUS_WEBASSEMBLY_BUILD)
+				// The Roboto-Regular.ttf font used in WebAssembly builds doesn't provide a glyph for the \u3398 codepoint.
+				// We could pass the Global._quantityFontLoader font file to this method via getDisplayText()
+				// and then use QFontMetrics::inFont(QChar('\u3398')) to determine its existence programmatically,
+				// but then if there were per-language fallbacks this wouldn't work.  So, just hardcode a workaround.
+				quantity.unit = QStringLiteral("k\u2113");
+#else
```
blammit, line 318: Is it easier to just use "k\u2113" on all platforms to be consistent?

blammit (approved):
Wonder if we should just use the same codepoint everywhere, but LGTM either way

src/units.cpp
chriadam: It would be easier, yes.  Should we just do that?  I vote yes, fwiw...

chriadam:
> Is not the Roboto-Regular.ttf built into the application resources? If there is a difference between the this font and the WASM case, then is it reasonable to suggest this could be aligned? In any case, as has been commented in the code, this glyph needs to exist in all the different fonts used for the other languages too - which don't look like they are in the resources at all so without familiarising myself with the code for switching fonts on language changes, the risk is even higher here.

Yeah, I am not 100% certain what is happening (i.e. which font file the kL glyph is coming from, on desktop).  All I know is that - at least for the font file we use on wasm (which I believe is loaded on demand from the VRM cache, when launching from VRM) it seems to lack the needed glyph.

src/units.cpp
chriadam: done.

chriadam:
Done.  Merging.

PR #1902, opened 2025-02-11
Change PagePowerDebug to use QuantityTable and load in mock mode

blammit:
Use QuantityTable as ListTextGroup will be obsoleted.

Use BackendConnection.uidPrefix() to load the page in mock mode.

Part of #1338

### DIFF
--- a/pages/settings/debug/PagePowerDebug.qml
+++ b/pages/settings/debug/PagePowerDebug.qml
@@ -9,25 +9,16 @@ import Victron.VenusOS
 Page {
 	id: root
 
-	property string qwacsPvInverterPrefix: "com.victronenergy.pvinverter.qwacs_di1"
-	property string sensorsPvInverterPrefix: "com.victronenergy.pvinverter.vebusacsensor_output"
+	property string qwacsPvInverterPrefix: "%1/com.victronenergy.pvinverter.qwacs_di1".arg(BackendConnection.uidPrefix())
+	property string sensorsPvInverterPrefix: "%1/com.victronenergy.pvinverter.vebusacsensor_output".arg(BackendConnection.uidPrefix())
 	property string vebusPrefix: Global.system.veBus.serviceUid
 
 	function powerDiff(a, b) {
-		if (!a.isValid || !b.isValid)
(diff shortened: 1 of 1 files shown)


pages/settings/debug/PagePowerDebug.qml
```diff
@@ -84,57 +82,36 @@ Page {
...
+					{ title: "Sensors", unit: VenusOS.Units_Watt },
+					{ title: "Diff", unit: VenusOS.Units_Watt },
 				]
+				valueForModelIndex: function(phaseIndex, column) {
+					const phaseName = `L${phaseIndex + 1}`
+					switch (column) {
+					case 0:
+						return phaseName
+					case 1:
+						return acOut[`power${phaseName}`].value ?? NaN
```
MikeTrahearn-Qinetic, line 102: Personally not a massive fan of treating a proper object as a JSON object using keys as it isn't very safe but I can see how it is convenient here.

pages/settings/debug/PagePowerDebug.qml
```diff
@@ -9,25 +9,16 @@ import Victron.VenusOS
 Page {
 	id: root
 
-	property string qwacsPvInverterPrefix: "com.victronenergy.pvinverter.qwacs_di1"
-	property string sensorsPvInverterPrefix: "com.victronenergy.pvinverter.vebusacsensor_output"
+	property string qwacsPvInverterPrefix: "%1/com.victronenergy.pvinverter.qwacs_di1".arg(BackendConnection.uidPrefix())
```
MikeTrahearn-Qinetic, line 12: I didn't know you could .arg() a string without enclosing it in brackets.

pages/settings/debug/PagePowerDebug.qml
blammit: We do use it in various places in the code base, as it's a convenient way to access values when the property name is dynamic.

blammit: Yeah that's fair enough 👍

chriadam (approved):

PR #1903, opened 2025-02-13
Avoid rebuilding models for ListQuantityGroup and ListTextGroup

blammit:


pages/settings/PageCanbusStatus.qml
```diff
@@ -92,34 +78,105 @@ Page {
...
+			ListItem {
 				id: stateGroup
 
 				//% "State"
 				text: qsTrId("settings_state")
-				bottomContentChildren: ListTextGroup {
-					id: busOffCounters
-				}
+				content.children: [
+					ColumnLayout {
```
MikeTrahearn-Qinetic, line 119: You don't normally reach for a layout - did you mean to use ColumnLayout deliberately over Column? Same below.

tests/quantityobjectmodel/tst_quantityobjectmodel.cpp
```diff
@@ -0,0 +1,18 @@
...
+*/
+
+#include <QtQuickTest/quicktest.h>
+#include <QtQml/QQmlEngine>
+#include "quantityobject.h"
+#include "quantityobjectmodel.h"
+
+int main(int argc, char **argv) \
+{
+    qmlRegisterType<Victron::VenusOS::QuantityObject>("Victron.VenusOS", 2, 0, "QuantityObject");
```
MikeTrahearn-Qinetic, line 13: These (and in other tests) would not be needed if the test executable was correctly linked to the module.

tests/quantityobjectmodel/tst_quantityobjectmodel.qml
```diff
@@ -0,0 +1,120 @@
...
+ * Copyright (C) 2025 Victron Energy B.V.
+ * See LICENSE.txt for license information.
+*/
+
+import QtTest
+import QtQuick
+import Victron.VenusOS
+import QtQuick.Window
+
+TestCase {
```
MikeTrahearn-Qinetic, line 11: does the test cover the append/remove cases or are they not applicable because of the usage of QQmlListProperty?

pages/settings/PageCanbusStatus.qml
blammit: Thanks, the ColumnLayout was from WIP code so I've reverted it to a Column, as there aren't any particular benefits here from using the layout.

tests/quantityobjectmodel/tst_quantityobjectmodel.cpp
blammit: Yes, it would be worth cleaning that up, as a separate task

tests/quantityobjectmodel/tst_quantityobjectmodel.qml
blammit: Yes, not applicable as you said. We can have add/remove support in QuantityObjectModel in the future if needed.

CMakeLists.txt
```diff
@@ -703,6 +702,11 @@ list(APPEND VenusQMLModule_CPP_SOURCES
     src/productinfo.h
     src/quantityinfo.h
     src/quantityinfo.cpp
+    src/quantityinfo.cpp
```
chriadam, line 705: duplicated file reference

blammit: Removed

chriadam (approved):

blammit:
Rebased.

PR #1904, opened 2025-02-14
Move ListItem bottom padding into GradientListView and Column spacing

### LINKED ISSUE #1827: ListItems do not appear to be the right height
Any ListItem derived class with a height of `n` appears to be `n - 4` pixels high.

See ListItem.qml:50: `height: root.height - root.spacing     // spacing === 4` 

If I create a 'ListNavigation' object and set its height to 64, I expect it to look 64 pixels high.
It looks like the `spacing` is there for convenience, so that you can have:
```
GradientListView {
    model: ObjectModel {
        ListItem { }
        ListItem { }
    }
}
```
... with a 4 pixel spacing between the 2 ListItems

blammit:
Remove the built-in bottom padding from ListItem. This was previously required as a way to add spacing between items, instead of setting explicit 'spacing' on GradientListView or Column views, due to the fact that this would produce additional unwanted spacing for non-visible items within the view.

Now that visible items are no longer rendered by the view if VisibleItemModel is used, this built-in padding can be removed, and the spacing can be inserted directly into ListView and Column views instead.

Fixes #1827

### DIFF
--- a/pages/settings/PageSettingsDisplayBrief.qml
+++ b/pages/settings/PageSettingsDisplayBrief.qml
@@ -47,19 +47,24 @@ Page {
 			}
 		}
 
-		footer: ListRadioButtonGroup {
-			//: Show percentage values in Brief view
-			//% "Brief view unit"
-			text: qsTrId("settings_briefview_unit")
-			optionModel: [
-				//% "No labels"
-				{ display: qsTrId("settings_briefview_unit_none"), value: VenusOS.BriefView_Unit_None },
-				//% "Show tank volumes"
-				{ display: qsTrId("settings_briefview_unit_absolute"), value: VenusOS.BriefView_Unit_Absolute },
-				//% "Show percentages"
-				{ display: qsTrId("settings_briefview_unit_percentages"), value: VenusOS.BriefView_Unit_Percentage },
-			]
-			dataItem.uid: Global.systemSettings.serviceUid + "/Settings/Gui/BriefView/Unit"
+		footer: Column {
+			width: parent.width
+			topPadding: Theme.geometry_gradientList_spacing
+
+			ListRadioButtonGroup {
+				//: Show percentage values in Brief view
+				//% "Brief view unit"
+				text: qsTrId("settings_briefview_unit")
+				optionModel: [
+					//% "No labels"
+					{ display: qsTrId("settings_briefview_unit_none"), value: VenusOS.BriefView_Unit_None },
+					//% "Show tank volumes"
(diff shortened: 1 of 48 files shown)


MikeTrahearn-Qinetic (approved):

MikeTrahearn-Qinetic (approved):

pages/settings/PageSettingsDisplayBrief.qml
```diff
@@ -47,19 +47,24 @@ Page {
...
-				{ display: qsTrId("settings_briefview_unit_none"), value: VenusOS.BriefView_Unit_None },
-				//% "Show tank volumes"
-				{ display: qsTrId("settings_briefview_unit_absolute"), value: VenusOS.BriefView_Unit_Absolute },
-				//% "Show percentages"
-				{ display: qsTrId("settings_briefview_unit_percentages"), value: VenusOS.BriefView_Unit_Percentage },
-			]
-			dataItem.uid: Global.systemSettings.serviceUid + "/Settings/Gui/BriefView/Unit"
+		footer: Column {
+			width: parent.width
+			topPadding: Theme.geometry_gradientList_spacing
```
chriadam, line 52: This pattern is a bit ugly.  I wonder if there's some more elegant solution?  (Maybe add a "topSpacing" property to ListItem?  but maybe that's just as bad or worse...)

chriadam (approved):
LGTM although I had one nitpick/question

pages/settings/PageSettingsDisplayBrief.qml
blammit: Yeah I was wondering if ListItem should get a topPadding property, but wasn't sure ... I suppose it does already have left/right padding.

blammit: I suppose this would have to be topMargin as it should be outside of the coloured background, rather than within it, unlike the padding. I'll leave it as-is for now, as it feels odd to have to build the margin back into the ListItem, when we've just removed it. @MikeTrahearn-Qinetic suggested that ListItem could eventually become a QtQuick control, which would also provide a built-in API for margin/padding.

MikeTrahearn-Qinetic: We should definitely take a look as as separate topic, where it makes actual sense to use QtQuick.Controls Control as a base (or even some of the primitive Qt Quick Controls Templates) to bring a much needed architectural simplification across the whole design. This task would be made much easier if the Figma design was componentised like it should be.

PR #1909, opened 2025-02-17
Use ObjectAcConnection to fetch /Ac/PvOn* phase totals

### LINKED ISSUE #1908: Use ObjectAcConnection to fetch /Ac/PvOn* phase totals
The `/Ac/PvOnGrid`, `/Ac/PvOnGenset` and `/Ac/PvOnOutput` values may change frequently, so the PvMonitor implementation is sub-optimal as it re-calculates the phase totals for these values whenever a single phase value has changed.

Use ObjectAcConnection to monitor these values instead, as that is only done once per second.

blammit:
The /Ac/PvOnGrid, /Ac/PvOnGenset and /Ac/PvOnOutput values may change frequently, so the PvMonitor implementation was sub-optimal as it re-calculated the phase totals for these values whenever a single phase value was changed.

Use ObjectAcConnection to monitor these values instead, as that only calculates the totals once per second.

Fixes #1908

### DIFF
--- a/components/ObjectAcConnection.qml
+++ b/components/ObjectAcConnection.qml
@@ -12,6 +12,7 @@ QtObject {
 	property string bindPrefix
 	property string powerKey: "Power"
 	property string currentKey: "Current"
+	readonly property alias phaseCount: _phases.phaseCount
 
 	readonly property VeQuickItem powerL1: VeQuickItem {
 		uid: bindPrefix ? bindPrefix + "/L1/" + root.powerKey : ""
@@ -53,6 +54,7 @@ QtObject {
 	readonly property real current: _phaseCount.value === 1 && _currentL1.isValid ? _currentL1.value : NaN
 
 	readonly property PhaseModel phases: PhaseModel {
+		id: _phases
 		l2AndL1OutSummed: root.l2AndL1OutSummed
 		phaseCount: _phaseCount.value || 0
 	}
--- a/data/common/PvMonitor.qml
+++ b/data/common/PvMonitor.qml
@@ -15,21 +15,15 @@ Instantiator {
 
 	// AC power is the total power from Ac/PvOnGrid/L*/Power, Ac/PvOnGenset/L*/Power
 	// and Ac/PvOnOutput/L*/Power.
-	function updateAcTotals() {
+	function _updateAcTotals() {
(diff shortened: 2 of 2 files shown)


data/common/PvMonitor.qml
```diff
@@ -53,48 +47,12 @@ Instantiator {
 		Global.system.serviceUid + "/Ac/PvOnOutput"
 	]
 
-	delegate: QtObject {
-		id: acPvDelegate
+	delegate: ObjectAcConnection {
+		required property string modelData
```
MikeTrahearn-Qinetic, line 51: Wouldn't it be nice if you could use required property alias but alas you can't.

components/ObjectAcConnection.qml
```diff
@@ -12,6 +12,7 @@ QtObject {
 	property string bindPrefix
 	property string powerKey: "Power"
 	property string currentKey: "Current"
+	readonly property int phaseCount: phases.phaseCount
```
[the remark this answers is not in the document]
blammit: Done

data/common/PvMonitor.qml
blammit: Yeah that would definitely be nice to have.

data/common/PvMonitor.qml
```diff
@@ -53,48 +47,12 @@ Instantiator {
...
-					uid: acPvDelegate.serviceUid + "/L" + (model.index + 1) + "/Current"
-					onValueChanged: {
-						phase.current = value === undefined ? NaN : value
-						Qt.callLater(root.updateAcTotals)
-					}
-				}
-			}
-		}
+		onPhaseCountChanged: Qt.callLater(root._updateMaximumPhaseCount)
+		onPowerChanged: Qt.callLater(root._updateAcTotals)
```
chriadam, line 56: is there an assumption that if the current changes, the power will also change, which is why you don't also call _updateAcTotals onCurrentChanged?  (I can imagine a hypothetical example where the power stays the same if the voltage drops very low and current goes very high, which might need to be visualized or whatever)

blammit: Oops, yes this was meant to hook into onCurrentChanged as well. Fixed.

chriadam (approved):
LGTM

PR #1910, opened 2025-02-17
Minor fixes

blammit:


components/CommonWords.qml
```diff
@@ -522,6 +522,9 @@ QtObject {
 	//% "Zero feed-in power limit"
 	readonly property string zero_feed_in_power_limit: qsTrId("common_words_zero_feed_in_power_limit")
 
+	//% "This feature is disabled, since \"All modifications enabled\" under \"Settings -> General -> Modification checks\" is disabled."
+	readonly property string all_modifications_disabled: qsTrId("settings_large_features_currently_disabled")
+
 	function acInput(index) {
 			   //% "AC Input 1"
```
MikeTrahearn-Qinetic, line 529: Is the indentation correct here?

components/CommonWords.qml
```diff
@@ -522,6 +522,9 @@ QtObject {
 	//% "Zero feed-in power limit"
 	readonly property string zero_feed_in_power_limit: qsTrId("common_words_zero_feed_in_power_limit")
 
+	//% "This feature is disabled, since \"All modifications enabled\" under \"Settings -> General -> Modification checks\" is disabled."
+	readonly property string all_modifications_disabled: qsTrId("settings_large_features_currently_disabled")
```
MikeTrahearn-Qinetic, line 526: should the translation id take on the "common_words" prefix now?

components/CommonWords.qml
blammit: Yeah that seems off, but looks like it's intended to align with the other translation texts below, so I'll leave this alone.

components/CommonWords.qml
blammit: Oops yes, fixed

MikeTrahearn-Qinetic (approved):

PR #1911, opened 2025-02-17
Explicitly require C++20 to build

blammit:
Otherwise the build fails on Windows due to use of designated initializers in solarinputmodel.cpp.

Also fix -Wdeprecated-this-capture warnings in backendconnection.cpp and basedevicemodel.cpp.


src/basedevicemodel.cpp
```diff
@@ -50,7 +50,7 @@ int BaseDevice::deviceInstance() const
 void BaseDevice::setDeviceInstance(int deviceInstance)
 {
 	if (m_deviceInstance != deviceInstance) {
-		maybeEmitValidChanged([=]() {
+		maybeEmitValidChanged([this]() {
```
chriadam, line 53: also need to capture deviceInstance

src/basedevicemodel.cpp
```diff
@@ -35,7 +35,7 @@ void BaseDevice::setServiceUid(const QString &serviceUid)
 	}
 
 	if (m_serviceUid != serviceUid) {
-		maybeEmitValidChanged([=]() {
+		maybeEmitValidChanged([this]() {
```
chriadam, line 38: also need to capture serviceUid

src/basedevicemodel.cpp
```diff
@@ -78,7 +78,7 @@ QString BaseDevice::productName() const
 void BaseDevice::setProductName(const QString &productName)
 {
 	if (m_productName != productName) {
-		maybeEmitValidChanged([=]() {
+		maybeEmitValidChanged([this]() {
```
chriadam, line 81: also need to capture productName

src/basedevicemodel.cpp
```diff
@@ -93,7 +93,7 @@ QString BaseDevice::customName() const
 void BaseDevice::setCustomName(const QString &customName)
 {
 	if (m_customName != customName) {
-		maybeEmitValidChanged([=]() {
+		maybeEmitValidChanged([this]() {
```
chriadam, line 96: also need to capture customName

src/basedevicemodel.cpp
blammit: Oops, I think this should be `=, this`

src/basedevicemodel.cpp
blammit: Done

src/basedevicemodel.cpp
blammit: Done as above

src/basedevicemodel.cpp
blammit: Done as above

src/basedevicemodel.cpp
blammit: Updated to capture the local variables.

chriadam (approved):

PR #1915, opened 2025-02-18
Allow Brief view levels to be set to None

### LINKED ISSUE #541: Can't set brief view level to none
There is no way to select None for any of the levels. There are systems that don't have that many tank sensors and/or you do not want to display something on the Brief page but do want it on the tanks page.

blammit:
Add 'None' to the Brief view level options. If this is selected, the level will be omitted from the Brief page's center gauges. If all levels are set to None, the Brief page will only show the single- gauge view with battery details.

Rework GaugeModel to load the model based on the preferred gauge order, instead of loading all available tank types and then finding the preferred order for each type.

Also simplify the settings implementation for saving and loading the Brief level settings.

Fixes #541

### DIFF
--- a/components/GaugeModel.qml
+++ b/components/GaugeModel.qml
@@ -5,7 +5,6 @@
 
 import QtQuick
 import Victron.VenusOS
-import QtQuick.Controls.impl as CP
 import Victron.Gauges
 
 /*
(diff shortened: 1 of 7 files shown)


components/GaugeModel.qml
```diff
@@ -15,73 +14,68 @@ import Victron.Gauges
...
+			} else if (preferredGaugeTypes[i] === VenusOS.Tank_Type_Battery) {
+				// Assume battery is always available.
+				gaugeTypes.push(VenusOS.Tank_Type_Battery)
+			} else if (preferredGaugeTypes[i] !== VenusOS.Tank_Type_None) {
+				// If there is a tank of this type, then add the type to the displayed gauges.
+				// Otherwise, add another tank that is available on the system.
+				const preferredTankModel = Global.tanks.tankModel(preferredGaugeTypes[i])
+				if (preferredTankModel.count > 0) {
+					gaugeTypes.push(preferredGaugeTypes[i])
+				} else {
```
chriadam, line 34: do we really need the else case?  I mean, if the user has specified a particular tank to be shown in the gauge, I'm not sure why we would fall back to showing a different one if their selected tank isn't available - but maybe I'm wrong.

pages/BriefPage.qml
```diff
@@ -160,25 +160,26 @@ SwipeViewPage {
...
+
 	Loader {
 		id: mainGauge
 
 		y: (root._unexpandedHeight - height) / 2
 		width: Theme.geometry_mainGauge_size
 		height: width
 		x: sidePanel.x/2 - width/2
-		sourceComponent: Global.tanks.totalTankCount === 0 ? singleGauge : multiGauge
+		sourceComponent: gaugeModel.count === 0 ? singleGauge : multiGauge
```
chriadam, line 174: If I change all levels to "None" except the second level which I set to "Battery", I expected the UI to load the CircularSingleGauge but it appeared to load a CircularMultiGauge with just a single (battery) gauge shown.

chriadam:
Mostly looks great, nice work!  Two comments / questions.

pages/BriefPage.qml
blammit: I wasn't entirely sure about this case either... my conclusion was that the "level" settings would appear to be configuring the multi-gauge UI, rather than the single-gauge UI. So if you set a battery as a level then it would seem that you would expect to see the battery in the multi-gauge UI. Otherwise if you just wanted to see the CircularSingleGauge with the battery, you could just set all the levels to "None".

components/GaugeModel.qml
blammit: Wonder if we should add another setting like "Show other tanks if available" as a switch, and if toggled on (the default) then it will keep the existing behaviour to ensure that tanks are shown if this configuration hasn't been set, and otherwise it can be used to ensure the user only sees the configured tanks.

blammit: In any case I've updated the code comment for now, to explain why this `else` case is needed.

Issue #1754, opened 2024-11-22
DVCC temperature selection menu: Use customname instead of product name


ReinvdZee:
In the DVCC menu, there is an option to select which temperature sensor is used for STS:

![image](https://github.com/user-attachments/assets/ab95bfe1-77d7-4340-a524-281fc0bcf890)

However, for temperature sensors connected to the SmartShunt, this menu shows the product names of the SmartShunts which can be selected:
![image](https://github.com/user-attachments/assets/4bade8e4-644f-4c7e-995c-404b8231669c)
For systems with multiple SmartShunts with temperature sensors, it is not possible for the user to distinguish between SmartShunts in order to select the desired SmartShunt to use the temperature data from.

The Solution is to show the customname of the devices instead of the product name.

This issue can be reproduced on [this sytem](https://vrm.victronenergy.com/installation/323213/dashboard), navigate to Settings->DVCC->Temperature sensor (STS needs to be enabled).

chriadam:
@ReinvdZee could you help to test PR#1917 please?  I checked with the installation you mentioned, and see "SmartShunt" in the radio button list now (which I assume is a custom name, since it's not the entire product name).

chriadam:
PR merged.  Closing as fixed - can be reopened if there's a case which doesn't work still :-)

PR #1917, opened 2025-02-18
Expose custom name for DVCC STS temperature sensors if available

chriadam:
Contributes to issue #1754

### DIFF
--- a/pages/settings/PageSettingsDvcc.qml
+++ b/pages/settings/PageSettingsDvcc.qml
@@ -89,11 +89,29 @@ Page {
 							return
 						}
 						const modelArray = Utils.jsonSettingsToModel(value)
-						if (modelArray) {
-							temperatureServiceRadioButtons.optionModel = modelArray
-						} else {
-							console.warn("Unable to parse data from", uid)
+						let serviceUids = []
+						for (let i = 0; i < modelArray.length; ++i) {
+							const serviceId = modelArray[i].value
+							if (serviceId.startsWith("com.victronenergy.")) {
+								const firstIndexOfSlash = serviceId.indexOf('/')
+								const secondIndexOfSlash = serviceId.indexOf('/', firstIndexOfSlash + 1)
+								const deviceInstanceSubstring = serviceId.substring(firstIndexOfSlash + 1, secondIndexOfSlash)
+								serviceUids.push({
+									optionIndex: i,
+									serviceUid: BackendConnection.serviceUidFromName(serviceId.substr(0, firstIndexOfSlash), parseInt(deviceInstanceSubstring))
+								})
+							}
 						}
+						temperatureServiceRadioButtons.optionModel = modelArray
+						temperatureServiceInstantiator.model = serviceUids
+					}
+				}
+
+				Instantiator {
+					id: temperatureServiceInstantiator
+					delegate: Device {
+						serviceUid: modelData.serviceUid
+						onNameChanged: temperatureServiceRadioButtons.optionModel[modelData.optionIndex].display = name
 					}
 				}
 			}


pages/settings/PageSettingsDvcc.qml
```diff
@@ -83,19 +83,59 @@ Page {
...
+					id: available
 					uid: Global.system.serviceUid + "/AvailableTemperatureServices"
 					onValueChanged: {
 						if (value === undefined) {
 							return
 						}
 						const modelArray = Utils.jsonSettingsToModel(value)
 						if (modelArray) {
-							temperatureServiceRadioButtons.optionModel = modelArray
+							temperatureServiceInstantiator.displays = modelArray.map(function(d) { return d.display; })
```
MikeTrahearn-Qinetic, line 94: I see how the the modelArray is being split up here and re-accessed by index in the tempSensorData delegate. 
Is this the most effective way to get into the data from the delegate (not appreciating the structure of modelArray) using required properties? Perhaps not, but I thought I'd ask the question.

chriadam: I originally attempted to have two required properties (`required property string value` and `required property string display` and just use the modelArray as the instantiator.model - but this didn't work, and I was not sure why, so I used this ugly approach instead.  If someone can tell me what my mistake was, I'm definitely happy to use that solution!

MikeTrahearn-Qinetic: Did you use the "display" and "value" at all? There's a lot going on here with multiple "hidden" calls to .map in the mix, so without really understanding the VeQuickItem's raw value is to begin with, it is difficult to advise further. The current method is OK.

blammit: The `required` property doesn't work for keys of JSON objects so you just have to use `modelData` instead. 

Perhaps a simpler way to do this is to only instantiate objects for the list entries that actually refer to devices, and then you could do the string parsing-for-device-instance when composing the Instantiator model instead of from the instantiated objects. Also, looks like you need `firstIndexOfSlash + 1` when finding the second slash. I tried this:

```
VeQuickItem {
    uid: Global.system.serviceUid + "/AvailableTemperatureServices"
    onValueChanged: {
        if (value === undefined) {
            return
        }
        const modelArray = Utils.jsonSettingsToModel(value)
        let serviceUids = []
        for (let i = 0; i < modelArray.length; ++i) {
            const serviceId = modelArray[i].value
            if (serviceId.startsWith("com.victronenergy.")) {
                const firstIndexOfSlash = serviceId.indexOf('/')
                const secondIndexOfSlash = serviceId.indexOf('/', firstIndexOfSlash + 1)
                const deviceInstanceSubstring = serviceId.substring(firstIndexOfSlash + 1, secondIndexOfSlash)
                serviceUids.push({
                    optionIndex: i,
                    serviceUid: BackendConnection.serviceUidFromName(serviceId.substr(0, firstIndexOfSlash), deviceInstanceSubstring)
                })
            }
        }
        temperatureServiceRadioButtons.optionModel = modelArray
        temperatureServiceInstantiator.model = serviceUids
    }
}

Instantiator {
    id: temperatureServiceInstantiator
    delegate: Device {
        serviceUid: modelData.serviceUid
        onNameChanged: temperatureServiceRadioButtons.optionModel[modelData.optionIndex].display = name
    }
}
```

I also added cropped the service name further when calling `BackendConnection.serviceUidFromName()` as there may be more characters after the device instance; with "Boat and Motorhome Demo 2" there was an entry named "com.victronenergy.alternator/288/Dc/0/Temperature".

pages/settings/PageSettingsDvcc.qml
```diff
@@ -83,19 +83,59 @@ Page {
 				preferredVisible: commonSettings.dvccActive && sharedTempSense.checked
 
 				VeQuickItem {
+					id: available
```
blammit, line 86: `id` seems unused?

pages/settings/PageSettingsDvcc.qml
MikeTrahearn-Qinetic: >> "The required property doesn't work for keys of JSON objects so you just have to use modelData instead."
I was curious about this because at first I thought "surely that is not the case", but I confirmed in a small test app that this is the case in Qt 6.6. However I can pleasantly confirm that this has been fixed in Qt 6.8 for your information.

blammit: Oh nice! Glad it's been fixed.

chriadam: Good points, thanks for catching the substr index bug!  I have updated the PR with your suggestions.

pages/settings/PageSettingsDvcc.qml
```diff
@@ -89,11 +89,29 @@ Page {
...
+						let serviceUids = []
+						for (let i = 0; i < modelArray.length; ++i) {
+							const serviceId = modelArray[i].value
+							if (serviceId.startsWith("com.victronenergy.")) {
+								const firstIndexOfSlash = serviceId.indexOf('/')
+								const secondIndexOfSlash = serviceId.indexOf('/', firstIndexOfSlash + 1)
+								const deviceInstanceSubstring = serviceId.substring(firstIndexOfSlash + 1, secondIndexOfSlash)
+								serviceUids.push({
+									optionIndex: i,
+									serviceUid: BackendConnection.serviceUidFromName(serviceId.substr(0, firstIndexOfSlash), deviceInstanceSubstring)
```
blammit, line 101: Oops, realised that my code didn't call `parseInt()` on the `deviceInstanceSubstring`, so it will probably need that to work with the `int` argument to `BackendConnection.serviceUidFromName()`.

blammit (approved):
Just a comment on the `int` argument, LGTM

pages/settings/PageSettingsDvcc.qml
chriadam: thanks, will update.  seemed to work regardless, but I agree it should be done.

PR #1921, opened 2025-02-19
Add QmlObject for future simplification of "private" properties

### LINKED ISSUE #1920: Introduce "QmlObject" to simplify "_private" properties
Proposing a new Victron.VenusOS / Core / QmlObject: 

A QtObject but with a default property alias to a property list<QtObject> - preferably in C++ as a [default] QQmlListProperty<QtObject>.
This would remove much of the “pseudo-private properties” and make the QML much more readable. It would also remove the ambiguity between accessing QtObject properties with its _propertyName or its id; only the id remains which causes a correct qualification which can be detected and validated by qmllint going forward.

MikeTrahearn-Qinetic:
Fixes #1920

### DIFF
--- a/src/qmlobject.cpp
+++ b/src/qmlobject.cpp
@@ -1,0 +1,59 @@
+/*
+** Copyright (C) 2024 Victron Energy B.V.
+** See LICENSE.txt for license information.
+*/
+
+#include "qmlobject.h"
+
+using namespace Victron::VenusOS;
+
+QmlObject::QmlObject(QObject *parent)
+	: QObject{parent}
+{}
+
+QQmlListProperty<QObject> QmlObject::data()
+{
+	return QQmlListProperty<QObject>(this, this,
+										QmlObject::qmlObjectAppend,
+										QmlObject::qmlObjectCount,
+										QmlObject::qmlObjectAt,
+										QmlObject::qmlObjectClear,
+										nullptr, // no replace
+										QmlObject::qmlObjectRemoveLast);
+}
+
+void QmlObject::qmlObjectAppend(QQmlListProperty<QObject> *prop, QObject *object)
+{
+	QmlObject *qmlObject = static_cast<QmlObject *>(prop->data);
+	qmlObject->m_data.append(object);
+	emit qmlObject->dataChanged();
+}
+
+qsizetype QmlObject::qmlObjectCount(QQmlListProperty<QObject> *prop)
+{
+	QmlObject *qmlObject = static_cast<QmlObject *>(prop->data);
+	return qmlObject->m_data.count();
+}
+
+QObject *QmlObject::qmlObjectAt(QQmlListProperty<QObject> *prop, qsizetype index)
+{
+	QmlObject *qmlObject = static_cast<QmlObject *>(prop->data);
+	if (index >= 0 && index < qmlObject->m_data.count()) {
+		return qmlObject->m_data.at(index);
+	}
+	return nullptr;
+}
+
+void QmlObject::qmlObjectClear(QQmlListProperty<QObject> *prop)
+{
+	QmlObject *qmlObject = static_cast<QmlObject *>(prop->data);
+	qmlObject->m_data.clear();
+	emit qmlObject->dataChanged();
+}
+
+void QmlObject::qmlObjectRemoveLast(QQmlListProperty<QObject> *prop)
+{
+	QmlObject *qmlObject = static_cast<QmlObject *>(prop->data);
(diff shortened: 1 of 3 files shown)


MikeTrahearn-Qinetic:
The following is now possible:

```
QmlObject {

   QtObject {
      id: heyImPrivateHaha
   }

   QtObject {
      id: soAmI
   }

   QmlObject {
      id: alsoImEvenNested

      QtObject {
         id: arentYouLucky
      }
   }
}
```

MikeTrahearn-Qinetic:
If there are any suggestions for a better name for "QmlObject" now's a good time to change it!

src/qmlobject.h
```diff
@@ -0,0 +1,50 @@
...
+	static qsizetype qmlObjectCount(QQmlListProperty<QObject> *prop);
+	static QObject *qmlObjectAt(QQmlListProperty<QObject> *prop, qsizetype index);
+	static void qmlObjectClear(QQmlListProperty<QObject> *prop);
+	static void qmlObjectRemoveLast(QQmlListProperty<QObject> *prop);
+
+	void append(QObject *object);
+	int count() const;
+	QObject *at(const qsizetype index);
+	void clear();
+	void removeLast();
```
blammit, line 38: Assuming these member functions will never be used outside of this class, I think it'd be reasonable to remove them, and just do the implementation within the `static` functions.

blammit (commented):
Looks reasonable enough :)

blammit:
> The following is now possible:

I think I hear the Minions when reading those `id` values

src/qmlobject.h
MikeTrahearn-Qinetic: DONE

src/qmlobject.h
```diff
@@ -0,0 +1,44 @@
+/*
+** Copyright (C) 2024 Victron Energy B.V.
+** See LICENSE.txt for license information.
+*/
+
+#ifndef QMLOBJECT_H
```
blammit, line 6: Use `VICTRON_GUIV2_QMLOBJECT_H` instead? @chriadam suggested that for some other C++ headers in another PR

src/qmlobject.cpp
```diff
@@ -0,0 +1,63 @@
...
+qsizetype QmlObject::qmlObjectCount(QQmlListProperty<QObject> *prop)
+{
+	QmlObject *qmlObject = static_cast<QmlObject *>(prop->data);
+	return qmlObject->m_data.count();
+}
+
+QObject *QmlObject::qmlObjectAt(QQmlListProperty<QObject> *prop, qsizetype index)
+{
+	QmlObject *qmlObject = static_cast<QmlObject *>(prop->data);
+	if(index >= 0 && index < qmlObject->m_data.count()) {
```
blammit, line 43: Add space after `if`

blammit (approved):
Commented about some minor things but LGTM

src/qmlobject.h
MikeTrahearn-Qinetic: DONE

src/qmlobject.cpp
MikeTrahearn-Qinetic: DONE

MikeTrahearn-Qinetic:
You are only allowed to use this class if you can say "Batooay to quer da ads a banana" beforehand.

src/qmlobject.cpp
```diff
@@ -0,0 +1,63 @@
...
+										QmlObject::qmlObjectAt,
+										QmlObject::qmlObjectClear,
+										nullptr, // no replace
+										QmlObject::qmlObjectRemoveLast);
+}
+
+void QmlObject::qmlObjectAppend(QQmlListProperty<QObject> *prop, QObject *object)
+{
+	// no attempt is made to reparent the object,
+	// however it is likely (but not guaranteed)) that it is a child anyway.
```
chriadam, line 28: Could you expand this comment to explain in which circumstances the object will be child, versus when it will not be?  I assume that if you just declare the object in-line, it WILL be a child, but if you imperatively manually modify the list property you can inject (by id) objects which aren't children?

MikeTrahearn-Qinetic: Declared children are .. children. This is a given. That they end up in the list property is not... this depends on whether the list property is a default property or not.
You may also bind the list property to another list somewhere else. Unlikely, but possible.
I added this comment for my own clarification because not only isn't it immediately obvious but I ran into this some years ago and discovered that internally QQmlListProperty is doing some smarts behind the scenes in terms of object ownership. However, since this comment is largely moot, I will remove it.

PR #1923, opened 2025-02-20
Fix value display in Settings > Debug > Values

blammit:
Need to use 'interactive' instead of 'enabled' as per 911f826df2019739c99759695d65cec60e979dc0.


MikeTrahearn-Qinetic (commented):
Could this whole thing be a simpler binding?

blammit:
Yes that's true, I've simplified it to ` secondaryText: interactive ? "" : (model.value ?? NaN)`.

MikeTrahearn-Qinetic (approved):

PR #1926, opened 2025-02-21
Device List: ensure service-specific delegates are created

### LINKED ISSUE #1925: Device list sometimes shows gaps between items
The Device List sometimes shows a gap between items. For example, this happened in mock mode after adding and removing various tanks:

Image

This appears related to the changes for https://github.com/victronenergy/gui-v2/pull/1904.

blammit:
Instead of only setting the Loader source from
on_LoadCustomDelegateChanged, also set it from
Component.onCompleted. Otherwise, if the value of loadCustomDelegate never changes from its initial value, the Loader source will never be set, and the service-specific delegate (or the 'Not connected' default delegate) will never be set, and the item will be blank.

Fixes #1925

### DIFF
--- a/pages/settings/devicelist/DeviceListPage.qml
+++ b/pages/settings/devicelist/DeviceListPage.qml
@@ -23,29 +23,35 @@ Page {
 			required property string cachedDeviceName
 
 			readonly property bool _loadCustomDelegate: connected && !!device
+			property bool _usingCustomDelegate
 
-			// Only set width; height is sized to the loaded item, in case preferredVisible=false and the
-			// item should not be visible.
-			width: parent ? parent.width : 0
-
-			on_LoadCustomDelegateChanged: {
-				let delegateUri
+			function _resetSource() {
 				if (_loadCustomDelegate) {
 					const serviceType = BackendConnection.serviceTypeFromUid(device.serviceUid)
 					if (!serviceType) {
 						console.warn("DeviceList: cannot load delegate, cannot read service type from serviceUid:", device.serviceUid)
 						return
 					}
-					setSource("delegates/DeviceListDelegate_%1.qml".arg(serviceType), {
-						device: Qt.binding(function() { return delegateLoader.device }),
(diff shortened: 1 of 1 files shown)


pages/settings/devicelist/DeviceListPage.qml
```diff
@@ -24,11 +24,7 @@ Page {
 
 			readonly property bool _loadCustomDelegate: connected && !!device
 
-			// Only set width; height is sized to the loaded item, in case preferredVisible=false and the
-			// item should not be visible.
-			width: parent ? parent.width : 0
-
-			on_LoadCustomDelegateChanged: {
+			function _resetSource() {
 				let delegateUri
```
MikeTrahearn-Qinetic, line 28: delegateUri seems to be unused?

pages/settings/devicelist/DeviceListPage.qml
```diff
@@ -47,13 +43,20 @@ Page {
...
+
 			onStatusChanged: {
 				if (status === Loader.Error) {
 					console.log("Failed to load Device List delegate for '%1' service from file: %2"
 						.arg(BackendConnection.serviceTypeFromUid(device.serviceUid))
 						.arg(source))
 				}
 			}
+
+			on_LoadCustomDelegateChanged: _resetSource()
```
MikeTrahearn-Qinetic, line 58: you may find that _resetSource() function calls itself whenever _loadCustomDelegate changes due to a rather sneaky-yet-useful binding quirk with functions using properties.

blammit: Oh yeah, good point, the delegate is now being reset multiple times, which is unnecessary. Have updated it to ensure it's only loaded once, thanks!

pages/settings/devicelist/DeviceListPage.qml
blammit: Removed

MikeTrahearn-Qinetic (approved):

PR #1931, opened 2025-02-25
Fix build issue

chriadam:
Don't rely on a forward type declaration used in a template type specialisation, instead include the header directly.


blammit (approved):

DanielMcInnes (approved):

blammit:
Merging now to fix the issue in `main`.

PR #1934, opened 2025-02-25
Rein/generator tank level condition

ReinvdZee:
https://github.com/victronenergy/venus/issues/1420

### DIFF
--- a/pages/settings/GeneratorCondition.qml
+++ b/pages/settings/GeneratorCondition.qml
@@ -15,7 +15,9 @@ ListNavigation {
 	property int decimals: 1
 	property bool startValueIsGreater: true
 	property bool supportsWarmup: false
+	property bool showTankServices:false
 	property string name: text
+	property string startStopBindPrefix
 
 	//% "Use %1 value to start/stop"
 	property string enableDescription: qsTrId("generator_condition_use_value_to_start_stop").arg(name)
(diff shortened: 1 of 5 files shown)


src/enums.h
```diff
@@ -269,6 +269,7 @@ class Enums : public QObject
 		Generators_State_WarmUp = 2,
 		Generators_State_CoolDown = 3,
 		Generators_State_Stopping = 4,
+		Generators_State_Stopped_By_Tank_Level = 5,
```
chriadam, line 272: I think this should be: Generators_State_StoppedByTankLevel

ReinvdZee: Thanks for spotting that. Should be fixed.

pages/settings/GeneratorCondition.qml
```diff
@@ -15,11 +15,16 @@ ListNavigation {
...
 	property bool startValueIsGreater: true
 	property bool supportsWarmup: false
+	property bool showTankServices:false
 	property string name: text
+	property string startStopBindPrefix
 
 	//% "Use %1 value to start/stop"
 	property string enableDescription: qsTrId("generator_condition_use_value_to_start_stop").arg(name)
 
+	//% "Prevent start until %1 is higher than"
```
chriadam, line 25: nitpick: could also add a translator-info comment like: "//: %1 is the name of the tank level sensor" (unless I've misunderstood what the name arg refers to here)

pages/settings/GeneratorCondition.qml
```diff
@@ -63,6 +68,35 @@ ListNavigation {
...
+						defaultSecondaryText: qsTrId("page_generator_conditions_unavailable_tank_service_set_another")
+						dataItem.uid: bindPrefix + "/TankService"
+
+						VeQuickItem {
+							id: availableTankServices
+
+							uid: startStopBindPrefix + "/AvailableTankServices"
+							onValueChanged: {
+								if (value === undefined) {
+									print("No value:", uid)
```
chriadam, line 87: instead of print(), maybe console.warn() instead?

chriadam (approved):
A couple of nitpicks, but aside from those it LGTM

pages/settings/GeneratorCondition.qml
ReinvdZee: Removed the print(), as it adds no value.

chriadam:
Thanks, LGTM, feel free to merge

PR #1936, opened 2025-02-26
Key navigation preparations

blammit:
Shuffle some code around in the MainView and add some FocusScopes so that we can add key navigation later on.

### DIFF
--- a/components/StatusBar.qml
+++ b/components/StatusBar.qml
@@ -8,11 +8,12 @@ import QtQuick.Controls as C
 import QtQuick.Controls.impl as CP
 import Victron.VenusOS
 
-Rectangle {
+FocusScope {
 	id: root
 
 	required property PageStack pageStack
 	property string title
+	property alias backgroundColor: backgroundRect.color
 
 	property int leftButton: VenusOS.StatusBar_LeftButton_None
 	property int rightButton: VenusOS.StatusBar_RightButton_None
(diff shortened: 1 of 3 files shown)


components/StatusBar.qml
```diff
@@ -8,11 +8,12 @@ import QtQuick.Controls as C
 import QtQuick.Controls.impl as CP
 import Victron.VenusOS
 
-Rectangle {
+FocusScope {
 	id: root
 
 	required property PageStack pageStack
 	property string title
+	property color backgroundColor
```
chriadam, line 16: possibly could be an alias property

blammit: Done

blammit:
On second thoughts, the intermedia FocusScope for containing the control cards loader and page stack are not actually necessary since the root MainView will become a FocusScope, so I've removed it.

components/StatusBar.qml
chriadam: this change might have been clobbered by your later one?

chriadam (approved):
one nitpick but LGTM.
Was able to test via: mock mode, click notifications, press tab, then could press left/right to navigate between pages.

Does the updated design for key navigation have focus halos or something?  I think we need to visualise focus somehow.

blammit:
> Does the updated design for key navigation have focus halos or something? I think we need to visualise focus somehow.

Yes, there are focus halos in the design; will add those when adding the proper key navigation feature.

PR #1937, opened 2025-02-27
Add option to override mqtt param for wasm

blammit:
Now you can do something like:
http://venus.local/gui-v2?mqtt=ws://123.123.1.2/websocket-mqtt

This also means we no longer need a manual patching of wasm/index.html when connecting from desktop.

### DIFF
--- a/wasm/index.html
+++ b/wasm/index.html
@@ -69,12 +69,20 @@
                 ui.style.display = 'block';
             }
 
+            const searchParams = new URLSearchParams(window.location.search);
+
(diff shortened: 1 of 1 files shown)


chriadam (approved):
style nit: it could be `const needDefaultArgs = !searchParams.has("mqtt") && !searchParams.has("id") && !searchParams.has("shard");` but anyway LGTM

martinbosma:
Great! I've closed https://github.com/victronenergy/gui-v2/pull/1861

Issue #1912, opened 2025-02-17
Some adjustments around PV power limiting


philipptrenz:
I discussed with @izak and we would propose the following enhancements:

- Reformat the "Zero feed-in power limit" value for PV inverters by stripping decimal places and adding it's unit "W"
- Change the "Zero feed-in power limit" text for PV inverters in the device list into "Dynamic power limit", as it actually not only applies to zero feed-in
- For uniformity, do the same in _Settings_ > _Integrations_ > _PV Inverters_ > _Inverters_ > _..._ and rename it from "Power limiting" to "Dynamic power limiting"

![Image](https://github.com/user-attachments/assets/78799701-bb93-40a8-90b6-f775db8a7b73)
![Image](https://github.com/user-attachments/assets/40921bb9-4e47-4217-a9ac-3a4b9b2db3cb)

chriadam:
Merged fix to main.  Closing as fixed.

PR #1940, opened 2025-02-27
Rename zero feed-in limit to dynamic power limit

chriadam:
Also use appropriate units and precision for displaying the limit.

Contributes to issue #1912

### DIFF
--- a/pages/settings/PageSettingsFroniusInverter.qml
+++ b/pages/settings/PageSettingsFroniusInverter.qml
@@ -61,8 +61,8 @@ Page {
 			}
 
 			ListRadioButtonGroup {
-				//% "Power limiting"
-				text: qsTrId("page_settings_fronius_inverter_power_limiting")
+				//% "Dynamic power limiting"
+				text: qsTrId("page_settings_fronius_inverter_dynamic_power_limiting")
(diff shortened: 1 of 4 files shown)


pages/settings/PageSettingsFroniusInverter.qml
```diff
@@ -61,8 +61,8 @@ Page {
 			}
 
 			ListRadioButtonGroup {
-				//% "Power limiting"
-				text: qsTrId("page_settings_fronius_inverter_power_limiting")
+				//% "Dynamic power limiting"
+				text: qsTrId("page_settings_fronius_inverter_dynamicpowerlimiting")
```
MikeTrahearn-Qinetic, line 65: query on the id: page_settings_fronius_inverter_dynamicpowerlimiting > page_settings_fronius_inverter_dynamic_power_limiting? Not sure what the general preference is here, but tapping for consistency.

chriadam: Oops, thanks, was just a mistake.  Fixed.

blammit (approved):
Could use "Fixes" instead of "Contributes to", if it's the only fix for this task? Then it would auto-close the issue as well.

PR #1941, opened 2025-02-27
Fix loading of Device Info page from PageAcInModel

blammit:
VisibleItemModel does not have an append() method, so add the item using a footer instead.


blammit:
@chriadam could you check if this fixes the issue you mentioned?

chriadam:
Yes, it does.  Thanks!

chriadam (approved):

chriadam:
Updated with mock data value for testing as discussed

Issue #1756, opened 2024-11-22
Levels, Notifications and Settings buttons visibility on wide screen


dirkjanfaber:
Granted that the monitor is extremely wide when I tested this in full screen mode, but noticing that Levels, Notifications and Settings buttons are still visible in the lower left corner when accessing Settings.

https://github.com/user-attachments/assets/8c571579-50d9-4b33-9ad0-1e97ed097e59

Tested this on a ultra wide screen, running the mac build. But I expect this also to be happening on long phones running the gui full screen in landscape mode.

mpvader:
discussed in todays gui-v2 meeting: most likely this is a regression and also not difficult to fix.

chriadam:
Merged.  Closing issue as fixed.

PR #1943, opened 2025-02-27
Clip main view on desktop as well as wasm

chriadam:
Otherwise, the NavBar is visible to the left of the content on ultrawide screens after opening a drilldown view.

For testing, you may run on desktop and resize the window horizontally larger, then open a drilldown from either the overview or settings pages.

We don't want to clip on GX devices, as this will incur a performance overhead, and offscreen content is not visible on that platform due to the physical screen.

Contributes to issue #1756

### DIFF
--- a/Main.qml
+++ b/Main.qml
@@ -88,7 +88,7 @@ Window {
 	Loader {
 		id: guiLoader
 
-		clip: Qt.platform.os == "wasm"
+		clip: Qt.platform.os == "wasm" || Global.isDesktop
 		width: Theme.geometry_screen_width
 		height: Theme.geometry_screen_height
 		anchors.horizontalCenter: parent.horizontalCenter


chriadam:
Note that with this fix, they both vanish rather suddenly.  A more comprehensive fix is attached, but considering it will only be seen in very strange circumstances, I believe the "comprehensive" fix would imply too much overhead for the gain (i.e. all the animators etc being constructed, etc).

[navbar-swipeview-fadeout.diff.txt](https://github.com/user-attachments/files/19006533/navbar-swipeview-fadeout.diff.txt)

pages/MainView.qml
```diff
@@ -240,6 +240,7 @@ Item {
 		color: root.backgroundColor
 		opacity: 0
 		model: swipeView ? swipeView.contentModel : null
+		visible: pageStack.swipeViewVisible
```
blammit, line 243: The issue doesn't occur on wasm due to the `clip: Qt.platform.os == "wasm"` in Main.qml. How about we just add `Global.isDesktop` to that condition? Otherwise, I suggest we remove that `clip` and fix it a different way on both wasm and desktop instead.

chriadam: Thanks, done

blammit (approved):
LGTM. Could juse use "Fixes #1756" in the commit message if you want to auto-close it?

Main.qml
```diff
@@ -88,7 +88,7 @@ Window {
 	Loader {
 		id: guiLoader
 
-		clip: Qt.platform.os == "wasm"
+		clip: Qt.platform.os == "wasm" || Global.isDesktop
```
MikeTrahearn-Qinetic, line 91: ===

PR #1948, opened 2025-02-28
Always show SOC or temperature in center of brief page

blammit:
Note this requires the local setting `/Settings/Gui2/BriefView/CenterService`, which hasn't been added on device yet. So to make it work on a device you need to add the setting, e.g. `dbus -y com.victronenergy.settings /Settings AddSettings '%[{"path": "Gui2/BriefView/CenterService", "default": ""}]'`.


components/listitems/ListBriefCenterDetails.qml
```diff
@@ -0,0 +1,100 @@
...
+				value: device.serviceUid,
+				//% "Temperature services"
+				section: qsTrId("settings_briefview_center_temperature_services")
+			})
+			if (selectedIndex < 0 && device.serviceUid === centerService.value) {
+				selectedIndex = i
+			}
+		}
+		Global.pageManager.pushPage(deviceOptionsComponent, {
+			"title": text,
```
chriadam, line 41: maybe just `title: text` rather than quoting title, for consistency with the other two params

components/BriefCenterDisplay.qml
```diff
@@ -0,0 +1,120 @@
...
+			Row {
+				anchors.horizontalCenter: parent.horizontalCenter
+				spacing: Theme.geometry_briefPage_centerGauge_centerText_horizontalSpacing
+
+				QuantityLabel {
+					valueColor: Theme.color_briefPage_battery_value_text_color
+					unitColor: Theme.color_briefPage_battery_unit_text_color
+					font.pixelSize: Theme.font_briefPage_battery_voltage_pixelSize
+					unit: VenusOS.Units_Volt_DC
+					value: Global.system.battery.voltage
```
chriadam, line 97: in mock mode, this was always "--" for some reason, but when I tested against the office cerbo (running a demo) it showed the actual value.

components/listitems/ListBriefCenterDetails.qml
```diff
@@ -0,0 +1,100 @@
+/*
+** Copyright (C) 2025 Victron Energy B.V.
+** See LICENSE.txt for license information.
+*/
+
+import QtQuick
+import Victron.VenusOS
+
+ListNavigation {
+	id: root
```
chriadam, line 10: On device, I did `dbus -y com.victronenergy.settings /Settings AddSettings '%[{"path": "Gui2/BriefView/CenterService", "default": ""}]'` and then I connected via WASM to office cerbo.  I then set all levels to "none" and selected the Ruuvi tag (temp sensor) as the center value.  Then, the Brief page is not very understandable, since the (single) full-circular-gauge doesn't have a label, and its value doesn't correspond to the center value (the temperature).

I would suggest that if the center value is a temperature sensor and all levels are "none", we force the first level to be the system battery.

chriadam: (or, maybe, hide the single gauge altogether in that case?)

blammit: Hmm yes ... I'll check double what should be done in this case.

components/BriefCenterDisplay.qml
```diff
@@ -0,0 +1,120 @@
...
+		uid: Global.systemSettings.serviceUid + "/Settings/Gui2/BriefView/CenterService"
+	}
+
+	VeQuickItem {
+		id: temperature
+		uid: root._useTemperature ? root._serviceUid + "/Temperature" : ""
+	}
+
+	Row {
+		anchors {
```
MikeTrahearn-Qinetic, line 31: I would normally flag using anchors inside Row*Layouts* since this is always problematic. So just checking the use of anchors in a Row *Positioner* in this case. I can see the intention, but is it the best approach?

components/Utils.js
```diff
@@ -117,7 +117,7 @@ function formatGeneratorRuntime(seconds) {
 }
 
 function formatBatteryTimeToGo(seconds) {
-	if (seconds == 0) {
+	if (isNaN(seconds) || seconds == 0) {
```
MikeTrahearn-Qinetic, line 120: ===

components/BriefCenterDisplay.qml
blammit: Hm, I am not sure, I can see the voltage appearing in mock mode.

components/listitems/ListBriefCenterDetails.qml
blammit: Oops, fixed

components/listitems/ListBriefCenterDetails.qml
blammit: I asked about this behaviour in Slack, and the conclusion was that it is fine for now, and we can change it later if needed.

components/Utils.js
blammit: Done

components/BriefCenterDisplay.qml
blammit: Sure, what alternative approaches were you thinking of? I commonly use anchors inside positioners in this manner, as it doesn't conflict with the positioner's layout direction, and is more readable than using x/y positioning.

MikeTrahearn-Qinetic: Whatever you are comfortable doing.

components/listitems/ListBriefCenterDetails.qml
chriadam: Ok :-)

chriadam (approved):
LGTM

PR #1951, opened 2025-03-03
Mock: cycle configurations with , + .

chriadam:
Previously, the left/right keys were used to cycle between configurations.  With KeyNavigation support coming, we need to use different keys to trigger mock configuration cycling.

Use comma and period keys (i.e. < and > keys) for this.

### DIFF
--- a/data/mock/config/MockDataSimulator.qml
+++ b/data/mock/config/MockDataSimulator.qml
@@ -88,13 +88,15 @@ QtObject {
 				event.accepted = true
 			}
 			break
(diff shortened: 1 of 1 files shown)


chriadam:
This one was actually in response to https://github.com/victronenergy/gui-v2/pull/1948 where for some reason I couldn't cycle configurations with left+right keys (instead that would navigate the swipe view...)

blammit (approved):
Feel free to remove the Key_Left and Key_Right bindings altogether, if you like.

Issue #1710, opened 2024-11-08
Improve error messaging / information in VRM use case


mpvader:
The system in this screenshot is offline, yet gui-v2 now says connected.

i think I know why, (even though wasm cannot connect to the device itself., it can connect to mqtt cloud) so need to come up with something for this.

![IMG_3435](https://github.com/user-attachments/assets/c8af516f-5c51-444e-8dbd-681ebd2c557e)

there are two related issues:
1) perhaps vrm should show some warning or something about system being offline before hand, however we don’t want to delay making it possible to open gui-v2 by first doing a mqtt online check..
2) the wording and terms shown during that connection process could be made less technical
3) the whole loading process when loading locally can also be better

chriadam:
I have created the above PR which exposes the heartbeat state as additional error information on the splash view.  It also returns the user to the splash view if the device heartbeat becomes inactive, and then reloads the UI once heartbeat returns.  Let me know if there are some changes wanted to this (e.g. perhaps returning to the splash view is a bit abrupt?)

mpvader:
Hi @chriadam can you share a screenrecording showing the result?

chriadam:
Here's what it looks like when connecting to a VRM installation where the remote device isn't available (i.e. isn't currently "realtime" connected to VRM, and isn't sending heartbeats to VRM):

https://github.com/user-attachments/assets/6c0ac0ef-b465-4b37-b674-fca701ccea20

And here is what it looks like when an existing wasm gui-v2 instance which is connected to VRM loses its heartbeat and then regains it after some time (i.e. because I rebooted the cerbo in the background):

https://github.com/user-attachments/assets/68ede07a-dd8f-42ff-b691-feb8056de791

mpvader:
good improvement, lets put this in with one change: change [4] Connected to ... to :

`[4] Connection to the device has been lost, awaiting reconnection`

chriadam:
Merged change, closing as fixed.

PR #1952, opened 2025-03-03
Display VRM device heartbeat information in splash view

chriadam:
Even if we manage to connect to the VRM broker, this doesn't mean that we will receive data from the device (e.g. if the device itself has lost internet connectivity or power, and thus is not connected to VRM).

In this case, the splash screen will simply tell the user that they are Connected (implicitly: to VRM) but will never progress to Ready state since no data from the device will be received.

This commit ensures that we tell the user that the heartbeat is missing in this case.

It also ensures we return to the splash view if we are connected to VRM and the heartbeat becomes inactive, since we don't want to continue to display stale data to the user.

Contributes to issue #1710

### DIFF
--- a/Global.qml
+++ b/Global.qml
@@ -19,6 +19,10 @@ QtObject {
 	property var allDevicesModel
 	property bool applicationActive: true
 
+	readonly property bool backendReady: BackendConnection.state === BackendConnection.Ready
+		&& (Qt.platform.os !== "wasm"
+			|| !BackendConnection.vrm
+			|| BackendConnection.heartbeatState !== BackendConnection.HeartbeatInactive)
 	readonly property string fontFamily: _defaultFontLoader.name
 	readonly property string quantityFontFamily: _quantityFontLoader.name
 	property var dialogLayer
@@ -59,6 +63,9 @@ QtObject {
 
 	readonly property int int32Max: _intValidator.top
 	readonly property int int32Min: _intValidator.bottom
+
+	property bool backendReadyLatched
+	onBackendReadyChanged: if (backendReady) backendReadyLatched = true
 
 	signal aboutToFocusTextField(var textField, var textFieldContainer, var flickable)
 	signal keyPressed(var event)
--- a/components/SplashView.qml
+++ b/components/SplashView.qml
@@ -118,7 +118,7 @@ Rectangle {
 			onRunningChanged: {
 				if (running) {
 					logoIconFadeOutAnim.running = true
-				} else if (BackendConnection.state === BackendConnection.Ready) {
+				} else if (Global.backendReady) {
 					animatedLogo.playing = true
 				}
 			}
@@ -200,7 +200,9 @@ Rectangle {
 
 			states: State {
 				name: "alarm"
-				when: BackendConnection.state >= BackendConnection.Disconnected || mqttErrorLabel.text.length > 0
+				when: BackendConnection.state >= BackendConnection.Disconnected
+					|| mqttErrorLabel.text.length > 0
+					|| mqttHeartbeatLabel.visible
 				PropertyChanges {
 					target: alarmIconContainer
 					opacity: 1.0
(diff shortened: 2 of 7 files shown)


components/SplashView.qml
```diff
@@ -273,6 +284,49 @@ Rectangle {
...
+
+			visible: false
+			width: parent.width
+			horizontalAlignment: Text.AlignHCenter
+			font.pixelSize: Theme.font_splashView_progressText_size
+			color: Theme.color_font_secondary
+			wrapMode: Text.Wrap
+			text: "[" + BackendConnection.heartbeatState + "] "
+			    + (BackendConnection.heartbeatState === BackendConnection.HeartbeatMissing
+				  //% "Device possibly lost connectivity to VRM"
```
blammit, line 299: How about "may have" instead of "possibly"?

components/SplashView.qml
```diff
@@ -238,14 +244,19 @@ Rectangle {
 				: BackendConnection.state === BackendConnection.Connected ? qsTrId("splash_view_connected")
 				  //% "Connected, receiving broker messages"
 				: BackendConnection.state === BackendConnection.Initializing ? qsTrId("splash_view_initializing")
-				  //% "Connected, loading user interface"
-				: BackendConnection.state === BackendConnection.Ready ? qsTrId("splash_view_ready")
+				: BackendConnection.state === BackendConnection.Ready
+					? (BackendConnection.vrm && BackendConnection.heartbeatState !== BackendConnection.HeartbeatActive)
+						  //% "Connected to VRM, awaiting device"
```
blammit, line 249: What does "awaiting device" mean here, is it waiting for the device to respond to VRM? If so could it be "awaiting device response" or "waiting for device connection" instead?

components/SplashView.qml
chriadam: Done

components/SplashView.qml
chriadam: Well, it's awaiting a particular "heartbeat" topic value update from the device to VRM.  I think simply "awaiting device" is descriptive enough, personally.

components/SplashView.qml
```diff
@@ -215,7 +217,11 @@ Rectangle {
 			CP.ColorImage {
 				anchors.centerIn: parent
 				source: "qrc:/images/icon_warning_24.svg"
-				color: Theme.color_red
+				color: (BackendConnection.state >= BackendConnection.Disconnected
+						|| mqttErrorLabel.text.length > 0
+						|| (BackendConnection.vrm && BackendConnection.heartbeatState === BackendConnection.HeartbeatInactive))
+					? Theme.color_red
```
MikeTrahearn-Qinetic, line 223: Should these be Theme.color_critcal / warning?

chriadam: Good point, thanks, fixed

components/SplashView.qml
blammit: Okay, no worries

blammit (approved):

Global.qml
```diff
@@ -19,6 +19,10 @@ QtObject {
 	property var allDevicesModel
 	property bool applicationActive: true
 
+	readonly property bool backendReady: BackendConnection.state === BackendConnection.Ready
+		&& (Qt.platform.os != "wasm"
```
MikeTrahearn-Qinetic, line 23: !== ;-)

components/SplashView.qml
```diff
@@ -215,7 +217,11 @@ Rectangle {
 			CP.ColorImage {
 				anchors.centerIn: parent
 				source: "qrc:/images/icon_warning_24.svg"
-				color: Theme.color_red
+				color: (BackendConnection.state >= BackendConnection.Disconnected
+						|| mqttErrorLabel.text.length > 0
```
MikeTrahearn-Qinetic, line 221: `>>>` or `>>=`
is a new operator which means "greater than (or equal to) operation ensuring no type co-ercion, returning nullish on incompatible operands"
- haha just kidding

src/backendconnection.h
```diff
@@ -61,6 +62,14 @@ class BackendConnection : public QObject
 	};
 	Q_ENUM(State)
 
+	// Same as VeQItemMqttProducer::HeartbeatState
```
MikeTrahearn-Qinetic, line 65: Would there be a problem if these got out of date? Or is there a way to bring them in here directly?

Global.qml
chriadam: done

src/backendconnection.h
chriadam: Well, if they change in the producer, we will need to change our handling here in the BackendConnection, so I don't know that there's much value in trying to couple them more tightly.

chriadam:
Rebased and added the translation string as discussed in https://github.com/victronenergy/gui-v2/issues/1710

blammit (approved):

PR #1957, opened 2025-03-06
Support showing individual batteries and tanks in brief page centre gauges

blammit:
This needs local settings `/Settings/Gui2/BriefView/Level/<0-3>`, which may not yet be available on the build. So to make it work on a device you need to add the setting, e.g. `dbus -y com.victronenergy.settings /Settings AddSettings '%[{"path": "Gui2/BriefView/Level/0", "default": ""}]'` and repeat for level 1-3.

### DIFF
--- a/components/GaugeModel.qml
+++ b/components/GaugeModel.qml
@@ -14,84 +14,119 @@ import Victron.Gauges
 ListModel {
 	id: root
 
-	function gaugesToDisplay() {
-		let gaugeTypes = []
-		const preferredGaugeTypes = Global.systemSettings.briefView.centralGauges.preferredOrder
-
-		for (let i = 0; i < preferredGaugeTypes.length; ++i) {
-			if (gaugeTypes.indexOf(preferredGaugeTypes[i]) >= 0) {
-				// Do not show more than one gauge for the same tank type.
-				continue
-			} else if (preferredGaugeTypes[i] === VenusOS.Tank_Type_Battery) {
-				// Assume battery is always available.
-				gaugeTypes.push(VenusOS.Tank_Type_Battery)
-			} else if (preferredGaugeTypes[i] !== VenusOS.Tank_Type_None) {
-				// If there is a tank of this type, then add the type to the displayed gauges. If
-				// not, add another tank that is available on the system. Otherwise, if the user has
-				// not selected any preferred tanks, but does not have any of the tanks from the
-				// default preferences, then the center display would not show any tanks at all.
-				const preferredTankModel = Global.tanks.tankModel(preferredGaugeTypes[i])
-				if (preferredTankModel.count > 0) {
-					gaugeTypes.push(preferredGaugeTypes[i])
-				} else {
-					for (const tankModel of Global.tanks.allTankModels) {
-						if (tankModel.count > 0 && gaugeTypes.indexOf(tankModel.type) < 0) {
-							gaugeTypes.push(tankModel.type)
-							break
-						}
+	function _loadDefaultGauges() {
+		let gauges = []
+		if (Global.tanks.totalTankCount === 0) {
+			// There are no tanks, so there are no gauges to be shown.
+			return []
+		} else {
+			// Show the system battery, followed by gauges with aggregated tank levels for whichever
+			// tank types are available on the system.
+			gauges.push({ centerGaugeType: VenusOS.BriefView_CentralGauge_SystemBattery, value: "" })
+			for (const tankModel of Global.tanks.allTankModels) {
+				if (tankModel.count > 0) {
+					gauges.push({ centerGaugeType: VenusOS.BriefView_CentralGauge_TankAggregate, value: tankModel.type })
+					if (gauges.length >= Theme.geometry_briefPage_centerGauge_maximumGaugeCount) {
+						break
 					}
 				}
 			}
 		}
-		return gaugeTypes
+		return gauges
 	}
 
 	// Rebuild the gauge model, based on the preferred gauges and the tanks available.
 	function _reset() {
-		const gaugeTypes = gaugesToDisplay()
-		if (gaugeTypes != _gaugeObjects.model) {
-			_gaugeObjects.model = []
-
+		let useDefaultGauges = false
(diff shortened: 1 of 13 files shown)


components/GaugeModel.qml
```diff
@@ -14,104 +14,221 @@ import Victron.Gauges
...
+	function _loadDefaultGauges() {
+		let gauges = []
+		if (Global.tanks.totalTankCount === 0) {
+			// There are no tanks, so there are no gauges to be shown.
+			return []
+		} else {
+			// Show the system battery, followed by gauges with aggregated tank levels for whichever
+			// tank types are available on the system.
+			gauges.push({ centerGaugeType: VenusOS.BriefView_CentralGauge_SystemBattery, value: "" })
+			for (const tankModel of Global.tanks.allTankModels) {
```
MikeTrahearn-Qinetic, line 26: Nice!

components/GaugeModel.qml
```diff
@@ -14,104 +14,221 @@ import Victron.Gauges
...
-			clear()
-			for (let i = 0; i < gaugeTypes.length; ++i) {
-				append(Object.assign({}, Gauges.tankProperties(gaugeTypes[i]), { tankType: gaugeTypes[i], level: 0, value: 0 }))
+			_gaugeObjects.model = []
+			let hasBatteryId = false
+			for (let i = 0; i < gauges.length; ++i) {
+				if (!hasBatteryId && gauges[i].centerGaugeType === VenusOS.BriefView_CentralGauge_BatteryId) {
+					hasBatteryId = true
+				}
+				set(i, {
```
MikeTrahearn-Qinetic, line 56: it wasn't immediately clear what set (and remove) functions were until I realised they are on the ListItem. Perhaps consider qualifying these.

components/listitems/core/ListRadioButton.qml
```diff
@@ -12,10 +12,16 @@ ListItem {
 
 	property alias checked: radioButton.checked
 	property alias radioButton: radioButton
+	property string secondaryText
 
 	interactive: true
 
 	content.children: [
+		SecondaryListLabel {
```
MikeTrahearn-Qinetic, line 20: Is this ok in all use cases of a ListRadioButton? Does it interfere with any primary text?

components/GaugeModel.qml
blammit: Qualifier added.

components/GaugeModel.qml
```diff
@@ -14,104 +14,221 @@ import Victron.Gauges
...
+				level: gauge.level,
+				value: value
+			})
 		}
 	}
 
-	property Instantiator _gaugeObjects: Instantiator {
+	readonly property VeQuickItem _batteriesItem: VeQuickItem {
+		property bool active
+
```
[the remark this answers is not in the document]
blammit: It's nice to have a newline after properties, though I'm not entirely sure it's necessary when the whole block only has a couple of lines. Will be nice to use qmllint or such one day to align these things...

components/listitems/core/ListRadioButton.qml
blammit: It's not visible when there is no text, so it shouldn't interefere.

components/GaugeModel.qml
MikeTrahearn-Qinetic: You can also do `Global.tanks.allTankModels.forEach(function(tankModel) { ... })` as well.

MikeTrahearn-Qinetic (approved):

blammit:
Will wait for some further backend verification before merging this.

chriadam:
I also just saw the following: "qml: Unknown tank type NaN" again, no line number info :-/

pages/settings/PageSettingsDisplayBrief.qml
```diff
@@ -10,35 +10,203 @@ import Victron.Gauges
...
+					levelIndex: index,
+					title: text,
+					optionModel: optionModel,
+					currentIndex: root._findSelectedLevel(optionModel, centerGaugeType, centerGaugeValue),
+				})
+			}
+
+			VeQuickItem {
+				id: tankFluidType
+				uid: levelDelegate.centerGaugeType === VenusOS.BriefView_CentralGauge_TankUid
```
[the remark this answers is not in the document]
blammit: Thanks, should be fixed now, and while fixing this I also updated the PR to save the uid in a backend-neutral format like 'com.victronenergy.tank/0' instead of instead of the backend-specific uid like  'dbus/com.victronenergy.tank.tty0' or 'mqtt/tank/1'. I also added another commit to fix this save value for the Brief centre temperature service as well.

pages/settings/PageSettingsDisplayBrief.qml
```diff
@@ -10,35 +10,206 @@ import Victron.Gauges
...
+				}
+				break
+			case VenusOS.BriefView_CentralGauge_TankId:
+				const tankServiceId = selectedCenterGaugeValue
+				if (levelOptionModel[i].value === tankServiceId) {
+					return i
+				}
+				break
+			default:
+				//: No option has been selected
```
chriadam, line 129: typo: colon

pages/settings/PageSettingsDisplayBrief.qml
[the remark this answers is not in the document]
chriadam: Yep, I ran on wasm and device simultaneously, and can confirm that changes to the levels settings on one is immediately reflected on the other, and vice versa, including for individual tanks or batteries.  Nice work.

One possible issue: the hydropack battery SoC always shows as "--" on both device and wasm, with the demo I'm currently running on the office cerbo:

victron-gauge-hydropack-battery

Also, I still see a transient warning: `qml: Unknown tank type NaN` but that shouldn't block this PR I think.

chriadam (approved):
LGTM, one nitpick remaining.

pages/settings/PageSettingsDisplayBrief.qml
blammit: Thanks, fixed.

PR #1958, opened 2025-03-06
Add SystemBatteryDeviceModel and SystemBatteryDelegate

blammit:
Do some refactoring to avoid duplicating model/delegate code between BatteryListPage and PageSettingsBatteries.


components/SystemBatteryDelegate.qml
```diff
@@ -0,0 +1,141 @@
...
+	required property SystemBatteryDeviceModel.Battery device
+	readonly property string serviceType: BackendConnection.serviceTypeFromUid(device.serviceUid)
+
+	height: Theme.geometry_batteryListPage_item_height
+
+	RowLayout {
+		width: parent.width
+		height: Theme.geometry_batteryListPage_item_height
+
+		Column {
```
MikeTrahearn-Qinetic, line 23: Mixing Positioners and Layouts - I assume this is intentional

components/SystemBatteryDelegate.qml
```diff
@@ -0,0 +1,141 @@
...
+				}
+			}
+		}
+
+		CP.ColorImage {
+			Layout.rightMargin: Theme.geometry_listItem_content_horizontalMargin
+			source: "qrc:/images/icon_arrow_32.svg"
+			rotation: 180
+			color: pressArea.containsPress ? Theme.color_listItem_down_forwardIcon : Theme.color_listItem_forwardIcon
+			opacity: pressArea.enabled ? 1 : 0
```
MikeTrahearn-Qinetic, line 113: Assume visible isn't used here because it would interfere with the layout?

components/SystemBatteryDeviceModel.qml
```diff
@@ -0,0 +1,101 @@
...
+		uid: Global.system.serviceUid + "/Batteries"
+		onValueChanged: {
+			if (!isValid) {
+				root.deleteAllAndClear()
+				return
+			}
+
+			const batteryList = value
+			let batteryUids = []
+			let i
```
MikeTrahearn-Qinetic, line 26: move let i into the respective for loops

pages/settings/PageSettingsBatteries.qml
```diff
@@ -50,155 +51,14 @@ Page {
...
-						Device {
-							id: genericDevice
-							serviceUid: batteryDelegate.device.instance >= 0 ? batteryDelegate.device.serviceUid : ""
-						}
-					}
-				}
-			}
+		footer: SettingsColumn {
+			width: parent ? parent.width : 0
+			topPadding: Theme.geometry_gradientList_spacing
```
MikeTrahearn-Qinetic, line 61: does the container not have a spacing property for this purpose?

components/SystemBatteryDelegate.qml
blammit: Yes, that's right

components/SystemBatteryDelegate.qml
blammit: Yes, the layout fillWidth is convenient for doing the layout here.

pages/settings/PageSettingsBatteries.qml
blammit: Ah yes, this should bind to `spacing` instead. Fixed.

components/SystemBatteryDeviceModel.qml
blammit: Qt Creator complains about the `i` scoping? Is it still a valid issue?

blammit:
Rebased.

components/SystemBatteryDelegate.qml
MikeTrahearn-Qinetic: I'll bet you money that if we got rid of all the fixed sizing everywhere, it would be convenient everywhere.

components/SystemBatteryDelegate.qml
MikeTrahearn-Qinetic: Using Layout Types would adjust the layout for invisible children... just pinning that here.

components/SystemBatteryDeviceModel.qml
MikeTrahearn-Qinetic: Hmm so it does. This is a surprising to me - but I did check it out further. Even though the i variables are technically scoped to their respective for loop blocks, the Javascript engine sees that within the outer block, the variable i has been declared twice, which is not allowed.

MikeTrahearn-Qinetic: Perhaps then may be name it something like `batterIndex` to demonstrate its purpose. 
`let i` by itself isn't very descriptive.

MikeTrahearn-Qinetic: Question though: can you collapse the two for loops into one, or is there a dependency on one loop finishing before the next starts?

components/SystemBatteryDeviceModel.qml
```diff
@@ -0,0 +1,103 @@
...
+		property real soc: NaN
+		property real temperature: NaN
+		property int timetogo: 0
+		property real voltage: NaN
+		readonly property int mode: VenusOS.battery_modeFromPower(power)
+
+		name: customName
+	}
+
+	readonly property Component _batteryComponent: Component {
```
MikeTrahearn-Qinetic, line 89: Is there a reason this is a Component property rather than an inline components above? Do you need to call .createObject() on it I guess? And readonly?

blammit: Yes it's because we need to create instances of it from the JS function, using createObject(). It's readonly since it doesn't need to be modified.

components/SystemBatteryDeviceModel.qml
blammit: It could be `batteryIndex` though I think `i` is okay as a generic index name. The existing `batteryIndex` is used specifically for "the index of the battery object in the list.

There's a dependency on the first loop finishing first, to remove the dead entries before updating the list.

components/SystemBatteryDelegate.qml
```diff
@@ -0,0 +1,141 @@
...
+				width: parent.width
+				horizontalAlignment: Text.AlignRight
+				elide: Text.ElideRight
+				font.pixelSize: Theme.font_size_body2
+				color: Theme.color_listItem_secondaryText
+				visible: !isNaN(root.device.power)
+				text: {
+					const modeText = VenusOS.battery_modeToText(root.device.mode)
+					if (root.device.mode === VenusOS.Battery_Mode_Discharging
+							&& root.device.timetogo > 0) {
```
chriadam, line 99: should timetogo be camel-cased as timeToGo?

chriadam (approved):
One nitpick but LGTM

components/SystemBatteryDelegate.qml
blammit: We could, it's just that this makes it convenient for updating the property directly from the "timetogo" json value

blammit: But maybe it's better to use the camel case for consistency... will update

CMakeLists.txt
```diff
@@ -263,6 +263,8 @@ set (VENUS_QML_MODULE_SOURCES
     components/SwipePageModel.qml
     components/SwipeViewPage.qml
     components/SystemReason.qml
+    components/SystemBatteryDeviceModel.qml
```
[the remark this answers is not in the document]
blammit: Fixed

PR #1960, opened 2025-03-07
Add Makefile to build

mr-manuel:
@jhofstee

### DIFF
--- a/scripts/build-wasm.sh
+++ b/scripts/build-wasm.sh
@@ -75,7 +75,7 @@ fi
 # Prepare files that are needed for the GUIv2 to run
 mkdir -p artifacts/wasm
 mv venus-gui-v2.{html,js,wasm} qtloader.js artifacts/wasm/
-cp ../images/victronenergy.svg ../LICENSE.txt artifacts/wasm/
+cp ../images/victronenergy.svg ../LICENSE.txt ../.github/patches/Makefile artifacts/wasm/
 mv artifacts/wasm/venus-gui-v2.html artifacts/wasm/index.html
 grep -q -E '^var createQtAppInstance' artifacts/wasm/venus-gui-v2.js
 sed -i "s%^var \(createQtAppInstance\)%window.\1%" artifacts/wasm/venus-gui-v2.js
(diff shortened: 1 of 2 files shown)


mr-manuel:
@chriadam can you merge that? I have no permissions to do so.

scripts/build-wasm.sh
```diff
@@ -75,7 +75,7 @@ fi
 # Prepare files that are needed for the GUIv2 to run
 mkdir -p artifacts/wasm
 mv venus-gui-v2.{html,js,wasm} qtloader.js artifacts/wasm/
-cp ../images/victronenergy.svg ../LICENSE.txt artifacts/wasm/
+cp ../images/victronenergy.svg ../LICENSE.txt ../.github/patches/Makefile artifacts/wasm/
```
chriadam, line 78: does the Makefile need to be modified to remove the gzip and sha256sum steps, since we do those here in this build script already?

mr-manuel: Good catch, yes.

jhofstee: It doesn't have to be adjusted to work, since the venus-gui-v2.wasm.gz doesn't depend on its input file. It can't hurt to clean up the Makefile though.

chriadam: I haven't been able to test personally, since I am not building an image, and `make` isn't available on device via opkg (as far as I can tell) in order to test on a live filesystem.  So, I hope @jhofstee can verify your changes.

jhofstee: I pushed a patch to this branch, which basically reduces the Makefile to just always copy the files to the right place. Feel free to make that a "fixup" of the previous commit. The commit message is just there to explain a bit how the Makefile works, but isn't really interesting to keep in the gui-v2 history.

chriadam (approved):

mr-manuel:
@chriadam can you merge it? I have no permissions to do so.

blammit:
@mr-manuel for future reference, can you wrap the commit messages to 72 chars?

PR #1963, opened 2025-03-10
Rename VeQuickItem 'isValid' to 'valid'

### LINKED ISSUE #1820: Rename VeQuickItem 'isValid' to 'valid'
Originally, VeQuickItem's `isValid` property was named as such for consistency with other properties in that type. However, this naming has led to various porting errors, and it's also inconsistent with general QML property naming conventions that avoids starting names with `is`.

So, rename the property to `valid`.

MikeTrahearn-Qinetic:
705 manually verified edits (this was not a find-replace)
Fixes #1820

See also https://github.com/victronenergy/veutil/pull/7 for the submodule change in veutils to complete this fix.


blammit (approved):
Excellent :)

PR #1972, opened 2025-03-12
Fix User Access Password can be revealed

### LINKED ISSUE #1971: User Access Password is Revealed
When entering the user access password and pressing the confirm button, the password is erroneously revealed momentarily before the page is closed.
This can be reproduced in Mock Mode.

MikeTrahearn-Qinetic:
- Remove ListTextField enabled binding which causes the change of userHasWriteAccess to be reflected in the SecondaryListLabel (ListTextField's interactive binding is sufficient)
- Update SecondaryListLabel visible binding to use interactive

Fixes #1971

### DIFF
--- a/components/listitems/core/ListTextField.qml
+++ b/components/listitems/core/ListTextField.qml
@@ -190,7 +190,6 @@ ListItem {
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


blammit (approved):
Great :)

Issue #1973, opened 2025-03-12
Device list doesn’t use custom name for digital inputs


mpvader:
Reported here:

https://community.victronenergy.com/t/venus-3-60-25-gui-v2-shows-digital-type-not-digital-input-name/22201

And reproduced, onscreen - haven’t checked wasm - running Venus OS v3.60~39 which is gui-v2 v1.1.5.

chriadam:
Merged to main.  Closing as fixed.

PR #1977, opened 2025-03-13
Show custom name of digital input in device list

chriadam:
Contributes to issue #1973

### DIFF
--- a/pages/settings/devicelist/PageDigitalInput.qml
+++ b/pages/settings/devicelist/PageDigitalInput.qml
@@ -12,15 +12,25 @@ Page {
 
 	property string bindPrefix
 
-	title: VenusOS.digitalInput_typeToText(inputType.value)
+	title: device.customName || VenusOS.digitalInput_typeToText(inputType.value)
 
 	VeQuickItem {
 		id: inputType
 		uid: root.bindPrefix + "/Type"
 	}
 
+	Device {
+		id: device
+		serviceUid: root.bindPrefix
+	}
+
 	GradientListView {
 		model: VisibleItemModel {
+			ListText {
+				text: CommonWords.type
+				secondaryText: VenusOS.digitalInput_typeToText(inputType.value)
+			}
+
 			ListText {
 				text: CommonWords.state
 				dataItem.uid: root.bindPrefix + "/State"
--- a/pages/settings/devicelist/delegates/DeviceListDelegate_digitalinput.qml
+++ b/pages/settings/devicelist/delegates/DeviceListDelegate_digitalinput.qml
@@ -9,7 +9,7 @@ import Victron.VenusOS
 DeviceListDelegate {
 	id: root
 
-	text: VenusOS.digitalInput_typeToText(type.value)
+	text: root.device.customName || VenusOS.digitalInput_typeToText(type.value)
 	secondaryText: VenusOS.digitalInput_stateToText(state.value)
 
 	onClicked: {


pages/settings/devicelist/PageDigitalInput.qml
```diff
@@ -20,6 +20,10 @@ Page {
 	}
 
 	GradientListView {
+		header: PrimaryListLabel {
+			visible: text !== root.title
```
chriadam, line 24: I can't seem to test this with mock mode, as setting the custom name to an empty string seems to "disconnect" the device for some reason.

pages/settings/devicelist/delegates/DeviceListDelegate_digitalinput.qml
```diff
@@ -9,12 +9,12 @@ import Victron.VenusOS
...
 	id: root
 
-	text: VenusOS.digitalInput_typeToText(type.value)
+	text: root.device.customName.length > 0 ? root.device.customName : VenusOS.digitalInput_typeToText(type.value)
 	secondaryText: VenusOS.digitalInput_stateToText(state.value)
 
 	onClicked: {
 		Global.pageManager.pushPage("/pages/settings/devicelist/PageDigitalInput.qml",
-				{ bindPrefix : root.device.serviceUid })
+				{ title: root.text, bindPrefix: root.device.serviceUid })
```
MikeTrahearn-Qinetic, line 17: Just checking if the title as pushed here into the new page, is an assignment or a binding - and what the intention vs. reality is?

chriadam: In this case, it's a value.  Maybe we should try to bind it somehow; I'm not sure...

pages/settings/devicelist/PageDigitalInput.qml
blammit: I would just add another `ListText` to the main model like `ListText { text: "Type"; text: typeToText(type) }`. This is what we do for other pages where we need to show device types, roles etc.

pages/settings/devicelist/delegates/DeviceListDelegate_digitalinput.qml
blammit: Instead of doing this you can change PageDigitalInput.qml to show the correct title using a `Device`, like we do in PageVeBus and so on. E.g.

```
// PageDigitalInput.qml
Page {
     id: root

     title: device.customName || input_typeToText(type)

     Device { id: device; serviceId: root.bindPrefx }
```

(Really, I would prefer to get rid of all the cases where we pass the page `title` in the push parameters; it's just convenient and a way to avoid making another qsTrId() call in the destination Page with the same text.)

pages/settings/devicelist/delegates/DeviceListDelegate_digitalinput.qml
```diff
@@ -9,12 +9,12 @@ import Victron.VenusOS
 DeviceListDelegate {
 	id: root
 
-	text: VenusOS.digitalInput_typeToText(type.value)
+	text: root.device.customName.length > 0 ? root.device.customName : VenusOS.digitalInput_typeToText(type.value)
```
blammit, line 12: Can simplify to `root.device.customName || VenusOS.digitalInput_typeToText(type.value)`

pages/settings/devicelist/PageDigitalInput.qml
chriadam: Sure, done

pages/settings/devicelist/delegates/DeviceListDelegate_digitalinput.qml
chriadam: Indeed, done

pages/settings/devicelist/delegates/DeviceListDelegate_digitalinput.qml
chriadam: Yep, makes sense, we should.  Thanks, done.

chriadam: I've created https://github.com/victronenergy/gui-v2/pull/1983 which resolves the issue in most cases, at least.

pages/settings/devicelist/PageDigitalInput.qml
```diff
@@ -12,15 +12,25 @@ Page {
 
 	property string bindPrefix
 
-	title: VenusOS.digitalInput_typeToText(inputType.value)
+	title: device.customName || VenusOS.digitalInput_typeToText(inputType.value)
```
blammit, line 15: Can use `device.name` instead of `device.customName`, since the `name` will fall back to a product name if available. Same below.

Otherwise LGTM

blammit (approved):

pages/settings/devicelist/PageDigitalInput.qml
chriadam: But don't we want it to explicitly fall back to a "digital input type to text" string instead of a product name, in the case where there is no custom name?

blammit: I was thinking that we would prefer the product name to the type name, but if that's not the case, then LGTM.

MikeTrahearn-Qinetic (approved):

Issue #1735, opened 2024-11-13
Hide position setting for energy meters that don't support it.


ldoesvictron:
The VM-3P75CT is configured over VictronConnect and has an internal setting for the meter position. Therefore it doesn't support the position setting in the GUI.

At the moment a position setting does appear in gui-v2, but it doesn't actually do anything. 
![image](https://github.com/user-attachments/assets/b6b74cd5-a05a-428b-a758-3a2b5178c42c)

In gui-v1 the position setting is conditional depending on the availability of allowed roles on dbus, see [here](https://github.com/victronenergy/gui/blob/776ec5bc2323baddaa7467ea8d59123832bc5e3a/qml/PageAcInModel.qml#L152).

blammit:
Note: this is referring to the PV inverter settings page, which is shown for Energy Meters that are configured to show up as `pvinverter` services.

chriadam:
Is there a specific installation I can test this with?  I tried testing with https://vrm.victronenergy.com/installation/394636/dashboard however it seems to have an SMA Solar Inverter and the energy meter must not be configured to show up as a pvinverter, I believe.

Also, interestingly, looking at the raw values for the SMA Solar Inverter device (i.e. settings -> debug & develop -> values -> pvinverter -> 20) we see that the /AllowedRoles path has an invalid value (`--`), even though I believe that this is NOT an energy meter but instead is a PV inverter (or am I wrong about this?).  Should we hide the position setting in this case also, or only in the case where the pvinverter is actually an energy meter (and if so, how can this condition be detected programmatically?)  Note that there appears to be a valid value for the /Position path (`0`) despite the /AllowedRoles value being null/invalid.

chriadam:
@ReinvdZee perhaps this is something you might know the answer to ^

chriadam:
Hi Martin, please see the discussion in https://github.com/victronenergy/gui-v2/pull/1980 - I'm not sure what to implement in gui-v2 precisely.

mpvader:
Unblocked - thanks Martin

chriadam:
Updated PR now.

chriadam:
Merged to main, closing as fixed.

PR #1980, opened 2025-03-14
Hide PvInverter position setting if unavailable

chriadam:
If an energy meter is configured to show up as a PV inverter, the PV inverter page should hide the position setting, as such devices have an internal setting for the meter position which does not track the position setting used by gui-v2.

The /AllowedRoles path will be invalid in the case where the energy meter has internal settings, and so we can use this as a condition to determine whether to hide or show the position setting.

Contributes to issue #1735

### DIFF
--- a/pages/solar/PvInverterPage.qml
+++ b/pages/solar/PvInverterPage.qml
@@ -78,6 +78,15 @@ Page {
 
 			ListPvInverterPositionRadioButtonGroup {
 				dataItem.uid: root.pvInverter.serviceUid + "/Position"
+				preferredVisible: (!positionIsAdjustable.valid || positionIsAdjustable.value === 1) ? dataItem.valid : false
+
+				// Datapoint will exist in VM-3P75CT energy meters, but usually will not exist.
+				// In cases where the data point does not exist, assume position IS adjustable.
+				// Value will be zero if the position setting is not adjustable via gui-v2.
+				VeQuickItem {
+					id: positionIsAdjustable
+					uid: root.pvInverter.serviceUid + "/PositionIsAdjustable"
+				}
 			}
 
 			ListQuantity {


MikeTrahearn-Qinetic (approved):

pages/solar/PvInverterPage.qml
```diff
@@ -78,6 +78,12 @@ Page {
 
 			ListPvInverterPositionRadioButtonGroup {
 				dataItem.uid: root.pvInverter.serviceUid + "/Position"
+				preferredVisible: dataItem.valid && allowedRoles.valid
```
chriadam, line 81: Actually, testing shows that this seems to hide it for all PV inverters, even ones which aren't energy meters configured to show up as pvinverters.  I wonder if I need to set this to something like:
```
preferredVisible: isEnergyMeter ? allowedRoles.valid : dataItem.valid
```
but that begs the question, how can I determine the isEnergyMeter condition?

blammit: I believe that the `isEnergyMeter` condition can be checked by seeing whether `/Role` is present, assuming that actual PV inverters cannot be changed to have other roles like grid, genset, etc. Perhaps @ReinvdZee can confirm.

ReinvdZee: @blammit, @chriadam, 

> I believe that the isEnergyMeter condition can be checked by seeing whether /Role is present

I agree, I would expect the `/Role` path to not be present for the other pv inverters, but I can't confirm. @izak probably can.

I also just noticed that the path also is not present when the grid meter is connected to the GX over VE.CAN, but I expect this to be a bug.

ReinvdZee: @Izak and I discussed this just now; you should display the setting only if the `/Position` path is valid. On the backend, we make sure to validate the path when its supposed to.

chriadam: Do you mean: I don't need to check the /AllowedRoles path at all, but instead should just do `preferredVisible: dataItem.valid` inside the /Position ListPvInverterPositionRadioButtonGroup ?

izak:
Hi, @ReinvdZee tells me that there are meters where /Position is apparently not writable. I have an idea which ones those might be, but just to not duplicate work: What meter did you observe this with?

This works on all the Carlo Gavazzi meters and the ABB meters. I am not sure if the VM-3P75CT works, and that may well be the meter you observed this on?

chriadam:
It is written in the issue #1735 "The VM-3P75CT is configured over VictronConnect and has an internal setting for the meter position. Therefore it doesn't support the position setting in the GUI."  (Or maybe I am misunderstanding what was meant?)

chriadam:
Perfect, thanks, I will update the PR based on this.

chriadam:
Ready for review/testing.  You can check Overview -> Solar Yield -> SMA Inverter on https://vrm.victronenergy.com/installation/55013/dashboard for an example where position setting is writable / visible, and you can check Overview -> Solar Yield -> VM-3P75CT on https://vrm.victronenergy.com/installation/473156/overview for an example where position setting is hidden.

chriadam:
@ReinvdZee perhaps you could help me with testing this one also ^

ReinvdZee:
@chriadam
Tested this, LGTM.

blammit (approved):

chriadam:
Thank you!

Issue #1966, opened 2025-03-11
Display DC Loads status change not directly visible on Overview page


mr-manuel:
Triggering `Display DC Loads` under Settings -> System Setup requires a reload of the webpage or restart of the GUI service for the local display to show the changes on the Overview page.

![Image](https://github.com/user-attachments/assets/c580801d-13c7-47a1-b1b2-8ed4c58ab8a9)

baudje:
ah that solved my issue of not seeing DC loads: https://github.com/victronenergy/gui-v2/issues/1962

chriadam:
Merged to main, closing as fixed.

mr-manuel:
Can confirm, that it works now.

PR #1981, opened 2025-03-14
Respect hasDcSystem setting value in system dc power

chriadam:
Set the Global.system.dc.power value to NaN if the hasDcSystem setting gets toggled to false.

Also update the trigger for _resetWidgets() in OverviewPage to ensure that it triggers if either the above dc.power value transitions to/from NaN, or if the count of combined dc load devices transitions to/from a positive value (i.e. properly handle potential negative values).

Contributes to issue #1966

### DIFF
--- a/data/System.qml
+++ b/data/System.qml
@@ -29,7 +29,7 @@ QtObject {
 	}
 
 	readonly property QtObject dc: QtObject {
-		readonly property real power: _dcSystemPower.valid ? _dcSystemPower.value : NaN
+		readonly property real power: (_hasDcSystem.valid && _hasDcSystem.value && _dcSystemPower.valid) ? _dcSystemPower.value : NaN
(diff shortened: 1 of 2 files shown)


data/System.qml
```diff
@@ -29,7 +29,7 @@ QtObject {
 	}
 
 	readonly property QtObject dc: QtObject {
-		readonly property real power: _dcSystemPower.valid ? _dcSystemPower.value : NaN
+		readonly property real power: (_hasDcSystem.valid && _hasDcSystem.value && _dcSystemPower.valid) ? _dcSystemPower.value : NaN
```
blammit, line 32: Was thinking it might be nice if setting HasDcSystem=0 would cause systemcalc or such to set /Dc/System/Power=NaN/invalid instead of checking the value in the UI here. But I suppose it's really a UI option and not something that should affect the actual power measurement.

blammit (approved):

Issue #1894, opened 2025-02-05
Settings->Integrations: Physical I/O section header label is shown even if no physical I/O ports exist on device


chriadam:
e.g. raspberry pi or gx card devices have no physical I/O ports.

In this case, the section and its header should be omitted / not visible.

blammit:
Marked as low priority but sounds easy enough to fix

chriadam:
Merged to main, but won't close issue until we can test to confirm the fix.  @ReinvdZee would you be able to help with testing this one?  Needs a GX Card device to test.

chriadam:
Thanks!  Verified.  Closing as fixed.

PR #1982, opened 2025-03-14
Hide Physical I/O section header if section is empty

chriadam:
GX card devices and Raspberry Pi devices don't have physical I/O ports.

Contributes to issue #1894


chriadam:
I don't know how to test this, as we need to test against a GX card device or a raspi device...

MikeTrahearn-Qinetic (approved):

Issue #1838, opened 2025-01-13
Breadcrumb language does not change instantly


mr-manuel:
If I change the language, then the language of the current position in the breadcrumb does not change. Going back and opening the same menu works like it should.

Would be nice, if the breadcrumb gets updated instantly, when the language changes.

![venus-os-language-breadcrumb](https://github.com/user-attachments/assets/e4d14019-c7e4-48c6-85da-4e7af77ab6f5)

chriadam:
Merged to main, closing as fixed.

PR #1983, opened 2025-03-14
Bind breadcrumb titles rather than initializing by value

chriadam:
Contributes to issue #1838

### DIFF
--- a/components/listitems/core/SettingsListNavigation.qml
+++ b/components/listitems/core/SettingsListNavigation.qml
@@ -16,7 +16,7 @@ ListNavigation {
 	property alias secondaryText: secondary.text
 
 	height: Theme.geometry_settingsListNavigation_height
-	onClicked: Global.pageManager.pushPage(root.pageSource, {"title": root.text })
+	onClicked: Global.pageManager.pushPage(root.pageSource, {"title": Qt.binding(function() { return root.text }) })
 
 	CP.ColorImage {
 		id: icon
--- a/components/listitems/core/ListRadioButtonGroup.qml
+++ b/components/listitems/core/ListRadioButtonGroup.qml
@@ -54,7 +54,7 @@ ListNavigation {
 
 	onClicked: {
 		// onClicked is not emitted if interactive: false
-		Global.pageManager.pushPage(optionsPageComponent, { title: text })
+		Global.pageManager.pushPage(optionsPageComponent, { title: Qt.binding(function() { return root.text }) })
 	}
 
 	VeQuickItem {


MikeTrahearn-Qinetic (approved):
Are there any other cases like this that aren't covered by ListNavigation derivatives?

chriadam:
Yes, there are a bunch of other places where we pushPage with a string title.  I will add a follow up commit which fixes those cases.

components/PageGensetModel.qml
```diff
@@ -202,10 +202,11 @@ VisibleItemModel {
 	}
 
 	ListNavigation {
+		id: engineItem
 		//% "Engine"
 		text: qsTrId("ac-in-genset_engine")
 		onClicked: {
-			Global.pageManager.pushPage(engineComponent, {"title": text})
+			Global.pageManager.pushPage(engineComponent, {"title": Qt.binding(function() { return engineItem.text })})
```
blammit, line 209: These don't seem necessary as the text never changes.

Ideally the titles would all be provided by the Page objects themselves. Previously we had an issue where pages with titles that were device names would not update their page titles as expected, due to the title being passed in without a binding like this, so we changed the Pages to internally have `title: device.name` instead. So would be good to do this similarly for cases where the title is expected to change dynamically for some reason.

This widespread practice of passing in a title when pushing a page, came from when I was lazy and didn't want to make another translation id in the Page itself, and it was easier to just pass the text in. So it's coming back to bite me now :(

MikeTrahearn-Qinetic: While I agree that when you see all these changes in one place it does seem rather unnecessarily verbose, but alas when you have dynamic object creation like this where the title binding could not be known in advance and added to the Component declaration (which would in subsequent Qt versions require the addition of pragma Component Behavior: Bound) then I don't think there's a better way.

MikeTrahearn-Qinetic (approved):

components/PageGensetModel.qml
blammit: Yes, if we don't know the title in advance then it makes sense to use this binding. For cases like this "Engine" page, we do know the title in advance, so the `Page` can set `title: "Engine"` accordingly.

chriadam: We expect all of the strings to change dynamically (i.e. if the system language changes in the background, e.g. due to setting change via VRM).  Yes, I agree we should define it as a declared binding in the page itself, where possible.  I will do this.

components/PageGensetModel.qml
```diff
@@ -295,22 +297,20 @@ VisibleItemModel {
...
-												   settingsBindPrefix: root.settingsBindPrefix,
-												   startStopBindPrefix: root.startStopBindPrefix,
-												   gensetBindPrefix: root.bindPrefix
-											   })
+		text: CommonWords.run_time_and_service
+		onClicked: Global.pageManager.pushPage(pageGeneratorRuntimeService,
+				{ settingsBindPrefix: root.settingsBindPrefix,
+				  startStopBindPrefix: root.startStopBindPrefix,
+				  gensetBindPrefix: root.bindPrefix })
+		Component { id: pageGeneratorRuntimeService; PageGeneratorRuntimeService { } }
```
blammit, line 305: Is there a particular reason for changing this one to a Component?

components/listitems/core/SettingsListNavigation.qml
```diff
@@ -16,7 +16,7 @@ ListNavigation {
 	property alias secondaryText: secondary.text
 
 	height: Theme.geometry_settingsListNavigation_height
-	onClicked: Global.pageManager.pushPage(root.pageSource, {"title": root.text })
+	onClicked: Global.pageManager.pushPage(root.pageSource, { title: Qt.binding(function() { return root.text }) })
```
blammit, line 19: Is it still necessary to pass in titles if the pages have their own titles now?

pages/settings/PageSettingsCGwacsOverview.qml
```diff
@@ -40,11 +40,14 @@ Page {
 		}
 
 		delegate: ListNavigation {
+			id: cgwacsDelegate
 			readonly property string devicePath: Global.systemSettings.serviceUid + "/Settings/Devices/cgwacs_" + modelData
 
 			text: getDescription(customNameItem.value, modelData)
 			secondaryText: getMenuName(serviceType.value, l2ServiceType.value)
-			onClicked: Global.pageManager.pushPage("/pages/settings/PageSettingsCGwacs.qml", { title: text, devicePath: devicePath })
+			onClicked: Global.pageManager.pushPage("/pages/settings/PageSettingsCGwacs.qml",
```
blammit, line 48: Not sure if this Qt.binding() is needed, as the description uses the device custom/product name which shouldn't dependent on the language

pages/settings/PageSettingsDisplayAndAppearance.qml
```diff
@@ -72,33 +72,35 @@ Page {
 			}
 
 			ListNavigation {
+				id: displayUnitsItem
 				//% "Data units"
 				text: qsTrId("pagesettingsgeneral_data_units")
-				onClicked: Global.pageManager.pushPage("/pages/settings/PageSettingsDisplayUnits.qml", {"title": text})
+				onClicked: Global.pageManager.pushPage(pageSettingsDisplayUnits)
+				Component { id: pageSettingsDisplayUnits; PageSettingsDisplayUnits { title: displayUnitsItem.text } }
```
blammit, line 79: I can see you added CommonWords for some of the page titles, so was wondering why you opted to create a Component instead of adding a `title: blah` to the QML files for these pages?

I'm not against adding Components for these and I know it reduces the runtime speed of loading these pages, but I don't know what this means going forward in the codebase -- do we want to always declare components of sub-pages where possible, instead of just launching them via url? Also, if we want to lower the runtime speed then I understand this change, but if not, why not set the title from the page instead?

chriadam: It basically would mean putting ALL strings into CommonWords (as they would be used in two places: 1) the page in which the list entry exists, 2) the page which is the target of that list entry).  Maybe this isn't a problem, and we should do it - I don't mind.

But, yes, the performance advantage is also something to bear in mind (although it will increase startup time also, which is a tradeoff worth considering).

blammit: I'm not sure of the best solution, maybe we can discuss it more offline. This kind of duplication really makes me want to move to qsTr() instead of moving so many of the translations to CommonWords.

chriadam: Yeah, fair point.

blammit (approved):

PR #1984, opened 2025-03-14
Reduce rendering cost of load graphs on GX devices

chriadam:


src/theme.cpp
```diff
@@ -80,6 +81,17 @@ bool Theme::objectHasQObjectParent(QObject *obj) const
 	return obj && obj->parent();
 }
 
+// this properly belongs in a utils class, but there is no cpp utils currently.
```
blammit, line 84: Would be great if you could add a utils-type class. There are some features in BackendConnection that should be moved to a utils class, e.g. maybe applicationVisible, demoImageFileName, needsWasmKeyboardHandler.

chriadam: There's already a Utils.js which has a bunch of javascripty stuff in it, so I don't want to clobber it.  Which name should I choose for the CPP utils?

blammit: I'm not sure... AppUtils?

chriadam:
This only moves the needle by a tiny amount (maybe 1% or 2% total load).  I will investigate how we can improve it more dramatically (e.g. by setting the "orange" parts of the shape paths to visible=false in the case where they are not needed).

components/LoadGraphShapePath.qml
```diff
@@ -21,35 +21,39 @@ Shape {
...
-		PathCubic { x: 6 * segWidth - offset; y: calcY(model[6]); relativeControl1X: rc1x; control1Y: calcY(model[5]); relativeControl2X: rc2x; control2Y: y; }
-		PathCubic { x: 7 * segWidth - offset; y: calcY(model[7]); relativeControl1X: rc1x; control1Y: calcY(model[6]); relativeControl2X: rc2x; control2Y: y; }
-		PathCubic { x: 8 * segWidth - offset; y: calcY(model[8]); relativeControl1X: rc1x; control1Y: calcY(model[7]); relativeControl2X: rc2x; control2Y: y; }
-		PathCubic { x: 9 * segWidth - offset; y: calcY(model[9]); relativeControl1X: rc1x; control1Y: calcY(model[8]); relativeControl2X: rc2x; control2Y: y; }
-		PathCubic { x: 10 * segWidth - offset; y: calcY(model[10]); relativeControl1X: rc1x; control1Y: calcY(model[9]); relativeControl2X: rc2x; control2Y: y; }
-		PathCubic { x: 11 * segWidth - offset; y: calcY(model[11]); relativeControl1X: rc1x; control1Y: calcY(model[10]); relativeControl2X: rc2x; control2Y: y; }
-
-		PathLine { x: root.width + root.strokeWidth; y: calcY(model[11])}
+		PathCubic { x: 1 * segWidth - offset; y: yValues[1]; relativeControl1X: rc1x - offset; control1Y: yValues[0]; relativeControl2X: rc2x - offset; control2Y: y; }
```
MikeTrahearn-Qinetic, line 44: n * segWidth - offset could be pre-calculated

components/LoadGraphShapePath.qml
```diff
@@ -21,35 +21,39 @@ Shape {
...
+		PathCubic { x: 2 * segWidth - offset; y: yValues[2]; relativeControl1X: rc1x; control1Y: yValues[1]; relativeControl2X: rc2x; control2Y: y; }
```
MikeTrahearn-Qinetic, line 45: Could a PathArc be used instead if any quicker?

chriadam: Possibly, I haven't benchmarked.  For what it's worth, the central gauges are FAR more expensive than these load graphs, and they use PathArc.  I'm not sure.

src/fastutils.cpp
```diff
@@ -0,0 +1,32 @@
...
+{
+}
+
+// this properly belongs in a utils class, but there is no cpp utils currently.
+QList<qreal> FastUtils::calculateLoadGraphYValues(const QList<qreal> &data, int dataLen, qreal height) const
+{
+	QList<qreal> ret;
+	ret.reserve(dataLen);
+	for (int i = 0; i < dataLen; ++i) {
+		ret.append((1.0 - (data.count() <= i ? 0.0 : data[i])) * height);
```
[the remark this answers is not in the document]
chriadam: I agree, but from my benchmarking this has basically zero effect.  The rendering cost dominates.

components/LoadGraph.qml
```diff
@@ -25,12 +25,13 @@ Item {
...
-		model.shift()
-		orangePath.model = []
-		orangePath.model = root.model
+		let temp = model
+		temp.push(value)
+		temp.shift()
+		model = temp
 	}
 
+	clip: true
```
chriadam, line 34: this is required if we don't use the supersample layer (which naturally causes clip) in the LoadGraphShapePath.

chriadam:
Saving between 4% and 7% CPU load on average (according to top) with all of these changes, on GX device.  The vast majority of the cost of rendering is still the central gauges, so the "best" thing you can do for performance currently is to disable all the levels if possible.

src/theme.cpp
chriadam: I went with FastUtils

components/LoadGraph.qml
```diff
@@ -71,20 +74,27 @@ Item {
...
 		}
 	}
+
 	Rectangle {
 		anchors.bottom: parent.bottom
 		width: parent.width
-		height: parent.height * threshold + 1 // '+1' to conceal a 1 pixel orange line from the orange section
-		clip: true
+		height: root.height * threshold
+		clip: true // we have to clip this, because we can't rely on setting myYValue of bluePath.
```
DanielMcInnes, line 82: myYValue?

components/LoadGraphShapePath.qml
```diff
@@ -21,35 +21,63 @@ Shape {
...
-	function calcY(data) { return (1 - (data || 0)) * height }
+	property bool calculateMinYValue: false // calculate vs clamp
+	property list<real> yValues: [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
+	property real minYValue: 0 // used to determine visibility for orange graphs, or clamp min value for blue graphs.
+	onModelChanged: {
+		const newYValues = FastUtils.calculateLoadGraphYValues(model, 12, height)
 
-	smooth: true
+		if (calculateMinYValue) {
+			let tempMin = 999
```
DanielMcInnes, line 31: use something better than 999. Theme.geometry_screen_height?

components/LoadGraphShapePath.qml
```diff
@@ -21,35 +21,63 @@ Shape {
...
+			for (let i = 0; i < 12; ++i)  {
```
DanielMcInnes, line 32: 12 -> newYValues.length
here and elsewhere

DanielMcInnes (approved):

components/LoadGraphShapePath.qml
chriadam: done

components/LoadGraphShapePath.qml
chriadam: done

Issue #1030, opened 2024-05-02
Add keyboard navigation to gui-v2


jpetrell:
In gui-v1, there is keyboard navigation when using it on Remote console. And that is used everyday by many technicians that sit behind their laptop and sort issues out or configure systems. Enable keyboard navigation on gui-v2 also.

![image](https://github.com/victronenergy/gui-v2/assets/2203667/331e4f10-6921-4290-9e91-bb331fb64798)

### Dependencies

This requires the following tasks to be resolved first:

- Implement the settings layout update #1232 
- Hide non-visible items in a list, otherwise they will receive focus during key navigation #1732 
- ListItem enabled must be "true" even when an item is not clickable, otherwise the item cannot be focused during key navigation #1876

jpetrell:
Arrow navigation e.g. between main views and vertically on the settings pages seems pretty straight-forward (i.e. the keyboard focus is moved to the adjacent item that is located in the arrow direction), and for the common UI components (dialogs, sliders, spin boxes, etc.) I can initially use gui-v1 behavior as an example.

Some open items to decide:
- How should the selected item look like and behave? Always there, or fade out after time out (as not meaningful for the touch case)? Visually mimick press effect or have it more clearly highlighted like on gui-v1 that displays blue background?
- Should there be key shortcut for opening Control cards. Could be that pressing Esc (back) on the main views would toggle between Control cards and the main carousel?
- Should we allow selecting boxes in the Overview page, or reserve left/right arrows for navigating to the adjacent Overview/Notifications pages
- How to switch between Tanks and Environment tabs in the Levels page? Clicking arrow up focuses tab bar?
- How to access status bar buttons? Similar to the tab bar, clicking arrow up focuses the first status bar button? And after clicking the right arrow key moves the focus to the status bar button on the right
- Should we show the keyboard layout somewhere like shown above in the Remote Console case?

jpetrell:
Not for 1.0.0.

jpetrell:
Updated the prototype to match Serj's design.

Todo:
- While the key navigation works on most sub-pages some of the main pages like the Control cards, Overview and Notifications pages don't support key navigation yet.
- Many settings list items that are not valid in the current customer's configuration are hidden. Take a variant of gui-v1's ObjectModel into use in gui-v2 that removes hidden settings list items from the model so they stop interfering with the key navigation
[...]

PR #1985, opened 2025-03-17
Control cards: exclude energy meters from EVCS model

blammit:
The evcharger services on the system may include energy meters acting as EV chargers, which just provide measurements and are not real EV chargers that actually charge. We know they are energy meters if the /Mode value is not valid.

In the control cards, these energy meters are excluded from the display as they are not controllable EV chargers. Instead of excluding them by setting visible=false on the relevant EVCSCard, use a different model that excludes these energy meters.

This simplifies the upcoming key navigation implementation, as the key navigation feature will not need to attempt to skip over invisible cards.

Part of #1030

### DIFF
--- a/pages/ControlCardsPage.qml
+++ b/pages/ControlCardsPage.qml
@@ -56,11 +56,11 @@ Page {
 				spacing: Theme.geometry_controlCardsPage_spacing
 
 				Repeater {
-					model: Global.evChargers.model
+					model: evChargerModel
 
 					EVCSCard {
 						width: root.cardWidth
-						evCharger: model.device
+						serviceUid: model.device.serviceUid
 					}
 				}
 			}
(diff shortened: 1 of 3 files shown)


pages/ControlCardsPage.qml
```diff
@@ -127,4 +127,33 @@ Page {
 			}
 		}
 	}
+
+	// A model of evcharger services that represent controllable EV chargers, i.e. those with a
+	// valid /Mode value. Global.evChargers.model cannot be used in the control cards, as it
```
chriadam, line 132: is this place (ControlCardsPage.qml) the only place where this distinction is useful?  If not, should we add a Global.controllableEvChargers.model or similar?

pages/controlcards/EVCSCard.qml
```diff
@@ -60,11 +65,20 @@ ControlCard {
...
+			dataItem.uid: root.serviceUid + "/SetCurrent"
 			preferredVisible: dataItem.valid
 			interactive: dataItem.valid && modeItem.value === VenusOS.Evcs_Mode_Manual
+
+			VeQuickItem {
+				id: maxCurrent
+				uid: root.serviceUid + "/MaxCurrent"
+				onValueChanged: {
+					if (valid) {
+						chargeCurrentSpinBox.to = value
```
chriadam, line 78: seems a bit unrelated to the rest of the PR, but I guess that's ok

chriadam: ah, no, I see, it's because the root.evCharger device no longer exists, as the device might be a pretend one.  ignore me.

chriadam (approved):

pages/ControlCardsPage.qml
blammit: Only here at the moment, but yes if we need this elsewhere, we can move it to a globally-acccesible location.

PR #1986, opened 2025-03-17
Device List: fix source check to ensure delegate is loaded

### LINKED ISSUE #1925: Device list sometimes shows gaps between items

blammit:
Since source is a url and not a string, the boolean coercion returns true even if the source is empty because an empty url does not equate to an empty string. So, convert the source to a string before doing the boolean check.

Fixes #1925


chriadam (approved):
LGTM.  Could even use `source.toString().length === 0` for complete self-documentation

blammit:
Updated to check the `length` as suggested

PR #1990, opened 2025-03-18
Control cards: change ListView to use per-device delegate cards, using AggregateDeviceModel

blammit:


pages/ControlCardsPage.qml
```diff
@@ -36,98 +40,102 @@ Page {
...
 		flickDeceleration: Theme.geometry_flickable_flickDeceleration
 
-		model: VisibleItemModel {
-			Loader {
-				active: systemType.value === "ESS" || systemType.value === "Hub-4"
-				width: active ? root.cardWidth : -cardsView.spacing
-				sourceComponent: ESSCard {
+		header: Loader {
+			active: systemType.value === "ESS" || systemType.value === "Hub-4"
+			sourceComponent: Item {
```
MikeTrahearn-Qinetic, line 45: Consider using DelegateChooser instead.
https://doc.qt.io/qt-6/qml-qt-labs-qmlmodels-delegatechooser.html

blammit: It would be nice to use something like that. Maybe not right now as it's still experimental, but would be worth considering if it moves out of the labs module.

pages/ControlCardsPage.qml
```diff
@@ -36,98 +40,102 @@ Page {
...
-				spacing: Theme.geometry_controlCardsPage_spacing
-
-				Repeater {
-					model: Global.generators.model
 
-					GeneratorCard {
-						width: root.cardWidth
-						generator: model.device
-					}
+				SwitchesCard {
```
chriadam, line 67: This isn't related to SmartSwitch is it?

pages/ControlCardsPage.qml
chriadam: I personally don't mind using DelegateChooser, although it would mean bundling a new plugin with the wasm build, and asking Jeroen to include that plugin on the firmware image.

I think let's keep this as is for now, but if we ever have to put a new type of card here, best to just switch to delegate chooser then.

src/aggregatedevicemodel.cpp
```diff
@@ -356,21 +434,25 @@ void AggregateDeviceModel::deviceNameChanged()
...
-			toIndex = i > fromIndex ? i - 1 : i;
-			break;
-		}
+	if (m_sortBy == NoSort) {
+		return;
 	}
 
+	// Update the list order if needed
+	const int sortedIndex = sortedDeviceIndex(device, m_deviceInfos[fromIndex].sourceModel, -1);
+	if (sortedIndex > 0 && m_deviceInfos[sortedIndex - 1].device == device) {
```
chriadam, line 443: I guess this doesn't need to check that `sortedIndex < m_deviceInfos.size()` (since `sortedDeviceIndex()` returns a sortedIndex which is definitely less than m_deviceInfos.size(), since it iterates over m_deviceInfos and returns matching i etc) but that's not immediately obvious from this code location.  I would be inclined to add the check here anyway, in case the implementation of sortedDeviceIndex() changes in future, remembering the TigerBeetle's exhortation "assertions come in pairs" (https://tigerbeetle.com/blog/2023-12-27-it-takes-two-to-contract/).

chriadam (approved):
One nit, feel free to ignore.  LGTM.  Great work!

pages/ControlCardsPage.qml
blammit: Not related - it's a card of switch toggles for the relays. Perhaps it will need to be renamed later on.

pages/ControlCardsPage.qml
MikeTrahearn-Qinetic: I will check with Qt on the status of this API in the meantime and whether it is going to be promoted out of Labs any time soon.

src/aggregatedevicemodel.cpp
blammit: Sounds okay to have the extra sanity check 👍 Updated.

PR #1997, opened 2025-03-24
Fix changing of access level

### LINKED ISSUE #1996: Can't change access from User to User & Installer
If you change to User access level, you cannot change back to User & Installer. The "Setting locked for access level" toast appears:

Image

blammit:
RadioButtonListPage needs to propagate the show/read access levels down to the radio button and text fields for each delegate. Otherwise, the components will not have the 'user' write access level, and ListItem will prevent the button from being clicked. The validatePassword function also needs to be propagated to the page so that the password can be validated.

Fixes #1996

### DIFF
--- a/components/RadioButtonListPage.qml
+++ b/components/RadioButtonListPage.qml
@@ -13,6 +13,9 @@ Page {
 	required property var optionModel
 	readonly property alias optionView: optionsListView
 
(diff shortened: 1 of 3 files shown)


MikeTrahearn-Qinetic:
+1 for inheriting access level eh? ;-)

MikeTrahearn-Qinetic (approved):

PR #2004, opened 2025-03-26
Tank model rework

blammit:
To prepare for key navigation support, rework TanksTab.qml to load all tank gauges and gauge groups through a model+delegate pattern, rather than using sets of Row+Repeaters to load each tank type. This allows the ListView to focus one gauge (or gauge group) at a time and auto-scroll to show the current gauge/group.

All merging/separation of gauge groups is now done in AggregateTankModel rather than in the UI level, which is a lot cleaner.


src/basetankdevice.cpp
```diff
@@ -0,0 +1,89 @@
...
+using namespace Victron::VenusOS;
+
+BaseTankDevice::BaseTankDevice(QObject *parent)
+	: BaseDevice(parent)
+{
+}
+
+void BaseTankDevice::updateMeasurements()
+{
+	// If one of the measurements are unavailable, calculate it from the other two (if available).
```
chriadam, line 19: nitpick: should be "is unavailable" as you're preferring to (singular) one of the measurements

src/basetankdevice.cpp
```diff
@@ -0,0 +1,89 @@
...
+qreal BaseTankDevice::level() const
+{
+	return m_level;
+}
+
+void BaseTankDevice::setLevel(qreal level)
+{
+	if (m_level != level) {
+		m_level = level;
+		updateMeasurements();
```
chriadam, line 58: can this result in a change to e.g. remaining()?  If so, it would need to emit remainingChanged() also, I think, to ensure binding re-evaluation.  Maybe detect and do ALL of the signal emission in updateMeasurements(), rather than emitting levelChanged etc below?

src/basetankdevice.cpp
```diff
@@ -20,7 +20,7 @@ void BaseTankDevice::updateMeasurements()
 	if (qIsNaN(m_level) && !qIsNaN(m_capacity) && !qIsNaN(m_remaining)) {
 		// Calculate level
 		if (m_capacity > 0) {
-			m_level = m_remaining / m_capacity * 100;
+			m_level = (m_remaining / m_capacity) * 100;
```
chriadam, line 23: nitpick: this line change should be squashed into the previous commit, I guess

src/basetankdevicemodel.cpp
```diff
@@ -0,0 +1,71 @@
...
+{
+	if (m_totalRemaining != totalRemaining) {
+		m_totalRemaining = totalRemaining;
+		emit totalRemainingChanged();
+	}
+}
+
+BaseTankDevice *BaseTankDeviceModel::tankAt(int index) const
+{
+	return qobject_cast<BaseTankDevice*>(deviceAt(index));
```
chriadam, line 72: just double checking: do each of the devices in the model have a QObject parent set (i.e. to the model itself)? Alternatively, is it worth calling QJSEngine::setObjectOwnership(CppOwnership) here?

tests/basetankdevice/tst_basetankdevice.qml
```diff
@@ -25,32 +25,38 @@ TestCase {
 				expected: { level: 75, capacity: 160, remaining: 120 },
 			},
 			{
+				// level=NaN, so calculate it (remaining / capacity*100)
```
chriadam, line 28: nitpick: I guess these changes should be squashed into the previous commit

src/aggregatetankmodel.cpp
```diff
@@ -0,0 +1,437 @@
...
+	}
+	return fromIndex + consecutiveMatches;
+}
+
+void AggregateTankModel::convertToGroup(int index)
+{
+	if (index >= 0 && index < m_entries.count()) {
+		m_entries[index].isGroup = true;
+		m_entries[index].tank.clear();
+		emit dataChanged(createIndex(index, 0), createIndex(index, 0), { TankRole, IsGroupRole });
```
chriadam, line 233: I don't fully understand this.  I assumed that this should update `m_entries[index].tankModel` to point to a (new?) tank model which contains a single entry (i.e. the current `m_entries[index].tank` value), and then that the dataChanged roles would include TankModelRole.  Why isn't this the case?  Is there an assumption that an entry at a previous index will have the appropriate tank model entry?  Or?

src/aggregatetankmodel.cpp
```diff
@@ -0,0 +1,437 @@
...
+					convertToGroup(i);
+					needsTankInsertion = false;
+					break;
+				}
+			}
+		}
+	}
+
+	if (needsTankInsertion) {
+		Entry newEntry = { tank, tankModel };
```
chriadam, line 302: doesn't explicitly initialise isGroup?

src/aggregatetankmodel.cpp
```diff
@@ -0,0 +1,437 @@
...
+		emit countChanged();
+	}
+}
+
+void AggregateTankModel::modelRowsAboutToBeMoved(const QModelIndex &, int sourceStart, int sourceEnd, const QModelIndex &, int destinationRow)
+{
+	if (sourceStart != sourceEnd) {
+		// Normally only a single row will be moved, due to a device being renamed. If multiple
+		// moves have occurred, reload the model.
+		reload();
```
chriadam, line 415: wondering: does this need to be triggered asynchronously?  Probably not, but just wondering what happens if a reload results in a model reset or something?

src/basetankdevicemodel.cpp
```diff
@@ -69,3 +71,76 @@ BaseTankDevice *BaseTankDeviceModel::tankAt(int index) const
...
+void BaseTankDeviceModel::modelRowsInserted(const QModelIndex &parent, int first, int last)
+{
+	for (int i = first; i <= last; ++i) {
+		// Update the totals whenever the measurements change for each tank. A device typically
+		// changes all of these values at the same time for each device, so delay the update until
+		// the end of the event loop to minimize unnecessary recalculations.
+		if (BaseTankDevice *tank = tankAt(i)) {
+			connect(tank, &BaseTankDevice::levelChanged,
+					this, &BaseTankDeviceModel::updateTotals,
+					Qt::QueuedConnection);
```
chriadam, line 84: Does this also need to be a Qt::UniqueConnection to avoid connecting the same tank multiple times, if the merge limit is exceeded and then underrun, at some point after this called, for example?  Maybe not, just checking..

chriadam: Ah, maybe not, since you disconnect in modelRowsRemoved(), I guess

chriadam:
Mostly looks great.  A few questions.

tests/basetankdevice/tst_basetankdevice.qml
blammit: Fixed

src/basetankdevicemodel.cpp
blammit: These are the TankModel objects that are declared per tank type in Tanks.qml, so they should never be destroyed.

src/aggregatetankmodel.cpp
blammit: The tank model stays the same; the list entry is just changing from a grouped to non-grouped entry. 
E.g. if you have a series of non group entries like this:

[ 0=Fuel, 1=Fuel, 2=Fuel]

And we want to merge these, then index 0 changes to a group entry, and index 1 and index 2 are removed. The tank model (i.e. the Fuel Tank Model) that is referenced by the entry will remain the same.

src/aggregatetankmodel.cpp
blammit: Fixed

src/aggregatetankmodel.cpp
blammit: If there are multiple moves in the source aggregate model, then at worst, this aggregate tank model will do multiple reloads. But the source aggregate model does not do multiple moves anyway, so this handling was added just to be safe.

chriadam:
Possibly unrelated to this PR, but testing in mock mode, with tanks configuration #13 selected, if I click on the combined "fuel" tank to expand it into the exploded view, I cannot scroll through the tanks in the exploded view.  I just see the following static screen:
victron-tanks-exploded-view-non-scrollable

blammit:
> Possibly unrelated to this PR, but testing in mock mode, with tanks configuration https://github.com/victronenergy/gui-v2/pull/13 selected, if I click on the combined "fuel" tank to expand it into the exploded view, I cannot scroll through the tanks in the exploded view. I just see the following static screen:

Yes this is a pre-existing problem. In the exploded view, we don't have a way to scroll, because any presses will cause it to exit the exploded view. Perhaps we can detect a flick vs a press to allow it to be scrolled.

chriadam (approved):
LGTM

PR #2008, opened 2025-03-28
Prepare for SmartSwitch support

blammit:


components/controls/Slider.qml
```diff
@@ -17,6 +17,7 @@ T.Slider {
 	property alias highlightColor: highlightRect.color
 	property alias showHandle: handleImg.visible
 	property bool animationEnabled
+	property Item maskSource: sourceItem
```
blammit, line 20: It's not ideal to expose the sourceItem in this way, so would be nice to improve this later on. @chriadam you probably know more about how to do that nicely.

data/Switches.qml
```diff
@@ -0,0 +1,17 @@
...
+** See LICENSE.txt for license information.
+*/
+
+import QtQuick
+import Victron.VenusOS
+
+QtObject {
+	id: root
+
+	property DeviceModel model: DeviceModel {
```
blammit, line 12: Actually, will remove this model since it will only be used in the Device List.

data/mock/MockDataManager.qml
```diff
@@ -40,6 +40,7 @@ QtObject {
 			property var pulseMeters: PulseMetersImpl { }
 			property var pvInverters: PvInvertersImpl {}
 			property var solarDevices: SolarDevicesImpl {}
+			property var switches: SwitchesImpl{}
```
chriadam, line 43: doesn't exist?

components/StatusBar.qml
```diff
@@ -93,19 +93,32 @@ FocusScope {
...
 	}
 
+	StatusBarButton {
+		id: auxButton
+
+		anchors {
+			left: leftButton.right
+			verticalCenter: parent.verticalCenter
+		}
+		icon.source: "qrc:/images/icon_auxpage_on_32.svg"
```
chriadam, line 112: not to block this PR, but I think we should add an icon_auxpage_off also, and toggle between the two icons, as we do for controls.

chriadam (approved):
LGTM

components/StatusBar.qml
blammit: We don't have a proper "off" version yet, but will ask Serj for one.

data/mock/MockDataManager.qml
blammit: Oops, fixed

PR #2014, opened 2025-04-02
Add SmartSwitch model and UI for cards and Device List

blammit:


data/common/SwitchableOutputName.qml
```diff
@@ -0,0 +1,52 @@
...
+				return `${device.customName} | ${channelName}`
+			} else {
+				return `${device.productName} ${device.deviceInstance} | ${channelName}`
+			}
+		} else {
+			// When the output is in the default group for the device, instead of in a named group,
+			// then only the channel id needs to be in the name.
+			//: %1 = channel name/number
+			//% "Channel %1"
+			return qsTrId("switchable_output_channel").arg(root.outputId)
```
blammit, line 45: @CJhuckstepp in your initial implementation, the first channel is shown as "Channel 1", but here I've use the channel identifier directly so it is "Channel 0". This is because https://github.com/victronenergy/venus/wiki/dbus#switch says the identifier might not be an integer. Do you think this will be problematic?

pages/settings/devicelist/PageSwitchableOutput.qml
```diff
@@ -0,0 +1,75 @@
...
+
+			ListRadioButtonGroup {
+				//% "Type"
+				text: qsTrId("page_switchable_output_type")
+				dataItem.uid: root.outputUid + "/Settings/Type"
+				preferredVisible: dataItem.valid
+				optionModel: validTypesItem.options
+			}
+
+			ListSpinBox {
```
blammit, line 64: @CJhuckstepp I changed the Fuse Rating to be modified by a spin box rather than a text field, as the existing UI typically uses a spin box for this kind of data. (The spin box takes care of the min/max clamping as well.) Is this UI change okay?

data/common/SwitchableOutputName.qml
CJhuckstepp: I had not appreciated the effect of this as Victron only changed this in there last internal review of the dbus api.  Changing this to be zero based index will causes as issues. The only way I can see round this is to check if valid number then treat as channel index and create the label as "Channel {index+1}"  or Ch {index+1}.  Or if is contain alfa characters just use the string as the label and not add the channel text. As the word channel will probably not be relevant

pages/settings/devicelist/PageSwitchableOutput.qml
CJhuckstepp: Ok it seem to work with the smartSwitch part vecan diver and an actual device. As this will will adjust the  set value on the dbus to the actual fuse values supported by the smartSwitch.  This method allows any other device to the 0.1A resolution.

data/common/SwitchableOutputName.qml
blammit: That makes sense - I've added a commit to show "Channel {index+1}" or "Ch {index+1}" as suggested, unless the id is not a number, and in that case it will be shown directly without the "Ch" or "Channel" prefix.

pages/settings/devicelist/PageSwitchableOutput.qml
blammit: Great, thanks for verifying.

ReinvdZee:
I did another round of testing on the `blam/smartswitch-model` branch.

I tested with two ES smartswitches, one GX IO extender and a virtual dbus-switch service created by node-red. 
Compared to the previous round of testing, some things are already fixed:
1. Spacing between output nr and the controls in the control card
2. Hide module voltage if its not there. 
3. GX IO Extender support


Open issues:
1. The control card icon must be updated. See here for the latest version: https://www.figma.com/design/GjCtz3A3jqSGvyjVvOm6RZ/GX-Touch-design?node-id=11183-230476&t=SIJkaGLU3iorLo6D-4
5. The Fuse rating setting was changed to a spinbox, but it still allows for decimal precision which will be rounded to the nearest integer when stored. I think its better if the spinbox also only allows setting the value with integers.
6. The path `/SwitchableOutput/x/Name` is added which holds the default channel name. Use that to display the name instead of "Ch <x>". This path has been added to the API recently so it may not be present yet in the current version of Venus OS. But you can consider this path mandatory.
7.   The virtual switch device is not working properly because it is missing paths. I'll discuss this with @dirkjanfaber 

And as is described in the [api](https://github.com/victronenergy/venus/wiki/dbus#switch), other services can create `/SwitchableOutput/..` paths as well. For now, it's not necessary to handle these paths.

blammit:
@ReinvdZee Thanks for the feedback. I've updated the control card icon, and the UI now shows `/SwitchableOutput/x/Name` where required, instead of something like "Channel <index+1>".

I've also updated the Fuse rating setting to disallow decimal precision, so now it can only be loaded and saved as an integer.

CJhuckstepp:
I have also tested this with an updated dbus-vcan driver (branch esSmartswitch) and it functions ok. Therefore we would be happy to release @blammit 

However on @ReinvdZee subject above the smart switch also has input function possibilities. which are configured used the  switchableOutput\Type path ,  when set to 3 or 4.  Therefore is it possible to add the below to the enum and exclude them from model.   
	case SwitchableOutput_Type_Dimmable:
		//% "Dimmable"
		return qtTrId("switchable_output_dimmable");
	case SwitchableOutput_Type_InputActiveHigh:  //=3
		//% "Input active high"
		return qtTrId("switchable_input_Active_High);
	case SwitchableOutput_Type_InputActiveHigh:  //=4
		//% "Input active low"
		return qtTrId("switchable_input_Active_Low);
	case SwitchableOutput_Type_Slave:

They currently would still not be available until the vecan driver supported them and publish the switchableOutput/ValidTypes path which included them. 

This would also help with the  GX IO extender  if the switchableOutput was set to SwitchableOutput_Type_InputActiveHigh before it was removed.

blammit:
@ReinvdZee Thanks, I have fixed the issue where outputs appear as "-1" after the GIO Extender is reconnected as you described.

The fix was to ensure the output would only be shown in the cards if the type was momentary/latchable/dimming (as it was previously shown even if the type was -1 or "invalid"), so @CJhuckstepp this should also fix the issue you mentioned, where it would be shown for other /Type values like 3 or 4.

chriadam (approved):
LGTM.  Would be good if @ReinvdZee could do another round of testing on his test system too.

blammit:
@ReinvdZee thanks for the testing -- sounds good to merge it for now, and I will look into this issue later this week.

PR #2015, opened 2025-04-02
Fix typos in Arctic and Antarctic time zone names

blammit:


MikeTrahearn-Qinetic (approved):

mr-manuel:
Can you also replace `Etc` with `Other` or `Others`? Would sound much better or what do you think?

blammit:
> Can you also replace Etc with Other or Others? Would sound much better or what do you think?

It definitely sounds better, but need to check the change is okay for the backend too; I'll comment in that bug.

PR #2028, opened 2025-04-04
Add key navigation support

blammit:
To navigate:
- Press any arrow key to enable the key navigation feature
- Use arrow keys to navigate between any UI controls. E.g. left/right amongst nav bar buttons, then up to the status bar. (Tab and Shift+Tab also works but is not the main supported navigation paradigm)
- Use 'space' to click a button, toggle a switch, or trigger whatever action is relevant for the focused item.
- Use 'escape' to pop to the previous page on the stack, or close control cards
- For pages with focusable child components (e.g. Overview or Settings page), the whole page is highlighted initially, and you need to press 'space' to change the focus to a child item. This is intentional, to allow the user to jump between the status bar and the nav bar in two hops, instead of having to traverse through all child items in the page first.

Additional features:
- Pressing any key will exit UI expanded mode in Overview/Tanks
- Modal dialogs: 
  - Enter to accept, Escape to cancel
  - Spin boxes: press Space to go into 'edit' mode, where up/down will change the value; press Enter to exit 'edit' mode.


blammit:
This is still WIP, but would appreciate a high-level look and any feedback on the general architecture and behaviour.

Known bugs:
- Key navigation is inadvertently enabled when a text field receives focus, and probably when arrow keys are used in a text field

Key navigation is not yet implemented for:
- Levels page
- Notifications page
- Solar history combo box, chart and dialog
- Probably some custom list items, need to check 
- Toast notifications (should probably close on Escape/Enter)

Things to consider:
- How to exit key navigation mode? Perhaps when app becomes inactive as per system settings timeout

chriadam:
This seems to work nicely for wasm builds, but desktop/device rendering seems not to work currently.

mr-manuel:
A made a few tests and I very like it.

Is there a reason why it's possible to go up a menu level with the left arrow key, but you can't enter e submenu if you press the right arrow key?

Here an example. Pressing the left arrow key brings me to the `General` page, pressing the righ arrow key does nothing. I have to press the space bar to enter the `Online updates` page. The GUIv1 behavior worked with the right arrow key to enter the submenu.

grafik

blammit:
> Is there a reason why it's possible to go up a menu level with the left arrow key, but you can't enter e submenu if you press the right arrow key?

Nicely spotted, thanks, that should be implemented too -- will update.

mr-manuel:
Could you check how that worked then in the old GUIv1?

blammit:
It's not a problem in gui-v1, as the right arrow key does not open combo-box option pages there. However in gui-v2, combo box list items have a ">" arrow, just like list items with sub-pages, so in gui-v2 it makes sense for the right arrow key to open sub-pages for combo box items as well.

DanielMcInnes (commented):
very cool

components/StatusBar.qml
```diff
@@ -275,16 +321,37 @@ FocusScope {
...
+			enabled: Global.screenBlanker?.supported
+					&& Global.screenBlanker?.enabled
+					&& Global.pageManager?.interactivity === VenusOS.PageManager_InteractionMode_Interactive
 			onClicked: Global.screenBlanker.setDisplayOff()
 		}
 	}
+
+	// The status bar should never become the focused item; if it does, it means there was no
+	// previously focused button in the status bar, or the last focused button is now disabled and
+	// not focusable. So, for the first available button and focus that instead.
```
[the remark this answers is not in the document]
blammit: Fixed

components/KeyNavigationListHelper.qml
```diff
@@ -0,0 +1,142 @@
...
+		if (Utils.acceptsKeyNavigation(headerItem)) {
+			headerItem.focus = true
+			return true
+		} else {
+			return focusNextItem(-1)
+		}
+	}
+
+	// Moves the focus to the next item in the list.
+	function focusNextItem(fromIndex) {
```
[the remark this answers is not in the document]
blammit: Have refactored, looks nicer now :)

Main.qml
```diff
@@ -23,6 +23,7 @@ Window {
 	property bool isDesktop: false
 	property real scaleFactor: 1.0
 	onIsDesktopChanged: Global.isDesktop = root.isDesktop
+	onActiveFocusItemChanged: console.log("** Active focused:", activeFocusItem, activeFocusItem?.title ?? activeFocusItem?.text ?? "")
```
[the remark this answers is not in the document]
blammit: Oops, forgot to remove this debug

blammit:
> It's not a problem in gui-v1, as the right arrow key does not open combo-box option pages there. However in gui-v2, combo box list items have a ">" arrow, just like list items with sub-pages, so in gui-v2 it makes sense for the right arrow key to open sub-pages for combo box items as well.

@mr-manuel updated to add the right arrow shortcut; access shortcuts are now in the access sub-page, as per https://github.com/victronenergy/gui-v2/pull/2118

DanielMcInnes (approved):

components/controls/DimmingSlider.qml
```diff
@@ -25,13 +26,24 @@ Slider {
...
 			Dot { }
 			Dot { }
 			Dot { }
 		}
 	}
 
+	// Declare this highlight outside the handle, else it is not shown due to the handle's mask source.
+	SliderHandleHighlight {
+		id: handleHighlight
+		x: handleItem.x + handleItem.width - (width / 2)
```
chriadam, line 40: I know we've probably done similar things a thousand times already in the codebase, but technically this code can cause a lot of problems (e.g. the client can specify a custom Item for the handle property, in which case this binding will be referring to an orphan item.  QML will still instantiate handleItem because it has an id and that id is referred to, so the item will be instantiated but will be orphaned).  The qmltc will also emit an error when encountering a construct like this.

You should be able to refer to root.handle.x and root.handle.width instead (as x/y/width/height/opacity/enabled are all "base" Item properties).  But, well, perhaps it's too much effort and we should fix it for all types where we do something similar at some point in the future, but thought I'd point it out.

components/listitems/core/ListTextField.qml
```diff
@@ -167,6 +167,13 @@ ListItem {
 			root.editingFinished()
 		}
 
+		// Consume arrow key events so that they do not trigger key navigation events while the text
+		// field is focused.
```
chriadam, line 171: does the textfield itself still get to handle the key events (e.g. so that the cursor position changes?)

components/controls/DimmingSlider.qml
blammit: Yep, this should have referred to `root.handle` instead. Fixed this, and would be good to fix the other ones in future too.

components/listitems/core/ListTextField.qml
blammit: Yes, this will only catch the case where the cursor is on the end, and the event would go up the item hierarchy.

chriadam (approved):
LGTM. Excellent work.

PR #2029, opened 2025-04-04
Distinguish between Atlantic Standard and Daylight time zones

### LINKED ISSUE #1999: incorrect time being displayed / wrong time zone info
The time zone for Halifax is actually "Atlantic Daylight Time", not "Atlantic Standard Time".
The time zone of the city Lower_Princes is "Atlantic Standard Time" and should be added to make it work correctly.

blammit:
Fixes #1999


chriadam:
LGTM but the commit message has a type, Atlanta -> Atlantic

blammit:
Thanks! Fixed

PR #2034, opened 2025-04-09
New Security Profile Selection UX flow

### LINKED ISSUE #1993: Improve Security Profile settings UI flow
Philipp has updated the design to separate the password and profile settings.
(Double check that the password setting is indeed a separate path, and isn't stored as a JSON array value in the profile setting.)

MikeTrahearn-Qinetic:
Added new Security Profile Selection UX Flow
Tested to work as expected with Cerbo.
No additional DBUS APIs required.
Fixes #1993


components/dialogs/SecurityProfileConfirmationDialog.qml
```diff
@@ -0,0 +1,79 @@
...
+		verticalAlignment: Label.AlignVCenter
+		horizontalAlignment: Label.AlignHCenter
+		font.pixelSize: Theme.font_size_body2
+		wrapMode: Text.Wrap
+		fontSizeMode: Text.Fit
+		text: {
+			if (root.currentPassword.length) {
+				switch (root.pendingProfile) {
+				case VenusOS.Security_Profile_Secured:
+					//% "• Local network services are password protected\n• The network communication is encrypted\n• A secure connection with VRM is enabled\n• Insecure settings cannot be enabled"
```
chriadam, line 63: nitpick, I'd prefer "\u2022" rather than copy-pasting the actual unicode character, it's more self-documenting.

components/dialogs/SecurityProfilePasswordDialog.qml
```diff
@@ -0,0 +1,158 @@
...
+
+	Component.onCompleted: resetValidation()
+
+	property string password
+
+	function valiate() : bool {
+
+		root.password = ""
+
+		if(firstPassword.length < 8) {
```
chriadam, line 29: code style nit, missing space between if and (

components/dialogs/SecurityProfilePasswordDialog.qml
```diff
@@ -0,0 +1,158 @@
...
+				echoMode: TextInput.Password
+
+				// As soon as the input in any input field changes,
+				// the borderColors turns back to blue and the red passwordHint turns invisible.
+				onTextEdited: root.resetValidation()
+
+				// When Return/Enter key is pressed (and as per current behaviour onFocusLost),
+				// the passwords shall be validated
+				onAccepted: root.acceptButton?.clicked()
+				onActiveFocusChanged: if(!focus) root.valiate()
```
chriadam, line 139: typo: should be validate()

pages/settings/PageSettingsAccessAndSecurity.qml
```diff
@@ -84,28 +84,18 @@ Page {
 					{ display: qsTrId("settings_access_service"), value: VenusOS.User_AccessType_Service, readOnly: true },
 				]
 				validatePassword: (index, password) => {
-					if ((index === 0 || index === 1) && password === "ZZZ") {
-						return Utils.validationResult(VenusOS.InputValidation_Result_OK)
-					}
-					//% "Incorrect password"
-					return Utils.validationResult(VenusOS.InputValidation_Result_Error, qsTrId("settings_access_incorrect_password"))
-				}
+									  if ((index === 0 || index === 1) && password === "ZZZ") {
```
chriadam, line 87: nitpick: doesn't need the added indentation, I think

pages/settings/PageSettingsAccessAndSecurity.qml
```diff
@@ -128,91 +118,88 @@ Page {
...
+												 // use an empty password for the unsecure profile
+												 "SetPassword": (profile === VenusOS.Security_Profile_Unsecured ? "" : password),
+												 "SetSecurityProfile": profile
+											 }
+											 // NOTE: this restarts the webserver when changed
+											 securityApi.setValue(JSON.stringify(object))
+
+											 // This guards the wasm version to trigger a reload even if the reply isn't received.
+											 BackendConnection.securityProtocolChanged()
+											 // Global.pageManager.popPage()
```
chriadam, line 195: should this be commented out?

pages/settings/PageSettingsAccessAndSecurity.qml
```diff
@@ -277,7 +264,7 @@ Page {
 
 				// Cannot log out from GX devices, VRM or Unsecured profile with no password
 				preferredVisible: Qt.platform.os === "wasm" && !BackendConnection.vrm
-						 && securityProfile.dataItem.value !== VenusOS.Security_Profile_Unsecured
+								  && securityProfile.dataItem.value !== VenusOS.Security_Profile_Unsecured
```
chriadam, line 267: nitpick: intentional indentation change?

pages/settings/PageSettingsAccessAndSecurity.qml
MikeTrahearn-Qinetic: I can remove it, but it will come back again when I routinely clean up my indentation.

pages/settings/PageSettingsAccessAndSecurity.qml
MikeTrahearn-Qinetic: I'm going to put it back in, but... we'll see how they like it.

pages/settings/PageSettingsAccessAndSecurity.qml
MikeTrahearn-Qinetic: As with all indentations - done by the formatter. Habitual due to not spending needless time fighting it.

MikeTrahearn-Qinetic:
Updated and pushed for any additional review.

philipptrenz:
Hi @MikeTrahearn-Qinetic & @chriadam, I tested the UI using the WASM version and have the following observations:
- The "Change GX Password" modal looks nice!
- Also the check for matching passwords looks good
- The check for matching passwords seems to be triggered with any mouse click, including clicking into the second input field or the Cancel button. This leads to unnecessarily shown error messages. Please make sure, it's only triggered on "Confirm" button click
- After confirming the password, shortly before browser reload, a blue toast is shown stating "Page will automatically reload in 5 seconds" while it actually happens in a fraction of a second. Please either wait the 5 seconds or just reload the page instantly without the toast
- Even after setting a new GX password, changing the security profile is always rejected with the "Changing the security profile not possible - No GX password has been assigned yet"

MikeTrahearn-Qinetic:
Validation of the password is now only done on the Confirm Button click

MikeTrahearn-Qinetic:
@philipptrenz 
Please note that as per existing behaviour, the GUI does not store the password; there is no API to get the current password (or at least that it was set) either. There has been no change to the behaviour of the aforementioned security api in regard to sending an empty string for the password when the mode is _Unsecured_ (2). Please could you discuss internally and decide if a persisted password needs to be available at the UI (not recommended) or how the API might be changed to accommodate the new "requirements" potentially imposed by your new UI flow.

MikeTrahearn-Qinetic:
"Even after setting a new GX password, changing the security profile is always rejected with the "Changing the security profile not possible - No GX password has been assigned yet"
@philipptrenz could confirm then exact reproduction steps noting the statement earlier: _"the GUI does not store the password"_ and consider this in your internal discussion about the matter.

MikeTrahearn-Qinetic:
@philipptrenz The Page Reload Toast has been removed as requested.

MikeTrahearn-Qinetic:
@philipptrenz Please pull the associated branch for this change for the updates as of 19/04/2025.

philipptrenz:
Dear @MikeTrahearn-Qinetic, 

thanks, I will look into it and discuss with the team. 

So from what I understand: In order to truly separate security profile and GX password changes, the following additional (or replacement) functions would be required in the API:

1. Request to set a GX password, independently from a security profile change
2. Reques to set a Security profile, independently from a GX password change
3. Request whether a GX password is currently assigned

Can you please confirm?

MikeTrahearn-Qinetic:
@philipptrenz The new UI design indeed does agree with your assessment and rather suggests that changing the security profile and changing the password are separate activities - that could be done in different Gui-v2 sessions even.
Therefore, your summary of the requirements is correct.

In an idea situation we need (numbered as per your previous points):
1. an API to set the password (value type: string) (there is no need for the UI to know what the password is, just the ability to call .setValue("<newpassword>") on the VeQuickItem. Once this has been called, the security profile should be changeable without also sending a password.
2. a **new** API to set the security profile (type: integer, no password required, could fail (ok) if no password set if required. The UI should prevent this case with the appropriate warning popups so long as we have 3. below:
3. an API to get the status of the password having been set (value type: boolean, readonly)

You _might_ consider keeping the existing security profile API (which takes a JSON string) for backwards compatibility with other systems. A compromise here would be to use the existing security profile API and we still set a JSON string, but you make it able to cope with no password being sent and either pass or fail based on one having _been_ set previously with the new API as per 1.

Note that 1. and 3. could be combined if the password were made readable as a string to avoid additional API, but I didn't want to request visibility of the actual password seeing as we don't actually need to know what it is, only that one was set.

Once this is in place, you should be able set a password. shut down guy-v2, restart guy-v2 and then change the security profile without entering a password as one has been set already in the OS.

chriadam:
Obsoleted by https://github.com/victronenergy/gui-v2/pull/2097

PR #2039, opened 2025-04-11
Move the AllDevicesModel to the DataManager

chriadam:
At least the Mock backend requires access to the AllDevicesModel, and so the data manager loader will not transition to active state unless this AllDevicesModel is accessible via Global.

The GUI (including ApplicationContent) is not loaded until the data manager loader is finished.

Thus, previously, there were cases where the UI would never load because the data manager would never finish loading because it was waiting for something which was initialized by the UI.

This commit ensures that the AllDevicesModel is initialized within the data manager instead.


MikeTrahearn-Qinetic (approved):
While in principle this would indeed "fix" the described race condition, the general question of why the Global object's properties are assigned rather than bound is raised - something that would prevent this condition occurring in the first place. However, there may be reasons for why this was done prior to my present understanding so I leave the question here.

blammit (approved):

PR #2044, opened 2025-04-15
Dialog cleanup

blammit:


pages/settings/PageVrmDeviceInstances.qml
```diff
@@ -66,7 +61,8 @@ Page {
 
 		for (let i = 0; i < Global.allDevicesModel.count; ++i) {
 			const device = Global.allDevicesModel.deviceAt(i)
-			if (device.deviceInstance === deviceInstance
+			if (device
```
MikeTrahearn-Qinetic, line 64: Might be just the web view but the indentation looks odd here.

pages/settings/PageVrmDeviceInstances.qml
```diff
@@ -32,24 +32,19 @@ import Victron.VenusOS
...
-			_swapDialog.instanceA.uid = uid
-			_swapDialog.instanceB.uid = classAndVrmInstanceModel.get(conflictingDeviceIndex).uid
-			_swapDialog.temporaryVrmInstance = isNaN(maximumVrmInstance) ? 0 : maximumVrmInstance + 1
-			_swapDialog.errorFunc = errorFunc
-			_swapDialog.open()
+			const dialogParams = {
+				instanceAUid: uid,
+				instanceBUid: classAndVrmInstanceModel.get(conflictingDeviceIndex).uid,
+				temporaryVrmInstance: isNaN(maximumVrmInstance) ? 0 : maximumVrmInstance + 1,
+				errorFunc: errorFunc,
```
MikeTrahearn-Qinetic, line 45: Passing presumably a JS function to another component via a JS Object? Dangerous if not strange.

blammit: Do you mean, in case the error function disappears during the dialog lifetime? Yes, ideally the dialog would be declared within the delegate to avoid needing this callback function to be called when the error function fails. That can be fixed in a separate PR as it will require some refactoring.

pages/settings/PageVrmDeviceInstances.qml
blammit: Yes, looks odd in the web view if using the default 8-space rendering for tabs.

pages/settings/PageVrmDeviceInstances.qml
```diff
@@ -76,10 +72,6 @@ Page {
 
 	// If user changes the VRM instance, ask whether reboot should be done when page is popped.
 	tryPop: () => {
-		if (_swapDialog && _swapDialog.visible) {
```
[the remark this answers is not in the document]
blammit: Originally the dialogs weren't modal, so it was possible to try to pop the page without closing the dialog. But now the dialog must be closed before the page can be popped, so it's no longer possible to run into this situation.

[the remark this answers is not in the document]
chriadam: I see

chriadam (approved):

pages/settings/PageVrmDeviceInstances.qml
MikeTrahearn-Qinetic: Nothing about lifetime, but there's another thing. But no, it was more about what looks like passing a function "pointer" via a JS Object. Doing it with var properties is one thing (that can't be qmllinted) but going via a JS Object is on another level of type-unsafeness. If this is common practice in Gui-v2 then there's not much else to comment.

blammit: Is there a better way to specify the function in this case? In terms of type safety, there isn't a function type that we can specify for the property, but that would be nice to have.

PR #2046, opened 2025-04-15
Use a single list model/view for notifications

blammit:


chriadam:
Is this for key navigation support?

blammit:
Yes, this will simplify things a lot for key navigation.

MikeTrahearn-Qinetic:
I kind of see what you did here, but it is a huge compromise when you could simply manage the focus on two ListViews - it can't be that hard.

blammit:
> I kind of see what you did here, but it is a rather complex solution considering ListViews are themselves FocusScopes.

Using only a single ListView will be a general improvement as well, as it greatly simplifies the model/view here. Having the non-interactive ListView within the header of another ListView already causes other problems. For example:

- it cannot make use of the cacheBuffer and delegate creation optimizations, since it always creates its delegates regardless of whether they are shown, as the height is sized to ensure it always shows the available contentHeight
- if the header ListView has more items than what fits on-screen, and you set its currentIndex, it cannot auto-scroll to show the currentIndex as expected, since the height is fixed and the view is within the header of another ListView; you could add some hooks to scroll the parent ListView contentY, but it would be far simpler to only have one currentIndex to track within a single view

Used a nested ListView is a convenient solution if a single model/view is not possible, but here it is quite straightforward to have the single model/view so it would be nicer to move to use that instead.

chriadam (approved):
LGTM

PR #2049, opened 2025-04-17
Rework Environments tab and ExpandedTanksView

blammit:
This is part of the preparation for key navigation.

### DIFF
--- a/pages/EnvironmentTab.qml
+++ b/pages/EnvironmentTab.qml
@@ -5,51 +5,27 @@
 
 import QtQuick
 import Victron.VenusOS
-import QtQuick.Controls.impl as CP
 import Victron.Gauges
 
-Flickable {
+LevelsTab {
 	id: root
 
-	property bool animationEnabled: true
+	readonly property int twoGaugeWidth: Gauges.width(Global.environmentInputs.model.count, 4, Theme.geometry_screen_width)
+	readonly property int oneGaugeWidth: Gauges.width(Global.environmentInputs.model.count, 6, Theme.geometry_screen_width)
 
-	width: parent.width
-	height: parent.height
-	contentWidth: contentRow.width
-	leftMargin: levelsRepeater.count > 3
-			? Theme.geometry_levelsPage_environment_horizontalMargin
-			: (width - contentRow.width) / 2
-	rightMargin: Theme.geometry_levelsPage_environment_horizontalMargin
-	boundsBehavior: Flickable.StopAtBounds
-	contentX: -leftMargin   // shouldn't be needed, but initial value may be incorrect due to delegate resizing
+	model: Global.environmentInputs.model
+	delegate: EnvironmentGaugePanel {
+		required property Device device
 
-	property int twoGaugeWidth: Gauges.width(levelsRepeater.count, 4, root.width)
-	property int oneGaugeWidth: Gauges.width(levelsRepeater.count, 6, root.width)
-
-	Row {
-		id: contentRow
-
-		height: parent.height
-
-		spacing: Gauges.spacing(levelsRepeater.count)
-
-		Repeater {
-			id: levelsRepeater
-			model: Global.environmentInputs.model
-
-			delegate: EnvironmentGaugePanel {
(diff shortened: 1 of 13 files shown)


pages/EnvironmentTab.qml
```diff
@@ -5,51 +5,27 @@
...
-				title: model.device.name
-				temperature: Global.systemSettings.convertFromCelsius(model.device.temperature)
-				temperatureType: model.device.temperatureType
-				humidity: model.device.humidity
-				temperatureGaugeGradient: temperatureGradient
-				humidityGaugeGradient: humidityGradient
-			}
-		}
+		width: hasTwoGauges ? root.twoGaugeWidth : root.oneGaugeWidth
+		height: Gauges.height(!!Global.pageManager && Global.pageManager.expandLayout)
```
MikeTrahearn-Qinetic, line 21: Maybe use `Global.pageManager?.expandLayout` as per LevelsPage.qml if the logic is the same

pages/LevelsPage.qml
```diff
@@ -64,36 +64,11 @@ SwipeViewPage {
 
 		anchors {
 			top: tabBar.bottom
-			topMargin: (!!Global.pageManager && Global.pageManager.expandLayout)
```
MikeTrahearn-Qinetic, line 67: Are there any tokens that can be removed as a result of these changes?

pages/LevelsTab.qml
```diff
@@ -0,0 +1,37 @@
...
+BaseListView {
+	id: root
+
+	property bool animationEnabled: true
+
+	// QTBUG-102340: Horizontal ListView ignores topMargin, so set anchors.topMargin instead
+	anchors.topMargin: Global.pageManager?.expandLayout
+			? Theme.geometry_levelsPage_gaugesView_expanded_topMargin
+			: Theme.geometry_levelsPage_gaugesView_compact_topMargin
+	bottomMargin: Global.pageManager?.expandLayout
```
MikeTrahearn-Qinetic, line 19: missing `anchors.` for bottomMargin, leftMargin, rightMargin?

pages/EnvironmentTab.qml
blammit: Fixed, thanks

pages/LevelsPage.qml
blammit: Yes, updated to remove unused values

pages/LevelsTab.qml
blammit: This is setting the margins for the ListView rather than the anchors, since the anchor points are not set from LevelsTab (only the anchor margins). Ideally the top margin would be set as the ListView `topMargin` rather than `anchors.topMargin`, but this is not possible due to QTBUG-102340 (as noted).

pages/EnvironmentTab.qml
```diff
@@ -5,51 +5,27 @@
...
-				height: Gauges.height(!!Global.pageManager && Global.pageManager.expandLayout)
-				title: model.device.name
-				temperature: Global.systemSettings.convertFromCelsius(model.device.temperature)
-				temperatureType: model.device.temperatureType
-				humidity: model.device.humidity
-				temperatureGaugeGradient: temperatureGradient
-				humidityGaugeGradient: humidityGradient
-			}
-		}
+		width: hasTwoGauges ? root.twoGaugeWidth : root.oneGaugeWidth
```
chriadam, line 20: I see `qrc:/qt/qml/Victron/VenusOS/pages/EnvironmentTab.qml:20:3: Binding loop detected` when running in mock mode at least

chriadam:
Seems to work great, really nice that the expanded tanks view can be scrolled now!  But, one binding loop warning, because LevelsTab.qml uses contentWidth to determine leftMargin etc etc and thus affects width, but the width value is used to calculate the delegate width in EnvironmentTab.qml

blammit:
Thanks, fixed the binding loop issue by referring to the model count rather than the view count, as the latter would affect the contentWidth.

chriadam (approved):
Excellent, LGTM

PR #2056, opened 2025-04-22
Do not display TextField if read-only. Fixes #2055

### LINKED ISSUE #2055: Input fields not displayed correctly if read-only
The input fields are not displayed correctly, if the property `interactive` is `false`. Then the input box and the list item is shown.

See also #2016.

Image

mr-manuel:
Fixes #2055

Currently the text field and the read-only label are displayed, if `interactive` is `false`.

grafik

grafik

grafik

### DIFF
--- a/components/listitems/core/ListTextField.qml
+++ b/components/listitems/core/ListTextField.qml
@@ -122,10 +122,10 @@ ListItem {
 		property bool _validateBeforeSaving
 
 		enabled: root.clickable
+		visible: root.clickable
 		width: Math.max(Theme.geometry_listItem_textField_minimumWidth,
 						Math.min(Theme.geometry_listItem_textField_maximumWidth,
 								 implicitWidth + leftPadding + rightPadding))
-		visible: root.enabled
 		text: dataItem.valid ? dataItem.value : ""
 		rightPadding: suffixLabel.text.length ? suffixLabel.implicitWidth : leftPadding
 		horizontalAlignment: root.suffix ? Text.AlignRight : Text.AlignHCenter
@@ -200,7 +200,7 @@ ListItem {
 
 		text: textField.text.length > 0 ? textField.text : "--"
 		width: Math.min(implicitWidth, root.maximumContentWidth)
-		visible: !root.interactive
+		visible: !textField.visible
 	}
 
 	VeQuickItem {


components/listitems/core/ListTextField.qml
```diff
@@ -125,7 +125,7 @@ ListItem {
 		width: Math.max(Theme.geometry_listItem_textField_minimumWidth,
 						Math.min(Theme.geometry_listItem_textField_maximumWidth,
 								 implicitWidth + leftPadding + rightPadding))
-		visible: root.enabled
+		visible: root.enabled && root.interactive
```
blammit, line 128: @MikeTrahearn-Qinetic should this be `visible: root.interactive` instead?

Also related to 7f82b4f9a21bba95a544a7dcddaeb0f998d69b4c.

MikeTrahearn-Qinetic: It all depends on the requirements: if you want to make TextField's disappear across the board when not enabled, that's a pretty severe change and raises the question of whether you should be using a naturally readonly Label in such a case. I'm not sure that this is a good idea. @mr-manuel could you please ensure you understand which properties you want to set and their effect - I'm not quite sure if I should respond with "it works like it is supposed to" or not. But making every single ListTextField disappear is probably not the right way to "solve" this.

blammit: There is a naturally readonly Label that is displayed when the ListTextField is not enabled,  as per 7f82b4f9a21bba95a544a7dcddaeb0f998d69b4c. The problem here is that `ListTextField { interactive: false }` will result in both the `TextField` and the `SecondaryListLabel` being displayed, which is definitely not what we want. This bug appears to be a result of the enabled/interactive changes from 911f826df2019739c99759695d65cec60e979dc0 - should this reference to `enabled` also have been replaced with `interactive`?

mr-manuel: @blammit @MikeTrahearn-Qinetic any news on this?

chriadam: Just looking at the code and the changes Bea linked, I suggest this should be `visible: root.interactive` as making the primary label disappear when not interactive is indeed the desired behaviour (note that the secondary label will appear in that case).  I have pinged Mike, he's been away for a bit, but hopefully he has a chance to check this and confirm.

mr-manuel: I have also relooked it and tested it. Would now also go with `visible: root.interactive`. I updated the PR.

Tests were ok.

components/listitems/core/ListTextField.qml
```diff
@@ -125,7 +125,7 @@ ListItem {
 		width: Math.max(Theme.geometry_listItem_textField_minimumWidth,
 						Math.min(Theme.geometry_listItem_textField_maximumWidth,
 								 implicitWidth + leftPadding + rightPadding))
-		visible: root.enabled
+		visible: root.interactive
```
DanielMcInnes, line 128: Would you expect the primary label to appear if the user didn't have write access? 
If you don't want it to appear if the user doesn't have write access, I think it should be: 
`visible: root.clickable`

mr-manuel: If the user has write access and interactive is true this should appear:
```
	property TextField defaultContent: TextField {
		id: textField
```

If the user don't have write access or interactive is false, this should appear:
```
	SecondaryListLabel {
		id: readonlyLabel
```

So in this case `visible: root.clickable` is not the correct one (when access level is user, the fields disappear).

With `visible: root.interactive` the text field is visible as user, but I cannot edit it, which seems correct. Here the question is, should it be displayed as value that could be edited with the correct permissions or should it only display the text label instead of the input label?

grafik

mr-manuel:
Together with @DanielMcInnes I looked at that and fixed it now in a different way.

## Result

**interactive: false**

grafik

**interactive: true and user has no write permissions**

grafik

**interactive: true and user has write permissions**

grafik

DanielMcInnes (approved):

chriadam:
Excellent, thanks.  Merging.

PR #2057, opened 2025-04-22
Optimize build process

mr-manuel:
I optimized the build scripts and GitHub action workflow.

GitHub Actions: Instead of about 11 minutes, it takes now about 6 minutes
Local PC: Instead of about 11 minutes, it takes now about 2 minutes

Is there any reason for not using `--parallel $(nproc)` with `cmake --build`?

I also added an option to directly upload the compiled files to the GX device by using `-H venus-ip-or-hostname` or `--host venus-ip-or-hostname` This checks, if there is already an SSH key present and if not generates one and uploads it. If there is already one present it only uploads it, if missing.

It's also possible to give more than one host for file upload by providing a comma separated list:
`-H 192.168.1.10,192.168.1.11,192.168.1.12`

### DIFF
--- a/.github/workflows/build-wasm.yml
+++ b/.github/workflows/build-wasm.yml
@@ -39,20 +39,21 @@ jobs:
 
       - name: Check for a cached version
         uses: actions/cache@v4
-        id: cached_qt_emscripten
+        id: cached-versions
         with:
           path: |
+            # Cache the Qt and emscripten installation
             ${{ env.OUTPUTDIR }}
           # Adding version as cache key
-          key: ${{ runner.os }}-qt-${{ env.QT_VERSION }}-em-${{ env.EMSCRIPTEN }}-libegl1
+          key: ${{ runner.os }}__qt-${{ env.QT_VERSION }}__em-${{ env.EMSCRIPTEN }}__libegl1
 
-      - name: Install requirements
-        if: steps.cached_qt_emscripten.outputs.cache-hit != true
+      - name: Install requirements (cache miss)
+        if: steps.cached-versions.outputs.cache-hit != 'true'
         run: |
           bash scripts/build-wasm-install-requirements.sh
 
-      - name: Install libraries needed for building the wasm
-        if: steps.cached_qt_emscripten.outputs.cache-hit == true
+      - name: Install libraries needed for building the wasm (cache hit)
+        if: steps.cached-versions.outputs.cache-hit == 'true'
         run: |
           sudo apt-get update -yq
           sudo apt-get install -y libegl1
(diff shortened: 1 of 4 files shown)


.github/workflows/build-wasm.yml
```diff
@@ -39,20 +39,23 @@ jobs:
...
       - name: Check for a cached version
         uses: actions/cache@v4
-        id: cached_qt_emscripten
+        id: cached-versions
         with:
           path: |
+            # Cache the Qt and emscripten installation
             ${{ env.OUTPUTDIR }}
+            # Cache the build directory
+            ./build-wasm
```
chriadam, line 48: Will this cache gui-v2 (wasm) build artifacts from previous runs?  Or am I misunderstanding?  (I would prefer entirely clean builds of gui-v2.)

chriadam:
Parallel builds in the past did cause some problem on some target (with translation targets being serviced in the wrong order, if I remember correctly?)  @DanielMcInnes do you remember what that issue was, and whether it's completely resolved now or not?

.github/workflows/build-wasm.yml
mr-manuel: Yes thats correct. Then I will remove it again.

DanielMcInnes:
> Parallel builds in the past did cause some problem on some target (with translation targets being serviced in the wrong order, if I remember correctly?) @DanielMcInnes do you remember what that issue was, and whether it's completely resolved now or not?



> Parallel builds in the past did cause some problem on some target (with translation targets being serviced in the wrong order, if I remember correctly?) @DanielMcInnes do you remember what that issue was, and whether it's completely resolved now or not?

I think that issue is resolved now, via `add_custom_target(${PROJECT_NAME}_lupdate_sleep`.

scripts/build-gx.sh
```diff
@@ -33,23 +33,13 @@ echo "Changed to parent directory: $(pwd)"
...
-# git clean -fd
-
-# Cleanup old build directory
-if [ -d "build-gx" ]; then
-    rm -rf "build-gx"
-fi
-
 # Create build directory
-mkdir "build-gx"
+if [ ! -d "build-gx" ]; then
```
DanielMcInnes, line 40: does this give us possible stale builds?

mr-manuel: It's the same as @chriadam wrote for the `build-wasm`. If that is the case, I would remove both.

mr-manuel:
@chriadam @DanielMcInnes as default option I removed the cache for the build. If a user wants to use it, he can add the `-P` or `--preserve` argument.

.github/workflows/build-wasm.yml
```diff
@@ -39,20 +39,21 @@ jobs:
 
       - name: Check for a cached version
         uses: actions/cache@v4
-        id: cached_qt_emscripten
+        id: cached-versions
```
DanielMcInnes, line 42: @chriadam do you think these scripts should clean the qml disk cache? Possibly something for another ticket

DanielMcInnes (approved):

DanielMcInnes:
the changes look good to me, but I would like @dirkjanfaber to have a look, he knows more about this stuff

dirkjanfaber:
> the changes look good to me, but I would like @dirkjanfaber to have a look, he knows more about this stuff

I don't see anything obviously breaking anything. 
Only remark I do have is that I generally prefer the variables to be enclosed in curly brackets. So
```
        ssh root@${HOST} "/opt/victronenergy/swupdate-scripts/remount-rw.sh"
```
instead of
```
        ssh root@$HOST "/opt/victronenergy/swupdate-scripts/remount-rw.sh"
```

(And on all of the other lines).

mr-manuel:
@dirkjanfaber all variables are now in curly brackets

@DanielMcInnes could you merge this now? Then I can update the wiki.

PR #2058, opened 2025-04-23
Add support for OrionXS as .dcdc service

### LINKED ISSUE #2026: Add support for OrionXS as .dcdc service
[vedirect_interface commit 76b6848](https://github.com/victronenergy/vedirect_interface/commit/76b6848962a9ecc5ac01ac439633e9486d1483ee) adds support for the OrionXS to be connected with its input to the main battery and the output to some auxiliary battery. Whether this is the case or not is determined by a setting in localsettings (`/Setttings/Victron/Products/##serialnr##/OutputBattery`) which is mirrored by the service itself as `/Settings/OutputBattery`. When this setting is 0, there is no change in the behaviour, in the sense that it is presented as a .alternator service. When the setting is 1, the OrionXS will be presented on the dbus as .dcdc service and without the `/Link/*` dbus paths (so DVCC will not interfere). 

Compared to an OrionXS that is configured as PSU (which is also presented as a .dcdc service), an OrionXS configured as charger will have the following extra paths:
`/Settings/OutputBattery` => This will always be 1 for the .dcdc service, but can be set to 0 to configure the OrionXS as being connected to the main battery with its output. When this is changed to 0, it disconnects from the dbus and restarts as a .alternator
`/History/*` => This is contains the charge cycle history, just as when it is a .alternator service

When the OrionXS is configured as .alternator, the `/Settings/OutputBattery` setting is also available and is always 0, but can be changed to 1.

In summary, what needs to be done:

- [ ] Show and allow user to change /Settings/OutputBattery for a .alternator service
- [ ] Conditionally show and allow user to change /Settings/OutputBattery for a .dcdc service
- [ ] Conditionally show the charger history for a .dcdc service

This requires issue #2019 to be solved.

blammit:
The OrionXS can be configured to change from an .alternator service to a .dcdc service, by changing /Settings/OutputBattery from 0 to 1. When it is a .dcdc service, it is connected with its input to the main battery and the output to some auxiliary battery.

To support this:
- allow the /Settings/OutputBattery setting to be changed from alternator and dcdc settings pages
- since a dcdc service can now have a charger history, move the charger history implementation from the alternator settings page into a common DcHistorySettingsColumn that can be shown in both the alternator and dcdc settings pages
- add ListSwitch toggled() signal so that a toast can be shown when the service type changes between alternator and dcdc

Fixes #2026

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -338,6 +338,7 @@ set (VENUS_QML_MODULE_SOURCES
     components/listitems/ListLink.qml
     components/listitems/ListMqttAccessSwitch.qml
     components/listitems/ListMountStateButton.qml
+    components/listitems/ListOutputBatterySwitch.qml
     components/listitems/ListPvInverterPositionRadioButtonGroup.qml
     components/listitems/ListEvChargerPositionRadioButtonGroup.qml
     components/listitems/ListRebootButton.qml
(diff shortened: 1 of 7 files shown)


DanielMcInnes:
is there an installation where I can test this?

blammit:
@DanielMcInnes yes, https://vrm.victronenergy.com/installation/99087/dashboard

DanielMcInnes (approved):

Issue #2053, opened 2025-04-21
Offline tanks are included in tank "total" calculations


blammit:
If a tank is offline, it should not be included in the total capacity/remaining/level calculations.

See [thread](https://victrondevelopment.slack.com/archives/C020QG3JVEW/p1745062410770269) where are two Freshwater tanks with 2000 litre and 800 litre capacities respectively. The 2000 litre tank is at 100%, and the 800 litre tank is offline (as it is in an error state), so the total Freshwater % should be 100%, not 50%.

To fix:
- Move the 'state' property from Tank.qml to BaseTankDevice
- Change BaseTankDeviceModel to only include calculations for tanks with Status=0 (i.e. "OK")

For testing, verify:
- The total % shown in the Brief page
- The total %, remaining and capacity shown in the Tanks tab if the tanks are merged into a group

blammit:
Fixed by https://github.com/victronenergy/gui-v2/pull/2059

PR #2059, opened 2025-04-23
Don't include offline tanks in tank totals calculations

chriadam:
Also visualize the status of non-OK tanks in the levels page via dimmed surface color, plus warning-colorized icon.

Contributes to issue #2053


src/basetankdevicemodel.cpp
```diff
@@ -123,7 +126,8 @@ void BaseTankDeviceModel::updateTotals()
 	bool requireFallback = false;
 
 	for (int i = 0; i < count(); ++i) {
-		if (const BaseTankDevice *tank = tankAt(i)) {
+		if (const BaseTankDevice *tank = tankAt(i);
```
DanielMcInnes, line 129: is this intentional? I don't think I've ever seen this syntax before: `if (exprA; exprB) doSomething()`

DanielMcInnes (approved):

chriadam:
After rebase, I have broken a unit test.  Will investigate.

blammit (approved):

Issue #2009, opened 2025-03-31
Font size for the overview page


dirkjanfaber:
Noticed via https://community.victronenergy.com/t/venus-os-3-52-caractere-size-on-gui-v2 in combination with French. The "discharging" translation doesn't fit in the box and overflows to the next line.
Talked to the translator and we both agree that this should be fixed on the gui side, rather than the translation side. Shrinking the font a bit, based on the string length seems easiest. Do note that at the moment "External control" can be in the same tile, which does stretch over 2 lines and thus fits. I am sure there are standard solutions for this kind of thing. 

![Image](https://github.com/user-attachments/assets/0778e20b-af2e-481d-9b77-84f35fe5698e)

![Image](https://github.com/user-attachments/assets/64a063d0-c41d-461b-9abd-61e42fe9aebc)

(For the daily yield, there is an alternative translation).

chriadam:
Merged, closing as fixed

PR #2062, opened 2025-04-23
Inverter/Charger widget: wrap on word boundaries + fit font size

chriadam:
To prevent strange rendering of longer (especially translated) strings, only wrap at word boundaries.  Also, enable both horizontal and vertical font fitting.

We still allow eliding (for very long terms which still don't fit using the minimum-allowed font size).

Contributes to issue #2009


dirkjanfaber (approved):
Looks good to me.

Issue #2043, opened 2025-04-15
Delay showing warning icon on reload (especially affects tablet on wakeup)


chriadam:
As per slack discussion:

When a tablet goes to sleep, the web assembly tab will lose its websocket, and then when it is woken up the websocket will attempt to reconnect.

Currently, while it is attempting to reconnect it will show a warning icon at the splash screen, saying "device has lost connection to VRM" or similar.

Instead of showing that warning icon, it should only show the "reconnecting to VRM" for a few seconds, and only switch to the warning display if it isn't able to reconnect within a few seconds.

chriadam:
Merged, closing as fixed.

PR #2063, opened 2025-04-23
Delay showing error icons and labels on splash view

chriadam:
In some common cases (i.e. wasm tab being woken up, and having to re-establish websockets) the application will be in an "expected" error state.  We should delay showing the error icons and labels until some time has passed, to ensure that the error state isn't purely transient.

Contributes to issue #2043


blammit (approved):
The SplashView state changes/interactions are getting complicated. It would be nice to refactor this later on.

Issue #1987, opened 2025-03-17
Digital inputs: Cannot set multiplier or inverted setting when configured as Pulse Meter.


izak:
I confirmed this myself:

![Image](https://github.com/user-attachments/assets/9210e061-fe9b-4013-b93d-d18c7c81f7a7)

chriadam:
Speculative fix is in PR above.  I am not sure how to test this one without a pulse meter device.

blammit:
@chriadam  This can be tested with the boat/motorhome 2 demo. If you go to Settings > Integrations > Digital I/O, you can configure one of the digital inputs as a pulse meter.

blammit:
@chriadam I've updated the PR and also added a fix for the Multiplier precision.

PR #2064, opened 2025-04-23
Fix uid prefix in pulse meter setup page

chriadam:
Information about the appropriate path may be found here: https://github.com/victronenergy/dbus-digitalinputs/blob/master/README.md#running

Contributes to issue #1987

### DIFF
--- a/pages/settings/devicelist/pulsemeter/PagePulseCounterSetup.qml
+++ b/pages/settings/devicelist/pulsemeter/PagePulseCounterSetup.qml
@@ -26,6 +26,8 @@ Page {
 				//% "Multiplier"
 				text: qsTrId("pulsecounter_setup_multiplier")
 				dataItem.uid: Global.systemSettings.serviceUid + "/Settings/DigitalInput/" + root.inputNumber + "/Multiplier"
+				decimals: 6
+				stepSize: Math.pow(10, -decimals)
 			}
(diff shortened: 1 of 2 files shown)


chriadam:
Help testing this one would be appreciated...

pages/settings/devicelist/pulsemeter/PagePulseCounterSetup.qml
```diff
@@ -10,6 +10,7 @@ Page {
 	id: root
 
 	property string bindPrefix
+	property int inputNumber: parseInt(bindPrefix.substring(bindPrefix.indexOf(".input0")))
```
blammit, line 13: This won't work on MQTT/Wasm uids as the uid will be like `mqtt/pulsemeter/x` instead of `com.victronenergy.pulsemeter.input0x`. Looking at gui-v1, the input number can be determined by `/DeviceInstance`, with 0 as a fallback: https://github.com/victronenergy/gui/blob/master/qml/PagePulseCounter.qml#L10

blammit: @chriadam  I've updated the PR and also added a fix for the Multiplier precision.

chriadam:
Thanks Bea, LGTM

PR #2080, opened 2025-04-29
Adjust nav button width and fix Levels re-insertion

blammit:
- Set nav button width according to the number of items in the bar
- If Levels page is removed and reinserted while Boat page is visible, ensure the Levels page is removed rather than the Overview page, and always insert Levels before Notifications page


chriadam (approved):
LGTM

PR #2082, opened 2025-04-30
Aux cards: update control values when backend values change

blammit:
The latching button currentIndex binding is cleared when the user clicks it to change the value. If the backend later changes the output state, ensure the currentIndex is updated again.

Simiarly, the dimming slider value binding is cleared when the user drags the slider, as that sets the value. If that happens, and the backend later changes the dimming value, ensure the slider value is updated again. Also, only update the dragged value when the slider is not pressed, to avoid conflicting data between the user input and the backend value.

### DIFF
--- a/components/SwitchableOutputDelegate.qml
+++ b/components/SwitchableOutputDelegate.qml
@@ -107,18 +107,47 @@ Item {
 		id: dimmingSlider
 
 		DimmingSlider {
+			id: slider
+
+			property real movedValue: NaN
+
 			width: root._buttonWidth
 			height: Theme.geometry_switchableoutput_button_height
 			highlightColor: output.state === 1 ? Theme.color_ok : Theme.color_button_down
 			from: 1
 			to: 100
-			value: output.hasDimming ? output.dimming : 0
-			onPositionChanged:{
-				value = Math.round(value) // break binding so that the delayed data update doesn't fight user input
-				output.setDimming(value)
-			}
 			onClicked: {
 				output.setState(output.state === 0 ? 1 : 0)
+			}
+			onMoved: {
+				value = Math.round(value)
+				output.setDimming(value)
+				movedValue = value
+			}
+
+			Connections {
+				target: output
+				Component.onCompleted: slider.value = output.dimming
+
+				// Update the slider value from the backend. If the backend value is changing in
+				// response to slider user input, the backend change may lag behind the user change,
+				// so do not update the slider value until the slider and backend values are in
+				// sync, else the two values will fight each other.
+				function onDimmingChanged() {
+					if (isNaN(slider.movedValue)) {
+						// The slider was not moved, so the value change must have come from
+						// the backend independently, and the slider value can updated directly.
+						slider.value = output.dimming
+					} else if (slider.movedValue === Math.round(output.dimming)) {
+						// The user moved the slider and the backend is now up-to-date with the
+						// user-moved value, so clear the flag.
+						slider.movedValue = NaN
+					} else {
+						// The slider was moved, and the backend value changed in response, but
+						// the backend still has an older moved value from the UI, so ignore
+						// this as the backend is not yet up-to-date with the UI value.
+					}
+				}
 			}
 
 			Label {
@@ -150,6 +179,7 @@ Item {
 		id: latchingButton
 
 		SegmentedButtonRow {
+			id: buttonRow
 			width: root._buttonWidth
 			height: Theme.geometry_switchableoutput_button_height
 			fontPixelSize: Theme.font_size_body2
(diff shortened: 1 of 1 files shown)


components/SwitchableOutputDelegate.qml
```diff
@@ -107,20 +107,36 @@ Item {
...
-			}
 			onClicked: {
 				output.setState(output.state === 0 ? 1 : 0)
 			}
 
+			// Only update the slider when the user is not dragging it, otherwise the backend data
+			// change fights the current dragged value.
+			onPressedChanged: {
+				if (!pressed) {
+					output.setDimming(Math.round(value))
```
chriadam, line 126: should this just be `output.dimming = Math.round(value)`?

components/SwitchableOutputDelegate.qml
```diff
@@ -107,20 +107,36 @@ Item {
...
 			value: output.hasDimming ? output.dimming : 0
-			onPositionChanged:{
-				value = Math.round(value) // break binding so that the delayed data update doesn't fight user input
-				output.setDimming(value)
-			}
 			onClicked: {
 				output.setState(output.state === 0 ? 1 : 0)
 			}
 
+			// Only update the slider when the user is not dragging it, otherwise the backend data
```
chriadam, line 122: Suggestion: change comment from "Only update the slider when the user is not dragging it" to "Only update the slider to match backend value updates when the user is not dragging it".  Secondly: I hope that we still do `output.dimming = Math.round(value)` every time the slider value changes while we are pressing it, in order to get instant feedback as we drag the slider?

components/SwitchableOutputDelegate.qml
CJhuckstepp: I feel this need to be closer to the original code as I suspect it is the only way both the users live update and dbus change update works.

      I.E.

 	DimmingSlider {
			id: slider
			width: root._buttonWidth
			height: Theme.geometry_switchableoutput_button_height
			highlightColor: output.state === 1 ? Theme.color_ok : Theme.color_button_down
			from: 1
			to: 100
			value: output.dimming //hasDimming checked in switchableOutput
			onClicked: {
				output.setState(output.state === 0 ? 1 : 0)
			}

			//keep value = output.dimming binding but prevent created loop by checking if changed
			onPositionChanged:{
				var newVal = Math.round(value)
				if (newVal !== output.dimming) output.setDimming(newVal )
			}

			Label {
				anchors.centerIn: parent
				text: output.state === 1 ? CommonWords.on : CommonWords.off
				font.pixelSize: Theme.font_size_body2
			}
		}

Tested with two Ekrano GX  and the SmartSwitch connected to the same CAN bus

blammit: > should this just be output.dimming = Math.round(value)?

The `dimming` property is read-only as it comes directly from the VeQuickItem, so `setDimming()` has to be used to modify the value.

@CJhuckstepp I tested this approach with the GIOExtender via MQTT/Wasm and found that the backend value change and the user input would still fight each other, as the backend is not able to keep up with the constant setValue() calls, so it lags behind and then the delayed update will change the slider value, which contradicts the user-dragged value (which is more recent). It would be nice to update the backend value while the user is dragging it, instead of waiting until the mouse is released, but I'm not sure how to do this when there is such a delay.

blammit: @CJhuckstepp we discussed this issue further in a meeting today, and came up with this proposal for changes:

- always set the backend value while dragging the slider 
- to avoid the conflict/fighting, do not update the slider value from the backend while the user is dragging
- when the user releases the slider, update the slider value to the one from the backend (we might need a short delay here, otherwise there will be a few delayed jumps in the slider value, as the backend value is settling)

chriadam: Discussed in the meeting - we will do instant updates to the backend from the slider value updates, but ignore value updates from the backend while the user is dragging the slider (and for a short period of time after release), to allow instant feedback etc, while preventing the source conflict.

CJhuckstepp: Ok that explains what is going on. It work with the smart switch as the CAN driver starts a lockout timer when the VE item value is updated and send the new value to the device. Any values received from the device are  then ignored until the lock out timer expires.  As old values are likely to be in transit. I we see it martin/Rein can implement a similar thing in the GX IO module driver

blammit: Right, I see, it makes sense to have the lock out timer. It would be nice to have this handled in the driver, but in case it is not done there, I have updated the PR to make sure the values are in sync so we avoid the conflict on the UI level as well.

components/SwitchableOutputDelegate.qml
```diff
@@ -107,18 +107,47 @@ Item {
...
 			highlightColor: output.state === 1 ? Theme.color_ok : Theme.color_button_down
 			from: 1
 			to: 100
-			value: output.hasDimming ? output.dimming : 0
-			onPositionChanged:{
+			onClicked: {
+				output.setState(output.state === 0 ? 1 : 0)
+			}
+			onMoved: {
 				value = Math.round(value) // break binding so that the delayed data update doesn't fight user input
```
chriadam, line 123: can remove the comment about breaking the binding, as there is no binding any more

chriadam (approved):

components/SwitchableOutputDelegate.qml
blammit: Thanks, removed

chriadam:
Thanks!  Actually, thinking of merging this now and testing can be done with the release we make this afternoon.

DanielMcInnes (approved):

PR #2092, opened 2025-05-02
Fix installation of boat page components

chriadam:


chriadam:
Philipp tested.  Self-approving and merging.

Issue #2093, opened 2025-05-04
boat overview: fix speed wrong & add decimal


mpvader:
https://victrondevelopment.slack.com/archives/C020QG3JVEW/p1746287018020769

mpvader:
I've marked this high prio, get it into v3.60 venus os official, and under the assumption that fixing this is super trivial.

the issue will how to test it

philipptrenz:
@DanielMcInnes you can use this site for testing, it has a simulated GPS: https://vrm.victronenergy.com/installation/717596/

chriadam:
PR merged to main, closing as fixed.

PR #2094, opened 2025-05-05
Fix bugs in center speed label in boat page

chriadam:
The default source unit is metres per second, not km/h. The label should allow 1 decimal of precision.
The label should use font-fitting to avoid overspill.

Contributes to issue #2093


philipptrenz:
@chriadam I tested and speed values are correct now. But there is one more thing: The label for the max speed must also be changed to `m/s`. It's a bit unintuitive to enter max speed in meters per second, but I think it's tolerable for v1, as we still have the auto-scaling feature and if done manually, it's most certainly a boat manufacturer and not an end customer.

192 168 178 96_gui-v2_(GX Touch 50) (1)

philipptrenz:
@mpvader What do you think, can we (for aesthetic reasons) omit the decimals for values ≥ 10?

192 168 178 96_gui-v2_(GX Touch 50)

mpvader:
Hi @philipptrenz yes sure.

DanielMcInnes (approved):
lgtm. During testing, I found a bug in PageSettingsDisplayMinMax. We weren't converting the max speed from m/s to system units. I pushed a fix onto this branch.

DanielMcInnes (approved):

philipptrenz:
> lgtm. During testing, I found a bug in PageSettingsDisplayMinMax. We weren't converting the max speed from m/s to system units. I pushed a fix onto this branch.

Hi @DanielMcInnes, wasn't planned to have this but super nice that it's now there, thanks!

Issue #2085, opened 2025-05-01
Brief page levels selection indexing off-by-one


chriadam:
See https://community.victronenergy.com/t/3-60-58-brief-setup-issue/28529

Customer reports that selecting the first temperature sensor as the center detail, and then selecting a battery as the center detail, results in the temperature continuing to be shown.  They managed to work around the issue by selecting the second temperature sensor as the center detail, and then selecting a battery as the center detail.

chriadam:
Merged to main, closing as fixed

PR #2096, opened 2025-05-05
Fix off-by-one in brief page center details selection dialog

chriadam:
The deviceOptionModel has an activeBatteryName option injected at index 0, so subsequent device indexes should be increased by one.

Contributes to issue #2085

### DIFF
--- a/components/listitems/ListBriefCenterDetails.qml
+++ b/components/listitems/ListBriefCenterDetails.qml
@@ -35,7 +35,7 @@ ListNavigation {
 				section: qsTrId("settings_briefview_center_temperature_services")
 			})
 			if (selectedIndex < 0 && portableServiceId === centerService.value) {
-				selectedIndex = i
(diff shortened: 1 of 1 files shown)


DanielMcInnes (approved):

components/listitems/ListBriefCenterDetails.qml
```diff
@@ -35,7 +35,7 @@ ListNavigation {
 				section: qsTrId("settings_briefview_center_temperature_services")
 			})
 			if (selectedIndex < 0 && portableServiceId === centerService.value) {
-				selectedIndex = i
+				selectedIndex = i+1
```
blammit, line 38: Could you add a brief comment for why the +1 is needed?

blammit (approved):
Nice, minor comment, LGTM.

components/listitems/ListBriefCenterDetails.qml
chriadam: done

PR #2097, opened 2025-05-05
Allow user to update the security profile password easily

chriadam:
Add a button which allows the user to update the password without changing the security profile selection.

Contributes to issue #1993

### DIFF
--- a/components/dialogs/SecurityProfilePasswordDialog.qml
+++ b/components/dialogs/SecurityProfilePasswordDialog.qml
@@ -1,0 +1,158 @@
+/*
+** Copyright (C) 2025 Victron Energy B.V.
+** See LICENSE.txt for license information.
+*/
+
+import QtQuick
+import QtQuick.Layouts
+import Victron.VenusOS
+
+ModalDialog {
+	id: root
+
+	property string password
+
+	function validate() : bool {
+
+		root.password = ""
+
+		if (firstPassword.length < 8) {
+
+			// First, validate that the firstPassword is at least 8 characters long.
+			// If not, the border of both firstPassword AND secondPassword turn red and
+			// the passwordHint warning about length, becomes visible.
+
+			firstPassword.borderColor = Theme.color_critical
+			secondPassword.borderColor = Theme.color_critical
+			passwordHint.visible = true
+			//% "Password needs to be at least 8 characters long"
+			passwordHint.text = qsTrId("settings_security_profile_password_incorrect_length")
+
+			return false
+		}
+
+		if (firstPassword.text !== secondPassword.text) {
+			// Secondly, validate that the passwords in both firstPassword AND secondPassword are equal.
+			// If not, the border of both firstPassword AND secondPassword turn red and
+			// the passwordHint warning about mismatch, becomes visible.
+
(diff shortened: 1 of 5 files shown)


chriadam:
@philipptrenz testing / feedback appreciated

chriadam:
I haven't yet tested on a real device, as the power is off at the office today, so cannot use our cerbo to test this properly unfortunately :-/  So, only tested with mock mode.  Testing appreciated!

chriadam:
Thanks for the fix, Mike!  I've now tested on CerboGX and it seems to work nicely.  More testing is of course appreciated.  I will send a build to Philipp for extra testing / feedback.

chriadam:
Philipp asked me to simplify the update dialog by only asking for a single password to be entered.  I have updated the PR with this change now, also.

philipptrenz:
@chriadam I tested the latest changes and it looks and works as expected 👍

philipptrenz (approved):

components/dialogs/SecurityProfilePasswordDialog.qml
```diff
@@ -0,0 +1,123 @@
...
+	//% "Confirm"
+	acceptText: qsTrId("modaldialog_confirm")
+
+	dialogDoneOptions: VenusOS.ModalDialog_DoneOptions_OkAndCancel
+
+	Component.onCompleted: resetValidation()
+
+	property string password
+
+	function valiate() : bool {
```
blammit, line 25: typo "valiate"

components/dialogs/SecurityProfilePasswordDialog.qml
```diff
@@ -0,0 +1,123 @@
...
+	title: qsTrId("settings_security_profile_change_password_title")
+
+	//% "Confirm"
+	acceptText: qsTrId("modaldialog_confirm")
+
+	dialogDoneOptions: VenusOS.ModalDialog_DoneOptions_OkAndCancel
+
+	Component.onCompleted: resetValidation()
+
+	property string password
```
blammit, line 23: Put properties and functions above property assignments?

components/listitems/core/ListRadioButtonGroup.qml
```diff
@@ -43,6 +43,8 @@ ListNavigation {
 	//% "Unknown"
 	property string defaultSecondaryText: qsTrId("settings_radio_button_group_unknown")
 
+	property Component footer
```
blammit, line 46: Should this be `optionFooter` to indicate that the footer is shown in the sub-page of options, rather than inside the initial list item?

chriadam: makes sense, will do

components/dialogs/SecurityProfilePasswordDialog.qml
chriadam: thanks, will fix

components/dialogs/SecurityProfilePasswordDialog.qml
chriadam: will do

DanielMcInnes (approved):

Issue #2098, opened 2025-05-05
Move "Demo Mode" back to Settings > General (high_prio)


philipptrenz:
Hi team (@chriadam),

Matthijs asked to move the "Demo mode" currently housed under _Settings_ > _Debug & Develop_ and thereby only accessible to superusers back to a place, where it is more easy to be accessed. 

Let's move it back to _Settings_ > _General_ and place it at the bottom, with a spacer above it (see screenshot, orange area).

As it's a blocker for v3.60, please make it high prio.

![Image](https://github.com/user-attachments/assets/338273c9-571e-4997-96b0-541e0000bc69)

chriadam:
https://github.com/victronenergy/gui-v2/pull/2099

PR #2099, opened 2025-05-06
Move Demo Mode setting back to the General settings page

chriadam:
Contributes to issue #2098

### DIFF
--- a/pages/settings/PageSettingsGeneral.qml
+++ b/pages/settings/PageSettingsGeneral.qml
@@ -166,6 +166,33 @@ Page {
 					uid: Global.venusPlatform.serviceUid + "/ModificationChecks/SystemHooksState"
 				}
 			}
+
+			SettingsListHeader { }
+
+			ListRadioButtonGroup {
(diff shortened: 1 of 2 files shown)


pages/settings/PageSettingsGeneral.qml
```diff
@@ -166,6 +166,39 @@ Page {
 					uid: Global.venusPlatform.serviceUid + "/ModificationChecks/SystemHooksState"
 				}
 			}
+
+			SettingsListHeader { }
+
+			Item {
+				id: spacer
```
blammit, line 173: I don't think this `spacer` is needed -- it looks like Philipp meant that the orange placeholder would be where the "Demo mode" item should go, rather than being the spacer itself.

philipptrenz: Indeed!

chriadam: Ah!  Ok, removed the spacer

blammit (approved):

PR #2102, opened 2025-05-06
Guard ac input references for null

chriadam:


data/AcInputs.qml
```diff
@@ -21,11 +21,13 @@ QtObject {
 			// If that path is not set, then as a backup, try to match /Ac/ActiveIn/ServiceType
 			// against /Ac/In/<0|1>/ServiceType to determine the active input.
 			if (_activeInSource.valid) {
-				return input1.source === _activeInSource.value ? input1
+				return (!input1 && !input2) ? null
+						: input1.source === _activeInSource.value ? input1
```
blammit, line 25: This would return null if `input1.source` matched but `input2` was null. How about using nullish operators instead, like:

```
            if (_activeInSource.valid) {
                return input1?.source === _activeInSource.value ? input1
                        : input2?.source === _activeInSource.value ? input2
                        : null
            } else {
                return input1?.serviceType === _activeInServiceType.value ? input1
                        : input2?.serviceType === _activeInServiceType.value ? input2
                        : null
            }
```

Then `input1?.source` would return undefined if `input1` was null.

chriadam: good point, will do

blammit (approved):

PR #2103, opened 2025-05-06
Only set switchable output dimming value if path is valid

chriadam:
Some switchable outputs are e.g. latchable only, so don't support dimming values.

### DIFF
--- a/data/common/SwitchableOutput.qml
+++ b/data/common/SwitchableOutput.qml
@@ -67,7 +67,9 @@ QtObject {
 	readonly property string customName: _customName.value ?? ""
 
 	function setDimming(value) {
-		_dimming.setValue(value)
+		if (hasDimming) {
+			_dimming.setValue(value)
+		}
 	}
 
 	function setState(value) {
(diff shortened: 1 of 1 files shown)


blammit:
Is this patch for avoiding the warning mentioned in Slack? It looks like we need to avoid the lookup for the uid on the /Dimming path, in the cases where it is not a dimming control. So:

```
diff --git a/data/common/SwitchableOutput.qml b/data/common/SwitchableOutput.qml
index 46633c33..98bdfbee 100644
--- a/data/common/SwitchableOutput.qml
+++ b/data/common/SwitchableOutput.qml
@@ -89,7 +89,7 @@ QtObject {
        }
 
        readonly property VeQuickItem _dimming: VeQuickItem {
-               uid: `${root.uid}/Dimming`
+               uid: root.type === VenusOS.SwitchableOutput_Type_Dimmable ? `${root.uid}/Dimming` : ""
        }
 
        readonly property VeQuickItem _customName: VeQuickItem {
```

chriadam:
Thanks, updated as you suggested

DanielMcInnes (approved):

PR #2111, opened 2025-05-08
Show correct AC input number in when changing current limit

### LINKED ISSUE #2095: ListCurrentLimitButton currentLimitDialogComponent secondaryText is wrong in some cases
From slack: "I have a MultiPlus in my offgrid system (with only AC in 1, and AC in 1 set to Generator) and I noticed that the input current screen incorrectly refers to AC in 2. I have already updated to v3.60~69."

I am not sure what is wrong here, unless the input is 1-based rather than 0-based (our code assumes 0-based input indexes).

The code looks correct to me assuming that the input indexes are zero-based rather than one-based.  i.e. Have secondaryTitle: CommonWords.acInput(root.inputNumber) where the root.inputNumber should be either 0 (for the first generator) or 1 (for the second generator).  The text becomes Ac Input 1  if inputNumber is 0, and becomes AC Input 2 if inputNumber is 1.

blammit:
The digit used to represent an AC input may be 0-based or 1-based depending on the context. For example, the first AC input current limit is configured by com.victronenergy.vebus.<suffix>/Ac/In/1/CurrentLimit while the same input's service name is available as com.victronenergy.system/Ac/In/0/ServiceName.

The current limit dialog showed a 0-based AC input title string instead of the correct 1-based number. Fix this and replace CommonWords.acInput() with acInputFromIndex() and acInputFromNumber() to make it more explicit as to whether a 0-based or 1-based text string should be returned.

Fixes #2095

### DIFF
--- a/components/CommonWords.qml
+++ b/components/CommonWords.qml
@@ -13,6 +13,9 @@ QtObject {
 	//% "AC In"
 	readonly property string ac_in: qsTrId("common_words_ac_in")
 
+	//% "AC Input"
+	readonly property string ac_input: qsTrId("common_words_ac_input")
+
 	//% "Add device"
 	readonly property string add_device: qsTrId("common_words_add_device")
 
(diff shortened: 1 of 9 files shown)


components/CommonWords.qml
```diff
@@ -528,13 +531,12 @@ QtObject {
...
-			   //% "AC Input 2"
-			 : index === 1 ? qsTrId("common_words_ac_input_2")
-			   //% "AC Input"
-			 : qsTrId("common_words_ac_input")
+	function acInputFromIndex(index) {
+		return acInputFromNumber(index + 1)
+	}
+
+	function acInputFromNumber(number) {
+		return `${ac_input} ${number}`
```
chriadam, line 539: I don't think we can do this.  In different languages, it might not appear like this.  It needs to be a fully translated string, I suspect.  Is there any problem with doing something similar to previous e.g. `return number === 1 ? qsTrId("common_words_ac_input_1") : number === 2 ? qsTrId("common_words_ac_input_2") : qsTrId("common_words_ac_input")` ?

blammit: Agreed, updated to pass the number as a translation arg. There are various places in the codebase where strings are concatenated, so those need to be fixed at some point too.

chriadam (approved):

PR #2118, opened 2025-05-08
Move access keyboard shortcuts into user access radio button list

blammit:
These shortcuts should only be enabled in the access radio button list, rather than in the page with the list of security options, to align with the gui-v1 behaviour.

This also prepares for the key navigation feature, by ensuring that the right arrow key can be used for key navigation, without conflicting with the right arrow key shortcut for changing user access.

Part of #1030


chriadam (approved):
Nice!

PR #2123, opened 2025-05-12
Cleanup: use Global.dialogLayer.open() for dialogs

blammit:
Use the convenience function as per
23a6fef93d0a4ccdfd62982ac5788ec7206d1e68.


blammit:
This also fixes key navigation for these dialogs, otherwise the focus disappears for key navigation and cannot be restored.

chriadam (approved):

PR #2126, opened 2025-05-12
Switches: support config to show/hide outputs in switch pane

### LINKED ISSUE #2107: Switching UI: Add configuration option to show/hide outputs in control cards
To allow users to control a switchable output only from NodeRed, we'd like to add the option to hide the controls of an output from the control card. 

A new button "Show controls" should be added here:

Image

Which reflects the state of, and writes to `/SwitchableOutput/x/Settings/ShowUIControl`. Show the corresponding output in the control card (switch pane) when the value is 1, and hide it when the value is 0.

Image

blammit:
Use the output /Settings/ShowUIControl setting to determine whether an output should be shown in the switch pane.

Fixes #2107

### DIFF
--- a/data/Switches.qml
+++ b/data/Switches.qml
@@ -45,6 +45,7 @@ QtObject {
 				onGroupChanged: outputObject.updateGroupModel()
 				onFormattedNameChanged: outputObject.updateSortToken()
 				onTypeChanged: outputObject.updateGroupModel()
+				onShowUIControlChanged: outputObject.updateGroupModel()
 			}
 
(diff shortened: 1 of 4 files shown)


chriadam (approved):
LGTM

ReinvdZee:
@blammit 

I just tested this, it works nicely. I have one request though; can you show the output by default if the path isn't there? So only hide it when the path is there and its 0.

blammit:
@ReinvdZee Thanks, I have updated the PR to do this.

PR #2128, opened 2025-05-12
Hide solar trackers with /Enabled=0

### LINKED ISSUE #2051: RS products: tracker data visibility
The RS and HS products have a setting to enable or disable a PV tracker input. If the tracker is disabled, it should be hidden in the gui.

Related path is `/Pv/n/Enabled`, where n is the tracker index. This also applies to single tracker models.

If `/Pv/n/Enabled` is 0, then the tracker should be hidden and when 1 it should be visible. 
If the item is not present, due to older firmware or unsupported products, then default to visible.

Related Venus issue: https://github.com/victronenergy/venus-private/issues/448

VictronConnect setting:
Image

Test sites:
- https://vrm.victronenergy.com/installation/55013/dashboard
- https://vrm.victronenergy.com/installation/394634/dashboard

blammit:
A solarcharger, multi or inverter service may set /Pv/<x>/Enabled = 0 (where <x> is the solar tracker identifier) to indicate that the solar tracker is disabled. If this is the case, hide the tracker in the UI display.

Ideally QuantityTable would use a filtered model instead of resetting its colours depending on the row count, but that will be fixed in future.

Fixes #2051

### DIFF
--- a/pages/settings/devicelist/rs/PageMultiRs.qml
+++ b/pages/settings/devicelist/rs/PageMultiRs.qml
@@ -247,6 +247,10 @@ Page {
 								return tracker.power
 							}
 						}
+						rowIsVisible: function(row) {
+							const tracker = trackerObjects.objectAt(row)
+							return tracker.enabled
+						}
 
 						Instantiator {
 							id: trackerObjects
--- a/components/QuantityTable.qml
+++ b/components/QuantityTable.qml
@@ -13,6 +13,7 @@ Column {
 	property var units: []
 	property int rowCount
 	property var valueForModelIndex  // function(row,column) -> data value for this row/column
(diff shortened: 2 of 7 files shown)


chriadam (approved):
LGTM

pages/settings/devicelist/rs/PageMultiRs.qml
```diff
@@ -247,6 +247,10 @@ Page {
 								return tracker.power
 							}
 						}
+						rowIsVisible: function(row) {
```
MikeTrahearn-Qinetic, line 250: Is there _any_ way you could avoid function injection into vars? It is really unsafe. Make isRowVisible a boolean and enforce declarative code.

Issue #2133, opened 2025-05-13
Improve "Display DC Loads" text


philipptrenz:
The subtext "Calculated estimate of DC consumption" caused some confusion, because it is actually not correct in case a SmartShunt is used to measure the DC consumption.

Therefore, the subtext "Calculated estimate of DC consumption" should just be removed, so that onle the "Display DC Loads" label remain on this entry.

![Image](https://github.com/user-attachments/assets/a0ad2a88-178f-49e9-a764-83284f172b29)

philipptrenz:
Hi @chriadam, I remember that we also discussed a wording like "Measured or estimated DC loads" during the settings redesign project and finally decided against it, as it might raise questions that are not being answered within the gui. Let's do as suggested above: Removing the subtitle and only keep the primary label.

chriadam:
Merged to main, closing as fixed.

PR #2136, opened 2025-05-15
Remove misleading secondary text related to DC Loads

chriadam:
Contributes to issue #2133


blammit (approved):

pages/settings/PageSettingsSystem.qml
```diff
@@ -107,14 +107,6 @@ Page {
 						//% "Display DC Loads"
 						text: qsTrId("settings_system_has_dc_system")
```
blammit, line 108: Looks like the parent column is no longer needed?

chriadam: Good point, fixed

blammit:
LGTM

PR #2141, opened 2025-05-15
Get service name when battery service is used for dbus-tempsensor-relay.

ReinvdZee:
Needed for https://github.com/victronenergy/venus/issues/1448


ReinvdZee:
@blammit, 

If approved, please merge before the next release.

blammit (approved):

PR #2104, opened 2025-05-06
Add auto dark/light scheme for wasm

mr-manuel:
Adds the possibility to automatically follow the system theme on WASM builds, which make the UX much better.

Needs this meta-victronenergy change first in order to work: https://github.com/victronenergy/meta-victronenergy/commit/b14a7f59a4776e3350e116b2655662f3a28e6e83

To test you need to manually add this path by executing this command until it's integrated into Venus OS:
```bash
dbus -y com.victronenergy.settings /Settings AddSetting Gui RemoteConsoleColorMode 1 i 0 1
```

venus-os-auto-color-mode

Fixes #2115

Issue #1882, opened 2025-01-29
Loading HTML screen not aligned properly


mman:
When the WASM initial loading screen with Victron logo is painted on screen... it's not aligned properly vertically...  and remains near the top of the screen.... where after several seconds (when 33MB gzipped wasm is loaded) it is replaced with centered Victron logo and MQTT loading progress bar... that are however centered nicely... to fill available space...

This can be seen on initial connection to GUI-v2 via browser that has not yet downloaded the WASM, and it is very visible on slower devices like Raymarine Axiom 12 where loading takes several seconds...

nice to have to make it look better...

![Image](https://github.com/user-attachments/assets/63e1ef85-50c7-4c00-b5f3-901f55e8d3fa)

![Image](https://github.com/user-attachments/assets/2efdd6cf-12c7-4425-8c21-162351404759)

mpvader:
discussed today: Manuel is working on something that fixes this. He'll make sure its also tested on an MFD and then close this

mr-manuel:
Will be fixed with https://github.com/victronenergy/gui-v2/pull/2236

![Image](https://github.com/user-attachments/assets/86e5e627-b4af-43b8-a462-8199e5310950)

mman:
That looks much better @mr-manuel, thanks for addressing this!

mr-manuel:
You're welcome

PR #2142, opened 2025-05-15
Add GX mockup for WASM

mr-manuel:
grafik

Can be merged after https://github.com/victronenergy/gui-v2/pull/2104.


philipptrenz:
This PR introduces ...
* the GX screen to be displayed horizontally and vertically centered,
* limited to a pixel size of 1024 x 600 px (resolution of a Ekrano GX and GX Touch 70),
* framed by a GX Touch mockup, if the viewport size is at least 1400 x 900 px, 
* automatically adjusting the background color to the system's light/dark mode setting

It also provides three new query parameters:
* `/gui-v2/?nomock` hides the GX Touch mockup at any resolution
* `/gui-v2/?mock` enforces the GX Touch mockup to be shown at any resolution
* `/gui-v2/?fullscreen` shows the gui-v2 without mockup and maximum size, as it is currently still the case

**CHANGES:**

By default (no query parameters), this does not significantly affect:
* VRM, because it does not use the `index.html` served by the GX
* Viewports with width <= 1024 px and height <= 600 px (typical for phones)
  * Only difference: UI will now also be centered vertically, if ratio is smaller than 1.7

By default (no query parameters), this does only slightly affect:
* Viewports with width <= 1400 px and height <= 900 px (typical for tablets)
  * Difference: UI is limited to 1024 x 600 px, so that there is a certain margin around the display
  * This also includes the Remote GX Display Android App for tablets, if the viewport is greater than 1024 x 600 px
  * But as already discussed with @jepefe, a newer release of the app will include the `?fullscreen` query parameter to enforce the fullscreen view, as it is to date


Attached are screnshots from an iPhone 16 (landscape and portrait) as well as from an iPad Air 11" (landscape) using the Safari browser.

IMG_3729
IMG_3728
IMG_3CAC0774359A-1

mr-manuel:
Test #1882 after merging

blammit (approved):
Looks great. I've rebased this against the latest changes from https://github.com/victronenergy/gui-v2/pull/2104.

PR #2145, opened 2025-05-16
Make StatusBar Notification Icon Clickable

### LINKED ISSUE #2072: Notification icon optimizations
1. Make the notifications icon in the top bar clickable. On click you get redirected to the notifications page.

    Image

2. ~The notification icon disappears as soon as the breadcrumb displays. Display the notifications icon at the top right, left of the display sleep button (if the button is displayed).~ Currently not needed

    Image
    Image

MikeTrahearn-Qinetic:
This PR to cover point 1) in #2072 which requests the Status Bar Notification Icon to be clickable and take you to the NotificationPage. Note, the use of Global.notificationLayer.popAndGoToNotifications() function was avoided as this is too heavy for this use case and causes the UI to refresh the NavBar in an undesirable manner. Instead, we just request to change the page, which invokes the correct swipe animation and keeps the NavBar intact.

Since the NotificationIcon is invisible when there are no notifications, the child TapHandler only works when there are any notifications.

Full Fixes #2072 completely as point 2 (noted by @mr-manuel) has been removed.

### DIFF
--- a/components/StatusBar.qml
+++ b/components/StatusBar.qml
@@ -193,7 +193,7 @@ FocusScope {
 			}
 		}
 
-		KeyNavigation.right: alarmButton
+		KeyNavigation.right: notificationButton
 
 		Connections {
 			target: root.pageStack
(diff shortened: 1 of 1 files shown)


mr-manuel:
I removed point 2 for now.

chriadam (approved):
LGTM, tested fine on desktop and wasm.  I am not in the office today, so didn't test on Cerbo.

components/StatusBar.qml
```diff
@@ -267,6 +267,10 @@ FocusScope {
 					   Global.notifications?.statusBarNotificationIconPriority === VenusOS.Notification_Info ? Theme.color_ok : notificationIcon.color
 			source: Global.notifications?.statusBarNotificationIconPriority === VenusOS.Notification_Info ?
 						"qrc:/images/icon_info_32.svg" : "qrc:/images/icon_warning_32.svg"
+
+			TapHandler {
+				onTapped: pageManager.navBar.setCurrentPage("NotificationsPage.qml")
```
blammit, line 272: Can you make this work with key navigation as well? (In that sense it might be easier to turn it into a `Button`, to get the space-to-click and navigation highlight for free.)

blammit: Now that key nav has been disabled in `main`, this would be fine to merge as-is. @chriadam any thoughts?

MikeTrahearn-Qinetic: @blammit if key navigation were added, it would simply be "disabled" because of #2161 until such time as it would be enabled again (or caught as part of further key navigation operational rework). If you think it would be wise to continue with the addition of "dormant" key navigation then I would value your assistance to understand the mechanism to adorn this button upgrade with this behaviour.

MikeTrahearn-Qinetic:
I pushed an update to bring the Notification Icon up to the same level of functionality of the other buttons:
- added support for KeyNavigation (tested by temporarily reverting Commit b4df1664 to re-enable it)
- fixed the layout location
- includes upgrade to StatusBarButton
- always shown like the clock, regardless if the opacity of the StatusBar
- clicking now shows the standard PressArea effect like other status bar buttons

Here is a preview of the button with the KeyNavigation also enabled:

NotificationButton-Dark
NotificationButton-Light

blammit (commented):
Works nicely with the key navigation.

components/StatusBar.qml
```diff
@@ -254,20 +254,29 @@ FocusScope {
...
 		}
+		// The notificationButton should always be shown, even when the StatusBar is not
+		opacity: 1
+		visible: !breadcrumbs.visible && (Global.notifications?.statusBarNotificationIconVisible ?? false)
+
+		color: Global.notifications?.statusBarNotificationIconPriority === VenusOS.Notification_Alarm
+			   ? Theme.color_critical
+			   : Global.notifications?.statusBarNotificationIconPriority === VenusOS.Notification_Warning
+				 ? Theme.color_warning :
+				   Global.notifications?.statusBarNotificationIconPriority === VenusOS.Notification_Info ? Theme.color_ok : notificationButton.color
```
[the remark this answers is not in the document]
MikeTrahearn-Qinetic: Such a binding is legal - no warning would be shown. I can however change it to "transparent" since I believe in this case the icon wouldn't be displayed. This would at least remove the "apparent" possibility for a value lock, even though it never would actually occur.

components/StatusBar.qml
```diff
@@ -254,20 +254,29 @@ FocusScope {
...
-					   Global.notifications?.statusBarNotificationIconPriority === VenusOS.Notification_Info ? Theme.color_ok : notificationIcon.color
-			source: Global.notifications?.statusBarNotificationIconPriority === VenusOS.Notification_Info ?
-						"qrc:/images/icon_info_32.svg" : "qrc:/images/icon_warning_32.svg"
+		anchors {
+			left: connectivityRow.right
+			leftMargin: Theme.geometry_statusBar_rightSideRow_horizontalMargin
+			verticalCenter: parent.verticalCenter
 		}
+		// The notificationButton should always be shown, even when the StatusBar is not
+		opacity: 1
```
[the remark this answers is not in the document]
blammit: If the status bar is not shown, then the button will not be shown either. The button can only be seen when the status bar has an opacity > 0.

PR #2150, opened 2025-05-20
Switch sync improvements

blammit:


blammit:
This should fix the sync issues described in the task, with the approach we discussed in the meeting:

- If user has changed the value of a button or slider, then prevent further button/slider changes until the backend is in sync
- If user has not changed the value, and the change came from elsewhere (i.e. from input by another user on VRM/device), then update the button/slider value.

I have also tested the scenario we discussed where multiple users are setting conflicting changes (across VRM and cerbo instances) at the same time.

chriadam (approved):
LGTM

ReinvdZee:
I tested this extensively with 4 GUI instances and it works flawless.
Really nice work!

ReinvdZee (approved):

blammit:
Great to hear! Thanks for verifying.

Issue #2139, opened 2025-05-15
Digital input menu fixes


ReinvdZee:
Currently, the digital inputs menu (Settings->Integrations->Digital I/O) does not work with the GX IO Extender. This is because the digital IO page expects only integer indexed digital inputs, but the inputs of the IO extender are prepended with the serial number so we can distinguish them:

This is the `com.victronenergy.digitalinputs` service:

![Image](https://github.com/user-attachments/assets/7790bb35-9a12-4ffa-9d7e-c174fe8065bd)

The top four are the internal inputs, and the latter four are inputs of a GX IO extender. 
Instead of displaying "Digital Input <index>" in the menu, the `/Label` path should be used to display the input. 
Note that dbus-digital-inputs currently only fills the label with a number, this will also be fixed to display what's in the screenshot above.

And there's another issue:
The menu compiles the list of digital inputs based on the inputs found in the settings service. For only the internal inputs, this is fine since the inputs are always there. But with the IO extender, inputs could come and go, but their settings paths remain, so the GUI needs to look at `com.victronenergy.digitalinputs/Devices/..` instead to compile the list. It is also fine to write the input type to `com.victronenergy.digitalinputs/Devices/x/Type` instead of writing to the settings service.

ReinvdZee:
I updated the labels of the built-in digital inputs and of the GX IO extender in the backend:

![Image](https://github.com/user-attachments/assets/65bef472-2691-47f4-bfa9-59a702a4deba)

ReinvdZee:
With the latest fixes in the GUI and in dbus-digital-inputs, it looks like:

![Image](https://github.com/user-attachments/assets/886cf05d-7d9b-4671-af28-574ca12a3c9a)

PR #2154, opened 2025-05-20
Digital inputs fixes

blammit:


blammit:
Note that the digital input settings will now just show digits for the inputs from the GIO Extender:

image

But this is okay, it will be fixed once Rein's changes to the labels are released, as per his comments in https://github.com/victronenergy/gui-v2/issues/2139#issuecomment-2890432261.

chriadam (approved):
LGTM, would be good if Rein could test also

blammit:
@ReinvdZee could you test this please?

ReinvdZee:
I tested this with the label changes on the digital inputs service, looks good:

image

ReinvdZee:
Unfortunately, I forgot to test with adding and removing IO extenders.

This does not work properly. When the IO extender is disconnected, the list items are cleared but do not disappear:
image

Even when backstepping and reentering the menu, the inputs are still listed as unknown.

blammit:
> This does not work properly. When the IO extender is disconnected, the list items are cleared but do not disappear:

Thanks, updated to fix this, and verified on cerbo/wasm.

chriadam:
LGTM

Issue #2089, opened 2025-05-01
Enable heatpump role for energy meters


martinbosma:
Finishing up the new roles as part of https://github.com/victronenergy/venus-private/issues/380.

The role is already supported, but is read-only, which can now be r/w.

https://github.com/victronenergy/gui-v2/blob/defeebef1e3202c12803fd2d3b3a1fc030d2a19e/pages/settings/devicelist/ac-in/PageAcInSetup.qml#L59-L71

We also need a drill down to the heatpump device when the user clicks the load box.

blammit:
> We also need a drill down to the heatpump device when the user clicks the load box.

@martinbosma Is that the "AC Loads" or "Essential Loads" box? If it can be either, how do you determine which one?

martinbosma:
I'd say a heatpump is an AC load. But perhaps @dirkjanfaber or @izak can comment on that, since they are more involved in the system context.

mpvader:
> @martinbosma Is that the "AC Loads" or "Essential Loads" box? If it can be either, how do you determine which one?

@blammit & @martinbosma: that depends on the position. Which is a setting of the meter, exposed in VictronConnect.

blammit:
Okay, sounds like this is the same as EVCS, where `/Position=1` means AC-in and "AC loads", and `/Position=0` means AC-out and "Essential Loads".

izak:
@martinbosma , remember I asked you about this? And we decided that unless it is asked for we won't change it now?

I added /Position to the cgwacs driver, somewhat by accident, but I don't think it was added to the Victron meter, or the others. So at the moment heat pumps don't have a position.

izak:
@blammit, further to my previous comment, heatpump (and acload) services don't have /Position right now, but we're going to add it.

martinbosma:
I've added this to vecan-dbus v3.53, which is on the todo but not in a release yet

blammit:
I see vecan-dbus v3.53 has now been released as part of Venus v3.60~75, thanks!

blammit:
@martinbosma should the "Position" setting be shown (and also editable) for AC inputs with the `heatpump` roles? Currently the setup page only does this for inputs with an `evcharger` role.

izak:
@blammit the same logic applies as for the EV charger. If PositionIsAdjustable (or does not exist), then the position can be edited, otherwise just show it.

Showing the total power for each is useful.

[...]

PR #2157, opened 2025-05-21
Heat pump features

blammit:


components/widgets/AcLoadsWidget.qml
```diff
@@ -19,7 +19,7 @@ AcWidget {
 	type: VenusOS.OverviewWidget_Type_AcLoads
 	quantityLabel.dataObject: root.measurements
 	phaseCount: root.measurements.phases.count
-	enabled: false
+	enabled: Global.allDevicesModel.heatPumpInputDevices.count > 0
```
chriadam, line 22: if this is meant to be an AcLoad shouldn't it look at heatPumpOutputDevices?  (i.e. heat pumps connected to the output side of the inverter, therefore being a load on the system?)  or am I misunderstanding?

chriadam: Ah, I see - there is a comment in #2089 saying that it should be an AcLoad if its position is input, and EssentialLoad if its position is output (i.e. will be kept powered even if the grid goes down, by draining battery).

chriadam: Maybe add a comment explaining this above this line?

chriadam:
I tried to find an installation to test this with, but without luck.  There are ones like e.g. https://vrm.victronenergy.com/installation/491097/dashboard which have AC loads and critical loads, but the critical load is an EV Charger, and no energy meter designated as a heat pump exists on that installation.

components/widgets/AcLoadsWidget.qml
blammit: Comment added, do you think it needs more detail? It's just about the position value at the moment.

chriadam: No more detail required, LGTM - thanks!

chriadam (approved):
Code LGTM but I couldn't figure out which installation I could use to test this with.

Issue #2159, opened 2025-05-22
Boat page - when there is only a single load (AC or DC), only display a single, full-height gauge on the RHS


DanielMcInnes:
@philipptrenz FYI

chriadam:
Necessary for release, since most electric boats will be simple and not have any AC loads

philipptrenz:
Changed to medium prio, does not block the release of v3.60.

DanielMcInnes:
Hi @philipptrenz , could you please link to a site where this is happening? I saw this happening on your screen during yesterday's meeting, but I can't find a site that reproduces the issue.

chriadam:
We double checked that it works correctly for the case where there are no dc loads (i.e. no "dcload", "dcsystem", or "dcdc" devices) -> in that case, only the ac load gauge is shown, and it is shown full-height.

However, the other case (where there are no ac loads) could be an issue.  Is there any installation we can test with which doesn't have any ac loads?  (I mean: not just that the consumption on a particular phase is zero, but that there is no inverter or other ac system in the installation.)  And, does systemcalc publish some value to let us know that there is no AC component to the system?  (We would need this to hide not only the ac load gauge on the boat page, but also the ac load widget on the overview page.)

@ReinvdZee perhaps you could provide some information here ^

Anyway, definitely not for 3.60.

philipptrenz:
Hi @chriadam & @DanielMcInnes, it took me a while to setup a system, sorry for the dalay. Systems with only AC loads are not really relevant for the Boat UI, DC-only and mixed ones are.

Here's a test site, it only has a battery and a GPS (latter with no connection currently): https://vrm.victronenergy.com/installation/366384/

philipptrenz:
> And, does systemcalc publish some value to let us know that there is no AC component to the system? (We would need this to hide not only the ac load gauge on the boat page, but also the ac load widget on the overview page.)

I think that's something @ReinvdZee can answer best, I'm not sure. If not, I'd propose to add a `Settings/SystemSetup/HasAcSystem` path, which indicates if there is any _vebus_ or _acsystem_ service available.

chriadam:
Created a prototype at https://github.com/victronenergy/gui-v2/pull/2165 which just checks if the System.load.ac.power (or System.load.acIn.power, if showInputLoads is set) value is NaN, to determine the "has ac loads" condition.

Obviously, shouldn't be merged until we have fully discussed the approach etc.  Not for 3.60.

[...]

PR #2165, opened 2025-05-23
Don't display AC Loads gauge or widget if no AC loads exist

chriadam:
Affects both the Boat Page and the Overview Page.

Contributes to issue #2159

### DIFF
--- a/data/System.qml
+++ b/data/System.qml
@@ -13,7 +13,10 @@ QtObject {
 	readonly property int state: _systemState.valid ? _systemState.value : VenusOS.System_State_Off
 
 	readonly property bool hasGridMeter: _gridDeviceType.valid
-	readonly property bool hasAcOutSystem: _hasAcOutSystem.valid && _hasAcOutSystem.value === 1
+	readonly property bool hasAcOutSystem: _hasAcOutSystem.valid && _hasAcOutSystem.value === 1 // i.e. Essential loads
+	readonly property bool hasAcLoads: showInputLoads
+				? (load.acIn !== null && load.acIn !== undefined && !isNaN(load.acIn.power))
+				: (load.ac !== null && load.ac !== undefined && !isNaN(load.ac.power))
 	readonly property bool hasVebusEss: _systemType.value === "ESS" || _systemType.value === "Hub-4"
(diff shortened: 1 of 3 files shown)


chriadam:
Not for 3.60.

This one needs discussion and very thorough testing, before we think about merging it.

chriadam:
Can test with https://vrm.victronenergy.com/installation/366384/ although obviously that's not going to be exhaustive (we also need to test all of the other possibilities e.g. EVCharger but no input loads, Essential Loads but no EVCharger, etc etc) to make sure we don't hide the AcLoads widget in some case where we need to show it, etc.

philipptrenz:
Nice, thanks Chris! I did some quick tests and there seem to be indeed a few issues with the overview page Let's look into this more closely as soon as v3.60 is out.

data/System.qml
```diff
@@ -13,7 +13,10 @@ QtObject {
 	readonly property int state: _systemState.valid ? _systemState.value : VenusOS.System_State_Off
 
 	readonly property bool hasGridMeter: _gridDeviceType.valid
-	readonly property bool hasAcOutSystem: _hasAcOutSystem.valid && _hasAcOutSystem.value === 1
+	readonly property bool hasAcOutSystem: _hasAcOutSystem.valid && _hasAcOutSystem.value === 1 // i.e. Essential loads
+	readonly property bool hasAcLoads: showInputLoads
+				? (load.acIn !== null && load.acIn !== undefined && !isNaN(load.acIn.power))
+				: (load.ac !== null && load.ac !== undefined && !isNaN(load.ac.power))
```
blammit, line 19: The `HasAcSystem` setting proposed in the task would definitely make things easier - I don't know if an invalid `load.acIn.power` guaranteeds the AC Loads is not available, and also it's not so nice for the UI to have to take dependencies into account (e.g. whether EVCS is connected to AC loads).

philipptrenz:
Hey @chriadam & @blammit, I discussed with Izak and having systemcalc to decide if there is a AC system is the best way to move forward. I created an issue for Izak about that: https://github.com/victronenergy/venus/issues/1547

Issue #2166, opened 2025-05-23
Writing state values to ES SmartSwitch is broken in current version


chriadam:
The code which added synchronisation improvements causes the backend to reject state value writes.

chriadam:
Merged to main, released in v1.1.17.  Closing as fixed.

PR #2167, opened 2025-05-23
Switches: ensure state value writes are integer values

chriadam:
Contributes to issue #2166


chriadam:
Self-approving and merging.

PR #2175, opened 2025-05-26
Digital inputs: fix list spacing and sorting 

blammit:


pages/settings/PageSettingsIntegrations.qml
```diff
@@ -155,22 +173,10 @@ Page {
 
 						GradientListView {
 							model: digitalModel
-
 							delegate: ListRadioButtonGroup {
-								text: inputLabel.value || ""
-								dataItem.uid: model.uid + "/Type"
+								text: model.item.value || ""
```
MikeTrahearn-Qinetic, line 177: could you use required properties here rather than the "model" context injected property?

blammit: Yep, done

data/mock/MockDataManager.qml
```diff
@@ -27,6 +27,7 @@ QtObject {
 			property var batteries: BatteriesImpl {}
 			property var dcInputs: DcInputsImpl {}
 			property var dcLoads: DcLoadsImpl { }
+			property var digitalInput: DigitalInputImpl {}
```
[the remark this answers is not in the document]
blammit: Hopefully, yes. We should change of all of these as a general clean-up, though last time I looked at this, there was an issue with making circular dependencies due to the Global.qml singleton being part of the module and being imported from elsewhere.

data/mock/DigitalInputImpl.qml
```diff
@@ -0,0 +1,39 @@
...
+				serviceUid: "mock/com.victronenergy.digitalinput.ttyUSB" + deviceInstanceNum,
+				deviceInstance: deviceInstanceNum,
+			})
+		}
+	}
+
+	property Component inputComponent: Component {
+		Device {
+			Component.onCompleted: {
+				_deviceInstance.setValue(deviceInstance)
```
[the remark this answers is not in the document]
blammit: This is a hacky workaround that needs to be cleaned up across the mock implementations. Basically, line 20 is to set a hardcoded number (to ensure there is a valid instance when the object is created) and then this line here is to make sure _deviceInstance is in sync with deviceInstance.

pages/settings/PageSettingsIntegrations.qml
```diff
@@ -155,22 +173,12 @@ Page {
 
 						GradientListView {
 							model: digitalModel
-
 							delegate: ListRadioButtonGroup {
-								text: inputLabel.value || ""
-								dataItem.uid: model.uid + "/Type"
-								optionModel: delegateOptionModel
+								required property var item
```
MikeTrahearn-Qinetic, line 177: I'm assuming you don't know what the type is or it is a JS object.

blammit: Yes, I think QML_UNCREATABLE would make sense for it. I'm not sure how much of that API is formalized, so it'd be good to clarify that in more detail at some point.

blammit: Updated to add QML_UNCREATABLE for VeQItem, and refer to the VeQItem type here.

pages/settings/PageSettingsIntegrations.qml
```diff
@@ -155,22 +173,12 @@ Page {
 
 						GradientListView {
 							model: digitalModel
-
 							delegate: ListRadioButtonGroup {
-								text: inputLabel.value || ""
-								dataItem.uid: model.uid + "/Type"
-								optionModel: delegateOptionModel
+								required property VeQItem item
```
MikeTrahearn-Qinetic, line 177: This is obviously better,  but I wonder for what benefit? To get the item value and the itemParent() - is it even possible to do something through the VeQuickItem by extending it somehow? As the comment says "Not creatable, use VeQuickItem instead". Could be opening an unwise can of worms by exposing what ought to be hidden by the layer explicitly designed for the purpose.

blammit: Do you mean, it's better to not expose VeQItem as an uncreatable type after all? VeQuickItem is not suitable here.

MikeTrahearn-Qinetic: It depends - if the requirements are that it shouldn't be, then no. If the requirements dictate that more work needs to be done on the C++ > QML interfaces, then obviously that's for another task. Just flagging this one as a "sore-thumb" example of "try not to do this again if possible".

chriadam (approved):
LGTM

PR #2176, opened 2025-05-26
Use settings proxy path for digital input settings

ReinvdZee:
Instead of looking at the settings service directly. The original way of compiling the settings paths
does not work for the IO extender digital inputs.


blammit (approved):

pages/settings/devicelist/PageDigitalInput.qml
```diff
@@ -47,27 +47,25 @@ Page {
...
-						readonly property string settingsBindPrefix: Global.systemSettings.serviceUid + "/Settings/DigitalInput/" + (deviceInstance.value || 0)
-
 						GradientListView {
 							model: VisibleItemModel {
 								ListSwitch {
 									id: alarmSwitch
 									//% "Enable alarm"
 									text: qsTrId("digitalinput_enable_alarm")
-									dataItem.uid: settingsBindPrefix + "/AlarmSetting"
+									dataItem.uid: bindPrefix + "/Settings/AlarmSetting"
```
blammit, line 56: Minor thing, would be more precise to use `root.bindPrefix` instead of `bindPrefix`.

PR #2191, opened 2025-05-29
Add BMS control reset button to OrionXS settings

### LINKED ISSUE #2190: Add BMS control Reset button to OrionXS settings pages
When an OrionXS charger is switched from charging the main battery to charging an aux battery, there is a chance that it will give an error "#67 BMS connection lost" after a while. This is caused by the fact that when the GX device is sending charge voltage setpoints or charge current limits, this is seen as a BMS. This is however only sent when the OrionXS is configured as charging the main battery. So when it is configured as charging an aux battery, the GX will no longer send these setpoints and the OrionXS after a while thinks the BMS is gone.

Currently, the only way to reset this is by resetting the settings of the OrionXS to default using VictronConnect (and then configuring it again), or by switching it to PSU mode briefly and then setting it to charger again. By adding a "BMS control" reset button, the user can clear the error without changing settings or needing VictronConnect.

This functionality is already available in the SolarCharger "Networked opeartion" menu (PageSolarChargerParallelOperation.qml) and it should behave the same for the OrionXS. As the OrionXS setting menus (PageAlternatorModel.qml and PageDcDcConverter.qml) do not have a networking sub menu and it does not really make sense to add one to PageDcDcConverter.qml (as the BMS reset will then be the only item), the reset button can be placed under the settings menu.

blammit:
Fixes #2190


pages/settings/devicelist/dc-in/DcBmsSettingsPage.qml
```diff
@@ -0,0 +1,70 @@
...
+	GradientListView {
+		model: VisibleItemModel {
+			ListOutputBatteryRadioButtonGroup {
+				bindPrefix: root.bindPrefix
+				settingsPage: root.settingsPage
+			}
+			ListText {
+				id: bmsControlled
+				//% "BMS controlled"
+				text: qsTrId("charger_network_bms_controlled")
```
chriadam, line 27: this id already exists in PageSolarParallelOperation.qml - could move to common words (or remove the eng eng text here, but that is a bit suboptimal)

Same for a couple of other charger_network_ ids

blammit: Fixed to use CommonWords instead.

chriadam (approved):

Issue #2189, opened 2025-05-28
Pulsemeter: Reset counter does not work


ReinvdZee:
Resetting the pulse counter does not work. In GUI-v1, this works properly.

![Image](https://github.com/user-attachments/assets/22e65cf4-284b-491c-acf3-58e8e4c3bd14)

This issue is also present in v3.55 so it's not a regression.
Can be reproduced with internal digital input as well as GX IO extender digital input.

chriadam:
gui-v1 writes a value of zero to the path: com.victronenergy.settings/Settings/DigitalInput/N/Count

whereas

gui-v2 attempts to write a value of zero to the path: com.victronenergy.pulsemeter.input_N/Count

It seems that the latter path is read-only, and writes to that path are ignored by the backend.

I have created PR: https://github.com/victronenergy/gui-v2/pull/2192
which resolves this by ensuring we attempt to write to the system settings prefixed digital input's /Count path, if that path exists.

Note that as per #2139 this will likely only resolve the issue for internal inputs configured as pulse meters, as the GX I/O extender settings path includes also a serial number (not just input number) and cannot be easily programmatically built (the serial number isn't exposed in Device.qml, for example).  In the future, the /Count path should also be added to the proxied settings paths, so we can write to that one directly.

chriadam:
I have also created this PR: https://github.com/victronenergy/gui-v2/pull/2193/files which implements it also for GX I/O Extender inputs.  Perhaps this is preferred?

chriadam:
Created https://github.com/victronenergy/gui-v2/issues/2194 to track adding (in future) the proxy settings count path.

chriadam:
PR #2193 merged to main, closing as fixed.

PR #2192, opened 2025-05-29
Fix resetCounter for PulseMeter inputs (excluding GX I/O Ext)

chriadam:
Write to the system settings prefixed input specific /Count path, instead of the pulse meter path.

Note that this does not work for inputs from the GX I/O extender, as the path includes a serial number and not just an input number (see #2139 for more details).  In future, the /Count path should be added to the proxied settings paths, to resolve this.

Contributes to issue #2189


chriadam:
Possibly obsoleted by https://github.com/victronenergy/gui-v2/pull/2193

PR #2207, opened 2025-06-05
Add read/write option for modbus-TCP

izak:
Related to https://github.com/victronenergy/venus/issues/1451

### DIFF
--- a/pages/settings/PageSettingsModbusTcp.qml
+++ b/pages/settings/PageSettingsModbusTcp.qml
@@ -31,6 +31,19 @@ Page {
 				dataItem.uid: Global.systemSettings.serviceUid + "/Settings/Services/Modbus"
 			}
 
+			ListRadioButtonGroup {
+				//% "Access permissions"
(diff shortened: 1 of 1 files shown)


pages/settings/PageSettingsModbusTcp.qml
```diff
@@ -31,6 +31,19 @@ Page {
 				dataItem.uid: Global.systemSettings.serviceUid + "/Settings/Services/Modbus"
 			}
 
+			ListRadioButtonGroup {
+				//% "Access permissions"
+				text: qsTrId("settings_modbus_access_rights")
+				preferredVisible: enableModbusTcp.checked
+				dataItem.uid: Global.systemSettings.serviceUid + "/Settings/ModbusServer/ReadWrite"
```
blammit, line 38: If the path does not exist, this option will still be shown, but with "Unknown" on the right-hand side - is that okay?

blammit (approved):

PR #2215, opened 2025-06-09
Rework QuantityTable and QuantityTableSummary to use QuantityRow

### LINKED ISSUE #2179: Rework QuantityTable and QuantityTableSummary to use QuantityRow
The implementation of QuantityTable and QuantityTableSummary duplicates various features of QuantityRow (in terms of showing a series of quantity labels), so they should juse use QuantityRow internally.

QuantityTable can be a ListView with a header, model and delegate, to avoid the need for the custom `valueForModelIndex` and `rowIsVisible` functions.

QuantityTableSummary is just a one-line table, so it could use QuantityTable internally.

blammit:
Remove the duplicated code between QuantityTable, QuantityTableSummary and QuantityRow. In QuantityTable and QuantityTableSummary, instead of providing valueForModelIndex() and rowIsVisible() functions that can be customized to provide the column data for each row, provide TableHeader and TableBody types that can be customized together with QuantityObjectModel to provide the column data.

Also simplify QuantityTableMetrics to remove most of its property configurations, which can be implemented in QuantityTable, QuantityTableSummary and QuantityRow instead.

Fixes #2179

### DIFF
--- a/components/QuantityGroupListHeader.qml
+++ b/components/QuantityGroupListHeader.qml
@@ -6,53 +6,56 @@
 import QtQuick
 import Victron.VenusOS
 
-Item {
-	property alias firstColumnText: firstTitleLabel.text
-	property alias quantityTitleModel: titleRepeater.model
+Row {
+	id: root
 
-	width: parent.width
-	height: Theme.geometry_listItem_height
+	property alias headerText: firstTitleLabel.text
+	property alias model: titleRepeater.model
+	property int fontSize: Theme.font_size_caption
+	readonly property int columnCount: titleRepeater.count + 1 // +1 for header column
+
+	// Column sizing parameters
+	property int metricsFontSize: fontSize // the QuantityTableMetrics font size, for calculating column size
+	property real fixedColumnWidth: NaN // if set, use this for column widths, instead of QuantityTableMetrics
+
+	height: Theme.geometry_quantityTable_row_height
+	leftPadding: Theme.geometry_listItem_content_horizontalMargin
+	rightPadding: Theme.geometry_listItem_content_horizontalMargin
+	spacing: Theme.geometry_quantityGroupRow_spacing
 
 	Label {
 		id: firstTitleLabel
-		anchors {
-			bottom: parent.bottom
-			bottomMargin: Theme.geometry_quantityTableSummary_verticalMargin
-		}
-		leftPadding: Theme.geometry_listItem_content_horizontalMargin
-		font.pixelSize: Theme.font_size_caption
+		anchors.verticalCenter: parent.verticalCenter
+		font.pixelSize: root.fontSize
 		color: Theme.color_solarListPage_header_text
 		elide: Text.ElideRight
-		width: parent.width - quantityRow.width - Theme.geometry_quantityGroupRow_spacing
+		width: isNaN(root.fixedColumnWidth)
+			   ? parent.width - parent.leftPadding - parent.rightPadding - quantityRow.width - root.spacing
+			   : root.fixedColumnWidth
 	}
 
 	Row {
 		id: quantityRow
 
-		anchors {
-			bottom: parent.bottom
-			bottomMargin: Theme.geometry_quantityTableSummary_verticalMargin
-			right: parent.right
-			rightMargin: Theme.geometry_listItem_content_horizontalMargin + Theme.geometry_icon_size_medium
-		}
-		width: Theme.geometry_solarListPage_quantityRow_width
+		anchors.verticalCenter: parent.verticalCenter
+		spacing: root.spacing
 
 		Repeater {
 			id: titleRepeater
 
 			delegate: Label {
-				width: quantityMetrics.columnWidth(modelData.unit)
+				width: isNaN(root.fixedColumnWidth)
+					   ? quantityMetrics.columnWidth(modelData.unit, implicitWidth)
+					   : root.fixedColumnWidth
 				text: modelData.text
-				font.pixelSize: Theme.font_size_caption
+				font.pixelSize: root.fontSize
 				color: Theme.color_solarListPage_header_text
 			}
 		}
 
 		QuantityTableMetrics {
 			id: quantityMetrics
-			count: titleRepeater.count
-			availableWidth: quantityRow.width
-			spacing: Theme.geometry_quantityGroupRow_spacing
+			font.pixelSize: root.metricsFontSize
 		}
 	}
 }
--- a/components/QuantityTable.qml
+++ b/components/QuantityTable.qml
@@ -6,153 +6,128 @@
 import QtQuick
 import Victron.VenusOS
 
-// TODO change this to use QuantityObjectModel for the model
-Column {
+/*
+	A table of quantity values.
+
+	The TableHeader and TableRow types can be used to implement header and delegate components that
+	automatically integrate with the look and feel of the table.
+*/
+ListView {
 	id: root
 
-	property var units: []
-	property int rowCount
-	property var valueForModelIndex  // function(row,column) -> data value for this row/column
-	property var rowIsVisible   // TODO remove when QuantityObjectModel is used as model instead.
-	property bool headerVisible: true
-	property Component headerComponent: defaultHeaderComponent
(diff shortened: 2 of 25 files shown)


blammit:
Updated with a minor improvement to auto-set the `table` property value for `TableRow` and `TableHeader`.

components/QuantityGroupListHeader.qml
```diff
@@ -6,53 +6,56 @@
...
-	property alias firstColumnText: firstTitleLabel.text
-	property alias quantityTitleModel: titleRepeater.model
+Row {
+	id: root
 
-	width: parent.width
-	height: Theme.geometry_listItem_height
+	property alias headerText: firstTitleLabel.text
+	property alias model: titleRepeater.model
+	property int fontSize: Theme.font_size_caption
```
MikeTrahearn-Qinetic, line 14: Is this property used to change the font size of multiple items in this scope? If not, consider a property alias.

MikeTrahearn-Qinetic: Consider renaming usages of "fontSize" to fontPixelSize" to match the intended font property.

components/QuantityGroupListHeader.qml
```diff
@@ -6,53 +6,56 @@
...
+	readonly property int columnCount: titleRepeater.count + 1 // +1 for header column
+
+	// Column sizing parameters
+	property int metricsFontSize: fontSize // the QuantityTableMetrics font size, for calculating column size
```
MikeTrahearn-Qinetic, line 18: could not the quantityMetrics.font.pixelSize bind directly to root.fontSize and remove this property?

components/QuantityTable.qml
```diff
@@ -6,153 +6,128 @@
...
+				visibleRowCount++
 			}
 		}
 	}
 
-	Loader {
-		id: header
-		sourceComponent: headerVisible ? headerComponent : null
+	function _fixedColumnWidth(columnCount) {
+		return equalWidthColumns
```
MikeTrahearn-Qinetic, line 44: Since this function is going to re-return whenever any of its other dependent properties) then it would might be more efficient to use a more declarative approach here else the function is going to get called multiple times. A binding update is going to be potentially faster, especially if the parts of the calculation that don't change too much are done ahead of time. At least annotate the function with _fixedColumnWidth(columnCount: int) : real/int {...

MikeTrahearn-Qinetic:
I'll need to re-profile the cerbo after this goes in as the performance of the Solar Yield pages will be affected by the changes here.

pages/evcs/EvChargerPage.qml
```diff
@@ -24,7 +24,7 @@ Page {
 				QuantityTableSummary {
 					id: chargerSummary
 
-					function _currentSummaryText() {
+					readonly property string currentSummaryText: {
```
MikeTrahearn-Qinetic, line 27: suggest: ```readonly property string currentSummaryText: root.energyMeterMode ? "---" : ""``` rather than a function

components/dialogs/SolarDailyHistoryDialog.qml
```diff
@@ -193,7 +191,8 @@ T.Dialog {
 				const history = root.solarHistory.dailyHistory(root.day)
 				return history ? history.errorModel : null
 			}
-			visible: model && model.count > 0
+			enabled: model && model.count > 0
```
MikeTrahearn-Qinetic, line 194: suggest ```model?.count > 0 ?? false```, might not even need the ?? on the end.

components/QuantityTable.qml
```diff
@@ -6,153 +6,128 @@
...
+
+	// Parameters for sizing the columns in the table.
+	property int metricsFontSize: bodyFontSize
+	property int columnSpacing: Theme.geometry_quantityTable_horizontalSpacing_large
+	property bool equalWidthColumns
+
+	property real leftPadding: Theme.geometry_listItem_content_horizontalMargin
+	property real rightPadding: Theme.geometry_listItem_content_horizontalMargin
+
+	function resetRowColors() {
```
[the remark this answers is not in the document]
blammit: Yeah, this is not ideal, but unfortunately it doesn't suit VisibleItemModel as the model is usually a QAbstractItemModel, rather than a visual model. I'm not sure of a better way to do it here.

components/QuantityTableSummary.qml
```diff
@@ -6,136 +6,34 @@
...
-							unit: root.model[model.index + 1].unit
-							precision: root.model[model.index + 1].precision || VenusOS.Units_Precision_Default
-							visible: unit !== VenusOS.Units_None
-						}
-					}
-				}
-			}
-		}
+	delegate: QuantityTable.TableRow {
+		color: "transparent"
```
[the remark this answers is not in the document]
blammit: We could have a Loader to optionally load the Rectangle, but I'm not sure if that's worthwhile?

components/QuantityGroupListHeader.qml
blammit: The `metricsFontSize` is overridden in some case to use a different font size than the one that is displayed, so we need two separate properties.

components/dialogs/SolarDailyHistoryDialog.qml
blammit: Done

components/QuantityTable.qml
MikeTrahearn-Qinetic: It depends what you are doing. But it is good practice so I would encourage you to adopt it.

src/quantityobject.h
```diff
@@ -41,6 +41,7 @@ class QuantityObject : public QObject
 	Q_PROPERTY(qreal numberValue READ numberValue NOTIFY numberValueChanged FINAL)
 	Q_PROPERTY(QString textValue READ textValue NOTIFY textValueChanged FINAL)
 	Q_PROPERTY(bool hasValue READ hasValue NOTIFY hasValueChanged FINAL)
+	Q_PROPERTY(bool hidden READ hidden WRITE setHidden NOTIFY hiddenChanged FINAL)
```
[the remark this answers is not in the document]
MikeTrahearn-Qinetic: I think visible is perfectly fine - this isn't a QQuickItem anyway. But whatever you think best.

pages/evcs/EvChargerPage.qml
MikeTrahearn-Qinetic: ah I had missed the next two lines. No matter.

src/quantityobject.h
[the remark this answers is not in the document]
blammit: Thinking about this again, "visible" is going to be confusing because normally when you set Item::visible=false, the UI layout space for the item is no longer reserved; here, we set opacity=0 when hidden=true, so the layout space is maintained.

blammit:
> Consider renaming usages of "fontSize" to fontPixelSize" to match the intended font property.

It would be more clear, though we should also rename other similar properties in that case (e.g. metricsFontSize -> metricsFontPixelSize) and that extra distinction seems unnecessary for now.

pages/evcs/EvChargerPage.qml
```diff
@@ -33,64 +33,49 @@ Page {
...
-							return 0
+					equalWidthColumns: true
+					model: root.evCharger.phases.count > 1 ? root.evCharger.phases.count : 0
+					delegate: QuantityTable.TableRow {
+						readonly property QtObject phase: root.evCharger.phases.get(index)
+
+						headerText: phase.name
+						model: QuantityObjectModel {
+							QuantityObject { object: phase; key: "power"; unit: VenusOS.Units_Watt }
+							QuantityObject {}
```
chriadam, line 76: These are a bit inscrutable - could a comment be added?

src/quantityobject.h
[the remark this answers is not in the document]
chriadam: Yeah, `hidden` is better, as it signals that the semantics don't match those of `visible`

pages/evcs/EvChargerPage.qml
blammit: Comment added -- these are empty because the columns are only relevant to the summary, not the table body.

PR #2220, opened 2025-06-11
Restore key navigation feature and update shortcuts

### LINKED ISSUE #2182: Enable key navigation and update shortcuts
When re-enabling key navigation:

- Allow Enter/Return to be used to activate a list item with a sub-menu
- If ESC is pressed when no other actions are available, remove the key navigation highlight
- If ESC is pressed when individual list items in the Notifications and Settings pages, remove the focus and show the full-list input blocker

blammit:
As the latest release has been made, restore the key navigation feature. Also add some key shortcuts:

- allow Enter/Return to activate a list item with a sub-menu
- on Settings and Notification pages: allow Escape to enable the full page blocker
- if Escape is pressed and the key is not already handled, disable key navigation

Also fix a bug where the current page would unexpectedly show the full page blocker, if the previously opened page had the blocker enabled.

Fixes #2182

### DIFF
--- a/Main.qml
+++ b/Main.qml
@@ -24,6 +24,9 @@ Window {
 	property real scaleFactor: 1.0
 	onIsDesktopChanged: Global.isDesktop = root.isDesktop
 
+	// Uncomment for key navigation debugging
+	// onActiveFocusItemChanged: console.info("** Active focused:", activeFocusItem, activeFocusItem?.title ?? activeFocusItem?.text ?? "")
+
 	function skipSplashScreen() {
 		Global.splashScreenVisible = false
 	}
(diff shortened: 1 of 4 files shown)


Main.qml
```diff
@@ -82,26 +85,31 @@ Window {
...
-		//     }
-		//     event.accepted = false
-		// }
+		Keys.onPressed: function(event) {
+			// If a key press is not handled by an item higher up in the hierarchy:
+			// Enable key navigation when an arrow or tab/backtab key is pressed.
+			// Disable it when the escape key is pressed.
+			if (Global.keyNavigationEnabled) {
+				if (event.key === Qt.Key_Escape) {
+					Global.keyNavigationEnabled = false
```
MikeTrahearn-Qinetic, line 94: This makes sense.

MikeTrahearn-Qinetic (approved):

PR #2223, opened 2025-06-11
Remove Mqtt and Dbus QML modules

### LINKED ISSUE #2217: Remove MqttDataManager and DBusDataManager
MqttDataManager and DBusDataManager were introduced at a time when it was difficult to share code between the MQTT, D-Bus and mock implementations due to the use of velib when veutil was not available. Now those limitations are gone, these manager types add an extra layer of complication with little benefit. In most cases, implementation files like GeneratorsImpl.qml, PvInvertersImpl.qml, etc. can be avoided by using ServiceDeviceModel to create device models directly in the main global object (e.g. Global.generators, Global.pvInverters).

The tasks are:
- Move MQTT and D-Bus Device object generation to the respective Global objects (e.g. move the contents of data/<mqtt|dbus>/EvChargersImpl.qml to EvChargers.qml)
- Remove MqttDataManager and DBusDataManager

MockDataManager can remain, as the mock implementations are used to coordinate various mock simulations.

blammit:
The modules are no longer needed, as MqttDataManager and DBusDataManager are no longer relevant. Their implementations have been moved into the service-specific global objects, with most backend-specific model implementations being abstracted by ServiceDeviceModel.

Fixes #2217


components/QuantityTableMetrics.qml
```diff
@@ -6,36 +6,24 @@
...
+		if (unit === VenusOS.Units_None) {
+			return defaultValue
 		}
 
 		// Give the unit symbol some extra space on the column.
 		const maxTextWidth = unit === VenusOS.Units_Energy_KiloWattHour
-						   ? advanceWidth("9999kWH")
-						   : advanceWidth("9999W")
-		return maxTextWidth + spacing
+						   ? advanceWidth("99.99kWH")
```
MikeTrahearn-Qinetic, line 22: _Strictly_ speaking, one would use the decimalPoint property of QLocale here.
Also, in some cases where the unit is shown in a separate Label, there is usually a small space between them, so I don't know if this is needed here.

data/Ess.qml
```diff
@@ -76,5 +68,91 @@ QtObject {
...
+		if (batteryLifeState !== null) {
+			_batteryLifeState.setValue(batteryLifeState)
+		}
+	}
+
+	function _getState() {
+		let hub4Mode = _hub4Mode.value
+		let currentState = _batteryLifeState.value
+		if (hub4Mode === undefined || currentState === undefined) {
+			return -1
```
MikeTrahearn-Qinetic, line 118: Is there no enum for Ess_State_Unknown?

data/Notifications.qml
```diff
@@ -193,6 +193,61 @@ QtObject {
 		}
 	}
 
+	readonly property Instantiator _notificationObjects: Instantiator {
+		model: root._modelLoader.item
+		delegate: Notification {
+			id: notification
+
+			required property string id
+			readonly property bool _canInitialize: _acknowledged.value !== undefined
```
MikeTrahearn-Qinetic, line 202: is this _canInitialise and _init bootstrap and the ctor/dtor necessary since Notification does this for you?

data/mock/DcInputsImpl.qml
```diff
@@ -27,9 +27,10 @@ QtObject {
 		target: Global.mockDataSimulator || null
 
 		function onSetDcInputsRequested(config) {
-			Global.dcInputs.reset()
 			while (_createdObjects.length > 0) {
-				_createdObjects.pop().destroy()
+				const obj = _createdObjects.pop()
+				Global.mockDataSimulator.setMockValue(obj.serviceUid + "/DeviceInstance", -1)
+				obj.destroy()
```
MikeTrahearn-Qinetic, line 33: Just a note: destroy() isn't immediate unlike a c++ dtor so a flag here to ensure that nothing depends on that behaviour.

data/mock/EvChargersImpl.qml
```diff
@@ -13,12 +13,12 @@ QtObject {
...
 	function populate() {
 		for (let i = 0; i < 3; ++i) {
-			createCharger({
+			const charger = createCharger({
 				position: i % 2 === 0 ? VenusOS.AcPosition_AcInput : VenusOS.AcPosition_AcOutput,
-				nrOfPhases: i + 1,
 				status: Math.floor(Math.random() * VenusOS.Evcs_Status_Charged),
 				mode: Math.floor(Math.random() * VenusOS.Evcs_Mode_Scheduled),
 			})
+			charger.phaseCount = i + 1
```
MikeTrahearn-Qinetic, line 21: move outside of the loop and do once.

data/mock/EvChargersImpl.qml
```diff
@@ -74,16 +79,16 @@ QtObject {
...
-						phases._phases.objectAt(i)._power.setValue(phasePower)
+						Global.mockDataSimulator.setMockValue(`${evCharger.serviceUid}/Ac/L${i+1}/Power`, phasePower)
 						totalPower += phasePower
 					}
 
-					_energy.setValue(1 + Math.random() * 10)
-					_current.setValue(1 + Math.random() * 20)
-					_power.setValue(totalPower)
-					_chargingTime.setValue(chargingTime + 60)
+					Global.mockDataSimulator.setMockValue(evCharger.serviceUid + "/Ac/Energy/Forward", 1 + Math.random() * 10)
```
MikeTrahearn-Qinetic, line 88: Side note for future project: it's ok to have singletons, but Global is too big and now causes a huge dependency tree between components. Suggest splitting it up into purpose-scoped separate singletons.

components/QuantityTable.qml
```diff
@@ -6,153 +6,128 @@
...
+			spacing: table.columnSpacing
+			fixedColumnWidth: table.equalWidthColumns ? (table.availableWidth - (table.columnSpacing * (columnCount-1))) / columnCount : NaN
+		}
+	}
 
-			readonly property var rowIndex: model.index
+	/*
+		A coloured rectangle with a QuantityRow for displaying a series of quantities.
+	*/
+	component TableRow: Rectangle {
```
MikeTrahearn-Qinetic, line 76: Put this into a separate QML file

pages/evcs/EvChargerPage.qml
```diff
@@ -33,64 +33,79 @@ Page {
...
-							return 0
+					width: chargerSummary.width
+					columnSpacing: chargerSummary.columnSpacing
+					equalWidthColumns: true
+					model: phaseModel.count > 1 ? phaseModel : null
+					delegate: QuantityTable.TableRow {
+						id: tableRow
+
+						required property string name
+						required property real power
```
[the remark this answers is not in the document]
MikeTrahearn-Qinetic: Given that model could be null, the use of required properties could _theoretically_ be problematic, although the delegate would not be created in such a case.

pages/evcs/EvChargerPage.qml
```diff
@@ -33,64 +33,79 @@ Page {
...
+						}
+						delegate: QtObject {
+							id: phaseObject
+
+							required property int index
+							required property string uid
+
+							readonly property VeQuickItem _power: VeQuickItem {
+								uid: phaseObject.uid + "/Power"
+								onValueChanged: phaseModel.setValue(index, PhaseModel.PowerRole, value)
```
MikeTrahearn-Qinetic, line 106: the use of index should be qualified with phaseObject just like the uid is.

pages/settings/debug/PagePowerDebug.qml
```diff
@@ -83,33 +83,37 @@ Page {
...
-						return qwacsPvInverter[`power${phaseName}`].value ?? NaN
-					case 4:
-						return qwacsPvInverter[`apparent${phaseName}`].value ?? NaN
-					case 5:
-						return sensorPvInverter[`power${phaseName}`].value ?? NaN
-					case 6:
-						return diffs[`power${phaseName}`]
+				id: quantityTable
+
+				width: parent?.width ?? 0
```
MikeTrahearn-Qinetic, line 88: Good: ```width: parent?.width ?? 0``` is a nice shortcut. However it also needs to be used everywhere else consistently.

pages/settings/debug/PagePowerDebug.qml
```diff
@@ -83,33 +83,37 @@ Page {
...
+						{ text: "AC Out", unit: VenusOS.Units_VoltAmpere },
+						{ text: "Qwacs", unit: VenusOS.Units_Watt },
+						{ text: "Qwacs", unit: VenusOS.Units_VoltAmpere },
+						{ text: "Sensors", unit: VenusOS.Units_Watt },
+						{ text: "Diff", unit: VenusOS.Units_Watt },
+					]
+				}
+				delegate: QuantityTable.TableRow {
+					id: tableRow
+					headerText: `L${index + 1}`
```
MikeTrahearn-Qinetic, line 106: I assume "L" is some kind of non-translatable prefix.

pages/solar/PvInverterPage.qml
```diff
@@ -58,21 +47,28 @@ Page {
...
-						return phase[columnProperties[column]]
+					metricsFontSize: phaseSummary.metricsFontSize
+					columnSpacing: phaseSummary.columnSpacing
+					model: root.pvInverter.phases.count > 1 ? root.pvInverter.phases : 0
+
+					delegate: QuantityTable.TableRow {
+						id: tableRow
+
+						required property string name
+						required property real energy
```
MikeTrahearn-Qinetic, line 60: All but name don't seem to be used.

pages/solar/SolarDevicePage.qml
```diff
@@ -22,90 +28,83 @@ Page {
...
+							QuantityObject { object: tracker; key: "voltage"; unit: VenusOS.Units_Volt_DC }
+							QuantityObject { object: tracker; key: "current"; unit: VenusOS.Units_Amp }
+							QuantityObject { object: tracker; key: "power"; unit: VenusOS.Units_Watt }
+						}
 
+						SolarTracker {
+							id: tracker
 							device: root.solarDevice
-							trackerIndex: index
+							trackerIndex: tableRow.index
```
MikeTrahearn-Qinetic, line 107: index here should be a required property on the tableRow delegate

components/QuantityTableMetrics.qml
blammit: The display will always use a decimal point, as the locale is not used to determine the decimal separator (as per some number of discussions about this). 

> Also, in some cases where the unit is shown in a separate Label, there is usually a small space between them, so I don't know if this is needed here.

Those spaces should be removed, so that the display is standardized across the app.

data/Notifications.qml
blammit: I haven't looked into this - the PR is mainly about moving the different implementations around, but would be nice to check whether this can be improved.

data/mock/EvChargersImpl.qml
blammit: I'm not sure what you mean, if the `i` comes from within the loop?

data/mock/EvChargersImpl.qml
blammit: Agreed, it should definitely be split up; that can be done separately.

components/QuantityTable.qml
blammit: It's quite tied to the QuantityTable implementation so perhaps it's okay to keep it in the same file for now. If it grows too much or the API between the three components are more formalized then it would be good to move elsewhere.

pages/evcs/EvChargerPage.qml
blammit: Sounds good, done

pages/settings/debug/PagePowerDebug.qml
blammit: Yes, each phase is L1-L3

pages/solar/PvInverterPage.qml
blammit: The properties are used by the QuantityObjects, so they do need to be added as required here.

data/mock/EvChargersImpl.qml
MikeTrahearn-Qinetic: Ah: OK - I had misunderstood as "the result of charger.phaseCount is surely always going to be 3 at the end right? This can be known outside the loop." but yes, the value being assigned here goes onto each separate instance being created so this is not an issue.

src/quantityobject.h
```diff
@@ -42,6 +42,10 @@ class QuantityObject : public QObject
 	Q_PROPERTY(QString textValue READ textValue NOTIFY textValueChanged FINAL)
 	Q_PROPERTY(bool hasValue READ hasValue NOTIFY hasValueChanged FINAL)
 
+	// If hidden=true, the UI should reserve the space for the quantity label in the layout, but
```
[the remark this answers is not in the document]
MikeTrahearn-Qinetic: Please make a separate ticket for this then so performance issues don't keep quietly building up.

components/QuantityTableMetrics.qml
MikeTrahearn-Qinetic: Please make a separate ticket for this.

blammit: https://github.com/victronenergy/gui-v2/issues/2253

data/Notifications.qml
blammit: https://github.com/victronenergy/gui-v2/issues/2254

src/quantityobject.h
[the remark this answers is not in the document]
blammit: https://github.com/victronenergy/gui-v2/issues/2255

Would be useful to see whether profiling shows any issues with QuantityRow that could be avoided if a RowLayout was used instead.

data/Ess.qml
```diff
@@ -76,5 +68,91 @@ QtObject {
...
+			break
+		case VenusOS.Ess_State_ExternalControl:
+			batteryLifeState = VenusOS.Ess_BatteryLifeState_Disabled
+			break
+		default:
+			console.warn("Unrecognised ESS state:", essState)
+			break
+		}
+		if (batteryLifeState !== null) {
+			_batteryLifeState.setValue(batteryLifeState)
```
chriadam, line 110: is the ordering important here (i.e. you do _batteryLifeState.setValue() after _hub4Mode.setValue(), and I'm wondering if systemcalc consumes these, and whether the ordering of value changes is important or not).

data/DcInputs.qml
```diff
@@ -122,5 +105,46 @@ QtObject {
...
+			}
+
+			onVoltageChanged: Qt.callLater(root.updateTotals)
+			onCurrentChanged: Qt.callLater(root.updateTotals)
+			onPowerChanged: Qt.callLater(root.updateTotals)
+		}
+	}
+
+	readonly property Loader _modelLoader: Loader {
+		sourceComponent: BackendConnection.type === BackendConnection.MqttSource ? mqttModelComponent : dbusOrMockModelComponent
```
chriadam, line 129: can you briefly describe why these can't be fully unified?

chriadam (approved):
Two questions, but LGTM.

data/Ess.qml
blammit: I don't know about that - this is just following the implementation from gui-v1, which sets the Hub 4 mode, then the BatteryLife state.

data/DcInputs.qml
blammit: If serviceType='tank', the D-Bus child uids might be like: [ "dbus/com.victronenergy.tank.adc_builtin1_1", "dbus/com.victronenergy.tank.adc_builtin0_4" ]

Whereas for MQTT, the child ids might be like: [ "mqtt/tank/22", "mqtt/tank/23" ]

So, D-Bus uids, we need to fetch service uids using a regex pattern, since the uid might be like "dbus/com.victronenergy.tank.adc_builtin0_4" with a random device id on the end, which means we need a VeQItemSortTableModel. On the other hand, for MQTT uids, we can fetch the uids using a VeQItemTableModel, since the root path is just "mqtt/tank" and no regex patterns are required. 

We could simplify this abstraction by using a VeQItemSortTableModel for both cases, if for the MQTT case we use a VeQItemSortTableModel with an empty filter, removing the need for a separate Loader. VeQItemSortTableModel is a little heavier as it is a QSortFilterProxyModel, but maybe it's okay if the filtering will not be used anyway.

chriadam: Yep, separate PR is fine.  Thanks!

PR #2224, opened 2025-06-12
Disable the watchdog from reloading wasm builds until initialized.

### LINKED ISSUE #2201: Look into watchdog timeout tripping on local lan
When trying to open gui-v2 WASM on local LAN, having the browser set to 3G throttling, you can never open it.

A watchdog timeout expires causing a reload; so there is never enough time to fully download the wasm file.

Perhaps the browser/javascript can see that a download is ongoing and not stuck - and then not trip?

Cause to look into this: https://community.victronenergy.com/t/new-ui-issues-due-to-low-bandwidth/34224/7

I'll first discuss this with Dirk-Jan

DanielMcInnes:
This gives enough time for slow connections to load wasm builds. Fixes #2201


src/backendconnection.cpp
```diff
@@ -269,7 +269,12 @@ void BackendConnection::openUrl(const QString &url)
 
 void BackendConnection::hitWatchdog()
 {
-	emscripten_run_script("watchdogHit = true"); // 'watchdogHit' is defined in index.html, which checks it periodically and reloads the page if not hit regularly.
+    // 'watchdogHit' and 'initialized' are defined in index.html.
```
chriadam, line 272: whitespace issue: use tabs instead of spaces in cpp code

chriadam (approved):

PR #2234, opened 2025-06-13
Brief & Overview: Show gauge colours when current is < 1.0

### LINKED ISSUE #2233: Gauge colours not shown when AC load current is < 1.0
For example, see https://vrm.victronenergy.com/installation/151755/dashboard:

Image

The AC Load is 0.9 A, with min=0 and max=5.4 according to the min/max gauge preferences, but the gauge does not have any colour.

blammit:
There are some checks to guard against showing gauge values if they are between -1.0 and 1.0, in case these values are occurring due to noise signals rather than significant value changes. These checks make sense for power values, which are only significant when < -1.0 or > 1.0, but current values can often be in the -1 to 1.0 range.

The changes are:
- Remove "phaseModelProperty" as it always refers to "current" anyway
- Always require a PhaseModel as the model input, to ensure the "power" and "current" role names are availalable
- Only check for noise/insigificant values for the "power" property, and do not check this for the "current" property

Fixes #2233

### DIFF
--- a/components/SideMultiGauge.qml
+++ b/components/SideMultiGauge.qml
@@ -9,6 +9,7 @@ import Victron.VenusOS
 Item {
 	id: root
 
+	required property PhaseModel phaseModel
 	property int valueType: VenusOS.Gauges_ValueType_NeutralPercentage
 	property int direction
 	property real startAngle
@@ -17,8 +18,6 @@ Item {
 	property real arcVerticalCenterOffset
 	property real phaseLabelHorizontalMargin
 	property bool animationEnabled
-	property var phaseModel
-	property string phaseModelProperty
 	property real minimumValue
 	property real maximumValue
 	property bool inputMode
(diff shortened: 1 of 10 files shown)


components/SideMultiGauge.qml
```diff
@@ -48,20 +40,21 @@ Item {
...
+			required property real current
+
 			width: Theme.geometry_briefPage_edgeGauge_width
 			height: root.height
 
-			// ignore noise values (close to zero)
-			readonly property real modelValue: Math.floor(Math.abs(model[root.phaseModelProperty] || 0)) < 1.0 ? 0.0
-					: model[root.phaseModelProperty]
-			readonly property bool feedingToGrid: root.inputMode && modelValue < 0.0
+			readonly property bool feedingToGrid: root.inputMode && power < -1.0 // ignore noise values (close to zero)
```
chriadam, line 50: should this condition be `Math.abs(power) < 1.0` ?

chriadam: Ah, no, I see, negative power values means that it's feeding to grid, but we want to ignore values close to 0.0 since they might just be noise.  Ignore me.

Issue #2218, opened 2025-06-10
Also show centre gauge for four ring setups


mpvader:
https://victrondevelopment.slack.com/archives/C020QG3JVEW/p1746679117743579

PR #2243, opened 2025-06-16
Show center details on Brief Page when 4 gauges are visible

chriadam:
Use a normal Label with font fitting to ensure that the text is sized appropriately to the available space.

Contributes to issue #2218

### DIFF
--- a/components/BriefCenterDisplay.qml
+++ b/components/BriefCenterDisplay.qml
@@ -12,7 +12,6 @@ Column {
 	id: root
 
 	property bool showFullDetails
-	property bool smallTextMode
 
 	readonly property bool _useTemperature: BackendConnection.portableIdInfo(centerService.value).type === "temperature"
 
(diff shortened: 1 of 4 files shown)


chriadam:
It would perhaps be better if I fixed QuantityLabel to allow proper font fitting / size modes - but I'm not sure how (as Text doesn't report which font size is "actually" used by the fitting algorithm (i.e. font.pixelSize is used as the maximum value; it isn't updated to reflect the utilised value).

components/BriefCenterDisplay.qml
```diff
@@ -62,29 +61,22 @@ Column {
...
+		horizontalAlignment: Text.AlignHCenter
+		verticalAlignment: Text.AlignVCenter
+		text: quantityLabel.valueText + quantityLabel.unitText
+		font.pixelSize: Theme.font_briefPage_battery_percentage_maximumPixelSize
+		fontSizeMode: Text.HorizontalFit
+	}
 
+	QuantityLabel {
+		id: quantityLabel
+		visible: false
```
MikeTrahearn-Qinetic, line 76: Not sure of the context here but is this always visible: false? If so can it be removed?

chriadam: It's used to determine the appropriate text to display, since quantity label has some QuantityInfo-derived units/precision magic in it.

chriadam:
Another (probably nicer) option is: https://github.com/victronenergy/gui-v2/pull/2247

PR #2247, opened 2025-06-17
Show center details on Brief Page when 4 gauges are visible

chriadam:
Use a QuantityLabel with font fitting to ensure that the text is sized appropriately to the available space.

Contributes to issue #2218

### DIFF
--- a/components/FittedQuantityLabel.qml
+++ b/components/FittedQuantityLabel.qml
@@ -1,0 +1,27 @@
+/*
+** Copyright (C) 2025 Victron Energy B.V.
+** See LICENSE.txt for license information.
+*/
+
+import QtQuick
+import Victron.VenusOS
+
+QuantityLabel {
+	id: root
+
+	property int minimumPixelSize: -1
+	property int maximumPixelSize: -1
+
+	readonly property real _maximumWidth: width - Theme.geometry_quantityLabel_spacing
+	readonly property string _combinedText: valueText + unitText
+
+	on_MaximumWidthChanged: Qt.callLater(_calculateFontFitting)
+	on_CombinedTextChanged: Qt.callLater(_calculateFontFitting)
+	onVisibleChanged:       Qt.callLater(_calculateFontFitting)
+
+	function _calculateFontFitting() {
+		if (root.visible) {
+			root.font.pixelSize = FastUtils.fittedPixelSize(root._combinedText, root._maximumWidth, root.minimumPixelSize, root.maximumPixelSize, root.font, Theme)
+		}
+	}
+}
--- a/src/fastutils.cpp
+++ b/src/fastutils.cpp
@@ -5,6 +5,8 @@
 
 #include "fastutils.h"
 
+#include <QFontMetricsF>
+
 namespace Victron {
 namespace VenusOS {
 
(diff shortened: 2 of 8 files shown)


components/FittedQuantityLabel.qml
```diff
@@ -0,0 +1,21 @@
...
+
+	property int minimumPixelSize: -1
+	property int maximumPixelSize: -1
+
+	readonly property real _maximumWidth: width - Theme.geometry_quantityLabel_spacing
+	readonly property string _combinedText: valueText + unitText
+
+	on_MaximumWidthChanged: if (root.visible) root.font.pixelSize = FastUtils.fittedPixelSize(root._combinedText, root._maximumWidth, root.minimumPixelSize, root.maximumPixelSize, root.font, Theme)
+	on_CombinedTextChanged: if (root.visible) root.font.pixelSize = FastUtils.fittedPixelSize(root._combinedText, root._maximumWidth, root.minimumPixelSize, root.maximumPixelSize, root.font, Theme)
+	onVisibleChanged:       if (root.visible) root.font.pixelSize = FastUtils.fittedPixelSize(root._combinedText, root._maximumWidth, root.minimumPixelSize, root.maximumPixelSize, root.font, Theme)
```
blammit, line 20: Can move the repeated code into a function?

I'm wondering if this is called when `visible` changes, is it called unnecessarily when switching to/from the Brief page?

src/fastutils.cpp
```diff
@@ -33,5 +35,66 @@ qreal FastUtils::degreesToRadians(const qreal degrees) const
...
+// If the Theme singleton is specified, check the various theme values from largest to smallest.
+// Otherwise, use binary search to find an arbitrary font size which works.
+int FastUtils::fittedPixelSize(const QString &text, const qreal maxWidth, int minPixelSize, int maxPixelSize, const QFont &font, ThemeSingleton *theme) const
+{
+	if (maxWidth <= 0 || maxPixelSize <= 0 || minPixelSize == maxPixelSize || text.isEmpty()) {
+		return minPixelSize;
+	}
+
+	int curr;
+	QFont f(font);
```
blammit, line 48: Use more descriptive variable names?

blammit (commented):
Works really nicely, just have some minor comments

src/fastutils.cpp
chriadam: done

components/FittedQuantityLabel.qml
blammit: What I mean is, does BriefPage become visible=false when we switch to the Overview page, or when a page is added to the stack, or the control cards are shown? If so then the font size would be recalculated when switching back to the Brief page, which would seem unnecessary, since the text may not have changed.

chriadam: I hope that visible becomes false once we switch to another page (and the animation transition has fully completed).  We don't want to be rendering anything offscreen.  When we switch back to the Brief page from the Overview page, it will become visible, and we will have to recalculate the fitted pixel size at that point (because the value may have changed while visible was false, but we don't recalculated the fitted pixel size in that case since visible is false).

blammit: Ah, yeah I see. Yep, LGTM

PR #2252, opened 2025-06-18
Show used sensor in DVCC settings as expected

blammit:
The "Used sensor" setting should be shown when
/Settings/SystemSetup/TemperatureService = "default", not when the "/AvailableTemperatureServices" selected value is "default". The latter will never be true.


blammit:
This was an error in the original porting. Original gui-v1 code is here, showing the correct check: https://github.com/victronenergy/gui/blob/master/qml/PageSettingsDVCC.qml#L157

chriadam (approved):

Issue #2226, opened 2025-06-12
Add a setting to disable overview + gauge animations


chriadam:
Animations cause a lot of rendering.  Currently, we have an "automatic" pausing of animations when CPU usage exceeds a high watermark temporarily.  However, for some installations, the user may prefer to ALWAYS disable animations, via the Settings page.

This should pause animations of electrons, gauge value changes, and battery charge animation on the Overview page, and also gauge value changes on Brief page.  (Need to decide what to do with Brief Page Side Panel graph animations...)

chriadam:
Implemented in https://github.com/victronenergy/gui-v2/pull/2258 however we need a setting to be added to systemsettings: "/Settings/Gui2/AnimationDisabled" with a default value of 0 (i.e. animations are not disabled).  @ReinvdZee is this something you can help with?

chriadam:
Need to switch the image used for the overview electrons to be arrows when animations are paused or disabled.  Asked Serj whether he could provide an icon; he is working on it :-)

PR #2258, opened 2025-06-18
Implement UIAnimations system setting

chriadam:
When this setting value is 1, all animations (including Overview page electron animations, gauge value change animations, page transition animations, and control card pane or side panel opening and closing animations) will be enabled; otherwise they will be disabled.

Disabling animations greatly reduces CPU usage on device.

Contributes to issue #2226

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1137,6 +1137,7 @@ qt_add_resources(${PROJECT_NAME} "${PROJECT_NAME}_large_resources"
         images/dcloads.svg
         images/dropdown.svg
         images/electron.svg
+        images/electron_arrow.svg
         images/ess.svg
         images/freshWater.svg
         images/key_navigation_highlight_dark.svg
--- a/Global.qml
+++ b/Global.qml
@@ -30,7 +30,7 @@ QtObject {
 	property var notificationLayer
 	property ScreenBlanker screenBlanker
 	property bool displayCpuUsage
-	property bool pauseElectronAnimations
+	readonly property bool animationEnabled: (systemSettings?.animationEnabled ?? true) && BackendConnection.applicationVisible
 
 	// data sources
 	property var acInputs
@@ -58,8 +58,6 @@ QtObject {
 	property bool isDesktop
 	property bool isGxDevice: Qt.platform.os === "linux" && !isDesktop
 	property real scalingRatio: 1.0
-
-	property bool animationEnabled: true // for mock mode only.
 
 	readonly property int int32Max: _intValidator.top
 	readonly property int int32Min: _intValidator.bottom
--- a/components/CardViewLoader.qml
+++ b/components/CardViewLoader.qml
@@ -12,10 +12,11 @@ Loader {
 	required property Item statusBarItem
 	required property Item navBarItem
 	required property Item swipeViewItem
+	required property bool animationEnabled
 	property color backgroundColor: Theme.color_page_background
 	readonly property bool animationRunning: inAnimation.running || outAnimation.running
-	property int animationDuration: Theme.animation_controlCards_slide_duration
 	property bool viewActive: false
+	readonly property int _animationDuration: animationEnabled ? Theme.animation_controlCards_slide_duration : 1
 
 	active: viewActive
 	onActiveChanged: if (active) active = viewActive // remove binding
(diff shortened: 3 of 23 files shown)


Global.qml
```diff
@@ -30,7 +30,7 @@ QtObject {
 	property var notificationLayer
 	property ScreenBlanker screenBlanker
 	property bool displayCpuUsage
-	property bool pauseElectronAnimations
+	property bool animationDisabled: !!systemSettings ? systemSettings.animationDisabled : false
```
blammit, line 33: Could the name be kept as `animationEnabled` instead of a negative `animationDisabled`? (Personally I find it easier to read)

Also this could use dot notation like `systemSettings?.animationDisabled` instead of a separate `systemSettings` check.

pages/MainView.qml
```diff
@@ -408,7 +410,7 @@ FocusScope {
 			return customButton
 		}
 		rightButton: !!root.currentPage ? root.currentPage.topRightButton : VenusOS.StatusBar_RightButton_None
-		animationEnabled: BackendConnection.applicationVisible
+		animationEnabled: BackendConnection.applicationVisible && !Global.animationDisabled
```
blammit, line 413: Should animations always be disabled if `BackendConnection.applicationVisible == false`? If so maybe `Global.animationDisabled` should check `BackendConnection.applicationVisible` so that we don't need to do it elsewhere.

components/CardViewLoader.qml
```diff
@@ -12,9 +12,10 @@ Loader {
 	required property Item statusBarItem
 	required property Item navBarItem
 	required property Item swipeViewItem
+	required property bool animationEnabled
 	property color backgroundColor: Theme.color_page_background
 	readonly property bool animationRunning: inAnimation.running || outAnimation.running
-	property int animationDuration: Theme.animation_controlCards_slide_duration
+	readonly property int animationDuration: animationEnabled ? Theme.animation_controlCards_slide_duration : 1
```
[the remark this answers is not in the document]
chriadam: It's just used within this file.  renamed with underscore.
Duration is 1 instead of 0 because when it is 1 some of the animators don't run and the transition never occurs properly.

CMakeLists.txt
```diff
@@ -1130,6 +1130,7 @@ qt_add_resources(${PROJECT_NAME} "${PROJECT_NAME}_large_resources"
         images/dcloads.svg
         images/dropdown.svg
         images/electron.svg
+        images/electron_arrow.svg
```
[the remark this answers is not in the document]
MikeTrahearn-Qinetic: Could you not get PNGs for this? Just wondering if the drawing of SVGs for the electrons might be just a tad quicker with a PNG rather than rendering the SVGs. Pixmap caching probably gets around this, but it may be worth a thought, especially if there are a lot of them.

components/CardViewLoader.qml
```diff
@@ -12,10 +12,11 @@ Loader {
 	required property Item statusBarItem
 	required property Item navBarItem
 	required property Item swipeViewItem
+	required property bool animationEnabled
```
MikeTrahearn-Qinetic, line 15: Does this property need to even exist if the status is now available from ```Global.animationEnabled```?

components/LoadGraph.qml
```diff
@@ -20,7 +20,7 @@ Item {
 	property color horizontalGradientColor1: Theme.color_briefPage_background
 	property color horizontalGradientColor2: "transparent"
 	property bool zeroCentered
-	property alias active: graphAnimation.running
+	property alias animationEnabled: graphAnimation.running
```
MikeTrahearn-Qinetic, line 23: same here: the graphAnimation would be drivable from ```Global.animationEnabled``` so this property alias could be removed

CMakeLists.txt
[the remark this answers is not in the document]
chriadam: I believe that after the first one is loaded, the pixmap cache will be hit for the rest.  But yes, given that we probably don't resize them for different display sizes, PNGs probably makes more sense.

Global.qml
```diff
@@ -30,7 +30,7 @@ QtObject {
 	property var notificationLayer
 	property ScreenBlanker screenBlanker
 	property bool displayCpuUsage
-	property bool pauseElectronAnimations
+	property bool animationEnabled: (systemSettings?.animationEnabled ?? true) && BackendConnection.applicationVisible
```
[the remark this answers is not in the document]
chriadam: I will double check this.  At least rendering definitely doesn't occur.  I assume I stopped the animations also, but will double check.

chriadam:
I also need to edit the commit message and PR description, as the setting is now `/Settings/Gui2/UIAnimationsEnabled` rather than `/Settings/Gui2/AnimationsDisabled`.  But this might change - need to double check with Rein.

Global.qml
blammit, line 33: Should this be `readonly`?

blammit (commented):
Great, this works really nicely.

These are still animated, should they also be disabled or are they exempt?

- When the Overview and Levels pages expand/contract
- When a toast notification fades into view
- When a dialog shows/hides

Global.qml
chriadam: Yes

blammit (approved):

PR #2260, opened 2025-06-19
Text input fixes

blammit:


components/listitems/core/ListTextField.qml
```diff
@@ -167,10 +167,15 @@ ListItem {
 			root.editingFinished()
 		}
 
-		// Consume arrow key events so that they do not go higher up the item hierarchy and trigger
-		// key navigation events while the text field is focused.
-		Keys.onLeftPressed: {}
-		Keys.onRightPressed: {}
+		// When the cursor is on the left/right edges, consume left/right key events so that they do
+		// not travel higher up the item hierarchy and activate key navigation.
+		Keys.onLeftPressed: (event) => { event.accepted = textField.cursorPosition === 0 }
```
chriadam, line 172: should this also require textField.activeFocus?

chriadam (approved):
One question, but tested in mock mode and it seems to work perfectly as-is.

components/listitems/core/ListTextField.qml
blammit: Done, it's not strictly required but it's nice for the consistency.

PR #2263, opened 2025-06-19
Miketrahearn/performance fixes

MikeTrahearn-Qinetic:
Various non-related commits in no particular sequence for review.

In particular stats as measured:

Before: Solar Device Page loading time: 452ms vs After: 361ms
Before: Page Solor Charger: 932ms vs After: 838ms

### DIFF
--- a/components/listitems/core/ListItem.qml
+++ b/components/listitems/core/ListItem.qml
@@ -99,18 +99,18 @@ BaseListItem {
 	Keys.enabled: Global.keyNavigationEnabled
 
 	// Show thin colored indicator on left side if settings is only visible to super/service users
-	Rectangle {
-		visible: root.showAccessLevel >= VenusOS.User_AccessType_SuperUser
-		width: Theme.geometry_listItem_radius * 2
-		height: parent.height
-		color: Theme.color_listItem_highAccessLevel
-		radius: Theme.geometry_listItem_radius
-
-		Rectangle {
-			x: Theme.geometry_listItem_radius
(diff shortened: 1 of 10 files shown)


pages/solar/SolarInputListPage.qml
```diff
@@ -70,13 +70,15 @@ Page {
 	// inject it into solarInputModel.
 	Instantiator {
 		model: Global.solarDevices.model
-		delegate: Instantiator {
```
blammit, line 73: Great - it looks like the `trackerObjects` Instantiator can actually be turned into the root type for this delegate, so you don't even need the extra QtObject.

components/listitems/core/ListItem.qml
```diff
@@ -123,13 +123,6 @@ BaseListItem {
 		radius: root.background.radius
 		effectEnabled: root.interactive
 		onClicked: root.activate()
-
-		Connections {
-			target: pressArea.toast
```
blammit, line 128: Do you recall why this was needed previously?

blammit (approved):
Great to see these improvements. Some minor comments, otherwise LGTM.

pages/solar/SolarInputListPage.qml
MikeTrahearn-Qinetic: I updated the file to remove the extra delegate level - could you check it please in case in so doing, I might have broken something?

blammit: I checked and it's all still working fine. Looks good to merge.

PR #2271, opened 2025-06-19
Service model improvements

blammit:


components/ServiceDeviceModel.qml
```diff
@@ -6,14 +6,14 @@
...
+// Provides a generic DeviceModel for the specified service types.
 //
-// All services found for the specified type will be added and removed as Device objects.
+// All services found for the specified types will be added and removed as Device objects.
 
 DeviceModel {
 	id: root
 
-	required property string serviceType
+	property alias serviceTypes: serviceModel.serviceTypes
```
MikeTrahearn-Qinetic, line 16: ServiceModel's serviceTypes property is required, but ServiceDeviceModel's alias to that property is not - this could cause DeviceModel to be attempted to be created without its serviceTypes property being set which would result in a run-time error. DeviceModel must therefore also define a required property to satisfy this condition. Note: you cannot write ```required property alias serviceTypes: serviceModel.servicesTypes``` so you must write ```required property list<string> serviceTypes``` instead, and then bind the ServiceModel instance's serviceTypes property to it.

components/SolarDeviceModel.qml
```diff
@@ -17,11 +17,8 @@ DeviceModel {
 			id: serviceInstantiator
 
 			required property string modelData
-			readonly property ServiceModelLoader modelLoader: ServiceModelLoader {
-				serviceType: modelData
-			}
 
-			model: modelLoader.item
+			model: ServiceModel { serviceTypes: [modelData] }
```
MikeTrahearn-Qinetic, line 21: square brackets not needed around modelData as it is declared a ```string``` type **OR** the ```required property string modelData should be ```required property var modelData``` because it is array. Either way, the square brackets isn't needed. But check the type of the modelData - I feel it is a string ARRAY not a string.

blammit:
@chriadam updated this to avoid the errors on ControlCardsPage about trying to add a device that was already added.

blammit:
@chriadam can you check this one please?

data/DcInputs.qml
```diff
@@ -106,7 +106,7 @@ QtObject {
 	}
 
 	readonly property Instantiator _dcInputObjects: Instantiator {
-		model: root._modelLoader.item
+		model: ServiceModel { serviceTypes: ["alternator","fuelcell","dcsource","dcgenset"] }
```
DanielMcInnes, line 109: Do all ServiceModels need to have a modelId specified?

I see in AggregateDeviceModel::setSourceModels(const QVariantList &models):

		BaseDeviceModel *model = modelVariant.value<BaseDeviceModel*>();
		if (model->modelId().isEmpty()) {
			qmlInfo(this) << "cannot use BaseDeviceModel* with empty modelId value:" << modelVariant.userType() << " " << modelVariant.typeName();
			continue;
		}

blammit: Ah yes - you're right, normally they would, but it turns out we don't need the `modelId` anymore. It was only needed when we would sometimes have the same service added from different DeviceModels to the same AggregateDeviceModel, and thus we needed a way to distinguish the model entries, as they had the same serviceUid. 

But we don't do that in the Device List anymore, so the modelId can be removed. I'll do that in a separate PR.

DanielMcInnes (approved):

PR #2278, opened 2025-06-22
PageDebugVeQItems: dump stringified tree/value and show value type

blammit:
When the P key is pressed on a highlighted item in the value list, print the stringified value if it is a leaf node, otherwise dump the stringified object tree.

Also, the debug information for each path:
- show the value type below the path
- stringify the displayed values, so that map and object data are displayed, instead of just showing as empty strings

This helps with seeing the exact data that is stored for each path.


blammit:
This expands the debug information, so that instead of seeing:

image

You see: 
image

PR #2285, opened 2025-06-25
Further improvements for modifications checks (support state) page

### LINKED ISSUE #2238: Improve modifications checks page
The current page seems to be unclear, therefore it needs to be made more clear and easier to understand.

Some proposals of @mpvader:

- group modifications things together. So: move "Modifications loaded at boot" and "Modifications enabled yes/no" close together.
- consider renaming that 2nd one to "Prevent installed modifications from running" or a better text.
- change that switch to no longer control node-red & signalk. Its double now, confusing.

So those two in their own box, clearly separated from other things.

- add indicator if node-red or signalk are running. and refer to other menu in case people want to disable that.
- add indicator if modbus-tcp server is running, same thing.
- add indicator if mqtt is being used and is controlling the system from an external thing... but that's a difficult one. Hard to implement. MQTT is always running; and to figure out if its being used to control the system requires work by Wiebe etc.

mr-manuel:
Fixes #2238

### DIFF
--- a/pages/settings/PageSettingsGeneral.qml
+++ b/pages/settings/PageSettingsGeneral.qml
@@ -23,7 +23,7 @@ Page {
 		}
 
 		if (modbusTcpItem.value !== 0){
-			//% "Modbus-TCP"
+			//% "Modbus TCP Server"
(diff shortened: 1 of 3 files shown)


mr-manuel:
@blammit ready to review and merge

PR #2287, opened 2025-06-25
Remove import Qt.Quick.Controls

MikeTrahearn-Qinetic:
This PR is to:
- remove redundant ```import QtQuick.Controls``` and ```import QtQuick.Controls.impl```
- remove the **run-time** style selection (we want to use the style from Victron.VenusOS (components) as a **faster** [compile-time style selection](https://doc.qt.io/qt-6/qtquickcontrols-styles.html#compile-time-style-selection) which helps the code model immensely as well. 
- remove the possibility of visual inconsistencies on different platforms when using the "fallback" [default style](https://doc.qt.io/qt-6/qtquickcontrols-styles.html#default-styles)
- make sure that when you ```import Victron.VenusOS``` you can know exactly what you are getting: 
      Priority 1: all the components that are defined in Victron.VenusOS and _only then:_
      Priority 2: where such is not available (e.g. Page, Pane etc) you get is absolutely prescribed - the controls from the **QtQuick.Controls.Basic style**
- ensure that all Victron.VenusOS components use QtQuick.Controls.Templates and avoid type name clashes (RangeSlider)
- Only the Qt.Quick.Controls.Basic style library needs to be shipped to target - all the others if provided can be eliminated.
- a contrived palette can be added to the ApplicationWindow component such that "purple" is used for all palette roles. This means that if any Control be present from the QtQuick.Controls.Basic library, they will immediately look purple and alert to attention - i.e. a potential new customised addition to the Victron.VenusOS component set.

### DIFF
--- a/ApplicationContent.qml
+++ b/ApplicationContent.qml
@@ -5,7 +5,6 @@
 
 import QtQuick
 import QtQuick.Window
-import QtQuick.Controls as C
 import Victron.VenusOS
 
 FocusScope {
@@ -138,7 +137,7 @@ FocusScope {
 		source: Global.isGxDevice
 				? "qrc:/qt/qml/Victron/VenusOS/components/InputPanel.qml"
 				: "qrc:/qt/qml/Victron/VenusOS/components/WasmVirtualKeyboardHandler.qml"
-		parent: C.Overlay.overlay
+		parent: Overlay.overlay
 		z: 1
 	}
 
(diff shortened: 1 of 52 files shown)


chriadam:
I will merge this one after the Qt 6.8 change goes through - just to reduce surface area of possible issues, and don't want to delay the Qt 6.8 update.

chriadam (approved):
LGTM

PR #2292, opened 2025-06-26
WASM: improve watchdog and initialized variables

chriadam:
Use a prefixed name for the initialized variable to avoid clashes. Check that the variables are defined in the JS environment before attempting to set them, in case they are not provided by VRM, as VRM does not use our index.html.

### DIFF
--- a/src/backendconnection.cpp
+++ b/src/backendconnection.cpp
@@ -269,13 +269,13 @@ void BackendConnection::openUrl(const QString &url)
 
 void BackendConnection::hitWatchdog()
 {
-	// 'watchdogHit' and 'initialized' are defined in index.html.
-	// After this function sets 'initialized' to true, a timer in index.html periodically:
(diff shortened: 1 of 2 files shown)


DanielMcInnes (approved):
lgtm. Did you want to put a big comment in index.html warning that vrm has its own version of index.html, that needs to be updated independently?

chriadam:
Funnily enough, this PR doesn't actually work.  I still get:
```
ReferenceError: guiv2initialized is not defined
    at eval (eval at _emscripten_run_script (eval at completeLoadEmscriptenModule (https://vrm.victronenergy.com/js/app.811be72d.v1.js:9:71501)), <anonymous>:1:1)
    at _emscripten_run_script (eval at completeLoadEmscriptenModule (https://vrm.victronenergy.com/js/app.811be72d.v1.js:9:71501), <anonymous>:8:225081)
```

PR #2306, opened 2025-07-02
Use British unit spelling instead of American

### LINKED ISSUE #2288: Spelling: "Liters" to "Litres" on data units page
Settings > General > Display & Appearance > Data units, Liters should be spelt Litres

Thanks to Jono.

ybott-qinetic:
Use British unit spelling instead of American

Fixes #2288


src/units.cpp
```diff
@@ -16,9 +16,9 @@ static const QString DegreesSymbol = QStringLiteral("\u00b0");
 Unit::Type unitToVeUnit(Victron::VenusOS::Enums::Units_Type unit)
 {
 	switch (unit) {
-	case Victron::VenusOS::Enums::Units_Volume_CubicMeter:
-		return Unit::CubicMeter;
-	case Victron::VenusOS::Enums::Units_Volume_Liter:
+    case Victron::VenusOS::Enums::Units_Volume_CubicMetre:
```
blammit, line 19: The indentations in this file use four spaces but should be tabs instead. Maybe .cpp files haven't been added to your .git attributes to fix the whitespace automatically?

This is my `.git/info/attributes` file:
```
*.qml  filter=tabspace
src/*.h  filter=tabspace
src/*.hpp  filter=tabspace
src/*.cpp  filter=tabspace
components/*.js  filter=tabspace
```
And then in ~/.gitconfig I have:

```
[filter "tabspace"]
    clean = gunexpand --tabs=4 --first-only
    smudge = gexpand --tabs=4 --initial
```

i18n/venus-gui-v2.ts
```diff
@@ -3917,7 +3917,7 @@
     <message id="settings_gps_speed_kmh">
```
blammit, line 3917: This `ts` change should be removed

src/units.cpp
```diff
@@ -16,9 +16,9 @@ static const QString DegreesSymbol = QStringLiteral("\u00b0");
...
+        return Unit::CubicMetre;
```
blammit, line 20: The `Unit` enums in `veutil` also need to be updated for this to build.

ybott-qinetic: Fixed

src/units.cpp
ybott-qinetic: Thanks.

i18n/venus-gui-v2.ts
ybott-qinetic: Removed.

Issue #2170, opened 2025-05-24
Control cards: Battery life limit


mr-manuel:
1. ~The "Battery life limit" in the ESS menu is named "Active SOC limit" would be nice, if they have matching names.~
    Fixed with https://github.com/victronenergy/gui-v2/commit/975b11c330e63ec1a56d5fe391e71d7b07426cf2

2. Redesign the current "Battery life limit" item from 
    ![Image](https://github.com/user-attachments/assets/7cb3cf5d-a46a-4293-8b01-a04b32178d66)
    to
    ![Image](https://github.com/user-attachments/assets/4dcaec1f-a77b-4f05-86cb-5c7992147dbb)

3. Show a popup with this text, when clicking on "Battery life limit"
    ```
    BatteryLife dynamically adjusts the minimum battery state of charge to prevent deep discharges and ensure regular full charges, helping to prolong battery life and maintain system reliability.
    ```

mpvader:
@mr-manuel pls confirm that the remaining issue here is that there is an (i) icon that doesn't work

mpvader:
ok, changed this to high prio. Its really a bug that that (i) is not clickable.

Text you proposed is ok.

PR #2313, opened 2025-07-05
Show BatteryLife info toast in ESS control card

ybott-qinetic:
Modify battery life info layout and colours

Fixes 2170

### DIFF
--- a/data/Ess.qml
+++ b/data/Ess.qml
@@ -16,6 +16,13 @@ QtObject {
 
 	readonly property var stateModel: [
 		{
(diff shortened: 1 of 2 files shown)


chriadam:
Nitpick: in commit message, change it to "Fixes #2170" so that github will automatically link PR to issue.  Without the hash before the issue number, it won't.

ybott-qinetic:
@chriadam done, will make sure they link correctly in future ...

chriadam (approved):

Issue #984, opened 2024-04-23
Explore possibility of automating GUI v2 UI testing


jpetrell:
Why
- There are too many features to maintain by a small team
- Manually testing 169 pages after every code change is not practical

Investigate possibility of
- Toggling through all the main views and all the mock configs (26 brief/overview configs, 13 tank/sensor configs, sidePanel.showing true/false, Global.mainView.controlsActive true/false, etc.)
- Iterating through all the available sub-pages pages by recursively finding all ListNavigationItems in the current page
- Running the tests on different languages, color schemes (dark/light), layout configurations (five/seven inch) and system units (electric, temperature, volume, brief levels)
- Generating screenshots of each opened page, and reporting any loading errors and warnings

Automated UI testing can be a huge time sink so need to be (effort vs. issues found) so need to be careful with this. Prioritise testing with real backends, but different layout configurations are probably better tested with the mock backend. Bea has built nice mocking capabilities, which make testing a breeze.

chriadam:
Subtasks of this might be:

1) giving unique (e.g. filename-prefixed) object names to every object we need to interact with programmatically, and adding a way to get a pointer to an object with a given name (by walking the child object hierarchy from the root window, for example).

2) creating the mock profile and testing-mode config (i.e. with stable values which aren't being animated, to ensure that the test conditions are stable).  We will actually need multiple such profiles to test different system variants (e.g. 3 phase vs 1 phase AC, EVCS vs no EVCS, critical loads vs none, a variety of different inputs e.g. wind, solar, generator, alternator, grid, shore, etc etc), and a way to launch the app with that profile, e.g. command line arg or whatever.

3) figuring out how/where to store the baseline images (they should probably NOT be part of the main repository, IMO, as there are too many of them).  I suspect creating another git repro like gui-v2-test-baselines or something is the correct solution, but not 100% sure.

4) writing some little application which allows simple comparison of the results would also be useful, e.g. you run the full test and then it will save all of the failed results (along with the baseline) to some temporary folder.  Then you run the "comparison" app to cycle through them one-by-one, and allows you to toggle on/off highlights (in red or whatever) the different pixels between each image, and then choose "update baseline" if the new image is correct (clicking "update baseline" should push that image to the git repo from (3) overwriting the old baseline).  Otherwise, clicking "acknowledge failure" just proceeds to the next failure comparison.

[...]

PR #2316, opened 2025-07-07
Rework mock backend to allow static configurations

blammit:
Provide static mock configurations in data/mock/conf, as JSON files. These can be loaded with the --mock-conf option; if no configuration is specified when running in mock mode, gui-v2 defaults to the 'maximal' configuration, which loads a large amount of services.

The static configurations make it possible to configure the mock runtime without editing the source code. It means gui-v2 can be loaded with a predictable set of services and backend values, which also allows automated UI testing to be implemented in the future.

For example, now data/mock/conf/services/lynxbms.json is used by data/mock/conf/maximal.json to load a Lynx BMS battery into the UI, whereas previously, data/BatteriesImpl.qml included its own set of Lynx BMS battery values. Those values could not be changed (or removed altogether to avoid loading the battery) without modifying the source code; now, you can add batteries to data/mock/conf/services/*.json and edit the JSON values instead of modifying and re-building the source.

This patch changes the mock directory structure:
- mock/conf: contains JSON files with static configurations.
- mock/conf/services: contains JSON files with service-specific values.
- mock/page-conf: contains dynamic page configurations (previously under mock/config) - all page configurations are now implemented within these files, instead of calling out to *Impl files.
- mock/*Impl.qml: these now only configure various services for the mock  backend, and animate mock values. They no longer generate dummy services  on startup, as that is now provided by the static configurations.

Other improvements:
- Add MockManager, a central point for changing mock values. 
- Move timersActive to MockManager so that MockDataSimulator can be removed from Global
- Rename MockDataSimulator to MockShortcuts (as it now provides only key shortcut features) 
- Rename MockDataManager to MockSetup to avoid confusion with DataManager
- Add tools/simulation_csv_to_json.py to convert venus-docker simulation CSV files to JSON
- Provide --no-mock-timers command line option to disable mock timers on startup
- Change the Values debug page to dump the object tree in a format that can be used for service configuration files

Part of #984


components/listitems/core/ListItem.qml
```diff
@@ -150,7 +150,6 @@ BaseListItem {
 			Layout.alignment: Qt.AlignLeft | Qt.AlignVCenter
 			font.pixelSize: flat ? Theme.font_size_body1 : Theme.font_size_body2
 			wrapMode: Text.Wrap
-			width: root.availableWidth - content.width - Theme.geometry_listItem_content_spacing
```
chriadam, line 153: why remove this?

data/mock/TanksImpl.qml
```diff
@@ -5,167 +5,44 @@
...
 			}
-
-			if (hasCreatedObjects) {
-				let lastObject = _createdObjects.shift()
-				Global.mockDataSimulator.setMockValue(lastObject.serviceUid + "/DeviceInstance", -1)
-				lastObject.destroy()
+			VeQuickItem {
+				id: remaining
+				uid: tank.uid + "/Remaining"
 			}
```
chriadam, line 45: could also add a "/Status" path and set it to `VenusOS.Tank_Status_Ok` with occasional excursions to `VenusOS.Tank_Status_Error` or something?

data/mock/conf/services/notifications-alarm.json
```diff
@@ -0,0 +1,26 @@
+{
+    "com.victronenergy.platform":
```
chriadam, line 2: missing a { after :

DanielMcInnes:
Will need to update the wiki after merging this:

https://github.com/victronenergy/gui-v2/wiki/Development---Testing-Tips-And-Tricks

DanielMcInnes (approved):

data/mock/TanksImpl.qml
blammit: Yeah that's nice, added.

chriadam (approved):
Excellent.

PR #2319, opened 2025-07-08
Show system relays as "GX device relays" in Device List

### LINKED ISSUE #2204: Support ../SwitchableOutput/.. from all services
## Tasks
- [ ] Make the switch pane also show switcheable outputs from services other than com.victronenergy.switch
- [ ] Add the configuration options (name, group, etcetera) for the built-in relays into the relays menu under the Settings menu
- [x] Remove the relays from the control cards

Where / how to add Configuration options for relays from yet other services, like solarchargers, will be discussed later. First we want to get more experience.

## Details
The switchable output API specifies outputs that can be controlled from the GX device. Currently, the only service to implement the API is the dedicated `com.victronenergy.switch` service.

However, the API is written such that any service that has switchable outputs will create the paths as well. So the GX internal relays will also come forward as `/SwitchableOutput/x/..` under `com.victronenergy.system` (systemcalc) or `com.victronenergy.platform` (venus-platform).  

The GUI should be modified to scan all services for switchableoutputs conform the API and handle them as such.

Open question: how should the configuration of the outputs take place? For the `com.victronenergy.switch` service, the device represented by the service is present in the device list and the outputs can be configured there. How would we do this with switchable outputs part of e.g. `com.victronenergy.system`?

blammit:
When a GX device relay is set to 'manual' function, it becomes available under com.victronenergy.system/SwitchableOutput. Show all such relays under Settings > Device List > GX device relays, with the same options as for those for switch services listed in the Device List. These relays will also appear in the Switch Pane as a 'toggle' control.

Remove the relays card from the Control Cards as the relays are now accessible from the Switch Pane.

PageSwitch no longer needs the output.state check to see whether an output should be shown, as SwitchableOutputModel now automatically filters out entries that do not have a valid /SwitchableOutput/<x>/Name value.

Fixes #2204


blammit:
WIP - still need to fix the group name for the system relays in the Switch pane.

blammit:
To test this:

1. Open the Switch Pane and observe that there are no cards named "GX device relays"
2. Go to Settings > Integrations > Relays, and set one of the relay functions to "Manual"
3. Go to Settings > Devices > GX device relays (at the bottom of the list), click on the relay and ensure "Show controls" is toggled on
4. Return to the main view, open the Switch Pane and observe there is now a card named "GX device relays" with the manual relay. You can toggle this on/off to trigger the relay.

data/common/SwitchableOutput.qml
```diff
@@ -111,4 +116,9 @@ QtObject {
 	readonly property VeQuickItem _showUIControl: VeQuickItem {
 		uid: `${root.uid}/Settings/ShowUIControl`
 	}
+
+	// /Settings/Function is only valid for system relays.
+	readonly property VeQuickItem _relayFunction: VeQuickItem {
+		uid: root.device.serviceType === "system" ? `${root.uid}/Settings/Function` : ""
```
DanielMcInnes, line 122: Does `${root.uid}/Settings/Function` resolve to the same thing as: `root.uid + "/Settings/Function"` ?

Is there any reason to use one instead of the other? (I have no preference either way)

chriadam: JavaScript string interpolation.  In general, I suspect it will be slower because the QML engine can't compile the binding (it doesn't - at least in 6.6.x - support string interpolation natively).  It's cleaner / nicer to look at.  For things which don't run every frame, I don't have a strong preference.

src/switchableoutputmodel.cpp
```diff
@@ -0,0 +1,238 @@
+/*
```
DanielMcInnes, line 1: I can't see this file mentioned in CMakeLists.txt

chriadam (approved):
Tested fine, code LGTM.

PR #2323, opened 2025-07-09
Popup: Fix height layout in dialogs

### LINKED ISSUE #2294: ESS mode popup layout bug
Venus OS v3.70~5

Image

ybott-qinetic:
Fixes #2294

### DIFF
--- a/components/dialogs/ModalDialog.qml
+++ b/components/dialogs/ModalDialog.qml
@@ -79,7 +79,7 @@ T.Dialog {
 	makes it scale properly when you shrink the window. See https://bugreports.qt.io/browse/QTBUG-127068
 	*/
 	width: Theme.geometry_modalDialog_width
-	height: Theme.geometry_modalDialog_height
+	height: Math.max(Theme.geometry_modalDialog_height, (header ? header.height : 0) + contentHeight + (footer ? footer.height : 0))
 	verticalPadding: 0
 	horizontalPadding: 0
 	modal: true
--- a/components/dialogs/InverterChargerEssModeDialog.qml
+++ b/components/dialogs/InverterChargerEssModeDialog.qml
@@ -28,7 +28,7 @@ ModalDialog {
 
 		SettingsColumn {
 			id: contentColumn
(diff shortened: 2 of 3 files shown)


components/dialogs/ModalDialog.qml
```diff
@@ -79,7 +79,7 @@ T.Dialog {
 	makes it scale properly when you shrink the window. See https://bugreports.qt.io/browse/QTBUG-127068
 	*/
 	width: Theme.geometry_modalDialog_width
-	height: Theme.geometry_modalDialog_height
+	height: header.height + contentHeight + footer.height
```
blammit, line 82: We want to keep Theme.geometry_modalDialog_height as a minimum height; otherwise, this will shrink the height of dialogs where the overall height is smaller than that. 

`Settings > Debug & Develop > UI Library` can be used to check the height of most existing dialogs (e.g. "Spin box", "Date selection" and "Time selection", which internally open `NumberSelectorDialog`, `DateSelectorDialog` and `TimeSelectorDialog` internally). Also try some of the control cards buttons (e.g. "Minimum SOC", "Mode", or "Current limit") that also open dialogs.

components/dialogs/ModalDialog.qml
```diff
@@ -79,7 +79,7 @@ T.Dialog {
 	makes it scale properly when you shrink the window. See https://bugreports.qt.io/browse/QTBUG-127068
 	*/
 	width: Theme.geometry_modalDialog_width
-	height: Theme.geometry_modalDialog_height
+	height: Math.max(Theme.geometry_modalDialog_height, header.height + contentHeight + footer.height)
```
blammit, line 82: When I go to Settings > General, and press "Reboot now", the dialog looks like this:

image

Mostly likely it's because the header is not set here. There might also be cases where a footer is not set, so just need to check for those cases here.

ybott-qinetic: Done.

blammit (approved):

PR #2324, opened 2025-07-09
Integrations: add Shelly Devices page

### LINKED ISSUE #2301: Add new page for shelly integration
In any of the coming beta's, shelly gen 2+ device support will be added. This will add a new umbrella service which lists all discovered shelly devices on dbus, and exposes a path to enable each channel of the discovered device.

The structure of `com.victronenergy.shelly` is the following:

```
com.victronenergy.shelly
Devices/E465B8B6F444/0/Enabled -> Channel 0: disabled, 1: enabled
Devices/E465B8B6F444/1/Enabled -> Channel 1: disabled, 1: enabled
Devices/E465B8B6F444/n/Enabled -> Channel n: disabled, 1: enabled
Devices/E465B8B6F444/Mac       -> Mac address string
Devices/E465B8B6F444/Model     -> Device model. e.g.: "PlusPlugS"
Devices/E465B8B6F444/Name      -> Device custom name given by user. Can be empty (invalid)
Devices/E465B8B6F444/Server    -> Device server usually something like: ShellyPlusPlugS-E465B8B6F444.local
Scan                           -> Manually start mDNS scan. Will read 0 at all times. write 1 to start scan
```

A new page must be created in settings->integrations, right below 'Modbus devices'. Opening this page should show a manual 'Scan' button and a submenu with the discovered devices listed. The discovered devices can be listed similar as the modbus devices are:

Image

However, for the shelly devices, each channel can be enabled/disabled separately, so they should be listed separately as well. Each entry should consist of the following:

"<Name or Model [MAC]> - channel <n>"

So display the device name if the `/Name` field is populated. Else, display the Model with the MAC in brackets. Enabling the switch should write a 1 to the corresponding `/Enabled` path.

[This system](https://vrm.victronenergy.com/installation/722304/dashboard) can be used for testing. The new shelly service is running and at least 1 device is listed in the service.

blammit:
Add Settings > Integrations > Shelly Devices, to scan for and show available shelly devices and their channels.

Fixes #2301

### DIFF
--- a/pages/settings/PageSettingsShelly.qml
+++ b/pages/settings/PageSettingsShelly.qml
@@ -1,0 +1,104 @@
+/*
+** Copyright (C) 2025 Victron Energy B.V.
+** See LICENSE.txt for license information.
+*/
+
+import QtQuick
+import Victron.VenusOS
+
+Page {
+	id: root
+
+	readonly property string serviceUid: BackendConnection.serviceUidForType("shelly")
+
+	// Get a list of channels from all devices on the shelly service. For example, here we have
+	// a device "E465B8B6F444" with "Mac" and "Name" values, and three channels (0-2) with
+	// "Enabled" states that can be toggled by the delegates in this list view.
+	//  com.victronenergy.shelly/Devices/E465B8B6F444/Mac
+	//  com.victronenergy.shelly/Devices/E465B8B6F444/Name
+	//  com.victronenergy.shelly/Devices/E465B8B6F444/0/Enabled
+	//  com.victronenergy.shelly/Devices/E465B8B6F444/1/Enabled
+	//  com.victronenergy.shelly/Devices/E465B8B6F444/2/Enabled
+	VeQItemSortTableModel {
+		id: channelModel
+		dynamicSortFilter: true
+		filterFlags: VeQItemSortTableModel.FilterInvalid
+		filterRole: VeQItemTableModel.UniqueIdRole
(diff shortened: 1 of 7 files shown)


blammit:
Can be tested on https://vrm.victronenergy.com/installation/512796/dashboard.

pages/settings/PageSettingsShelly.qml
```diff
@@ -0,0 +1,86 @@
...
+			width: parent?.width ?? 0
+
+			ListButton {
+				text: CommonWords.scan_for_devices
+				secondaryText: scanItem.value === 1
+						  //% "Scanning"
+						? qsTrId("page_settings_shelly_scanning")
+						: CommonWords.press_to_scan
+				interactive: !scanItem.value
+				onClicked: scanItem.setValue(1)
```
chriadam, line 27: as discussed offline, it seems like this path isn't writable currently?  or, it requires some specific user level to access (if so, button should be marked non-interactive in that case I guess).  @ReinvdZee might know more.

chriadam: Discussed in meeting, the feedback about the scan click can be done separately/later.

ReinvdZee:
@blammit 
Can you please hide a device if the corresponding paths are invalid?
When a device disconnects, the paths in the `com.victronenergy.shelly` service are invalidated, not removed. Currently it shows "undefined" when that happens.

blammit:
@ReinvdZee I've updated the PR with those changes (Hide disconnected devices, Scan -> Refresh, and remove channel index when there is only one channel).

Issue #2327, opened 2025-07-11
VE.Can devices unique id formatting issue


martinbosma:
Large numbers have scientific notation, while they should be integer.
Applies to VE.Can devices menu only: https://github.com/victronenergy/gui-v2/blob/main/pages/settings/PageSettingsVecanDevices.qml#L27

<img width="1551" height="905" alt="Image" src="https://github.com/user-attachments/assets/c74af0c8-afdf-402b-b0ac-1f84b8126691" />

chriadam:
PR: https://github.com/victronenergy/gui-v2/pull/2328/files

chriadam:
Merged to main, closing as fixed

PR #2328, opened 2025-07-11
Fix bug in VE.Can device unique id formatting

chriadam:
QML string type's arg() function will format large numbers using scientific notation by default.

Avoid this by constructing a JS string with concatenation.

Contributes to #2327


blammit (approved):
Could you add a brief comment to the code? Otherwise LGTM

chriadam:
Done

blammit:
LGTM

PR #2332, opened 2025-07-14
Define dcload services based on Monitor Type

### LINKED ISSUE #2281: DC load / fridge labeled as DC generator
When a Smartshunt/BMV is configured as DC load of type fridge, it is labeled in the device page (via the overview load box or the device list) as DC generator

Image

A system available for testing is: https://vrm.victronenergy.com/installation/50884/dashboard

Reported via the community: https://community.victronenergy.com/t/smart-shunt-type-fridge-appears-as-dc-generator-in-remote-console-v3-60/36782

ybott-qinetic:
Change some DC Generator references to generic terminology
Adds translations
Fixes #2281


data/DcInputs.qml
```diff
@@ -53,48 +53,72 @@ QtObject {
...
 		}
 
 		try {
 			monitorMode = parseInt(monitorMode)
 		} catch (e) {
-			console.warn("Defaulting to DC generator type, invalid monitor mode specified!", monitorMode)
-			return VenusOS.DcInputs_InputType_DcGenerator
+
+			console.warn("Defaulting to generic load/source type, invalid monitor mode specified!", monitorMode)
+			return (serviceType === "dcload") ? VenusOS.DcInputs_InputType_GenericLoad : VenusOS.DcInputs_InputType_GenericSource
```
blammit, line 77: We could use "DC meter" for the "unknown" case, instead of returning either a load/source type. Can you add a `DcInputs_InputType_GenericMeter` that can be returned here, and then in `dcInput_typeToText()`, return that type as "DC Meter"?

src/enums.cpp
```diff
@@ -87,9 +87,30 @@ QString Enums::dcInput_typeToText(DcInputs_InputType type) const
 	case DcInputs_InputType_WaterGenerator:
 		//% "Water generator"
 		return qtTrId("dcInputs_water_generator");
-	case DcInputs_InputType_Wind:
-		//% "Wind"
-		return qtTrId("dcInputs_wind");
+	case DcInputs_InputType_WindCharger:
+		//% "Wind Charger"
```
blammit, line 91: There's inconsistent letter casing in the source table - it uses "Wind Charger" instead of "Wind charger", even though there is also "Solar charger" rather than "Solar Charger". Let's use lowercase for the second word here and below, to be more consistent with names elsewhere in gui-v2. 

Also change "DC charger" to "DC/DC charger".

blammit: I'm getting a compile warning:

```
victron/src/gui-v2/src/enums.cpp:65:13: warning: enumeration value 'DcInputs_InputType_SolarCharger' not handled in switch [-Wswitch]
   65 |     switch (type) {
      |             ^~~~

```

src/enums.h
```diff
@@ -196,15 +196,23 @@ class Enums : public QObject
...
-		DcInputs_InputType_DcGenerator,
+		DcInputs_InputType_GenericLoad,
+		DcInputs_InputType_GenericSource,
 		DcInputs_InputType_DcSystem,
 		DcInputs_InputType_FuelCell,
 		DcInputs_InputType_ShaftGenerator,
 		DcInputs_InputType_WaterGenerator,
-		DcInputs_InputType_Wind
+		DcInputs_InputType_WindCharger,
+		DcInputs_InputType_SolarCharger
```
blammit, line 215: Might be easier to maintain if the alphabetical ordering is maintained (and the cases in dcInput_typeToText() as well), given that the actual enum values are not important here, and that the MonitorMode number is only sometimes relevant to the type.

blammit (commented):
In the commit message, add newline above the "Fixes #" line.

src/enums.h
ybott-qinetic: Done.

data/DcInputs.qml
ybott-qinetic: Done

src/enums.cpp
ybott-qinetic: Done.

data/DcInputs.qml
```diff
@@ -54,48 +54,72 @@ QtObject {
...
+		case "dcsystem":
+			return VenusOS.DcInputs_InputType_DcSystem
+		case "dcload":
+			// use the monitor mode to determine a sub-type
+			break
 		case "dcsource":
 			// use the monitor mode to determine a sub-type
 			break
-		case "dcsystem":
-			return VenusOS.DcInputs_InputType_DcSystem
```
blammit, line 61: The default case on the next line should also return `VenusOS.DcInputs_InputType_GenericMeter`, instead of just breaking. Otherwise, it will drop to the new if-else condition below, which assumes the service type is either dcsource or dcload.

ybott-qinetic: Done

blammit (approved):

PR #2336, opened 2025-07-14
MotorDrive: add Sevcon setup page

citolen:
Adds a new page to control the settings of a Sevcon motor controller.

The page can be navigated to from the motordrive device's page when the `ProductId` matches the one from a Sevcon controller.

Driver: https://github.com/citolen/dbus-sevcon

Screenshot 2025-07-13 at 15 41 36  
Screenshot 2025-07-13 at 15 41 48

### DIFF
--- a/pages/settings/devicelist/PageMotorDrive.qml
+++ b/pages/settings/devicelist/PageMotorDrive.qml
@@ -18,6 +18,11 @@ Page {
 		serviceUid: root.bindPrefix
 	}
 
+	VeQuickItem {
+		id: productId
+		uid: root.bindPrefix + "/ProductId"
+	}
+
 	GradientListView {
 		model: VisibleItemModel {
 			ListQuantity {
@@ -69,6 +74,15 @@ Page {
 				text: qsTrId("devicelist_motordrive_controllertemperature")
 				dataItem.uid: root.bindPrefix + "/Controller/Temperature"
 				preferredVisible: dataItem.valid
+			}
+
+			ListNavigation {
+				text: CommonWords.setup
+				onClicked: {
+					Global.pageManager.pushPage("/pages/settings/devicelist/motordrive/PageSevconSetup.qml",
+							{ "title": text, "bindPrefix": root.bindPrefix })
+				}
+				preferredVisible: productId.value == ProductInfo.ProductId_MotorDrive_Sevcon
 			}
 
 			ListNavigation {
--- a/pages/settings/devicelist/motordrive/PageSevconSetup.qml
+++ b/pages/settings/devicelist/motordrive/PageSevconSetup.qml
@@ -1,0 +1,32 @@
+/*
+** Copyright (C) 2025 Victron Energy B.V.
+** See LICENSE.txt for license information.
+*/
+
+import QtQuick
+import Victron.VenusOS
+
+Page {
+	id: root
+
+	property string bindPrefix
+
+	title: device.name
+
+	Device {
+		id: device
+		serviceUid: root.bindPrefix
+	}
+
+	GradientListView {
+		model: VisibleItemModel {
+			ListSwitch {
+				//% "Motor Direction Inverted"
+				text: qsTrId("devicelist_motordrive_motordirectioninverted")
+				dataItem.uid: root.bindPrefix + "/Motor/DirectionInverted"
+				dataItem.invalidate: false
+				preferredVisible: dataItem.valid
+			}
+		}
+	}
+}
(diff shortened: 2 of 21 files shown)


pages/settings/devicelist/PageMotorDrive.qml
```diff
@@ -71,6 +76,15 @@ Page {
 				preferredVisible: dataItem.valid
 			}
 
+			ListNavigation {
+				text: CommonWords.setup
+				onClicked: {
+					Global.pageManager.pushPage("/pages/settings/devicelist/motordrive/PageSevconSetup.qml",
```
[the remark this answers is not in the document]
citolen: Yes totally, the switch could be moved to `PageMotorDrive.qml`.  
It's currently on a separate page because I didn't know what the correct pattern was and thought device specific settings didn't belong to the generic motordrive page?

The driver makes use of a single setting for now, but I imagine this may evolve to more.

>Is inverted motor direction something which might be supported by devices with different product id values?

I guess it could at some point, but currently `Motor/DirectionInverted` is a property specific to the product id mentioned.

pages/settings/devicelist/motordrive/PageSevconSetup.qml
```diff
@@ -0,0 +1,32 @@
...
+		id: device
+		serviceUid: root.bindPrefix
+	}
+
+	GradientListView {
+		model: VisibleItemModel {
+			ListSwitch {
+				//% "Motor Direction Inverted"
+				text: qsTrId("devicelist_motordrive_motordirectioninverted")
+				dataItem.uid: root.bindPrefix + "/Motor/DirectionInverted"
```
philipptrenz, line 26: @citolen This is what I wanted to check: As discussed, we would like to rename this path to `/Settings/Motor/DirectionInverted`. Can you take care of that here? Thanks!

citolen: >So you tried to access the localsettings service on gui-v2 directly?

No

I believe I've been creating the setting correctly, the same way `can-bus-bms` is doing it.

The setting proxy is created with:
```c
snprintf(settingsPath, sizeof(settingsPath), "Settings/Devices/%s", device->identifier);
veItemCreateSettingsProxy(
        localSettings, settingsPath, device->root, "Motor/DirectionInverted",
        veVariantFmt, &veUnitNone, &booleanType
);
```

I believe this results in two path.
`com.victronenergy.settings/Settings/Devices/<identifier>/Motor/DirectionInverted`
`com.victronenergy.motordriver.xxx/Motor/DirectionInverted`

What I was saying is that when calling `veItemCreateSettingsProxy`, if the word `Settings` is used in the id (4th paramater) so `Settings/Motor/DirectionInverted` instead of `Motor/DirectionInverted`, it doesn't work.
I'm not just speaking about gui-v2, if I look in the debug values UI, the setting is not working in either paths.

It is very specific to the word `Settings`, if I use `Setting/Motor/DirectionInverted` (no 's'), it works.  
I didn't look into it further, I assumed that was a limitation of the localsettings.

philipptrenz: Okay, I asked @martinbosma to take a look and come back to you about it.

martinbosma: @citolen Try using `veItemCreateSettingsProxyId` instead, I'll submit a PR on your repo, as it works on my test build.

https://github.com/user-attachments/assets/049ca9eb-6892-4d66-9f85-7e4d569ac122

pages/settings/devicelist/motordrive/PageSevconSetup.qml
```diff
@@ -0,0 +1,32 @@
...
+import Victron.VenusOS
+
+Page {
+	id: root
+
+	property string bindPrefix
+
+	title: device.name
+
+	Device {
```
blammit, line 16: @citolen Just a heads-up - I've merged [3c8c0e3](https://github.com/victronenergy/gui-v2/pull/2365/commits/3c8c0e3c451669b9b5c5a8ed084e9602f3c79d0d), which changes a few things for this PR:

- This page should now extend `DevicePage` instead of `Page`, with a binding of `serviceUid: bindPrefix`
- This `Device` can be removed, as `title` will be auto-set to the `device.name`
- The `GradientListView` below can be removed, and you can set `settingsModel: VisibleItemModel { ...` instead to directly set the model for the GradientListView within DevicePage.
-

citolen:
> Was this one obsoleted by https://github.com/victronenergy/gui-v2/pull/2468/files ?

Yes

PR #2339, opened 2025-07-15
Temporarily disable PressEffect

chriadam:
Requires a shader which must be baked using QtShaderTools which isn't available from aqt for all required platforms. We use aqt in the CI system and build scripts, so for now, we can't use this press effect.


chriadam:
Although this seems to cause problems for windows desktop - I'm not sure why, though.  I have already reached my quota of CMake wrangling for this week, so I'm going to ignore it for now.

chriadam:
Not needed, closing.

PR #2340, opened 2025-07-15
Overview: show correct drilldown for generic DC source inputs

### LINKED ISSUE #2338: Overview: clicking on DC input box shows the wrong page
When a DC input is connected and the type is a generic DC source:
- Click on the input box in the Overview. Observe it opens a page with "AC totals" - that is, it is a page for AC inputs, rather than DC inputs.
- From the Device List, clicking the same device will open the correct page, which shows DC totals.

This bug might be occurring for other DC input types as well.

Seen on https://vrm.victronenergy.com/installation/50884/dashboard.

blammit:
When the DC input is a generic source, show the DC meter page, rather than the Genset page which is for AC gensets.

Fixes #2338


blammit:
Merge after https://github.com/victronenergy/gui-v2/pull/2332 is merged

ybott-qinetic (approved):

PR #2350, opened 2025-07-17
PressEffect is only valid in WebAssembly builds

chriadam:
We don't support it on desktop or device.


chriadam:
Potentially we could support this also for desktop, but I don't see any reason to do so.

chriadam:
This is a temporary measure.  Once Jeroen has created the new device firmware image and SDK, we can look at adding support also for the device (i.e. if the SDK includes QtShaderTools).

blammit (approved):
Maybe make a task to re-enable this, and add code comments that refer to the task number?

chriadam:
Done.

PR #2364, opened 2025-07-21
Add translation context with lupdate include paths

ybott-qinetic:


DanielMcInnes (approved):

chriadam (approved):
I am not sure why it's necessary, but it does indeed resolve the warnings - thanks!

PR #2367, opened 2025-07-21
Add Modbus TCP server note about unit ids

### LINKED ISSUE #2362: Add note on top of Modbus-TCP servives list submenu
Add this note “See the Settings -> VRM -> VRM Device instances menu to change the Modbus-TCP unit IDs.” at the top of this page, with perhaps an (i) icon in front of it and proper visual separation to the list/table.

(I took the screenshot from this discussion: https://community.victronenergy.com/t/how-can-i-change-modbus-tcp-server-unit-ids/40077)

Image

ybott-qinetic:
Fixes #2362

### DIFF
--- a/pages/settings/PageSettingsModbusTcpServices.qml
+++ b/pages/settings/PageSettingsModbusTcpServices.qml
@@ -36,6 +36,11 @@ Page {
 	GradientListView {
 		model: serviceCount.value || 0
 
+		header: PrimaryListLabel {
+			//% "See the Settings → VRM → VRM device instances menu to change the Modbus-TCP unit IDs."
+			text: qsTrId("settings_modbus_tcp_unit_id_note")
+		}
+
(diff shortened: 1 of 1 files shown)


pages/settings/PageSettingsModbusTcpServices.qml
```diff
@@ -36,6 +36,11 @@ Page {
 	GradientListView {
 		model: serviceCount.value || 0
 
+		header: PrimaryListLabel {
+			//% "See the Settings -> VRM -> VRM Device instances menu to change the Modbus-TCP unit IDs."
```
blammit, line 40: Use consistent casing, i.e. either "VRM device instances" or "VRM Device Instances". 

Also you can use the → character instead of "->", like we do in `PageGensetModel.qml`.

blammit (approved):

Issue #2303, opened 2025-07-02
Switch pane design improvements and additions


ReinvdZee:
There are some updated and new designs to be added to the switching pane:

**Updated basic elements:**

![Image](https://github.com/user-attachments/assets/ee7fffc5-dd18-492a-b60a-fd91e6776126)
(Note that "Latching" is recently renamed to "Toggle")

**Update color for the 'Off' state**
Small UI improvement: The 'active' color of the 'Off' state of a toggle button should be different from that of the 'On' state:
![Image](https://github.com/user-attachments/assets/8989d69f-db90-4b02-861b-f392d5922229)

**Disabled state**
An output can be disabled. This is indicated by the `/SwitchableOutput/x/Status` being equal to `0x20`. If this is the case, the UI shows a 'Disabled' label to reflect this. To indicate that the output is not clickable, it should be greyed out as well:

![Image](https://github.com/user-attachments/assets/c991c7a7-aa35-4b1a-af62-3f71d1eaccb2)

**Control for sliders without min/max:**
Some dimmable outputs may not have a minimum or maximum value. In such cases, a slider cannot be used since the limits are unknown. 

![Image](https://github.com/user-attachments/assets/50fcd3db-7a37-4631-9384-11adc01bb830)

**New designs:**
![Image](https://github.com/user-attachments/assets/27fc4f55-f069-4dbc-83ad-3bbdfdfcfe7a)

The designs appear in the order of priority.

New: **Bilge pump control**

<img width="579" height="623" alt="Image" src="https://github.com/user-attachments/assets/01c57d22-3ed1-4732-bf27-294f83fec02c" />

Details are on the dbus wiki and in the slide deck

Tasks:

- [x] Implement updates for the dimming switch. https://github.com/victronenergy/gui-v2/pull/2368
- [x] Implement updates for the momentary switch. https://github.com/victronenergy/gui-v2/pull/2410
- [ ] (Rein) Update dbus API wiki.
- [ ] (Dirk-Jan) Update NR virtual switch to support the new types.
- [x] Update 'Off' state color https://github.com/victronenergy/gui-v2/pull/2346
- [x] Update disabled state for momentary, toggle and dimming controls https://github.com/victronenergy/gui-v2/pull/2346 (latching button needs further changes)
- [x] Implement control:  temperature setpoint https://github.com/victronenergy/gui-v2/pull/2397
- [x] Implement control: stepped switch https://github.com/victronenergy/gui-v2/pull/2433
[...]

PR #2386, opened 2025-07-28
Switch pane: add placeholder support for new control types

blammit:
Add SwitchableOutput_Type_* enum values for the new control types that will be supported. Update to SwitchableOutputDelegate.qml to just show placeholders for all controls that are yet to be implemented.

Add switch-pane.json mock configuration, which loads switch-controls-tester.json with all of the new control types that need to be supported.

Part of #2303


ybott-qinetic (approved):
Looks good.

PR #2389, opened 2025-07-29
Switch pane: key navigation fixes

blammit:


blammit:
Actually, I need to make some other momentary button updates, so I’ll put these into the same PR. Closing this one.

PR #2397, opened 2025-07-30
Switch pane: implement temperature slider control

ybott-qinetic:
The TemperatureSlider control is a Slider that sets left and right padding
with label contents Min/Max for each language.

/SwitchableOutput/<x>/Settings/DimmingMax|DimmingMin sets the sliders
range and StepSize sets the selectable increments.
/SwitchableOutput/<x>/Dimming sets the value of the selected option.

It supports layout mirroring.

Contributes to #2303

### DIFF
--- a/components/SwitchableOutputDelegate.qml
+++ b/components/SwitchableOutputDelegate.qml
@@ -3,6 +3,7 @@
 ** See LICENSE.txt for license information.
 */
 import QtQuick
+import QtQuick.Layouts
 import Victron.VenusOS
 import QtQuick.Templates as CT
 
@@ -22,7 +23,7 @@ BaseListItem {
 		anchors.verticalCenter: parent.verticalCenter
 		width: parent.width
 
-		Row {
+		RowLayout {
 			anchors {
 				left: switchWidgetLoader.left
 				right: switchWidgetLoader.right
@@ -30,23 +31,26 @@ BaseListItem {
 
 			Label {
 				id: nameLabel
-				anchors.bottom: parent.bottom
+				Layout.fillWidth: true
 				bottomPadding: Theme.geometry_switchableoutput_label_margin
 				rightPadding: Theme.geometry_switchableoutput_label_margin
 				text: output.formattedName
-				width: parent.width - (statusRect.visible ? statusRect.width : 0)
 				elide: Text.ElideMiddle // don't elide right, as it may obscure a trailing channel id
+			}
+
+			Label {
+				id: secondaryTitleLabel
+				bottomPadding: Theme.geometry_switchableoutput_label_margin
+				text: switchWidgetLoader.item.secondaryTitle ?? ""
 			}
 
 			Rectangle {
 				id: statusRect
 
-				anchors {
-					bottom: parent.bottom
-					bottomMargin: Theme.geometry_switchableoutput_label_margin
-				}
-				width: Math.max(Theme.geometry_switchableoutput_status_minimum_width,
-								Math.min(parent.width / 2, statusLabel.implicitWidth))
+				Layout.bottomMargin: Theme.geometry_switchableoutput_label_margin
+				Layout.maximumWidth: parent.width / 2
+				Layout.minimumWidth: statusLabel.implicitWidth
+				Layout.alignment: Qt.AlignRight
 				height: statusLabel.height
 				color: statusLabel.color === Theme.color_green ? Theme.color_darkGreen
 						: statusLabel.color === Theme.color_orange ? Theme.color_darkOrange
@@ -62,6 +66,7 @@ BaseListItem {
 				Label {
 					id: statusLabel
 
+					anchors.centerIn: parent
 					text: VenusOS.switchableOutput_statusToText(output.status)
 					width: parent.width
 					topPadding: Theme.geometry_switchableoutput_status_verticalPadding
@@ -90,6 +95,22 @@ BaseListItem {
 							return Theme.color_red
 						}
 					}
+				}
+
+				TextMetrics {
+					id: nameTextMetrics
+					font: nameLabel.font
+					text: nameLabel.text
+				}
+				TextMetrics {
+					id: secondaryTitleTextMetrics
+					font: secondaryTitleLabel.font
+					text: secondaryTitleLabel.text
+				}
+				TextMetrics {
+					id: statusTextMetrics
+					font: statusLabel.font
+					text: statusLabel.text
 				}
 			}
 		}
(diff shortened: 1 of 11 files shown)


components/SwitchableOutputDelegate.qml
```diff
@@ -374,7 +374,137 @@ BaseListItem {
...
+			TextMetrics {
+				id: fontMetricsMax
+				font.pixelSize: Theme.font_size_body1
+				text: CommonWords.max
+			}
+
+			TextMetrics {
+				id: fontMetricsMin
+				font.pixelSize: Theme.font_size_body1
+				text: CommonWords.min
```
blammit, line 499: Looks like these TextMetrics belong in TemperatureSlider instead?

components/controls/TemperatureSlider.qml
```diff
@@ -0,0 +1,156 @@
...
+
+		contentItem: Rectangle {
+			anchors.fill: parent
+			color: Theme.color_blue
+			radius: 10
+
+			Label {
+				anchors.centerIn: parent
+				font.pixelSize: Theme.font_size_h1
+				text: root.value.toFixed(0) + "°"
```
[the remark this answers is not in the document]
blammit: The temperature will need to be converted to Fahrenheit if that is preferred by the user settings.

You can do that using `Units.convert(value, VenusOS.Units_Temperature_Celsius, Global.systemSettings.temperatureUnit)`

Also, use the unicode string "\u00b0" for the circle, in case there are file encoding issues.

components/controls/TemperatureSlider.qml
```diff
@@ -0,0 +1,156 @@
...
+			GradientStop { position: 1.0; color: root.enabled ? Theme.color_temperatureslider_gradient_max_border : Theme.color_gray3 }
+		}
+
+		anchors {
+			left: parent.left
+			right: parent.right
+			verticalCenter: parent.verticalCenter
+		}
+
+		implicitWidth: 4*Theme.geometry_switch_indicator_width // suitably small.
```
blammit, line 36: What is the 4 for?

components/SwitchableOutputDelegate.qml
```diff
@@ -374,7 +374,137 @@ BaseListItem {
...
+				if (!dragging) {
+					dimmingValue.syncBackendValueToSlider()
+				}
+			}
+			onMoved: {
+				value = Math.round(value)
+				dimmingValue.writeValue(value)
+			}
+
+			Keys.onPressed: (event) => {
```
blammit, line 436: There's a lot of code duplication between here and the DimmingSlider usage in this file. It doesn't have to be refactored out right now but we will need to at some stage before all the controls are finalized.

components/controls/TemperatureSlider.qml
```diff
@@ -0,0 +1,156 @@
...
+			anchors.right: parent.right
+			anchors.verticalCenter: parent.verticalCenter
+			width: root.rightPadding
+			text: CommonWords.max
+			horizontalAlignment: Text.AlignHCenter
+			font.pixelSize: Theme.font_size_body1
+			color: root.enabled ? Theme.color_white : Theme.color_font_disabled
+		}
+	}
+
```
blammit, line 74: I noticed this doesn't have a  `XAnimator` for the value change, unlike the DimmingSlider. Is that deliberate?

components/controls/TemperatureSlider.qml
```diff
@@ -0,0 +1,156 @@
...
+import Victron.VenusOS
+import QtQuick.Effects as Effects
+
+T.Slider {
+	id: root
+
+	implicitHeight: Theme.geometry_dimmingSlider_height
+
+	property int currentSliderDots: ((to - from) / stepSize) * Theme.geometry_temperatureSlider_dot_size * 2 > availableWidth
+									? Math.floor(availableWidth / (4 * 2)) + 1 : ((to - from) / stepSize) + 1
```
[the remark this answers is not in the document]
ybott-qinetic: That is twice a dot size. I use this metric to determine how many dots will fit in the availableWidth and then limit the dot number to upper bound.
The other case is that the range specified (from - to) and stepSize allow for the number of dots to be displayed correctly.
I may be out by one on my upper bound? But hopefully it is ok. This should also work if the control is moved to different width constraints.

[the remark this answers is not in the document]
ybott-qinetic: I will add a comment.

components/controls/TemperatureSlider.qml
ybott-qinetic: I presumed that is a Slider based "minimum allowable size" for the control.

components/controls/TemperatureSlider.qml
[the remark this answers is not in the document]
blammit: Looks like this can be simplified to a `Math.min()` expression? And name it `dotCount` and make it `readonly`.

The expression would be clearer if the dot padding was in a separate property, instead of adding `*2`. (Also not sure what `+1` is for)

pages/settings/devicelist/switchable-outputs/SwitchableOutputListDelegate.qml
```diff
@@ -17,6 +17,7 @@ ListQuantityGroupNavigation {
 		filterType: QuantityObjectModel.HasValue
 
 		QuantityObject { object: outputCurrent; unit: VenusOS.Units_Amp }
+		QuantityObject { object: output.displayTemperature ? output : null; key: "dimming"; unit: VenusOS.Units_Temperature_Celsius }
```
blammit, line 20: This can't be always Celsius, as the user might prefer to see it in Fahrenheit; use `Global.systemSettings.temperatureUnit` instead. This means the displayed value also needs to be converted from Celsius to Fahrenheit if needed. So, something like this:

```
QuantityObject { 
    id: temperatureQuantity
    
    readonly property real temperature: output.displayTemperature ? Units.convert(output.dimming, VenusOS.Units_Temperature_Celsius, Global.systemSettings.temperatureUnit) : NaN
    
    object: output.displayTemperature ? temperatureQuantity : null
    key: "temperature"
    unit: Global.systemSettings.temperatureUnit
}
```

components/controls/TemperatureSlider.qml
```diff
@@ -0,0 +1,169 @@
...
+			horizontalAlignment: Text.AlignHCenter
+			font.pixelSize: Theme.font_size_body1
+			color: root.enabled ? Theme.color_white : Theme.color_font_disabled
+		}
+	}
+
+	handle: Rectangle {
+		id: handleItem
+		x: root.visible ? root.leftPadding + root.visualPosition * (root.availableWidth - width) : 0
+		y: root.height/2 - height/2
```
blammit, line 83: Please add parenthesis, i.e. `(root.height / 2) - (height / 2)`, and elsewhere.

components/controls/TemperatureSlider.qml
```diff
@@ -0,0 +1,169 @@
...
+		Repeater {
+			model: currentSliderDots
+			Dot {
+				x: modelData * (root.availableWidth - width)/(currentSliderDots - 1)
+				y: parent.height/2 - height/2
+			}
+		}
+	}
+
+	component Dot: Rectangle {
```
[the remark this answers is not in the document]
chriadam: Yes, very good point.  That will let it be an instanced texture which is cached in the pixmap cache.

components/controls/TemperatureSlider.qml
```diff
@@ -0,0 +1,174 @@
...
+		SliderHandleHighlight {
+			id: handleHighlight
+			x: (parent.width / 2) - (width / 2)
+			y: (parent.height / 2) - (height / 2)
+			width: root.handle.height
+			height: Theme.geometry_switch_groove_border_width * 2
+			visible: Global.keyNavigationEnabled && root.activeFocus
+		}
+
+		XAnimator {
```
chriadam, line 98: does something need to trigger this animator?  e.g. on value change?

components/controls/TemperatureSlider.qml
```diff
@@ -0,0 +1,174 @@
...
+		TextMetrics {
+			id: popupTextMetrics
+			font.pixelSize: Theme.font_size_h2
+			text: popup._popupText
+		}
+	}
+
+	TextMetrics {
+		id: textMetricsMax
+		font.pixelSize: Theme.font_size_body1
```
chriadam, line 166: also need to set the font.family and font.weight I guess?

components/SwitchableOutputDelegate.qml
```diff
@@ -388,7 +388,122 @@ BaseListItem {
...
+					// before syncing the slider to the backend value, else this may be done while
+					// user changes are still being written.
+					if (busy || slider.dragging || delayedSliderUpdate.running) {
+						delayedSliderUpdate.restart()
+					} else {
+						slider.value = backendValue
+					}
+				}
+
+				backendValue: output.dimming
```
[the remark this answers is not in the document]
ybott-qinetic: This is fixed.

components/SwitchableOutputDelegate.qml
```diff
@@ -391,7 +414,118 @@ BaseListItem {
 	Component {
 		id: temperatureSetpointComponent
 
-		PlaceholderDelegate {}
+		TemperatureSlider {
+			id: temperatureSlider
+
+			property bool valueChangeKeyPressed
+			readonly property bool dragging: pressed || valueChangeKeyPressed
+			readonly property string info: Math.round(value) + Global.systemSettings.temperatureUnitSuffix
```
blammit, line 422: This temperature needs the unit conversion as well.

The `info` name suggests some data object to me, perhaps `secondaryTitle` instead? (Since the main title is the name)

components/SwitchableOutputDelegate.qml
```diff
@@ -22,31 +23,37 @@ BaseListItem {
...
+				width: nameTextMetrics.tightBoundingRect.x + nameTextMetrics.tightBoundingRect.width
+				elide: Text.ElideMiddle // don't elide right, as it may obscure a trailing channel id
+			}
+
+			Label {
+				id: infoLabel
+				Layout.minimumWidth: width
+				bottomPadding: Theme.geometry_switchableoutput_label_margin
+				text: switchWidgetLoader.item.info ?? ""
+				width: infoTextMetrics.tightBoundingRect.x + infoTextMetrics.tightBoundingRect.width
```
blammit, line 47: The "Disabled" status text is now misaligned vertically:

image

blammit (commented):
Mostly looks good, @chriadam any comments?

components/controls/TemperatureSlider.qml
```diff
@@ -0,0 +1,184 @@
...
+** See LICENSE.txt for license information.
+*/
+
+import QtQuick
+import QtQuick.Templates as T
+import QtQuick.Controls
+import QtQuick.Controls.impl as CP
+import QtQuick.Shapes
+import Victron.VenusOS
+import QtQuick.Effects as Effects
```
blammit, line 12: Some of these imports aren't used anymore and can be removed

components/SwitchableOutputDelegate.qml
```diff
@@ -91,6 +96,22 @@ BaseListItem {
...
+				TextMetrics {
+					id: secondaryTitleTextMetrics
+					font: secondaryTitleLabel.font
+					text: secondaryTitleLabel.text
+				}
+				TextMetrics {
+					id: statusTextMetrics
+					font: statusLabel.font
+					text: statusLabel.text
+				}
```
blammit, line 114: These TextMetrics can be removed

components/SwitchableOutputDelegate.qml
```diff
@@ -391,7 +412,118 @@ BaseListItem {
...
+				case Qt.Key_Space:
+					focus = true
+					return true
+				case Qt.Key_Return:
+					focus = false
+					return true
+				case Qt.Key_Up:
+				case Qt.Key_Down:
+					// Remove focus and reject event to allow key navigation to delegates above/below.
+					focus = false
```
blammit, line 436: Don't remove the focus here, as it means we won't prevent key navigation as per the next line below.

Also can you extend the `Qt.Key_Return` case to also include `Key_Enter` and `Key_Escape`?

components/controls/TemperatureSlider.qml
ybott-qinetic: Done

components/SwitchableOutputDelegate.qml
ybott-qinetic: Done

components/SwitchableOutputDelegate.qml
ybott-qinetic: I have fixed these issues.

blammit (approved):
LGTM. @chriadam could you double check this?

chriadam (approved):
LGTM

PR #2399, opened 2025-07-31
Modify power flow indicator display

### LINKED ISSUE #2395: Electron flow direction is sometimes incorrect on Overview page
The geometry_overviewPage_connector_animationPowerThreshold theme constant is 30.
It is used in binding expressions of the form:

```
                        // For AC inputs, positive power means energy is flowing towards inverter/charger,
                        // and negative power means energy is flowing towards the input.
                        return power > Theme.geometry_overviewPage_connector_animationPowerThreshold
                                        ? VenusOS.WidgetConnector_AnimationMode_StartToEnd
                                        : VenusOS.WidgetConnector_AnimationMode_EndToStart
```

this ensures that the direction of the flow doesn't "flicker" when the power is close to zero.

However, the logic is incorrect, and it is particularly visible when animations are disabled (as the arrow is always shown, and in the error case, it points in the wrong direction).

These bindings should be improved to be something like:

```
return Math.abs(power) < Theme.geometry_overviewPage_connector_animationPowerThreshold ? NoDirection
    : power > 0.0 ? StartToEnd
    : EndToStart
```

ybott-qinetic:
Modify how the animated energy flow is displayed.
When not animated, only show an arrow when power is above 30 in either direction.

Fixes #2395

### DIFF
--- a/components/widgets/WidgetConnector.qml
+++ b/components/widgets/WidgetConnector.qml
@@ -214,6 +214,7 @@ Item {
 				delegate: Image {
 					opacity: 0.0
 					source: animationEnabled ? "qrc:/images/electron.svg" : "qrc:/images/electron_arrow.svg"
+					visible: root.animationMode !== VenusOS.WidgetConnector_AnimationMode_NotAnimated
 
 					Behavior on opacity {
 						enabled: root._animated
--- a/components/widgets/WidgetConnector.qml
+++ b/components/widgets/WidgetConnector.qml
@@ -215,6 +215,7 @@ Item {
 					opacity: 0.0
 					source: animationEnabled ? "qrc:/images/electron.svg" : "qrc:/images/electron_arrow.svg"
 					visible: root.animationMode !== VenusOS.WidgetConnector_AnimationMode_NotAnimated
+					rotation: animationEnabled ? 0.0 : pathUpdater.angleForArrow(pathUpdater.progress, pathUpdater.startToEnd)
 
 					Behavior on opacity {
 						enabled: root._animated
--- a/components/ObjectAcConnection.qml
+++ b/components/ObjectAcConnection.qml
@@ -52,7 +52,7 @@ QtObject {
 
 	readonly property bool singlePhaseCurrentValid: _phaseCount.value === 1 && currentL1.valid
 	// multi-phase systems don't have a total current
-	readonly property real current: singlePhaseCurrentValid ? currentL1.value : NaN
+	readonly property real current: singlePhaseCurrentValid && currentL1.value !== undefined ? currentL1.value : NaN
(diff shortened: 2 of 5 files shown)


chriadam:
LGTM, but would definitely be good to double check the algorithm, before we merge.  I will ping someone on slack about this.

chriadam:
Indeed, thank you for pointing this out!

The _resetDistance() function in WidgetConnector.qml won't re-evaluate if the animation mode changes.  So, better instead to bind the visibility of the Image delegate in electronRepeater (i.e. the electron) to `visible: root.animationMode !== VenusOS.WidgetConnector_AnimationMode_NotAnimated` instead, I think.

chriadam:
I also wonder if we need to modify the animationMode associated with other widgetConnectors in the OverviewPage?

e.g. the inverterToAcLoadsConnector -> maybe there should be possible "reverse" flows from AC Loads to the Inverter, in some cases?  If the power can go negative, I guess we should visualise it?

components/widgets/WidgetConnector.qml
```diff
@@ -60,8 +60,9 @@ Item {
 		const electronTravelDistance = Math.max(Theme.geometry_overviewPage_connector_electron_interval, _electronTravelDistance)
 		const modelCount = animationEnabled
 			? Math.floor(electronTravelDistance / Theme.geometry_overviewPage_connector_electron_interval)
-			: 1 // show just one arrow, if animations are disabled.
-
+			: animationMode !== VenusOS.WidgetConnector_AnimationMode_NotAnimated
```
blammit, line 63: Could avoid the double negative by changing the condition to check `animationMode === VenusOS.WidgetConnector_AnimationMode_NotAnimated` instead.

chriadam: we shouldn't modify this code, as this method won't re-evaluate when the animation mode changes anyway.  Instead, I suspect that the electron visible property should have a binding on the animationMode.

blammit (approved):
LGTM, but would like @chriadam to check as well.

chriadam:
Found one issue with initialising the electron arrow rotation.  Have added a commit for this - would be good if you could review that one @blammit if possible.

components/widgets/WidgetConnector.qml
```diff
@@ -214,6 +214,8 @@ Item {
 				delegate: Image {
 					opacity: 0.0
 					source: animationEnabled ? "qrc:/images/electron.svg" : "qrc:/images/electron_arrow.svg"
+					visible: root.animationMode !== VenusOS.WidgetConnector_AnimationMode_NotAnimated
+					rotation: animationEnabled ? undefined : pathUpdater.angleForArrow(pathUpdater.progress, pathUpdater.startToEnd)
```
chriadam, line 218: actually, I can't do this.  undefined is wrong.  I should just set to 0 in that case.  the path updater will update it during animations anyway.

chriadam: Fixed.

blammit (approved):

PR #2402, opened 2025-08-05
New drilldowns for AC Loads & Essential Loads

blammit:


components/widgets/EssentialLoadsWidget.qml
```diff
@@ -23,8 +23,37 @@ AcWidget {
...
+			measurements: Global.system.load.acOut,
+			model: essentialLoadDevices
+		})
+	}
+
+	AggregateDeviceModel {
+		id: essentialLoadDevices
+		sourceModels: [
+			acLoadOutputDevices,
+			evChargerOutputDevices,
```
DanielMcInnes, line 41: is it true that evChargerOutputDevices are essential loads? Seems odd to me. If I was running from a battery, I definitely wouldn't be charging an EV. Or a heat pump for that matter.

DanielMcInnes (approved):

PR #2404, opened 2025-08-05
Fix uploading translatable terms file

dirkjanfaber:
Before Qt 6.8.3 the system created a `build/i18n/venus-gui-v2.ts` file. This file no longer gets created, but we do have a `venus-gui-v2_en.ts` file, which contains all of the terms. So instead of passing the non-existing file to the upload script, pass the English source file to be uploaded to POEditor.

I believe this is a minor change and won't affect much else, but just to make sure I'd like it to be reviewed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1186,7 +1186,7 @@ add_custom_target(download_translations
 )
 
 add_custom_target(upload_translations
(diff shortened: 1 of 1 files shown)


dirkjanfaber:
I believe I was too hasty with this change. Need to do more checking before this change (or another one) is needed. Closing the PR meanwhile.

PR #2409, opened 2025-08-06
*** DO NOT MERGE ***

DanielMcInnes:
protoype changes to allow cerbo builds to load qml files from the filesystem, not from qrc:/....


mpvader:
So when people install a second copy on the GX they need to be super careful?

I assume this is a “last resort fix” - no better solution is available?

DanielMcInnes:
If you do a desktop build of this and run it, it will abort, as the qml files are not in the expected spot.
(`cmake --build .; ./bin/venus-gui-v2 --mqtt 10.0.0.2 `)

You have to do the install step as well, and run the installed binary.
(`cmake --build .; cmake --install .; ./install/bin/venus-gui-v2 --mqtt 10.0.0.2`)

@chriadam there are a few things we could do here:
1. the url interceptor could check that the file actually exists at the expected path before replacing the qrc resource with the file system path. If it doesn't exist, print an error and use the qrc resource instead?
2. We could disable LOAD_QML_FROM_FILESYSTEM by default, and only enable it for GX builds. That way desktop builds would just use qrc resources

src/urlinterceptor.cpp
```diff
@@ -0,0 +1,17 @@
+#include "urlinterceptor.h"
```
chriadam, line 1: missing copyright header

src/urlinterceptor.cpp
```diff
@@ -0,0 +1,17 @@
...
+#include <QCoreApplication>
+#include <QUrl>
+
```
chriadam, line 4: add some documentation explaining that it doesn't rewrite urls to images / fonts / translation catalogues etc, but only .qml files which are part of the VenusOS module (i.e. doesn't affect customisations)

src/urlinterceptor.h
```diff
@@ -0,0 +1,14 @@
+#ifndef URLINTERCEPTOR_H
```
chriadam, line 1: copyright header

src/urlinterceptor.h
```diff
@@ -0,0 +1,14 @@
...
+#define URLINTERCEPTOR_H
```
chriadam, line 2: please use VICTRON_VENUSOS_GUIV2_URLINTERCEPTOR or similar as the header guard, to avoid potential clashes with other libraries (e.g. qzxing etc)

scripts/build-gx.sh
```diff
@@ -87,10 +87,11 @@ cd "build-gx"
 
 
 # Configure the project with CMake, setting the build type to MinSizeRel (minimum size release)
-cmake -DCMAKE_BUILD_TYPE=MinSizeRel ..
+cmake -DCMAKE_BUILD_TYPE=MinSizeRel -DLOAD_QML_FROM_FILESYSTEM=ON ..
 
 # Build the project using CMake with the MinSizeRel configuration
-cmake --build . --config MinSizeRel --parallel $(nproc)
+cmake --build .
+cmake --install .
```
chriadam, line 94: do we need an installation prefix?  I guess not?

src/urlinterceptor.cpp
DanielMcInnes: fixed

src/urlinterceptor.cpp
DanielMcInnes: Done

src/urlinterceptor.h
DanielMcInnes: done

CMakeLists.txt
```diff
@@ -37,7 +37,7 @@ else()
     include(GNUInstallDirs)
 endif()
 
-option(LOAD_QML_FROM_FILESYSTEM "disable filesystem loading via cmake -DLOAD_QML_FROM_FILESYSTEM=OFF" ON) # Enabled by default
+option(LOAD_QML_FROM_FILESYSTEM "enable filesystem loading via cmake -DLOAD_QML_FROM_FILESYSTEM=ON" OFF) # Disabled by default
```
chriadam, line 40: I think we shouldn't make this an option.  Instead, we should ALWAYS enable it for GX builds, and always disable it for desktop or WASM builds.  Otherwise, we are relying on people setting the appropriate option at build time, for something which is very important for many customers.

DanielMcInnes:
@chriadam I think the only qml file that actually gets intercepted is InputPanel.qml, from a Loader in ApplicationContent.qml. It seems a bit heavy-handed to use a URL interceptor for this single case - what do you think?

PR #2410, opened 2025-08-07
Update momentary button look and feel, and re-enable shader effects

blammit:


components/controls/MomentaryButton.qml
```diff
@@ -0,0 +1,90 @@
...
+import QtQuick
+import QtQuick.Templates as CT
+import Victron.VenusOS.Shaders
+
+import Victron.VenusOS
+
+CT.Button {
+	id: root
+
+	implicitWidth: contentItem.implicitWidth + root.leftPadding + root.rightPadding
```
DanielMcInnes, line 15: maybe: `implicitWidth: contentItem.width + root.leftPadding + root.rightPadding`
i.e. we care about contentItem's actual width here, not its implicit width

blammit: Ah yes, fixed now, thanks.

blammit: I saw the other discussion just now, have reverted this to use the implicitWidth, and with the standard Math.max() that considers the background width as well.

PR #2413, opened 2025-08-11
Switch pane: implement auto toggle switch control

ybott-qinetic:
The AutoToggleSwitch control is a three way switch controlling two states.

/SwitchableOutput/<x>/State sets the default toggle.
 /SwitchableOutput/<x>/Auto sets the additional toggle.

Contributes to #2303

### DIFF
--- a/cmake/ModuleVenus_Sources.cmake
+++ b/cmake/ModuleVenus_Sources.cmake
@@ -130,6 +130,7 @@ set (VictronVenusOS_QML_MODULE_SOURCES
     components/WasmVirtualKeyboardHandler.qml
     components/WifiModel.qml
     components/ClassAndVrmInstance.qml
+    components/controls/AutoToggleButton.qml
     components/controls/Button.qml
     components/controls/ComboBox.qml
     components/controls/DimmingSlider.qml
@@ -555,6 +556,8 @@ list(APPEND VictronVenusOS_CPP_SOURCES
     src/aggregatetankmodel.cpp
     src/visibleitemmodel.h
     src/visibleitemmodel.cpp
+    src/baseautotoggleswitch.h
+    src/baseautotoggleswitch.cpp
     src/basedevice.h
     src/basedevice.cpp
     src/basedevicemodel.h
--- a/components/SwitchableOutputDelegate.qml
+++ b/components/SwitchableOutputDelegate.qml
@@ -566,6 +566,43 @@ BaseListItem {
 	Component {
 		id: threeStateSwitchComponent
 
-		PlaceholderDelegate {}
+		AutoToggleButton {
+			id: autoToggleButton
+
+			function handlePress(key) {
+				switch (key) {
+				case Qt.Key_Space:
+					focus = true
+					return true
+				}
+				return false
+			}
+
+			height: Theme.geometry_switchableoutput_button_height
+			width: root._buttonWidth
+			enabled: !toggleState.busy || !autoToggleState.busy
+			onChecked: toggleState.backendValue === 1
+			autoChecked: autoToggleState.backendValue
+			onOnClicked: toggleState.writeValue(1)
+			onOffClicked: toggleState.writeValue(0)
+			onAutoClicked: autoToggleState.writeValue(autoToggleState.backendValue === 1 ? 0 : 1)
+
+			SettingSync {
+				id: autoToggleState
+				backendValue: autoState.value
+				onUpdateToBackend: (value) => { autoState.setValue(value) }
+			}
+
+			VeQuickItem {
+				id: autoState
+				uid: root.outputUid + "/Auto"
+			}
+
+			SettingSync {
+				id: toggleState
+				backendValue: output.state
+				onUpdateToBackend: (value) => { output.setState(value) }
+			}
+		}
 	}
 }
--- a/src/baseautotoggleswitch.cpp
+++ b/src/baseautotoggleswitch.cpp
@@ -1,0 +1,39 @@
+#include "baseautotoggleswitch.h"
+
+BaseAutoToggleSwitch::BaseAutoToggleSwitch(QQuickItem *parent)
+    : QQuickControl(parent)
+{
+	setFlag(ItemIsFocusScope);
+	setFocusPolicy(Qt::StrongFocus);
+}
+
+bool BaseAutoToggleSwitch::onChecked() const
+{
+	return m_onChecked;
+}
+
+void BaseAutoToggleSwitch::setOnChecked(bool checked)
+{
+	if (m_onChecked != checked) {
+		m_onChecked = checked;
+		Q_EMIT onCheckedChanged();
+	}
+}
+
+bool BaseAutoToggleSwitch::autoChecked() const
+{
+	return m_autoChecked;
+}
+
+void BaseAutoToggleSwitch::setAutoChecked(bool checked)
+{
+	if (m_autoChecked != checked) {
+		m_autoChecked = checked;
+		Q_EMIT autoCheckedChanged();
+	}
+}
+
+int BaseAutoToggleSwitch::buttonCount() const
+{
+    return 3;
+}
--- a/src/baseautotoggleswitch.h
+++ b/src/baseautotoggleswitch.h
@@ -1,0 +1,40 @@
+#ifndef BASEAUTOTOGGLESWITCH_H
+#define BASEAUTOTOGGLESWITCH_H
+
+#include <QObject>
+#include <QQmlEngine>
+#include <QtQuickTemplates2/private/qquickcontrol_p.h>
+
+class BaseAutoToggleSwitch : public QQuickControl
+{
+	Q_OBJECT
+	QML_ELEMENT
+	Q_PROPERTY(bool onChecked READ onChecked WRITE setOnChecked NOTIFY onCheckedChanged FINAL)
+	Q_PROPERTY(bool autoChecked READ autoChecked WRITE setAutoChecked NOTIFY autoCheckedChanged FINAL)
+	Q_PROPERTY(int buttonCount READ buttonCount CONSTANT FINAL)
+
+public:
+    BaseAutoToggleSwitch(QQuickItem *parent = nullptr);
+
+	bool onChecked() const;
+	void setOnChecked(bool checked);
+
+	bool autoChecked() const;
+	void setAutoChecked(bool checked);
+
+	int buttonCount() const;
+
+signals:
+	void onCheckedChanged();
+	void autoCheckedChanged();
+
+	void onClicked();
+	void offClicked();
+	void autoClicked();
+
+private:
+	bool m_onChecked = false;
+	bool m_autoChecked = false;
+};
+
+#endif // BASEAUTOTOGGLESWITCH_H
--- a/cmake/ModuleVenus.cmake
+++ b/cmake/ModuleVenus.cmake
@@ -23,6 +23,7 @@ target_link_libraries(VictronVenusOS PRIVATE
     Qt6::Quick
     Qt6::Svg
     Qt6::QuickPrivate
+    Qt6::QuickTemplates2Private
     Qt6::Xml
     Qt6::Mqtt
 )
--- a/components/controls/AutoToggleButton.qml
+++ b/components/controls/AutoToggleButton.qml
@@ -1,0 +1,129 @@
+/*
+** Copyright (C) 2025 Victron Energy B.V.
+** See LICENSE.txt for license information.
+*/
+
+import QtQuick
+import Victron.VenusOS
+
+BaseAutoToggleSwitch {
+	id: root
+
+	property int buttonWidth: (root.width - (Theme.geometry_button_border_width * (root.buttonCount + 2)) - Theme.geometry_autotoggle_button_spacing) / root.buttonCount
+
+	implicitWidth: parent.width
+	implicitHeight: Theme.geometry_segmentedButtonRow_height
+
+	// background is the on/off buttons visual border
+	background: Rectangle {
+		anchors.left: parent.left
+		anchors.top: parent.top
+		anchors.bottom: parent.bottom
+		width: buttonWidth * 2 + Theme.geometry_button_border_width * 3
+		radius: Theme.geometry_button_radius
+		color: root.enabled ? Theme.color_ok : Theme.color_font_disabled
+	}
+
+	function notification() {
+		//% "Function is set on Auto Mode"
+		Global.showToastNotification(VenusOS.Notification_Info, qsTrId("autotoggleswitch_function_auto_mode_info"))
+	}
+
+	contentItem: FocusScope {
+		focus: true
+
+		Button {
+			id: offButton
+
+			anchors {
+				left: parent.left
+				leftMargin: Theme.geometry_button_border_width
+				top: parent.top
+				topMargin: Theme.geometry_button_border_width
+				bottom: parent.bottom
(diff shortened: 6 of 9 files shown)


components/AutoToggleButton.qml
```diff
@@ -0,0 +1,190 @@
...
+
+                Label {
+                    anchors.centerIn: autoButton
+                    font.pixelSize: root.fontPixelSize
+                    horizontalAlignment: Text.AlignHCenter
+                    verticalAlignment: Text.AlignVCenter
+                    x: Theme.geometry_tabBar_horizontalMargin
+                    width: parent.width - 2*x
+                    elide: Text.ElideRight
+                    text: "Auto"
```
ybott-qinetic, line 169: Replace with CommonWords.auto

components/AutoToggleButton.qml
```diff
@@ -0,0 +1,190 @@
...
+    }
+
+    T.Control {
+        anchors.fill: parent
+
+        // background is the control's visual border
+        background: Rectangle {
+            anchors.fill: parent
+            radius: Theme.geometry_button_radius
+            color: root.enabled === false ? Theme.color_font_disabled : Theme.color_ok
```
blammit, line 46: `enabled` is a bool so there's no type coercion involved when testing the condition, so just use `!root.enabled` instead of `root.enabled === false` (here and elsewhere, and also for `checked` usage).

components/AutoToggleButton.qml
```diff
@@ -0,0 +1,190 @@
...
+        }
+
+        contentItem: Row {
+            id: buttonRow
+
+            property real buttonWidth: (root.width - (Theme.geometry_button_border_width * 4)) / 3
```
blammit, line 52: Use theme values instead of constants? Also looks like this property can be `readonly`.

components/AutoToggleButton.qml
```diff
@@ -0,0 +1,190 @@
...
+                           : offButton.pressed || root.checked === false
+                             ? Theme.color_button_down_text
+                             : Theme.color_font_primary
+                }
+
+                MouseArea {
+                    anchors.fill: parent
+                    onPressed: {
+                        if (root.autoChecked === false) {
+                            root.toggled()
```
blammit, line 90: It shouldn't toggle between on and off; it should only select the button that you have clicked. Otherwise it's confusing if you select "off" but it changes to "on".

components/AutoToggleButton.qml
```diff
@@ -0,0 +1,190 @@
...
+                Label {
+                    anchors.centerIn: onButton
+                    font.pixelSize: root.fontPixelSize
+                    horizontalAlignment: Text.AlignHCenter
+                    verticalAlignment: Text.AlignVCenter
+                    x: Theme.geometry_tabBar_horizontalMargin
+                    width: parent.width - 2*x
+                    elide: Text.ElideRight
+                    text: CommonWords.on
+                    color: root.enabled === false
```
blammit, line 123: To avoid the code duplication between the three buttons, use an inline component like:

```
component InternalButton : Rectangle {
    property alias text: buttonLabel.text
    
    Label { id: buttonLabel }
    MouseArea { ... }
}
```

and then your content item can just be:

```
contentItem: Row {
    InternalButton {
        text: "Off"
        bottomLeftRadius: Theme.geometry_button_radius - Theme.geometry_button_border_width
        topLeftRadius: Theme.geometry_button_radius - Theme.geometry_button_border_width
    }
    InternalButton {
        text: "Off"
    }
    InternalButton {
        text: "Auto"
        bottomRightRadius: Theme.geometry_button_radius - Theme.geometry_button_border_width
        topRightRadius: Theme.geometry_button_radius - Theme.geometry_button_border_width        
    }
}
```


But also, would it be easier to just use `Button` instead of a custom component? That comes with a built-in press-effect, and that is missing from this current implementation. It also comes with the key navigation handling.

components/AutoToggleButton.qml
```diff
@@ -0,0 +1,190 @@
...
+    signal autoToggled()
+
+    implicitWidth: parent.width
+    implicitHeight: Theme.geometry_segmentedButtonRow_height
+
+    function handlePress(key) {
+        switch (key) {
+        case Qt.Key_Space:
+            if (activeFocus) {
+                offButton.focus = true
```
blammit, line 29: Move this `handlePress()` into the `SwitchableOutputDelegate`, as it's only used from that context.

And then it should be sufficient to `focus=true` on this whole AutoToggleButton control, rather than a specific button.

Currently, once you press Space and the focus is taken by this control, the focus cannot be removed, so you need a `Keys.onEnterPressed` and `Keys.onReturnPressed` handler here to remove the focus from the control.

components/AutoToggleButton.qml
```diff
@@ -0,0 +1,190 @@
...
+import QtQuick
+import QtQuick.Controls as C
+import QtQuick.Controls.impl as CP
+import QtQuick.Templates as T
+import Victron.VenusOS
+
+FocusScope {
+    id: root
+
+    property int fontPixelSize: Theme.font_size_body3
```
blammit, line 15: Looks like the indentation in this file is using spaces rather than tabs.

components/AutoToggleButton.qml
ybott-qinetic: Done

components/AutoToggleButton.qml
ybott-qinetic: I have added readonly.
I am using Theme values. Each buttons width is the total width of the control minus the borders (left and right as well as two inner borders, which makes four) divided by the total number of buttons that will be displayed in this control, which is fixed at 3. Should these buttons have more Theming in regards to layout? I made this generic so small increases/decreases in width control will not break the layout.

components/AutoToggleButton.qml
ybott-qinetic: Done

components/AutoToggleButton.qml
ybott-qinetic: Done.

components/SwitchableOutputDelegate.qml
```diff
@@ -498,6 +498,49 @@ BaseListItem {
...
+			enabled: !toggleState.busy
+			checked: toggleState.backendValue === 1 ? true : false
+			autoChecked: autoState.value === 1 ? true : false
+			onOnClicked: toggleState.writeValue(1)
+			onOffClicked: toggleState.writeValue(0)
+			onAutoClicked: autoState.setValue(autoState.value === 1 ? 0 : 1)
+
+			KeyNavigationHighlight.active: autoToggleButton.activeFocus
+
+			VeQuickItem {
```
blammit, line 534: Use a `SettingSync` instead of writing directly to a `VeQuickItem` here for the auto state. And then actually, the "auto" button should be enabled depending on whether its sync state is busy. And instead of enabling/disabling the entire control based on the `toggleState`, just enable the on/off buttons based on the toggle state, and the auto button based on the auto state.

components/SwitchableOutputDelegate.qml
```diff
@@ -498,6 +498,49 @@ BaseListItem {
...
+			width: root._buttonWidth
+			fontPixelSize: Theme.font_size_body1
+			enabled: !toggleState.busy
+			checked: toggleState.backendValue === 1 ? true : false
+			autoChecked: autoState.value === 1 ? true : false
+			onOnClicked: toggleState.writeValue(1)
+			onOffClicked: toggleState.writeValue(0)
+			onAutoClicked: autoState.setValue(autoState.value === 1 ? 0 : 1)
+
+			KeyNavigationHighlight.active: autoToggleButton.activeFocus
```
blammit, line 532: This is not needed as we don't want to focus the entire control, just the individual controllable buttons.

components/SwitchableOutputDelegate.qml
```diff
@@ -498,6 +498,49 @@ BaseListItem {
...
+				}
+				return false
+			}
+
+			height: Theme.geometry_switchableoutput_button_height
+			width: root._buttonWidth
+			fontPixelSize: Theme.font_size_body1
+			enabled: !toggleState.busy
+			checked: toggleState.backendValue === 1 ? true : false
+			autoChecked: autoState.value === 1 ? true : false
```
blammit, line 527: Don't need the `true` and `false`?

components/AutoToggleButton.qml
```diff
@@ -0,0 +1,144 @@
...
+			id: buttonRow
+
+			anchors.fill: parent
+			anchors.margins: Theme.geometry_button_border_width
+			spacing: Theme.geometry_button_border_width
+
+			InternalButton {
+				id: offButton
+
+				onPressed: if (!root.autoChecked) {
```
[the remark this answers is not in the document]
ybott-qinetic: Fixed.

components/AutoToggleButton.qml
```diff
@@ -0,0 +1,144 @@
...
+			anchors.fill: parent
+			onPressed: parent.pressed()
+		}
+
+		activeFocusOnTab : true
+		Keys.enabled: Global.keyNavigationEnabled
+		KeyNavigationHighlight.active: activeFocus
+	}
+
+	Keys.onSpacePressed: offButton.focus = true
```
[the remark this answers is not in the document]
ybott-qinetic: I think this is fixed with your suggestions.

components/SwitchableOutputDelegate.qml
ybott-qinetic: Removed.

components/SwitchableOutputDelegate.qml
ybott-qinetic: Removed.

components/AutoToggleButton.qml
```diff
@@ -0,0 +1,144 @@
...
+				Keys.onSpacePressed: if (!root.autoChecked) {
+					root.offClicked()
+				}
+				KeyNavigation.right: onButton
+			}
+
+			InternalButton {
+				id: onButton
+
+				onPressed: if (!root.autoChecked) {
```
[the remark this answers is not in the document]
ybott-qinetic: Waiting for disabled design feedback.

components/AutoToggleButton.qml
blammit: I see, in that case, add a comment above that explains what the 4 and 3 are for, or alternatively a `readonly property int buttonCount: 3` and then use that property.

components/AutoToggleButton.qml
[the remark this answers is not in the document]
blammit: Actually, on second thoughts, I think it should be this way instead: the button should still be clickable when autoChecked=true, but, if you click on/off in this state, then the Auto button becomes deselected. (And obviously, handle this behaviour from SwitchableOutputDelegate.qml rather than from within AutoToggleButton.qml.) Because if you click on/off, you are essentially removing the automatic behaviour.

components/AutoToggleButton.qml
```diff
@@ -0,0 +1,139 @@
...
+*/
+
+import QtQuick
+import QtQuick.Templates as T
+import Victron.VenusOS
+
+FocusScope {
+	id: root
+
+	property int fontPixelSize: Theme.font_size_body3
```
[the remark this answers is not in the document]
ybott-qinetic: Done

cmake/ModuleVenus_Sources.cmake
```diff
@@ -23,6 +23,7 @@ set (VictronVenusOS_QML_MODULE_SOURCES
     components/ArcGauge.qml
     components/ArcGaugeQuantityRow.qml
     components/AsymmetricRoundedRectangle.qml
+    components/AutoToggleButton.qml
```
blammit, line 26: Can you put this in `components/controls` instead?

components/AutoToggleButton.qml
```diff
@@ -0,0 +1,111 @@
...
+			Keys.onSpacePressed: root.autoClicked()
+		}
+
+		Keys.onEscapePressed: focus = false
+		Keys.onEnterPressed: focus = false
+		Keys.onReturnPressed: focus = false
+		Keys.onUpPressed: {}
+		Keys.onDownPressed: {}
+		Keys.onLeftPressed: {}
+		Keys.onRightPressed: {}
```
blammit, line 109: As discussed, these handlers need to be in the base item

src/baseautotoggleswitch.h
```diff
@@ -0,0 +1,42 @@
...
+class BaseAutoToggleSwitch : public QQuickControl
+{
+	Q_OBJECT
+	QML_ELEMENT
+	Q_PROPERTY(bool onChecked READ onChecked WRITE setOnChecked NOTIFY onCheckedChanged FINAL)
+	Q_PROPERTY(bool autoChecked READ autoChecked WRITE setAutoChecked NOTIFY autoCheckedChanged FINAL)
+	Q_PROPERTY(int buttonCount READ buttonCount CONSTANT FINAL)
+
+public:
+	BaseAutoToggleSwitch();
```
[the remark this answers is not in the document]
ybott-qinetic: Done

src/baseautotoggleswitch.h
```diff
@@ -0,0 +1,42 @@
...
+
+	void onClicked();
+	void offClicked();
+	void autoClicked();
+
+private:
+	bool m_onChecked = false;
+	bool m_autoChecked = false;
+
+	int m_buttonCount = 3;
```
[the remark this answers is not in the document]
ybott-qinetic: Done

cmake/ModuleVenus_Sources.cmake
ybott-qinetic: Done

blammit:
@chriadam could you check this one as well? The control now has a template implemented in C++, as discussed.

components/SwitchableOutputDelegate.qml
```diff
@@ -566,6 +566,49 @@ BaseListItem {
...
+				return false
+			}
+
+			height: Theme.geometry_switchableoutput_button_height
+			width: root._buttonWidth
+			enabled: !toggleState.busy || !autoToggleState.busy
+			onChecked: toggleState.backendValue === 1
+			autoChecked: autoToggleState.backendValue
+			onOnClicked: {
+				autoToggleState.writeValue(0)
```
chriadam, line 587: I think we should disallow clicks to on/off if auto is currently checked, but maybe I'm wrong?

src/baseautotoggleswitch.cpp
```diff
@@ -0,0 +1,38 @@
+#include "baseautotoggleswitch.h"
+
+BaseAutoToggleSwitch::BaseAutoToggleSwitch(QQuickItem *parent)
+    : QQuickControl(parent)
+{
+	setFlag(ItemIsFocusScope);
```
chriadam, line 6: might also need to set TabFocus policy but not sure - check with Bea

components/SwitchableOutputDelegate.qml
blammit: I wasn't sure at the time, but Rein just confirmed in the latest design update that we should disallow this, yes. If Auto is checked, then show a toast notification if the user tries to click on/off.

ybott-qinetic: I have implemented this as well as the toast notification.

src/baseautotoggleswitch.cpp
ybott-qinetic: I have set the policy for this.

blammit (approved):

PR #2415, opened 2025-08-12
Fix Mock mode performance

chriadam:


src/mockmanager.h
```diff
@@ -46,6 +53,509 @@ class MockManager : public QObject
 	bool m_timersActive = false;
 };
 
+// Repeatedly changes data values to a different value within the specified deltas.
+class MockDataRandomizer : public QObject, public QQmlParserStatus
```
blammit, line 57: Could you move these into a separate .h and .cpp?

blammit (commented):
Looks great to move this to C++. The animators aren't working for me though - e.g. in main, the battery and tank levels increase to the max, then decrease back to min, and does this continuously in an infinite loop (unless Global.timersActive=false). Also if you click on the Batteries list, each of the voltage, power and temperature values  should be changing continuously.

src/mockmanager.h
chriadam: moved to mockanimators.h, but no cpp as I think there's value in everything being inlined

blammit:
Yeah, would be great to move those to C++, or alternatively, have them implemented at the systemcalc level, which would be even better.

PR #2417, opened 2025-08-14
Switch from localStorage to cookies with fallback for index.html

mr-manuel:
On some MFD devices localStorage is not available and therefore the code stops with an error and the WASM is not loaded.

Improves compatibility with devices lacking localStorage by introducing cookie-based helper functions for display mode persistence and fallback if even cookies are not available. Ensures display options are retained even when localStorage is unavailable.

### DIFF
--- a/wasm/index.html
+++ b/wasm/index.html
@@ -219,6 +219,34 @@
             check(location.href, 30, 1000, (function() { location.reload(); }), (function() { console.log("Failed to reload gui-v2: web server unavailable"); }) );
         }
 
+        // Helper functions for cookies
+        function setCookie(name, value, days) {
+            // Check if document.cookie is available, since it might not be on some older MFD devices
+            // or if the browser has disabled cookies.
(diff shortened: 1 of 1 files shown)


wasm/index.html
```diff
@@ -308,25 +336,23 @@
 
             // Options to change display size and border
             // Force fullscreen
-            if (searchParams.has("fullscreen") || (!searchParams.has("default") && !searchParams.has("nomock") && !searchParams.has("mock") && localStorage.getItem('venus-gui-v2-css-display') === 'fullscreen')) {
+            if (searchParams.has("fullscreen") || (!searchParams.has("default") && !searchParams.has("nomock") && !searchParams.has("mock") && getCookie('venus-gui-v2-css-display') === 'fullscreen')) {
```
chriadam, line 339: should we try to use localStorage, and only fall back to cookie if that fails?

PR #2422, opened 2025-08-17
Switch pane: allow key navigation when DimmingSlider is selected

blammit:
The user should be prevented from navigation away from the DimmingSlider when the slider is in "edit" mode, but not when the user has highlighted the slider as part of moving around the grid view.


ybott-qinetic (approved):
Makes sense.

PR #2428, opened 2025-08-18
Key navigation improvements

blammit:


blammit:
Rebased.

chriadam (approved):
LGTM.  The escape/return behaviour seems a bit strange to me, but if it's the design requirement, then LGTM.

Issue #2411, opened 2025-08-07
Switchable output deadlock


martinbosma:
When an output is changed, but the new value is not allowed for some reason, the UI components gets in a deadlock.

Consider the following example with two toggle buttons:
- A main switch 
- A run switch that is only allowed to do something if the main switch is on

Initial state, main switch off and run switch off:
<img width="1174" height="777" alt="Image" src="https://github.com/user-attachments/assets/72800f99-856a-47e6-8a31-27a7f5e79815" />

Then I try to set the run switch to on:
<img width="1171" height="775" alt="Image" src="https://github.com/user-attachments/assets/1b82092c-1f81-4b1a-9044-35f45de54e17" />

The `1` value has been written to `SwitchableOutput/n/State` but not accepted and remains `0`. Since the value didn't change, no value is sent back to the UI. However, now there is no way to change the run switch anymore. Do note that the WASM build behaves differently compared to the local UI. The local UI doesn't get in the "blocked" state, but no new values are sent either.

When I do force sent the unchanged value, it does behave a bit differently. It does reflect the actual state instead of the desired state but is still blocked.
<img width="1174" height="775" alt="Image" src="https://github.com/user-attachments/assets/73742fc3-84bc-4330-9e61-1c36509a7749" />

It feels like perhaps it should at least always write the desired value in the button is pressed and/or there should be a timeout on the toggle state. Perhaps we also need to forcefully send the value from the dbus/mqtt service even if the value hasn't changed. This currently isn't the case. Also, this example applies to the toggle switch but likely affects other types as well.

This setup is available for testing: https://vrm.victronenergy.com/installation/496541/dashboard
However, the SwitchableOutput might not always be available. Please contact me when you need it.

blammit:
> I suppose preset the button state with the desired state is likely done because of fast feedback to the user. Otherwise, if you wait for a response from the device to set the active state, it might feel sluggish. 

Yes, exactly. The current settings list switches do not do this, and that does feel sluggish, especially in Wasm, so perhaps we'll need to consolidate the behaviours at some stage.

>  And would it help that, if the desired state is not accepted, to send the active value again (even if it didn't change)? Can both native and wasm builds handle that?

[...]

PR #2431, opened 2025-08-18
Switch pane: Add disabled toggle switch colours

ybott-qinetic:


blammit:
Discussed that this can be closed, and I will fold the changes into other fixes I need to make for #2411.

PR #2433, opened 2025-08-19
Switch pane: implement multi option switch control

ybott-qinetic:
The MultiStepButton control is a state (on/off) and user defined options switch.

/SwitchableOutput/<x>/Settings/DimmingMax sets the number of options available.
/SwitchableOutput/<x>/Dimming sets the value of the selected option. /SwitchableOutput/<x>/State sets the on/off state of the switch.

Contributes to #2303

### DIFF
--- a/components/MultiStepButton.qml
+++ b/components/MultiStepButton.qml
@@ -1,0 +1,136 @@
+/*
+** Copyright (C) 2025 Victron Energy B.V.
+** See LICENSE.txt for license information.
+*/
+
+import QtQuick
+import Victron.VenusOS
+
+BaseListView {
+	id: root
+
+	property int currentIndex
+	property bool checked
+	readonly property real _totalDelegateWidth: width - headerItem.width
+
+	signal indexClicked(index: int)
+	signal onClicked()
+	signal offClicked()
+
+	implicitWidth: parent.width
+	implicitHeight: Theme.geometry_segmentedButtonRow_height
+
+	orientation: ListView.Horizontal
+	focus: false
+
+	delegate: BaseListItem {
+
+		readonly property bool lastListItem: index === root.model.length - 1
+
+		height: root.height
+		width: root._totalDelegateWidth / root.model.length
+
+		// background border color
+		Rectangle {
+			anchors.fill: parent
+			topRightRadius: lastListItem ? Theme.geometry_button_radius : 0
+			bottomRightRadius: lastListItem ? Theme.geometry_button_radius : 0
+			color: !root.enabled ? Theme.color_gray4
+				: root.checked ? Theme.color_ok
+				: Theme.color_button_off_background
+		}
+
+		Button {
+			anchors {
+				fill: parent
+				topMargin: Theme.geometry_button_border_width
+				bottomMargin: Theme.geometry_button_border_width
+				rightMargin: lastListItem ? Theme.geometry_button_border_width : 0
+			}
+
+			radius: 0 // Override property value intitialisation from base class to ensure PressEffect renders correctly
+			topLeftRadius: 0
+			bottomLeftRadius: 0
+			topRightRadius: lastListItem ? Theme.geometry_button_radius - Theme.geometry_button_border_width : 0
+			bottomRightRadius: lastListItem ? Theme.geometry_button_radius - Theme.geometry_button_border_width : 0
+			borderWidth: 0
+
+			text: modelData.text
+			font.pixelSize: Theme.font_size_body1
+			color: root.enabled ? index === root.currentIndex ? Theme.color_button_down_text : Theme.color_font_primary : Theme.color_font_disabled
+			backgroundColor: !root.enabled ? Theme.color_background_disabled
+					: index === root.currentIndex ? (root.checked ? Theme.color_ok : Theme.color_button_off_background)
+					: Theme.color_darkOk
+
+			enabled: root.checked
+			focus: true
+
+			onClicked: root.indexClicked(index)
+		}
+	}
+
+	header: BaseListItem  {
+		width: Math.ceil(labelTextMetrics.tightBoundingRect.x + labelTextMetrics.tightBoundingRect.width)
+			   + Theme.geometry_dimmingSlider_separator_width + (Theme.geometry_dimmingSlider_text_padding * 2)
+		height: parent.height
+
+		// background border color
+		Rectangle {
+			anchors.fill: parent
+			topLeftRadius: Theme.geometry_slider_groove_radius
+			bottomLeftRadius: Theme.geometry_slider_groove_radius
+			topRightRadius: 0
+			bottomRightRadius: 0
+			color: !root.enabled ? Theme.color_gray4
+				: root.checked ? Theme.color_ok
+				: Theme.color_button_off_background
+		}
+
+		Button {
+			anchors {
+				fill: parent
+				topMargin: Theme.geometry_button_border_width
+				bottomMargin: Theme.geometry_button_border_width
+				leftMargin: Theme.geometry_button_border_width
+			}
+
+			radius: 0 // Override property value intitialisation from base class to ensure PressEffect renders correctly
+			topLeftRadius: Theme.geometry_button_radius - Theme.geometry_button_border_width
+			bottomLeftRadius: Theme.geometry_button_radius - Theme.geometry_button_border_width
+			topRightRadius: 0
+			bottomRightRadius: 0
+
+			text: root.checked ? CommonWords.on : CommonWords.off
+			font.pixelSize: Theme.font_size_body1
+			backgroundColor: root.enabled ? Theme.color_darkOk : Theme.color_background_disabled
+
+			focus: true
+
+			onClicked: root.checked ? root.offClicked() : root.onClicked()
+		}
+
+		Rectangle {
+			anchors.right: parent.right
+			anchors.verticalCenter: parent.verticalCenter
+			width: Theme.geometry_dimmingSlider_separator_width
+			height: parent.height - (Theme.geometry_dimmingSlider_decorator_vertical_padding * 2)
+			radius: Theme.geometry_dimmingSlider_separator_width / 2
+			visible: currentIndex > 0
+			color: root.enabled ? Theme.color_multistepbutton_separator : Theme.color_background_disabled
+		}
+
+		TextMetrics {
+			id: labelTextMetrics
+			font.pixelSize: Theme.font_size_body1
+			text: (CommonWords.on.length > CommonWords.off.length) ? CommonWords.on : CommonWords.off
+		}
+	}
+
+	Keys.onEnterPressed: focus = false
+	Keys.onEscapePressed: focus = false
+	Keys.onReturnPressed: focus = false
+	Keys.onLeftPressed: (event) => { event.accepted = true }
+	Keys.onRightPressed: (event) => { event.accepted = true }
+	Keys.onUpPressed: (event) => { event.accepted = true }
+	Keys.onDownPressed: (event) => { event.accepted = true }
+}
--- a/components/SwitchableOutputDelegate.qml
+++ b/components/SwitchableOutputDelegate.qml
@@ -512,7 +512,64 @@ BaseListItem {
 	Component {
 		id: steppedSwitchComponent
 
-		PlaceholderDelegate {}
+		MultiStepButton {
+			id: multiStep
+
+			function handlePress(key) {
+				// Enter edit mode when space is pressed.
+				switch (key) {
+				case Qt.Key_Space:
+					focus = true
+					return true
+				default:
+					return false
+				}
+			}
+
+			function generateModel() {
+				if (multiStepMax.value === undefined) {
+					multiStep.model = []
+				} else {
+					let items = []
+					// limit maximum number of options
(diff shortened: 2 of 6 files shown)


components/SwitchableOutputDelegate.qml
```diff
@@ -397,7 +397,64 @@ BaseListItem {
...
+					for (let i = 0; i < totalOptions; i++) {
+						items.push({ 'text': i, 'backgroundColor': i === multiStep.currentIndex ? Theme.color_ok : Theme.color_gray1 })
+					}
+					multiStep.model = items
+				}
+			}
+
+			width: root._buttonWidth
+			height: Theme.geometry_switchableoutput_button_height
+			currentIndex: multiStepSelection.valid ?? multiStepSelection.value
```
blammit, line 430: The `??` causes it to return undefined if the value is not present, which you don't want for a currentIndex that requires an int. Could do `multiStepSelection.valid ? multiStepSelection.value : -1` instead.

components/SwitchableOutputDelegate.qml
```diff
@@ -397,7 +397,64 @@ BaseListItem {
...
+					multiStep.currentIndex = Math.floor(multiStepSelection.value)
+					multiStep.generateModel()
+				}
+			}
+
+			SettingSync {
+				id: multiStepState
+				backendValue: output.state
+				onUpdateToBackend: (value) => { output.setState(value) }
+				onBackendValueChanged: multiStep.generateModel()
```
blammit, line 448: Why does generateModel() need to be called here? I don't see where this value is used in generating the model.

components/MultiStepButton.qml
```diff
@@ -0,0 +1,158 @@
+/*
+** Copyright (C) 2025 Victron Energy B.V.
+** See LICENSE.txt for license information.
+*/
+
+import QtQuick
+import QtQuick.Controls as C
+import QtQuick.Controls.impl as CP
+import QtQuick.Templates as T
```
blammit, line 9: The `Controls` and `Controls.impl` imports seem unused

components/SwitchableOutputDelegate.qml
```diff
@@ -397,7 +397,64 @@ BaseListItem {
...
+					multiStep.model = items
+				}
+			}
+
+			width: root._buttonWidth
+			height: Theme.geometry_switchableoutput_button_height
+			currentIndex: multiStepSelection.valid ?? multiStepSelection.value
+			checked: multiStepState.backendValue
+			onClicked: (index) => { multiStepSelection.setValue(index) }
+			onToggled: multiStepState.writeValue(!multiStepState.backendValue)
```
blammit, line 433: If the backend value is not yet in sync with the current displayed value, we might end up changing it to the wrong value here. If the `toggled` signal included bool argument to explicitly indicate whether the desired value is "off" or "on", then that could be used here to make sure the right value is written.

Also, the control needs enabled=false when the state or dimming values are being written.

components/MultiStepButton.qml
```diff
@@ -0,0 +1,158 @@
...
+
+FocusScope {
+	id: root
+
+	property int fontPixelSize: Theme.font_size_body3
+	property alias model: baseListView.model //buttonRepeater.model
+	property int currentIndex
+	property bool checked
+
+	signal clicked(index: int)
```
blammit, line 20: Maybe `onItemClicked` or `onIndexClicked` instead, `clicked()` sounds like when the entire control is clicked.

components/MultiStepButton.qml
```diff
@@ -0,0 +1,158 @@
...
+		property int totalDelegateWidth: width - headerItem.width
+		property int maxIndexCount: model.length - 1
+
+		height: parent.height
+		width: parent.width
+		orientation: ListView.Horizontal
+
+		delegate: BaseListItem {
+			height: root.height
+			width: baseListView.totalDelegateWidth / (baseListView.maxIndexCount + 1)
```
blammit, line 52: maybe `width: baseListView.totalDelegateWidth / baseListView.count` would be more clear

components/MultiStepButton.qml
```diff
@@ -0,0 +1,158 @@
...
+			// background color
+			Rectangle {
+				anchors.fill: parent
+				anchors.topMargin: Theme.geometry_button_border_width
+				anchors.bottomMargin: Theme.geometry_button_border_width
+				anchors.rightMargin: index === baseListView.maxIndexCount ? Theme.geometry_button_border_width : 0
+
+				topRightRadius: index === baseListView.maxIndexCount ? Theme.geometry_button_radius - Theme.geometry_button_border_width : 0
+				bottomRightRadius: index === baseListView.maxIndexCount ? Theme.geometry_button_radius - Theme.geometry_button_border_width : 0
+				color: root.enabled ? index === root.currentIndex ? root.checked ? Theme.color_ok : Theme.color_button_off_background : Theme.color_darkOk : Theme.color_background_disabled
```
blammit, line 71: Can you split this multiple lines and add parentheses, like:

```
color: root.enabled 
    ? index === root.currentIndex 
           ? (root.checked ? Theme.color_ok : Theme.color_button_off_background)
           : Theme.color_darkOk 
    : Theme.color_background_disabled
```

components/SwitchableOutputDelegate.qml
```diff
@@ -397,7 +397,64 @@ BaseListItem {
...
+				}
+			}
+
+			function generateModel() {
+				if (multiStepMax.value === undefined) {
+					multiStep.model = []
+				} else {
+					let items = []
+					// limit maximum number of options
+					let totalOptions = Math.min(10, multiStepMax.value)
```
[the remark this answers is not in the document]
ybott-qinetic: Rein said to limit the maximum to 7

components/MultiStepButton.qml
```diff
@@ -0,0 +1,158 @@
...
+					if (root.checked) {
+						root.clicked(index)
+					}
+				}
+			}
+
+			activeFocusOnTab : true
+			Keys.enabled: Global.keyNavigationEnabled
+			Keys.onSpacePressed: root.clicked(index)
+			KeyNavigationHighlight.active: activeFocus
```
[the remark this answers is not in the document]
ybott-qinetic: I think I have done that.

components/MultiStepButton.qml
ybott-qinetic: Done

components/SwitchableOutputDelegate.qml
ybott-qinetic: I have changed the signals to be more inline with baseautotoggleswitch

components/MultiStepButton.qml
ybott-qinetic: Done

components/MultiStepButton.qml
```diff
@@ -0,0 +1,121 @@
...
+
+import QtQuick
+import Victron.VenusOS
+
+BaseListView {
+	id: root
+
+	property int currentIndex
+	property bool checked
+	property int totalDelegateWidth: width - headerItem.width
```
blammit, line 14: Should this be a real if it's divided to get the individual delegate widths?

Also can be readonly and start with "_" as internal property.

components/MultiStepButton.qml
```diff
@@ -0,0 +1,121 @@
...
+			anchors.leftMargin: Theme.geometry_button_border_width
+
+			radius: 0 // this influences the press effect
+			topLeftRadius: Theme.geometry_button_radius - Theme.geometry_button_border_width
+			bottomLeftRadius: Theme.geometry_button_radius - Theme.geometry_button_border_width
+			topRightRadius: 0
+			bottomRightRadius: 0
+
+			text: root.checked ? CommonWords.on : CommonWords.off
+			font.pixelSize: Theme.font_size_body1
```
blammit, line 89: The text is not visible in light mode when the control is disabled:

image

components/MultiStepButton.qml
```diff
@@ -0,0 +1,121 @@
...
+		// background border color
+		Rectangle {
+			anchors.fill: parent
+			topRightRadius: lastListItem ? Theme.geometry_button_radius : 0
+			bottomRightRadius: lastListItem ? Theme.geometry_button_radius : 0
+			color: root.enabled ? root.checked ? Theme.color_ok : Theme.color_button_off_background : Theme.color_gray4
+		}
+
+		Button {
+			anchors.fill: parent
```
blammit, line 41: Since the Button is not the root item in the delegate, it needs to have `focus: true`, else it cannot be triggered with the Space key during keyboard navigation.

components/MultiStepButton.qml
```diff
@@ -0,0 +1,121 @@
...
+					: index === root.currentIndex ? (root.checked ? Theme.color_ok : Theme.color_button_off_background)
+					: Theme.color_darkOk
+
+			enabled: root.checked
+
+			onClicked: root.indexClicked(index)
+		}
+	}
+
+	header: Rectangle {
```
blammit, line 66: It's not possible to navigate to the header during keyboard navigation;  making it a BaseListItem should fix that.

components/MultiStepButton.qml
```diff
@@ -0,0 +1,121 @@
...
+
+	}
+
+	Keys.onEnterPressed: focus = false
+	Keys.onEscapePressed: focus = false
+	Keys.onReturnPressed: focus = false
+	Keys.onLeftPressed: {}
+	Keys.onRightPressed: {}
+	Keys.onUpPressed: {}
+	Keys.onDownPressed: {}
```
blammit, line 119: When key navigation is enabled, it's possible to navigate away from the control while it is focused. This is because BaseListView explicitly allows the focus to go elsewhere if there are no items left to scroll through in the list. You can avoid this by accepting the events:

``` 
    Keys.onLeftPressed: (event) => { event.accepted = true }
    Keys.onRightPressed: (event) => { event.accepted = true }
    Keys.onUpPressed: (event) => { event.accepted = true }
    Keys.onDownPressed: (event) => { event.accepted = true }
```

data/mock/conf/services/switch-controls-tester.json
```diff
@@ -145,9 +145,9 @@
         "/SwitchableOutput/stepped_switch_2/Status": 9,
 
         "/SwitchableOutput/stepped_switch_3/Dimming": 0,
-        "/SwitchableOutput/stepped_switch_3/DimmingMax": 3,
         "/SwitchableOutput/stepped_switch_3/Name": "4c. Stepped switch, disabled",
         "/SwitchableOutput/stepped_switch_3/Settings/CustomName": "",
+        "/SwitchableOutput/stepped_switch_3/Settings/DimmingMax": 3,
```
blammit, line 150: Oops, thanks for fixing these.

blammit (commented):
Works nicely, just some minor issues remaining

components/MultiStepButton.qml
```diff
@@ -0,0 +1,121 @@
...
+
+	header: Rectangle {
+		width: Math.ceil(labelTextMetrics.tightBoundingRect.x + labelTextMetrics.tightBoundingRect.width)
+			   + Theme.geometry_dimmingSlider_separator_width + (Theme.geometry_dimmingSlider_text_padding * 2)
+		height: parent.height
+		topLeftRadius: Theme.geometry_slider_groove_radius
+		bottomLeftRadius: Theme.geometry_slider_groove_radius
+		topRightRadius: 0
+		bottomRightRadius: 0
+		color: root.enabled ? root.checked ? Theme.color_ok : Theme.color_button_off_background : Theme.color_gray4
```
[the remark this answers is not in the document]
ybott-qinetic: I changed the logic flow slightly.

components/MultiStepButton.qml
ybott-qinetic: Fixed

components/MultiStepButton.qml
ybott-qinetic: Fixed

components/MultiStepButton.qml
```diff
@@ -0,0 +1,132 @@
...
+
+		Button {
+			anchors {
+				fill: parent
+				topMargin: Theme.geometry_button_border_width
+				bottomMargin: Theme.geometry_button_border_width
+				leftMargin: Theme.geometry_button_border_width
+			}
+
+			radius: 0 // this influences the press effect
```
chriadam, line 93: same here

components/MultiStepButton.qml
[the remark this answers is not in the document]
ybott-qinetic: Fixed

components/MultiStepButton.qml
```diff
@@ -0,0 +1,132 @@
...
+
+		Button {
+			anchors {
+				fill: parent
+				topMargin: Theme.geometry_button_border_width
+				bottomMargin: Theme.geometry_button_border_width
+				rightMargin: lastListItem ? Theme.geometry_button_border_width : 0
+			}
+
+			radius: 0 // this influences the press effect
```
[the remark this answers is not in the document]
ybott-qinetic: I have expanded this comment.

components/MultiStepButton.qml
ybott-qinetic: Here too.

chriadam (approved):

blammit (approved):

PR #2435, opened 2025-08-20
Clean up and prepare to remove Global.allDevicesModel

blammit:


src/filteredservicemodel.cpp
```diff
@@ -36,14 +36,33 @@ void FilteredServiceModel::setServiceTypes(const QStringList &serviceTypes)
 {
 	if (m_serviceTypes != serviceTypes) {
 		m_serviceTypes = serviceTypes;
+		if (m_completed) {
+			invalidateFilter();
```
chriadam, line 45: This changes the order (previously, serviceTypesChanged() was emitted before invalidateFilter()).  Is that intentional?

blammit: Yes, this was to bring it in line with NotificationSortFilterProxyModel and when the model is invalidated when filter properties are updated.

PR #2436, opened 2025-08-21
Fix concurrent builds

DanielMcInnes:


CMakeLists.txt
```diff
@@ -80,7 +80,6 @@ include(cmake/RunTests.cmake)
 
 add_dependencies(generate_translation_sources_file theme_parser)
 add_dependencies(${PROJECT_NAME} theme_parser)
-add_dependencies(${PROJECT_NAME} qm_files_exist)
```
DanielMcInnes, line 83: this is the line that caused the issue. `qm_files_exist` is a custom target that will always be built, it doesn't need the dependency added here. Adding the dependency causes the translations to build twice, leading to a race condition. Lupdate is called, then lrelease, then half way through the call to lrelease, lupdate gets called again, temporarily causing an empty .ts file, which confuses lrelease, and causes a build error.

chriadam (approved):

CMakeLists.txt
DanielMcInnes: Seems to still work. Yes, it is diabolical that this doesn't emit a warning.

mr-manuel:
Nice!

PR #2442, opened 2025-08-22
Avoid switchable output deadlock and allow writes even when value is not in sync

blammit:


ybott-qinetic (approved):
Looks good. 
Tested simultaneously with hardware, cerbo-gx and wasm interaction.

PR #2443, opened 2025-08-22
Start page: fix reference to pageManager

blammit:
It may be invalid if the start page is loaded before all Global references are initialized.


mr-manuel (approved):
LGTM

Issue #2311, opened 2025-07-03
Drill-down for "AC Loads" and "Essential Loads" on Overview page


philipptrenz:
See https://victrondevelopment.slack.com/archives/C020QG3JVEW/p1751545755283939

philipptrenz:
Also related to https://github.com/victronenergy/venus-private/issues/380

ReinvdZee:
As discussed in the meeting, the voltage and frequency per phase can be fetched from the inverter/charger service. 
Systemcalc provides these paths (this is on a system with a multi):

```
Ac/In/0/ServiceName   com.victronenergy.vebus.ttyS4
Ac/In/0/ServiceType    vebus
```

Which tells you where to look for the voltage and frequency values.

philipptrenz:
For when to enable the drill-down (clickable blue framed tile on the Overview), I have the following suggestions:

- For "AC Loads" and "Essential Loads" on 1-phase systems, only enable the drill-down if there is at least one service to show
- For "AC Loads" and "Essential Loads" on 3-phase systems, always enable them. The reason is, that the individual phase info might still be insightful
- For "DC Loads", enable them if there is either at least one `.dcload` or `.dcdc` system to show or more than one `.dcsystem` service, so that the table in the top "Total" is insightful again.

blammit:
> Which services to list for AC Loads and Essential Loads?
> com.victronenergy.acload
> com.victronenergy.heatpump
> com.victronenergy.evcharger

Currently, if there are both AC Loads and Essential Loads:
- AC Loads shows heat pumps with `/Position=1` (i.e. AC input position)
- Essential Loads shows heat pumps with `/Position=0` (i.e. AC output position)

If we add `acload` and `evcharger` to the services to be shown, then:
- I guess `acload` services are always shown in AC Loads instead of Essential Loads? Is that right?
- `evcharger` services can have a `/Position` value, so we can use that to determine whether to show each charger in AC Loads vs Essential Loads. (Perhaps fall back to AC Loads if `/Position` is invalid/unavailable)

philipptrenz:
> Currently, if there are both AC Loads and Essential Loads:
> 
> AC Loads shows heat pumps with /Position=1 (i.e. AC input position)
> Essential Loads shows heat pumps with /Position=0 (i.e. AC output position)
> If we add acload and evcharger to the services to be shown, then:
> 
> I guess acload services are always shown in AC Loads instead of Essential Loads? Is that right?
[...]

PR #2445, opened 2025-08-24
Overview: fix table column alignment in AC/Essential loads drilldowns

blammit:
Ensure the last column of each delegate aligns with the last column of the table summary.

Since the table summary uses fixed column widths - rather than a width calculated depending on the unit displayed, which could be calculated independently in each delegate - this width needs to be provided to each delegate.

Contributes to #2311


blammit:
@ybott-qinetic could you check this one please? It will also affect the DC Load updates.

PR #2447, opened 2025-08-25
Rework device models

blammit:
This reworks most of the device models in the codebase. The main changes are:

- AllDevicesModel is now a C++ singleton, rather than a QML object referenced as Global.allDevicesModel that contained various sub-model instances.
- Added the C++ FilteredDeviceModel class, which provides a model with the requested service types. This replaces all uses of ServiceDeviceModel.qml, and greatly simplifies the cases where we need to filter on child ids to decide whether to include a device in a device model.
- The Device List page now uses a new C++ RuntimeDeviceModel class instead of Global.allDevicesModel. This creates a clearer separation between AllDevicesModel and the model used by the Device List, instead of using the same model for both.


components/widgets/AcLoadsWidget.qml
```diff
@@ -31,50 +31,26 @@ AcWidget {
 	// accessible from this AC Loads widget.
 	// For 3-phase systems, the drilldown is always enabled.
 	// For 1-phase systems, only enable the drilldown if there are devices to be shown.
-	enabled: root.measurements.phaseCount > 1 || inputLoadDevices.count > 0
+	enabled: root.measurements.phaseCount > 1 || acLoadDevices.count > 0
```
DanielMcInnes, line 34: are `acLoadDevices` always loads that are powered by AC from grid / shore, or can they be loads powered by an inverter?

src/alldevicesmodel.cpp
```diff
@@ -0,0 +1,232 @@
...
+	return serviceItem
+			&& !knownSystemServices.contains(BaseDevice::serviceTypeFromUid(serviceItem->uniqueId()));
+}
+}
+
+AllDevicesModel::AllDevicesModel(QObject *parent)
+	: QAbstractListModel(parent)
+{
+	for (int i = 0; i < AllServicesModel::create()->count(); ++i) {
+		VeQItem *serviceItem = AllServicesModel::create()->itemAt(i);
```
chriadam, line 47: instead of calling create() in a loop, I'd just hoist it up to the top, e.g. `AllServicesModel *servicesModel = AllServicesModel::create();` then use that servicesModel here in the loop and below in the signal connections.

src/alldevicesmodel.cpp
```diff
@@ -0,0 +1,232 @@
...
+void AllDevicesModel::servicesAboutToBeReset()
+{
+	beginResetModel();
+	m_devices.clear();
+	cleanUp();
+}
+
+void AllDevicesModel::servicesReset()
+{
+	for (int i = 0; i < AllServicesModel::create()->count(); ++i) {
```
chriadam, line 117: again, I'd prefer to hoist the AllServicesModel::create() out of the loop

src/alldevicesmodel.cpp
```diff
@@ -0,0 +1,232 @@
...
+	Device *device = new Device(this, item);
+	m_allDeviceCandidates.insert(item->uniqueId(), device);
+	connect(device, &BaseDevice::validChanged, this, &AllDevicesModel::deviceValidChanged);
+
+	return device;
+}
+
+void AllDevicesModel::deviceValidChanged()
+{
+	Device *device = qobject_cast<Device *>(sender());
```
chriadam, line 146: not a massive fan of using sender(), but I guess we dynamically disconnect the signal connection when the device becomes invalid / disconnected, so this is probably the best way :-/

src/alldevicesmodel.cpp
```diff
@@ -0,0 +1,232 @@
...
+{
+	if (index >= 0 && index < m_devices.count()) {
+		return m_devices.at(index);
+	}
+	return nullptr;
+}
+
+QHash<int, QByteArray> AllDevicesModel::roleNames() const
+{
+	static QHash<int, QByteArray> roles = {
```
chriadam, line 222: nitpick: could be static const QHash

chriadam: also, no need for `=`

src/alldevicesmodel.cpp
```diff
@@ -0,0 +1,232 @@
...
+			return i;
+		}
+	}
+	return -1;
+}
+
+Device *AllDevicesModel::deviceAt(int index) const
+{
+	if (index >= 0 && index < m_devices.count()) {
+		return m_devices.at(index);
```
chriadam, line 215: it might not be required, since I guess the device object will always have a QObject parent (i.e. the VeItemProducer?) but just for self-documenting purposes, it might be best to set `QQmlEngine::setObjectOwnership(device, QQmlEngine::CppOwnership);`

src/runtimedevicemodel.cpp
```diff
@@ -0,0 +1,278 @@
...
+QVariant RuntimeDeviceModel::data(const QModelIndex &index, int role) const
+{
+	const int row = index.row();
+	if (row < 0 || row >= m_devices.count()) {
+		return QVariant();
+	}
+
+	switch (role)
+	{
+	case DeviceRole:
```
chriadam, line 63: might need to set Cpp ownership here too

src/filtereddevicemodel.cpp
```diff
@@ -0,0 +1,456 @@
...
+	}
+}
+
+void ChildFilterHandler::addFilteredChild(VeQItem *child)
+{
+	m_childItems.insert(child->uniqueId(), child);
+	connect(child, &VeQItem::valueChanged, m_model, &FilteredDeviceModel::invalidate, Qt::UniqueConnection);
+}
+
+void ChildFilterHandler::deviceChildAbouToBeRemoved(VeQItem *child)
```
chriadam, line 191: typo, AbouToBe

src/alldevicesmodel.cpp
```diff
@@ -0,0 +1,232 @@
...
+{
+	static QHash<int, QByteArray> roles = {
+		{ DeviceRole, "device" },
+	};
+	return roles;
+}
+
+AllDevicesModel* AllDevicesModel::create(QQmlEngine *engine, QJSEngine *)
+{
+	static AllDevicesModel* instance = new AllDevicesModel(engine);
```
DanielMcInnes, line 230: nit: remove 'static'. 
```
AllDevicesModel* AllDevicesModel::create(QQmlEngine *engine, QJSEngine *)
{
	AllDevicesModel* instance = new AllDevicesModel(engine);
	return instance;
}
```

chriadam (approved):
Various nits, but mostly LGTM.  Great work.

src/alldevicesmodel.cpp
chriadam: no, needs static.

DanielMcInnes: I recently learnt that this is kind of wrong. After the engine is destroyed you get left with a dangling pointer if you use static here. Also, there is an expectation of 1 singleton per engine (I forget why we had multiple engines, maybe unit testing?). Neither of those things is relevant for us, so I guess it's ok.

chriadam: Yes, that's correct - if you have separate engine instances, or the lifetime of the engine doesn't match the lifetime of the application, then it's problematic.  But for us, it's fine.

src/allservicesmodel.cpp
```diff
@@ -77,34 +77,41 @@ QHash<int, QByteArray> AllServicesModel::roleNames() const
...
+		} else {
+			// For D-Bus and Mock, the servicesRoot has children, which are the services to be added.
+			addServicesFromChildrenOf(servicesRoot);
+		}
+	}
+
+	endResetModel();
+
+	if (count() != prevCount) {
+		emit countChanged();
```
DanielMcInnes, line 114: in AllServicesModel::addServicesFromChildrenOf(...), 

`beginInsertRows(QModelIndex(), 0, services.count() - 1);` // is this right?

src/allservicesmodel.cpp
```diff
@@ -77,34 +77,41 @@ QHash<int, QByteArray> AllServicesModel::roleNames() const
...
-		connect(servicesRoot, &VeQItem::childAdded, this, &AllServicesModel::addServicesFromChildrenOf);
-
-		// When a service type is removed, remove all of its children from the service list.
-		connect(servicesRoot, &VeQItem::childAboutToBeRemoved, this, [this](VeQItem *child) {
-			child->disconnect(this);
-			for (auto it = child->itemChildren().constBegin(); it != child->itemChildren().constEnd(); ++it) {
-				removeServiceItem(it.value());
+	if (VeQItem *servicesRoot = BackendConnection::create()->producer()
+			? BackendConnection::create()->producer()->services()
+			: nullptr) {
```
DanielMcInnes, line 86: can this be:

```	if (VeQItem *servicesRoot = BackendConnection::create()->producer()
			? servicesRoot->services()
			: nullptr) {
			```

File: src/filtereddevicemodel.cpp (hunk 8 above)
Comment (DanielMcInnes) [inline src/filtereddevicemodel.cpp:191]: typo

File: components/widgets/AcLoadsWidget.qml [hunk 12]
```diff
@@ -31,50 +31,26 @@ AcWidget {
...
-	AcMeterModel {
-		id: heatPumpInputDevices
-		position: VenusOS.AcPosition_AcInput
-		serviceTypes: ["heatpump"]
-		modelId: "heatpump-input"
-	}
+		childFilterIds: Global.system.showInputLoads
+				? { "acload": ["Position"], "evcharger": ["Position"], "heatpump": ["Position"] }
+				: {}
+		childFilterFunction: (device, childItems) => {
```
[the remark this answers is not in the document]
blammit: In my mind it's like the QSortFilterProxyModel `filterAcceptsRow()` function, so it returns true to include it - maybe could be renamed to be more clear? How about `childFilterAllow`?

src/alldevicesmodel.cpp
```diff
@@ -0,0 +1,232 @@
...
+		if (m_devices.at(i)->serviceItem() == item) {
+			beginRemoveRows(QModelIndex(), i, i);
+			m_devices.removeAt(i);
+			endRemoveRows();
+			emit countChanged();
+			break;
+		}
+	}
+
+	delete device;
```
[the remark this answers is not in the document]
blammit: Is deleteLater necessary? If not, I would just delete it now instead

components/widgets/AcLoadsWidget.qml
blammit: I don't believe so, as the `acLoadDevices` only include "acload", "evcharger" and "heatpump" that are configured for input.

PR #2456, opened 2025-08-26
Key navigation: fixes for tab navigation

blammit:


blammit:
Moved these commits to https://github.com/victronenergy/gui-v2/pull/2428

PR #2468, opened 2025-08-29
Feat/canopen motordrive

citolen:


mpvader:
Can we not use the term Node ? Instead just motor drive?

- CANopen motor drives
- Scan for motor drives
- Discovered motor drives

mpvader:
@chriadam how does this work in gui-v2, pls confirm if it helps if @citolen cleans the commits up? (ie combines them).

chriadam:
> @chriadam how does this work in gui-v2, pls confirm if it helps if @citolen cleans the commits up? (ie combines them).

Yes, I would prefer if it was squashed into a single commit.  I have added some documentation in https://github.com/victronenergy/gui-v2/wiki/Versioning-and-Releasing#commit-messages which describes how we'd prefer our commit messages to be structured, going forward, if possible.

philipptrenz:
> > @chriadam The dependencies for this PR were released in 3.70~34. We can move this one forward.
> > A few things I would like to check with you.
> > 
> > 1. The location of the motor direction inverted setting.
> >    The setting is available on the dbus when the service dbus-canopen-motordrive is running and the motor was added from it.
> >    Initially it was discussed that this setting could go on a dedicated page assuming more settings would go in there, but instead those settings were moved to `pages/settings/PageSettingsCanOpenMotordrive.qml`.
> >    Are you happy for it to be on the page `pages/settings/devicelist/PageMotorDrive.qml` whenever the settings is present?
> 
> I have no strong opinion about it. On the one hand, if there are potentially a large number of similar properties which might or might not be available on a per-vendor basis, I worry that it might set a bad precedent to include in the PageMotorDrive.qml. However, if it's likely to be a very small number, perhaps it is fine as-is. In general, I will defer to @philipptrenz on all such questions.

I’m fine with keeping the toggle at the bottom of the list, like on the screenshot on top. Ideally, we could add our separator element above it and name it “Settings” to have a bit of a visual division, but it’s not mandatory and can also be added later.

chriadam (approved):

pages/settings/PageSettingsCanOpenMotordrive.qml
```diff
@@ -0,0 +1,45 @@
+/*
+** Copyright (C) 2024 Victron Energy B.V.
```
DanielMcInnes, line 2: nit, 2025

chriadam (approved):

PR #2474, opened 2025-09-02
Fix rendering issues on the Mobile Network page

### LINKED ISSUE #2467: Rendering issues on the Mobile Network page
On both v1.2.9 and v1.1.27 (Venus OS v3.64 official) there are some rendering issues with the _Settings -> Connectivity -> Mobile Network_ screen. 

Mainly when the GSM fails to connect, some texts start to overlap. 

Image

Image

The least it should do is first clear the existing text before showing new text.

Also not sure if the check on `com.victronenergy.modem/SimStatus` is correct. There is a GX GSM connected to the GX.

DanielMcInnes:
Fixes #2467.

Prior to this fix, if `simStatus.valid` flicked back and forth, both `VisibleItemModels` would be created and displayed on top of each other. If I used `ObejctModel` instead of `VisibleItemModel` for `modemConnected` and `notConnected`, the GradientListView would behave correctly, making me think that the 'real' bug is in in `VisibleItemModel` somewhere.

Fixing `VisibleItemModel` to behave properly, like `ObjectModel` is likely to be difficult. This fix is easy, and works fine.

### DIFF
--- a/pages/settings/PageSettingsGsm.qml
+++ b/pages/settings/PageSettingsGsm.qml
@@ -30,255 +30,265 @@ Page {
 	GradientListView {
 		id: settingsListView
 
-		model: simStatus.valid ? modemConnected : notConnected
+		model: loader.status === Loader.Ready ? loader.item : null
 
-		VisibleItemModel {
+		Loader {
+			id: loader
+
+			sourceComponent: simStatus.valid ? modemConnected : notConnected
+		}
+
+		Component {
 			id: notConnected
+			VisibleItemModel {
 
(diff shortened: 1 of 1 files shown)


blammit:
VisualItemModel will have to be fixed, since we rely on this pattern in multiple places. I think the issue is only seen when the model changes after the view is already created.

Discussed with @DanielMcInnes on how to reproduce this issue; I've created a fix for VisualItemModel at https://github.com/victronenergy/gui-v2/pull/2475 .

PR #2475, opened 2025-09-03
Models: ensure VisualItemModel items are unparented when unused

### LINKED ISSUE #2467: Rendering issues on the Mobile Network page

blammit:
Fix VisibleItemModel::release() to only return the Referenced flag for an item when the item reference count indicates it is no longer required by the view. Otherwise, the view contentItem will continue to be the parent of the item, which means the item will remain in the view even when it should no longer be visible. This is the same approach taken by ObjectModel.

Fixes #2467

### DIFF
--- a/src/visibleitemmodel.cpp
+++ b/src/visibleitemmodel.cpp
@@ -16,6 +16,17 @@ class VisibleItemModelPrivate : public QObjectPrivate
 {
 	Q_DECLARE_PUBLIC(VisibleItemModel)
 public:
+	class Item {
+	public:
+		Item(QQuickItem *i) : item(i) {}
+
+		void addRef() { ++ref; }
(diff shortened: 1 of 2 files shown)


blammit:
@DanielMcInnes could you test this fix?

blammit:
@chriadam okay if I cherry-pick this to v3.60 as well?

chriadam (approved):

Issue #2188, opened 2025-05-28
Temperature sensor scale step size is whole-integer, need fractional step support


chriadam:
See https://community.victronenergy.com/t/new-ui-setting-temperature-sensor-scale-bug-still-present-in-v3-60-80/33837

chriadam:
PR: https://github.com/victronenergy/gui-v2/pull/2496/files

PR #2496, opened 2025-09-16
Settings: use stepSize of 0.1 for temperature sensor scale setup

chriadam:
Allows setting scaling to non-integer values between 0 to 10.

Contributes to issue #2188

### DIFF
--- a/pages/settings/devicelist/temperature/PageTemperatureSensorSetup.qml
+++ b/pages/settings/devicelist/temperature/PageTemperatureSensorSetup.qml
@@ -54,6 +54,7 @@ Page {
 				preferredVisible: dataItem.valid
 				from: 0
 				to: 10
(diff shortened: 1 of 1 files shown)


pages/settings/devicelist/temperature/PageTemperatureSensorSetup.qml
```diff
@@ -54,6 +54,7 @@ Page {
 				preferredVisible: dataItem.valid
 				from: 0
 				to: 10
+				stepSize: 0.1
```
chriadam, line 57: Possible alternative: we could modify ListSpinBox.qml so that `stepSize: decimals <= 0 ? 1 : (1.0 / decimals)`
But that change is potentially more invasive, so went with this minimal one for now...
What do you think?

PR #2498, opened 2025-09-17
EV charger: show correct charge current (fix ListSpinBox clamped value)

### LINKED ISSUE #2491: EVSE charge current limit shows maximum instead of current value
Observed in latest main branch, but not in 3.6x. Might be in v3.70~34, I did not check.

When opening the EVSE component from the overview, the spinbox shows the maximum charge current instead of the current value it is set to. Example screenshot, compiled GUI app in front, browser (running 3.64) in the back:

Image

blammit:
Do not call QRangeModel::setRange() if the min or max is invalid, as that unexpectedly transforms the ListSpinBox value.

This is the case when EvChargerPage is loading: it creates an EvCharger object, with a maxCurrent value that is briefly set to NaN before it changes to the valid value from the backend. In this situation, the ListSpinBox "to" value is initially NaN, and when
QRangeModel::setRange() clamps the value using invalid min/max values, this leads to its minimumValue becoming NaN and so the minimumValue is auto-set to the maximumValue.

Avoid this issue by delaying the call to setRange() until the min/max are valid.

Also update evhcharger1.json SetCurrent so we can see this issue in mock mode.

Fixes #2491


src/qrangemodel.cpp
```diff
@@ -286,6 +286,9 @@ void QRangeModel::setRange(qreal min, qreal max)
 void QRangeModel::setMinimum(qreal min)
 {
 	Q_D(const QRangeModel);
+	if (qIsNaN(min) || qIsNaN(d->maximum)) {
```
DanielMcInnes, line 289: why is the check for `qIsNaN(d->maximum)` needed?

If `min` is valid, and `d->maximum` is briefly NaN, this will return. When `d->maximum` settles to a sensible value, I don't think this function will be called again, or?

izak:
When I was reading `qrangemodel.cpp` I got the idea that it wasn't designed to be fed Nan values. This code was ported over from gui-v1 according to the commit message, so this would have been a decision made a long time ago, by someone else.

The reason I don't think it is supposed to get NaN values, is because it is initialised to valid integers in `QRangeModelPrivate::init()`, and nowhere else in that code does it check for NaNs. So I got the feeling that the fix would be to check for the NaN at a higher level, potentially in EvChargerPage.qml itself.

Using 32A as a default upper end (until the real one is known) is not a bad fix. That is the typical max the [type-2 connector][1] can handle.

While we're at it, 6A is also the minimum that can be signaled by the analog protocol most EVSEs including our own use. It might be a good idea to also set that on the lower end.

Right now it allows a user to turn it down all the way to zero, but then it jumps back up to 6A later. Which isn't a big problem, but not doing that would be a tiny bit nicer. And I think such hardcoding can be excused for something this specific.

[1]: https://en.wikipedia.org/wiki/Type_2_connector

src/qrangemodel.cpp
blammit: Basically since QRangeModel doesn't seem equipped to handle NaN values for either min/max, I just added the check for both.

DanielMcInnes:
@DanielMcInnes what do you think? ^
sounds good to me

DanielMcInnes (approved):

PR #2528, opened 2025-10-01
Settings: allow user-access to open settings menu items on local UI

### LINKED ISSUE #2526: Access level user: None of the settings accessible from the local UI
On Venus OS 3.70~40.

When the access level is set to user, none of the menu items in the setting is accessible on the local UI. On WASM, things are accessible. On the local UI, not even the settings menu to change the access level is accessible. Also the user cannot open the device list. 

https://github.com/user-attachments/assets/f1dac53c-438b-4da5-9f3b-8b78ffdab93e

blammit:
The (instanceof ListNavigation) condition in ListItem fails when running on device, which means none of the ListNavigation items can be opened. This seems to be an issue with the Qt 6.8.3 device build, as it works fine on Wasm and Desktop.

Use a test for a hardcoded property name instead. It's not ideal but it also avoids the need to add yet another read/write/visibility configuration property to ListItem.

Fixes #2526

### DIFF
--- a/components/listitems/core/ListItem.qml
+++ b/components/listitems/core/ListItem.qml
@@ -81,7 +81,7 @@ BaseListItem {
 	function activate() {
 		if (root.interactive) {
 			// Issue #1964: userHasWriteAccess is ignored for ListNavigation
-			if (root instanceof ListNavigation || root.userHasWriteAccess) {
+			if (root.__is_venus_gui_list_navigation__ === true || root.userHasWriteAccess) {
 				root.clicked()
 			} else {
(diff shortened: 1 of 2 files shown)


blammit:
There are also uses of `instanceof` in PageDebugVeQItems.qml and OverviewPage.qml, but those uses are only for key navigation, which is not relevant on device. There are some Qt bugs related to instanceof like https://bugreports.qt.io/browse/QTBUG-137581 so I expect this will be fixed in the next version.

ybott-qinetic (approved):

PR #2530, opened 2025-10-02
UI Controls: reduce use of transparent colours and reduce colour set

blammit:
Only use transparent colours where necessary - for example, in BatteryWidget where the fill animation may partly cover the text.


chriadam (approved):
In general, LGTM, and always good to reduce usage of transparent colours.

Note that for font colours it's not too important, since font rendering antialiasing means that the result always contains transparency and the glyph will pretty much always be alpha-blended into the scene.

But, in general, always happy to reduce usage of transparency in the app.

PR #2531, opened 2025-10-02
Notifications improvements

chriadam:


chriadam:
This implements most of the improvements, however there are two things still required, I believe:
1) need to refactor the NotificationsModel so that its elements (i.e. BaseNotification instances) are "live" objects with connections to the backend (via VeQItem etc) so that we don't have the Notification.qml type which binds values at the QML level.
2) then we can add an invokable method to NotificationsModel: `acknowledgeNotifications(int type)`.  That one can be called to acknowledge all "Info" notifications known to the model (i.e. for the "automatically acknowledge all info notifications upon navigating away from the notifications page" subtask), or alternatively to acknowledge all non-info notifications (which I believe we should call when the user clicks the "Silence alarm" button on a Warning/Alarm Toast).

ReinvdZee:
Hi @chriadam,

I tried this build. Looks good. 

One question, I believe the user can never see the number of notification on the notification menu item, because the toast message is blocking it. The only way to get rid of the toast message is to silence the alarm, but that would also dismiss the number on the notifications menu item. Also, opening a second instance of the UI reissues the toast message. Am I correct? Or is there a way for the user to see the number on the notifications menu item?

Maybe we should dismiss the toast after a while, so that if the number of notifications pile up, the user will see the status bar icon and a larger number on the notifications menu item instead of a toast with just the latest notification.

chriadam:
If the toasts are info toasts, you should be able to dismiss without silence/acknowledge.  Then the bubble number should be visible.

But in general: yes, we definitely could automatically hide toasts after some period of time.  I wonder whether it's a good idea if the notification is an Alarm toast, though - that one maybe should stay on top to ensure the user sees it?

Also, one possibility: when we silence/acknowledge from either a Warning or Alarm toast, maybe we should only silence/acknowledge Warning/Alarm notifications, so that we don't reset the bubble count for Info notifications, so the user can see after dismissing that there might be other Info notifications worth seeing?

ReinvdZee:
>If the toasts are info toasts, you should be able to dismiss without silence/acknowledge. Then the bubble number should be visible.

Oh right. Yes that works.

> But in general: yes, we definitely could automatically hide toasts after some period of time. I wonder whether it's a good idea if the notification is an Alarm toast, though - that one maybe should stay on top to ensure the user sees it?

True, the toast definitely draws more attention than just the status bar icon and the bubble count. 

>Also, one possibility: when we silence/acknowledge from either a Warning or Alarm toast, maybe we should only silence/acknowledge Warning/Alarm notifications, so that we don't reset the bubble count for Info notifications, so the user can see after dismissing that there might be other Info notifications worth seeing?

We don't have many information type notifications, but I think this is a good idea. Makes sense to not acknowledge information type notification when silencing an alarm that is not caused by that notification in the first place.

chriadam:
Ok.  That latter one will have to wait until I finish the model improvements mentioned.

chriadam:
> Before merging, could you also test, if the injected notification is still shown correctly? Since this is a function that broke easily. When it broke no notification popup was shown, but only the warning/error/info symbol in the notification bar.

I tested this on device tonight, and it seems to work now.

e.g. can do:

```
dbus -y com.victronenergy.platform /Notifications/Inject SetValue "1$(echo -e '\t')Alarm Title$(echo -e '\t')Description 1"
dbus -y com.victronenergy.platform /Notifications/Inject SetValue "0$(echo -e '\t')Warning Title$(echo -e '\t')Description 2"
dbus -y com.victronenergy.platform /Notifications/Inject SetValue "2$(echo -e '\t')Info Title$(echo -e '\t')Description 3"
```

ApplicationContent.qml
```diff
@@ -98,8 +98,8 @@ FocusScope {
 
 	NotificationLayer {
 		id: notificationLayer
-
 		anchors.fill: parent
+		animationEnabled: Global.animationEnabled
```
DanielMcInnes: Are we supposed to leave a blank line after the id or not? Would be nice to keep it consistent

components/NotificationDelegate.qml
```diff
@@ -10,8 +10,8 @@ import Victron.VenusOS
 BaseListItem {
 	id: root
 
-	required property BaseNotification notification
-	required property int notificationId
+	required property int modelId
```
DanielMcInnes: does `modelId` mean the index into the model for this element, or something that identifies the model? If it's the former, can you change the name?

pages/NotificationLayer.qml
```diff
@@ -48,24 +65,121 @@ Item {
...
 			bottomMargin: Theme.geometry_toastNotification_bottomMargin
 		}
-		width: parent.width
-		height: childrenRect.height
-		spacing: Theme.geometry_toastNotification_bottomMargin
-		layoutDirection: Qt.RightToLeft     // layout from bottom to top
 
-		model: ObjectModel {
-			id: toastItemsModel
+		height: 2*Theme.geometry_toastNotification_minHeight // childrenRect.height
```
DanielMcInnes: leftover comment?

pages/NotificationLayer.qml
```diff
@@ -48,24 +65,121 @@ Item {
...
-	}
+		function resetAnimationEnabled() {
+			animationEnabled = true
+		}
+
+		delegate: Item {
+			id: toastContainer
+
+			required property int index
+
```
DanielMcInnes: bonus blank line

pages/NotificationsPage.qml
```diff
@@ -23,45 +23,51 @@ SwipeViewPage {
 		// over the top of the notificationsView
 		clip: true
 
-		header: Global.notifications.activeOrUnAcknowledgedCount === 0 ? noAlertsHeader : null
+		property int activeNotifications: NotificationModel.activeAlarms + NotificationModel.activeWarnings + NotificationModel.activeInfos
```
DanielMcInnes: readonly?

src/notificationmodel.cpp
```diff
@@ -0,0 +1,883 @@
...
+	} else {
+		QTimer::singleShot(2000, this, &NotificationModel::init);
+	}
+}
+
+void NotificationModel::watchSlot(VeQItem *slotItem)
+{
+	const QString id = slotItem->id();
+	const qsizetype idLen = id.length();
+	if (idLen == 0 || idLen > 2 || !id.at(0).isDigit() || (idLen == 2 && !id.at(1).isDigit())) {
```
DanielMcInnes: does this intentionally limit the number of notifications to 99? Sounds reasonable, but probably worth a comment

src/notificationmodel.cpp
```diff
@@ -0,0 +1,883 @@
...
+	toast.modelId = ++m_modelId;
+	toast.notificationModelId = notificationModelId;
+	toast.type = type;
+	toast.description = text;
+	toast.autoCloseInterval = autoCloseInterval;
+
+	// insert in appropriate sort order, highest prio first.
+	for (qsizetype i = 0; i < m_data.size(); ++i) {
+		if (type == Enums::Notification_Alarm
+				|| m_data[i].type == type
```
DanielMcInnes: Do we care about maintaining FIFO order within a given priority? E.g. I think new alarms go in front of old alarms.

DanielMcInnes (approved):
lgtm, nits. I didn't test this, code review only.

mr-manuel (approved):
Notification injection still works

pages/NotificationLayer.qml
chriadam: yes, will remove

pages/NotificationLayer.qml
chriadam: removed

src/notificationmodel.cpp
```diff
@@ -0,0 +1,883 @@
...
+	if (entry.type != Enums::Notification_Alarm
+			&& entry.type != Enums::Notification_Warning
+			&& entry.type != Enums::Notification_Info) {
+		qWarning() << "Adding unknown notification type:" << entry.type;
+		// TODO: early return / slot contains invalid notification?
+	}
+
+	// We don't want to store infinite history.
+	// Make some room by removing the oldest non-associated row.
+	if (m_data.size() >= m_maximumRows) {
```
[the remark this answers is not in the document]
chriadam: Updated comment for clarity.

src/notificationmodel.cpp
```diff
@@ -0,0 +1,883 @@
...
+			return m_data[row].modelId;
+		case static_cast<int>(NotificationRoles::Type):
+			return m_data[row].type;
+		case static_cast<int>(NotificationRoles::Acknowledged):
+			return m_data[row].acknowledged;
+		case static_cast<int>(NotificationRoles::Active):
+			return m_data[row].active;
+		case static_cast<int>(NotificationRoles::Silenced):
+			return m_data[row].silenced;
+		case static_cast<int>(NotificationRoles::Section):
```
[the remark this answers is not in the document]
chriadam: I thought about this, but it would require another linear-time traversal of m_data, I don't think it's worth it.

pages/NotificationsPage.qml
```diff
@@ -76,26 +82,40 @@ SwipeViewPage {
...
+					width: checkmarkIcon.width + 2*checkmarkIcon.anchors.leftMargin
 					height: parent.height
 
 					Image {
 						id: checkmarkIcon
-						anchors.centerIn: parent
+						anchors {
+							verticalCenter: parent.verticalCenter
+							left: parent.left
+							leftMargin: Theme.geometry_listItem_content_horizontalMargin - 8 // (48 - 32) / 2, to centre with delegate icons
```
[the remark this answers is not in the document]
DanielMcInnes: bump

src/notificationmodel.cpp
chriadam: not quite.  there are exactly 20 "slots".  that is, under the /Notifications path there are various child paths (including e.g. data items which provide summary of active or unacknowledged counts etc) but there are 20 paths (from "0" through "19") which are specific notification slots.  see https://github.com/victronenergy/venus-platform/blob/master/src/notifications.cpp#L91 for info.

src/notificationmodel.cpp
```diff
@@ -0,0 +1,687 @@
...
+}
+
+void NotificationModel::init()
+{
+	if (m_data.size()) {
+		qInfo() << "NotificationModel re-initialising!";
+		m_acknowledgeAll = nullptr;
+		qDeleteAll(m_slots);
+		m_slots.clear();
+		beginRemoveRows(QModelIndex(), 0, m_data.size() - 1);
```
blammit: Should this be a model reset rather than removing all rows?

src/notificationsortfilterproxymodel.h
```diff
@@ -23,37 +25,30 @@ class NotificationSortFilterProxyModel : public QSortFilterProxyModel
...
 	~NotificationSortFilterProxyModel();
 
 	int count() const;
-	QJSValue filterFunction() const;
-	void setFilterFunction(const QJSValue &callback);
 
-	QJSValue sortFunction() const;
-	void setSortFunction(const QJSValue &callback);
+	Q_INVOKABLE notificationData at(int row) const;
+	Q_INVOKABLE int sectionAt(int row) const;
```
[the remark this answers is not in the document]
chriadam: yeah, I added it for debugging, trying to figure out what was going wrong with the section headers.  I will remove it.

src/notificationmodel.cpp
chriadam: you're right, yes

src/notificationmodel.cpp
```diff
@@ -0,0 +1,687 @@
...
+		// This may be a NEW notification, so we need to
+		// create a new entry and associate it with this slot.
+		// However, it may just be the first-time-value-set
+		// for a previously newly-added slot.
+		// Reverse-iterate as newer data entries are more likely to have associated slots.
+		const QString slotId = slot->notificationId();
+		bool found = false;
+		for (qsizetype i = m_data.size() - 1; i >= 0; --i) {
+			if (m_data[i].notificationId == slotId) {
+				found = true;
```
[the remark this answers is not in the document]
chriadam: yep, done

PR #2536, opened 2025-10-08
EV Charger: hide Overview labels that convey no information

### LINKED ISSUE #2532: EVCS: Hide some of the details when not available
When using an energy meter to proxy for an EVCS, some values might not be available.

In the image below, the time and state (Onbekend/Unknown) can be hidden.

Image

ybott-qinetic:
Fixes #2532

### DIFF
--- a/components/widgets/EvcsWidget.qml
+++ b/components/widgets/EvcsWidget.qml
@@ -72,6 +72,7 @@ OverviewWidget {
 				elide: Text.ElideRight
 				text: Global.evChargers.chargerStatusToText(statusItem.value)
 				color: Theme.color_font_secondary
+				visible: statusItem.valid
 
 				VeQuickItem {
 					id: statusItem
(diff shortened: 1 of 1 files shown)


chriadam:
Please update the description in the commit message according to https://github.com/victronenergy/gui-v2/wiki/Versioning-and-Releasing#commit-messages especially include a link back to the task number which this relates to.

components/widgets/EvcsWidget.qml
```diff
@@ -100,6 +102,8 @@ OverviewWidget {
 
 					text: Utils.formatAsHHMM(chargingTimeItem.value, true)
 					color: Theme.color_font_secondary
+					// do not show value under a minute
+					visible: chargingTimeItem.value > 60
```
chriadam, line 106: I'm not certain, but I suspect we should only hide it if the status is invalid / unknown.  That is, we still show it if it's charging, even if it has only been charging for less than a minute.

chriadam: Or, perhaps hide it if chargingTimeItem.value === 0 ?

chriadam (approved):
LGTM

blammit (approved):

PR #2549, opened 2025-10-10
Switches: set multi-step and 3-state controls based on expected value

blammit:
If the backend has not yet updated its value, show the button as checked if it was clicked by the user; it will revert to the unchecked state when SettingsSync times out.

Also, allow the three-state switch to be clicked even while waiting for the backend to update, as the user may want to click to force a value to be set.

Contributes to victronenergy/venus#1561


chriadam (approved):
LGTM but Toby knows more about the switches backend synchronisation logic than I do, so would like him to review also

ybott-qinetic (approved):

PR #2571, opened 2025-10-23
Bump version to v1.2.16

ybott-qinetic:


chriadam:
Not valid any more, will create a new version bump today, closing.

PR #2582, opened 2025-10-28
Add menu for additional PV-inverter modbus settings

izak:
This adds a menu for adding additional port and unit ID settings for PV-inverters.

Please feel free to make any improvements you feel is necessary.

https://github.com/victronenergy/venus/issues/1531

### DIFF
--- a/pages/settings/PageSettingsFroniusModbus.qml
+++ b/pages/settings/PageSettingsFroniusModbus.qml
@@ -1,0 +1,112 @@
+/*
+** Copyright (C) 2025 Victron Energy B.V.
+** See LICENSE.txt for license information.
+*/
+
+import QtQuick
+import Victron.VenusOS
+
+Page {
+	id: root
+
+	property string settings: Global.systemSettings.serviceUid
+
+	topRightButton: Global.systemSettings.canAccess(VenusOS.User_AccessType_Installer)
+			? VenusOS.StatusBar_RightButton_Add
+			: VenusOS.StatusBar_RightButton_None
+
+	Connections {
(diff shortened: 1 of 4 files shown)


pages/settings/PageSettingsFroniusModbus.qml
```diff
@@ -0,0 +1,115 @@
...
+
+import QtQuick
+import Victron.VenusOS
+
+Page {
+	id: root
+
+	property string settings: Global.systemSettings.serviceUid
+
+	topRightButton: VenusOS.StatusBar_RightButton_Add
```
blammit: @izak I have an open PR at https://github.com/victronenergy/gui-v2/pull/2577 to only allow fronius IP address additions/removals when the access level is "installer" (to align with gui-v1 behaviour).

Should the same be done here? If so, I can make the changes directly to your commit.

izak: Yes, that sounds like a good idea, please make it so.

pages/settings/PageSettingsFroniusModbus.qml
```diff
@@ -0,0 +1,115 @@
...
+
+		uid: root.settings + "/Settings/Fronius/ModbusAlternates"
+		// eg: [[1501,1],[1502,2]]
+	}
+
+
+	GradientListView {
+		header: PrimaryListLabel {
+			horizontalAlignment: Text.AlignHCenter
+			//% "The default modbus port is 502 and the default unit ID is 126."
```
blammit: @izak is this supposed to be 1502 rather than 502? (Seeing as the default port from the "Add port" option is 1502)

Issue #2202, opened 2025-06-02
Tank sensor custom shape edit option


ReinvdZee:
There should be an option to edit custom tanks shape points. Currently, an incorrect point needs to be deleted and recreated to fix a mistake in the sensor level or volume.

![Image](https://github.com/user-attachments/assets/3dd8ed63-cb4f-46b7-84b2-c87776aa7e60)

I'd suggest that clicking on the menu item itself reopens the dialog box with the current values and lets the user change them.

PR #2588, opened 2025-10-29
Tank: Allow editing custom shape points

chriadam:
Previously, to edit a point one had to delete it and re-add it with the updated value.

This commit allows the user to press the entry in the list in order to edit it.

It also makes two minor improvements:
- increase the touch area of the remove button
- pre-resolve symbols in the remove button clicked handler to avoid reference error upon splicing the model data which results in the delegate being orphaned from the component context.

Contributes to issue #2202

### DIFF
--- a/pages/settings/devicelist/tank/PageTankShape.qml
+++ b/pages/settings/devicelist/tank/PageTankShape.qml
@@ -24,13 +24,11 @@ Page {
 		enabled: root._canEditPoints && root.isCurrentPage
 
 		function onRightButtonClicked() {
-			if (!root._editPointDialog) {
-				root._editPointDialog = editPointDialogComponent.createObject(root)
-			}
-			root._editPointDialog.sensorLevel = 1
-			root._editPointDialog.volume = 1
-			root._editPointDialog.modelIndex = -1
-			root._editPointDialog.open()
+			Global.dialogLayer.open(editPointDialogComponent, {
+					"sensorLevel": 1,
+					"volume": 1,
+					"modelIndex": -1
+				})
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


chriadam:
I agree, I will implement as you suggest

components/controls/SpinBoxInputArea.qml
```diff
@@ -99,8 +99,8 @@ TextInput {
 		event.accepted = false
 	}
 
-	leftPadding: Theme.geometry_textField_horizontalMargin
-	rightPadding: suffixLabel.width + Theme.geometry_quantityLabel_spacing + Theme.geometry_textField_horizontalMargin
+	leftPadding: 0
```
chriadam: This padding got moved into the contentArea of SpinBox.  I don't believe that MiniSpinBox required it here at all, and it was much cleaner to put it in SpinBox (so if the space gets squeezed, that margin can be used up without affecting the content of the SpinBoxInputArea.

@blammit can you please double check that I haven't broken something in MiniSpinBox without realising it?  I did some testing with the mock switches conf, and I believe they look correct still, but would like you to confirm if possible.

pages/settings/devicelist/tank/PageTankShape.qml
```diff
@@ -104,16 +104,53 @@ Page {
...
+							pointsItem.savePoints(pointList)
 						}
 					}
 				}
 			]
+
+			interactive: root._canEditPoints
+			onClicked: {
+				if (!root._editPointDialog) {
+					root._editPointDialog = editPointDialogComponent.createObject(root)
```
blammit: Can we just use `Global.dialogLayer.open(editPointDialogComponent)` instead of keeping a reference to `root._editPointDialog`? We used to do this (make a dialog instance and re-use it) but it shouldn't be necessary anymore.

pages/settings/devicelist/tank/PageTankShape.qml
```diff
@@ -241,16 +287,22 @@ Page {
 
 			tryAccept: function() {
 				_pointArrayToSave = ""
-				let pointList = pointsListView.model
+				let pointList = []
+				let point = []
```
blammit: looks like `point` is only used within the loop and could be declared there

PR #2599, opened 2025-11-03
Switches: color wheel - small fixes

blammit:


chriadam (approved):
LGTM, but I haven't tested on device, as I am WFH today

PR #2601, opened 2025-11-04
Fix color selector close button theme values

ybott-qinetic:


components/dialogs/ColorWheelDialog.qml
```diff
@@ -87,7 +87,13 @@ ModalDialog {
 		}
 
 		CloseButton {
-			anchors.right: parent.right
+			anchors {
+				top: parent.top
+				topMargin: 16
+				right: parent.right
+				rightMargin: 32
+			}
```
blammit: Can you fix this in `CloseButton` instead? I think it will need to use a fixed width/height instead of `icon.sourceSize.width/height`. Otherwise the bug is also occurring in other places where we use `CloseButton` - e.g. in `SolarDailyHistoryDialog.qml`, which is seen in Overview > Solar widget > (pick relevant device if in a list) > History > Chart (if more than two days of history is selected in the dropdown) > click on a day in the bar chart.

Also:
- use Theme values instead of constants
- add a topic prefix to the commit message, e.g. UI Controls

ybott-qinetic: Done

blammit (approved):
Remove the other commit that's unrelated to the close button, then LGTM

PR #2602, opened 2025-11-04
EVCS measurement updates

blammit:


chriadam (approved):
LGTM.  Are the backend changes already available in Venus OS?

blammit:
> LGTM. Are the backend changes already available in Venus OS?

Yes, they are available now, so we can merge.

PR #2609, opened 2025-11-05
Switches: hide controls when switch parameters are not valid

blammit:


pages/settings/devicelist/switchable-outputs/SwitchableOutputListDelegate.qml
```diff
@@ -26,13 +26,21 @@ ListQuantityGroupNavigation {
...
-		QuantityObject { object: output; key: "typeText" }
+		QuantityObject {
+			object: output
+			key: output.secondaryDataText ? "secondaryDataText" : ""
+			unit: VenusOS.Units_None
+		}
+		QuantityObject {
+			object: output
+			key: "typeText"
+			valueColor: output.hasValidType ? undefined : Theme.color_critical
```
chriadam: is the `undefined` deliberate (i.e. to call a reset method?)

src/switchableoutput.cpp
```diff
@@ -403,3 +444,4 @@ void SwitchableOutput::updateFormattedName()
 		emit formattedNameChanged();
 	}
 }
+\
```
chriadam: typo?

chriadam (approved):

PR #2616, opened 2025-11-06
Update CAN-bus profiles

martinbosma:
The `VE.Can & Lynx Ion BMS (250 kbit/s)` profile was renamed to `VE.Can`.
The `CANopen Motor drive (250 kbit/s)` profile now also adds the VE.Can service, so add that to the name.

While at it, also change motor drive to E-drive, as that is what we use for marketing. Besides, "motor drive" doesn't fit on the screen when using the classic UI.

image


blammit:
Renamed the commit message topics from "CanbusProfile" to "VE.Can".

PR #2621, opened 2025-11-06
add an Auto setting for ESS meter selection

izak:
This allows the system to use an energy meter if present, but fall back to using the Inverter/Charger if the meter is lost. Or in other words, it automatically uses a meter if available, otherwise the internal measurement.

https://github.com/victronenergy/venus-private/issues/595


blammit (approved):

blammit:
Added "ESS:" to the start of the commit message.

PR #2626, opened 2025-11-07
Switches: write rounded numbers to /LightControls

blammit:
Double precision is not required, so round the numbers.

Although the data format is a list of doubles, this is only because the backend does not distinguish between lists of doubles vs ints for the data types.


blammit:
Actually, we'll use a list of ints instead. Closing.

PR #2628, opened 2025-11-07
Switches: Use correct lower slider theme values

ybott-qinetic:


blammit:
@ybott-qinetic is this PR still relevant or can it be closed?

PR #2634, opened 2025-11-10
Switches: use int list for LightControl data

blammit:
The backend type has changed from a list of doubles to ints.

Note the implementation still uses QColor setHsvF() and getHsvf() instead of setHsv() and getHsv(), as the former functions use 0-1 ranges for the saturation and value components (which map more easily to the 0-100 range expected by the backend API), while the latter functions use 0-255.


blammit:
> Works good for me, though testing was a bit limited due to https://github.com/victronenergy/gui-v2/pull/2609 not being merged in yet.

I've merged that now, should make things easier.

Issue #2607, opened 2025-11-05
Notification center: Make toast text left-aligned.


ReinvdZee:
The text on the toast message should be left-aligned.

<img width="720" height="424" alt="Image" src="https://github.com/user-attachments/assets/830b529d-a633-48c3-aa62-4b79c15522ef" />

EDIT:
According to Serj, this applies to all toast notifications, not only the ones coming from notifications.
Has this been incorrect from the beginning then?

chriadam:
Adding back prioritisation, just to make filtering easier.

chriadam:
See PR https://github.com/victronenergy/gui-v2/pull/2640

PR #2640, opened 2025-11-11
Notifications: Don't acknowledge Info toasts on dismiss

chriadam:
Only Warning and Alarm toasts should be acknowledged on close or dismiss; Info toasts should remain unacknowledged.

Also, left align text in all toasts.

Contributes to issue #2607


ReinvdZee:
I quickly tested this yesterday btw, looks good.

PR #2651, opened 2025-11-13
RGB wheel fixes

blammit:


chriadam (approved):
Code LGTM.  Testing is tricky at the moment because of the image not including the preset stuff, but I think we can get it into the release for wider testing anyway.

Issue #2331, opened 2025-07-13
WASM: Icons sometimes become miss-sized


philipptrenz:
I noticed now a few times, that icons become miss-sized when resizing the browser, appearing much larger than initially. So far, I wasn't able to reproduce it, though. 

Happened in Chrome 138.0.7204.101 and earlier on MacOS (ARM64).

<img width="1024" height="600" alt="Image" src="https://github.com/user-attachments/assets/aab6a5bd-de7e-4d03-b40d-0ea437fd1263" />

chriadam:
I have seen cases like this in other projects, where some icon/image sourceSize.width and sourceSize.height are bound separately.  This should NEVER be done in code, as those bindings will be re-evaluated separately.  If we have cases where we do this, we need to ensure we bind `sourceSize: Qt.size(someWidthExpression, someHeightExpression)` instead.

PR #2653, opened 2025-11-13
UI Controls: fix erroneously resizing icons in WASM

chriadam:
Ensure that icon sourceSize is bound atomically rather than
separately binding sourceSize.width and sourceSize.height.

Ensure that we don't attempt to use anchors within the contentItem
of a Button, as the Control base-class handles that for us.

Contributes to issue #2331


chriadam:
Honestly, I couldn't reproduce the issue which Philipp reported, either on Chrome or Firefox.  But the Button anchors binding may very well have caused the issue, depending on order-of-evaluation and value clobbering etc.

blammit (approved):

ybott-qinetic (approved):

PR #2658, opened 2025-11-14
Controls: Add informative empty page

### LINKED ISSUE #2563: Improve empty control card when there are no devices with 'quick access controls'
In some cases, the control card on the top-left of the GUI can be empty, which will then look like this:

Image

This feels like a broken UI. In 99% of the systems this won't be the case because there usually is a device with quick controls. Still, it's something which needs to be looked at.

We can of course fix this by just hiding the control card when there's nothing to display. However, Instead of hiding it, maybe it's better to come up with some content to put here to showcase the possibilities of the control card. 

For now, the card can just display a sentence like:

> This is the control card.
> Useful settings of your Victron devices will show up here.

OR

> No control cards yet
> Add an inverter, charger, inverter/charger, EV charging station or genset to control it from here.

Things currently displayed in the control card:

- Relays when configured as manual (This will be removed per 3.70 as they will be migrated into the new switch pane)
- Inverter/charger controls (ESS, mode, grid limit, etc.)
- Genset
- Inverter
- etc.

@philipptrenz fyi

ybott-qinetic:
Adds an empty page to the controls panel if there are no devices that are defined on the system that expose controls.

Fixes #2563


pages/EmptyPageItem.qml
```diff
@@ -0,0 +1,43 @@
...
+    Label {
+        anchors.horizontalCenter: parent.horizontalCenter
+        height: implicitHeight + Theme.geometry_emptyPageItem_title_bottomMargin
+        text: root.titleText
+        font.pixelSize: Theme.font_size_body2
+    }
+
+    Image {
+        anchors.horizontalCenter: parent.horizontalCenter
+        source: root.imageSource
```
blammit: Looking at the Figma, the image does have different colours in dark/light mode, so this needs to be a `CP.ColoredImage` after all, with the `color` set as per the design.

pages/EmptyPageItem.qml
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    Label {
+        anchors.horizontalCenter: parent.horizontalCenter
+        text: root.primaryText
+        font.pixelSize: Theme.font_size_body2
```
blammit: In case the translated text doesn't fit on a single line, set `width: parent.width` and wrap the text. (Same with the other two labels)

pages/EmptyPageItem.qml
```diff
@@ -0,0 +1,43 @@
+/*
+** Copyright (C) 2025 Victron Energy B.V.
+** See LICENSE.txt for license information.
+*/
+
+import QtQuick
+import Victron.VenusOS
+
+Column {
```
blammit: This file appears to be using spaces instead of tab indentation

pages/EmptyPageItem.qml
```diff
@@ -0,0 +1,56 @@
...
+	id: root
+
+	property string titleText
+	property string primaryText
+	property string secondaryText
+	property string imageSource
+	property color imageColor
+
+	width: implicitWidth + Theme.geometry_page_content_horizontalMargin * 2
```
blammit: This would expand the Column if its contents are wider than the implicitWidth, but you want the Column to constrain the size of its children instead.

components/EmptyPageItem.qml
```diff
@@ -0,0 +1,57 @@
+/*
+** Copyright (C) 2025 Victron Energy B.V.
+** See LICENSE.txt for license information.
+*/
+
+import QtQuick
+import QtQuick.Controls.impl as CP
```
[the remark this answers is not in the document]
blammit: Sorry, ignore the comment about the import - I was looking at the wrong file

PR #2659, opened 2025-11-14
Switches: adapt to changes in /ShowUIControl value format

### LINKED ISSUE #2583: Switch pane: Update show controls field
The `/ShowUiControl` path will be updated to an integer value having a value between 0 and 3 so we can specify to show or hide the control separately in local and in the remote UI's.

Tasks:

- [ ] Change the setting in the switchable output setup to a dropdown
- [ ] Implement showing / hiding the control in the switch pane on the UI based on the backend connection type and the setting.

The "Show control" setting should be changed to a dropdown:
Path values:
* 0: Never show the control
* 1: Only show on the local UI
* 2: Only show on the remote UI
* 3: Show on all UI's

blammit:
The /SwitchableOutput/x/Settings/ShowUIControl value has changed from a binary 0 or 1 value to:

0 = Off
1 = Always
0x2 = Only local
0x4 = Only on VRM

Change the UI setting from a toggle switch to a radio button list, and update SwitchableOutput::allowedInGroupModel() to check for the new values that are possible for /ShowUIControl.

Fixes #2583

### DIFF
--- a/pages/settings/devicelist/switchable-outputs/PageSwitchableOutput.qml
+++ b/pages/settings/devicelist/switchable-outputs/PageSwitchableOutput.qml
@@ -110,13 +110,22 @@ Page {
 				defaultSecondaryText: VenusOS.switchableOutput_typeToText(root.switchableOutput.type, root.switchableOutput.outputId)
 			}
 
-			ListSwitch {
+			ListRadioButtonGroup {
 				//: Whether UI controls should be shown for this output
 				//% "Show controls"
 				text: qsTrId("page_switchable_show_controls")
 				dataItem.uid: root.switchableOutput.uid + "/Settings/ShowUIControl"
 				writeAccessLevel: VenusOS.User_AccessType_User
 				preferredVisible: dataItem.valid
+				optionModel: [
+					{ display: CommonWords.off, value: VenusOS.SwitchableOutput_ShowUiControl_Off },
+					//% "Always"
+					{ display: qsTrId("page_switchable_output_show_always"), value: VenusOS.SwitchableOutput_ShowUiControl_Always },
+					//% "Only local"
+					{ display: qsTrId("page_switchable_output_show_local"), value: VenusOS.SwitchableOutput_ShowUiControl_Local },
+					//% "Only on VRM"
+					{ display: qsTrId("page_switchable_output_show_vrm"), value: VenusOS.SwitchableOutput_ShowUiControl_Remote }
+				]
 			}
 
 			ListQuantity {
--- a/src/enums.h
+++ b/src/enums.h
@@ -535,6 +535,14 @@ public:
 	};
 	Q_ENUM(SwitchableOutput_Status)
 
+	enum SwitchableOutput_ShowUiControl {
+		SwitchableOutput_ShowUiControl_Off = 0x0,
+		SwitchableOutput_ShowUiControl_Always = 0x1,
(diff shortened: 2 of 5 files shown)


pages/settings/devicelist/switchable-outputs/PageSwitchableOutput.qml
```diff
@@ -110,13 +110,22 @@ Page {
...
-			ListSwitch {
+			ListRadioButtonGroup {
 				//: Whether UI controls should be shown for this output
 				//% "Show controls"
 				text: qsTrId("page_switchable_show_controls")
 				dataItem.uid: root.switchableOutput.uid + "/Settings/ShowUIControl"
 				writeAccessLevel: VenusOS.User_AccessType_User
 				preferredVisible: dataItem.valid
+				optionModel: [
+					{ display: CommonWords.off, value: VenusOS.SwitchableOutput_ShowUiControl_Off },
```
chriadam: nitpick: I feel like this should be "Never" or "Not shown" rather than "Off"

src/enums.h
```diff
@@ -535,6 +535,14 @@ class Enums : public QObject
 	};
 	Q_ENUM(SwitchableOutput_Status)
 
+	enum SwitchableOutput_ShowUiControl {
+		SwitchableOutput_ShowUiControl_Off = 0,
```
chriadam: style nit: would prefer that all values are specified in hex form if some of them are.
0x0, 0x1, 0x2, 0x4

src/switchableoutput.cpp
```diff
@@ -402,20 +403,50 @@ void SwitchableOutput::updateAllowedInGroupModel()
 	// Output is allowed in the group model if all these conditions are true:
 	// - /Type is valid and matches the /ValidTypes
 	// - /State or /Dimming are valid
-	// - /ShowUIControl is not set, or is set to 1
+	// - /ShowUIControl indicates the control should be shown
```
chriadam: "indicates whether the control should be shown" maybe?

chriadam: /edit - removed incorrect assumption

tests/switchableoutput/tst_switchableoutput.qml
```diff
@@ -333,6 +333,94 @@ TestCase {
...
+			{
+				tag: "ShowUIControl=Local+Remote, remote connection",
+				uid: "mock/com.victronenergy.test.a/SwitchableOutput/1",
+				outputProperties: {
+					"Settings/ShowUIControl": 6, // Local+Remote = 0x2 | 0x4
+					"State": 0, "Settings/Type": 0, "Settings/ValidTypes": 1 << 0
+				},
+				vrm: true,
+				allowedInGroupModel: true,
+			},
```
chriadam: maybe also add a test for 0x5 (i.e. I would expect this value to be considered invalid, the backend should not do this, but we should test that we handle it sanely and presumably assume it is 0x1 and not 0x4)

blammit: Done

src/enums.h
blammit: Done

src/switchableoutput.cpp
blammit: Changed to "indicates that the control should be shown" - using "whether" seems odd as the list is about all the "true" assumptions

chriadam (approved):

PR #2661, opened 2025-11-14
Settings: fix model for VRM instance list

### LINKED ISSUE #2660: VRM instances list is missing some devices or may show large gaps
Go to Settings > VRM > VRM device instances.

In [this](https://vrm.victronenergy.com/installation/133622/dashboard) installation, the list only shows unconnected devices, but there should be at least a GIO Extender, based on the value of com.victronenergy.settings/Settings/Devices.

In [this](https://vrm.victronenergy.com/installation/491097/dashboard) installation, there are large gaps:

Image

blammit:
Migrate the VRM instance list model to C++. This resolves a number of issues. For example:

- ensure all devices from the /Settings/Devices list are shown
- avoid including invalid devices in the list, as they were shown as empty list items
- sort the list correctly

Fixes #2660


src/classandvrminstancemodel.cpp
```diff
@@ -0,0 +1,385 @@
...
+
+
+ClassAndVrmInstanceModel::ClassAndVrmInstanceModel(QObject *parent)
+	: QAbstractListModel(parent)
+{
+	AllServicesModel *servicesModel = AllServicesModel::create();
+	for (int i = 0; i < servicesModel->count(); ++i) {
+		if (servicesModel->data(servicesModel->index(i, 0),
+					AllServicesModel::ServiceTypeRole).toString() == QStringLiteral("settings")) {
+			if (VeQItem *settingsItem = servicesModel->itemAt(i)) {
```
chriadam: should this be `VeQItem *serviceItem = servicesModel->itemAt(i)` ?  Then you can do `VeQItem *settingsItem = serviceItem->itemGet("Settings")` ?  Or am I misunderstanding?

src/classandvrminstancemodel.cpp
```diff
@@ -0,0 +1,385 @@
...
+	return m_item ? m_item->uniqueId() : QString();
+}
+
+void ClassAndVrmInstance::setUid(const QString &uid)
+{
+	if (uid != this->uid()) {
+		if (VeQItem *instanceItem = VeQItems::getRoot()->itemGet(uid)) {
+			initialize(instanceItem, instanceItem->itemParent()->itemGet(QStringLiteral("CustomName")));
+		}
+		emit uidChanged();
```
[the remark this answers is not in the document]
blammit: Fixed to not emit the signal if it didn't call initialize().

src/classandvrminstancemodel.cpp
blammit: It can't fetch the "Settings" item because the name differs on D-Bus ("com.victronenergy.settings") vs MQTT ("settings"). So it has to do this roundabout thing where it checks for a matching service type, and only fetches the VeQItem if the service type matches.

chriadam (approved):
LGTM but hopefully Daniel can give a tick, too

DanielMcInnes (approved):

PR #2663, opened 2025-11-17
Key navigation and scrolling fixes

blammit:


pages/ControlCardsPage.qml
```diff
@@ -36,6 +36,11 @@ Page {
 		spacing: Theme.geometry_controlCardsPage_spacing
 		orientation: ListView.Horizontal
 
+		// Use a velocity that roughly matches the flick velocity produced by AuxCardsPage's
+		// scrollToControl() when scrolling between controls in the Switch Pane.
```
chriadam: I'm confused: why don't we use that "scrollToControl" method also here?  Aren't both cards pages roughly similar in terms of needing scroll-to-focused-item behaviour?  Or am I misunderstanding?

chriadam (approved):

PR #2665, opened 2025-11-18
AC Load: Add PhaseSetting to acload model

### LINKED ISSUE #2554: Acload service: add phase setting for shelly devices
Single channel shelly devices with energy measuring capabilities will show up on dbus as `com.victronenergy.acload`. These devices can only switch and measure on one phase, which is not necessarily phase 1. The phase on which the device is connected should be configurable by the user. 

The shelly service will expose a new path for this: `/PhaseSetting`. If this path is present, a setting 'Phase' should be visible in the setup menu: 

Image

With which the user can select the phase the device is connected on. The GUI should write the phase (integer 1, 2, or 3) to the `/PhaseSetting` path.

Wrt the UI control element, a menu with three options: "Phase 1", "Phase 2", and "Phase 3" is fine with me, but other suggestions are welcome.

ybott-qinetic:
Shelly devices with energy measuring capabilities will show on dbus as an acload. The phase these devices switch and meter can be setup as required. Provide a mechanism to set this metered phase in the device settings.

Fixes #2554

### DIFF
--- a/data/mock/conf/services/em-acload.json
+++ b/data/mock/conf/services/em-acload.json
@@ -53,6 +53,7 @@
         "/N2kUniqueNumber": 435059,
         "/NrOfPhases": 3,
         "/PhaseSequence": 1,
+        "/PhaseSetting": 1,
         "/ProductId": 41393,
         "/ProductName": "Energy Meter VM-3P75CT",
         "/RefreshTime": 1000,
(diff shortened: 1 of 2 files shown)


blammit (commented):
The commit message refers to "AC System", but that is something different that relates to Multi RS devices and such. You could refer to "AC load" instead.

blammit (approved):

PR #2669, opened 2025-11-20
Mock: use correct phase counts when changing page configurations

blammit:
When changing Brief/Overview mock configurations, clear the power/current values for phases that are unused; also, update SystemAcImpl.qml to avoid adding data from phases beyond the desired phase count for a service. Otherwise, the power/current values for unused phases will be picked up by SystemAcImpl, and it will continue to use the old phase count for calculating /Ac/Consumption, /Ac/ConsumptionOnOutput and /Ac/ConsumptionOnInput.

This was causing the AC Loads widget to show the wrong number of phases when moving from a 3-phase to 1-phase configuration, and also causing the Grid widget to include state power/current data in its total when moving from a 3-phase to 1-phase configuration.


chriadam (approved):
Sure, LGTM

PR #2678, opened 2025-11-23
Switch: don't use Qt.binding for page titles

### LINKED ISSUE #2537: Bug: Specifying `CustomName` for a GX IO Device breaks devices navigation
Here is the attached video, latest beta v3.70~41

1. Have a system with GX IO Extender
2. Go to Settings -> Devices -> GX IO Extender -> Device
3. Specify `Custom Name`, Enter

At this moment top level breadcrumb changes and shows wrong path (tank sensor)... going back and deep diving into GX IO entry goes to tank sensor...

A data inconsistency somewhere...

Alternatively the same bug happens when already specified `Custom Name` is deleted, and GX IO transitions back to `Product Name`, as shown at the beginning of the video.

Reload of gui-v2 of course fixes it.

https://github.com/user-attachments/assets/1c6d4168-e015-4a8f-8152-ebf702c8be6f

DanielMcInnes:
Fixes #2537

### DIFF
--- a/pages/settings/devicelist/delegates/DeviceListDelegate_switch.qml
+++ b/pages/settings/devicelist/delegates/DeviceListDelegate_switch.qml
@@ -23,8 +23,7 @@ DeviceListDelegate {
 
 	onClicked: {
 		Global.pageManager.pushPage("/pages/settings/devicelist/PageSwitch.qml", {
-			serviceUid: root.device.serviceUid,
-			title: Qt.binding(function() { return root.text })
+			serviceUid: root.device.serviceUid
 		})
 	}
 
--- a/pages/settings/devicelist/PageSwitch.qml
+++ b/pages/settings/devicelist/PageSwitch.qml
@@ -9,6 +9,16 @@ import Victron.VenusOS
 DevicePage {
 	id: root
 
+	title: 	{
+		if (device.customName) {
+			return device.customName
+		} else if (device.deviceInstance >= 0 && device.productName) {
+			return `${device.productName} (${device.deviceInstance})`
+		} else {
+			return ""
+		}
+	}
+
 	settingsHeader: SettingsColumn {
 		width: parent?.width ?? 0
 		bottomPadding: spacing


pages/settings/devicelist/delegates/DeviceListDelegate_switch.qml
```diff
@@ -23,8 +23,7 @@ DeviceListDelegate {
 
 	onClicked: {
 		Global.pageManager.pushPage("/pages/settings/devicelist/PageSwitch.qml", {
-			serviceUid: root.device.serviceUid,
-			title: Qt.binding(function() { return root.text })
+			serviceUid: root.device.serviceUid
```
DanielMcInnes: @blammit do you remember why we had: `title: Qt.binding(function() { return root.text })`? I expect there was a good reason at the time

chriadam:
We previously used `"title": Qt.binding(function() { return root.text })` instead of just `"title": root.text` because otherwise it wouldn't automatically update when changing the system language.

I agree that we should have code in each page which binds the title manually.

As for why it causes a problem - is it possible that after changing the custom name, the device itself gets regenerated, causing the previous evaluation context to be deleted?  (Hmm, but then I'd expect the page hierarchy to be invalid also ... we'd need to pop back to the device list... and it seems this isn't necessary, so it probably isn't that?)

blammit:
Discussed this further with Daniel today. With the latest `main`:

1. Set a custom name for the GIO Extender; the breadcrumb changes to the new custom name.
2. Now delete the custom name; the breadcrumb text stays the same. It does not revert to the product name as expected, but it also does not change to the name of some other random service.

If an `onTitleChanged` signal handler is placed in `PageSwitch.qml`, it is triggered by step 1, but not by step 2, which indicates the `title` binding provided by `DeviceListDelegate_switch.qml` is not updating at that point.

In regards to the fix in this PR, it's not ideal that we now have duplicated definitions of the `title`, but we need to move in the direction of defining page titles within their pages (rather than in the previous page) anyway. So I think it's fine for this to go in.

PR #2688, opened 2025-12-01
Genset: check GensetServiceType when matching start/stop generator

### LINKED ISSUE #2686: Same generator auto start instance is shown for both DC and AC genset instances
Applies to: build-in display only
Seen on: gui-v2 v1.2.21

When there is a DC and a AC genset present, they both show the same generator auto start instance.
The WASM build does show only one generator auto start instance.

The `com.victronenergy.generator.startstop1` service is linked the `genset` type. 
So, the generator auto start should not be shown for the DC genset instance (=`dcgenset`).
```
GensetInstance                                                                                                                                            0
GensetProductId                                                                                                                                       45120
GensetService                                                                                           com.victronenergy.genset.socketcan_vecan0_vi0_uc456
GensetServiceType                                                                                                                                    genset
```

This is the AC genset
Image

This is the DC genset
Image

blammit:
If there is both a DC and AC genset present, they could both have device instances that match the generator startstop1 /GensetInstance value, but only one of them is the actual startstop1 generator. Check the service type matches /GensetServiceType as well, to differentiate between DC genset (serviceType=dcgenset) and AC genset (serviceType=genset).

Fixes #2686

### DIFF
--- a/components/PageGensetModel.qml
+++ b/components/PageGensetModel.qml
@@ -19,15 +19,23 @@ VisibleItemModel {
 															  : 3
 
 	// In case of multiple gensets, startstop will control the one with the lowest device instance.
-	// Check if this genset is controlled by startstop by checking if the instance is the same.
+	// Check if this genset is controlled by startstop by checking if the instance and service type
+	// are the same.
 	// When not controlled by startstop, the genset can only be monitored, so hide some controls.
-	readonly property bool isStartStopControlled: startStopGensetInstance.valid && gensetInstance.valid ?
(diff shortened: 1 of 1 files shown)


blammit:
@martinbosma could you verify whether this fixes the issue on your GX?

Also, can you confirm it is safe to assume that `GensetServiceType` is always set, if `GensetInstance` is set?

blammit:
@chriadam could you review this one please?

chriadam (approved):

PR #2698, opened 2025-12-04
VE.Can device connection status & N2K-out alert enable

martinbosma:


martinbosma:
This is now the online/offline status looks like on the devices page:
image

And the device page (device instance editing not allowed when offline):
image
image

And the switch for NMEA2000 outbound alerts:
image
Hidden if NMEA2000-out is disabled:
image

Issue #2696, opened 2025-12-04
DC Loads drill-down improvements


philipptrenz:
Some improvements as a follow-up of #2311:

1. Hide the status of `com.victronenergy.dcdc` service, as it's only showing a static configuration setting (it will always show "Power supply mode", see screenshot)
2. Add `com.victronenergy.motordrive` service to be shown in the DC Loads drill-down, which wasn't clearly enough  specified before, I assume

<img width="890" height="529" alt="Image" src="https://github.com/user-attachments/assets/77ac60e5-dc18-476a-bd44-357ffb739663" />

ybott-qinetic:
@philipptrenz For issue 1., is this intentional? Should the dcdc service be returning something meaningful? The System_State enum has two dozen different states available. https://github.com/victronenergy/venus/wiki/dbus#dcdc-converters

PR #2699, opened 2025-12-05
DC Load: add motordrive to DC Load device list

ybott-qinetic:
Contributes to #2696


blammit (approved):
LGTM but please change "Fixes #" to "Contributes to #" so that it doesn't auto-close the task, as the other point in the task requires more investigation.

Issue #2685, opened 2025-11-27
Microgrid parameters: Make sure that all values are always shown with two decimals


philipptrenz:
All parameter values of all three modes within the "Microgrid" submenu must be rounded to two decimal places at all times, even if the decimals are "00".

Current state:
<img width="1371" height="799" alt="Image" src="https://github.com/user-attachments/assets/399b262e-5f96-42a0-88e1-66525db4626c" />

chriadam:
Merged to main, closing as fixed.

PR #2705, opened 2025-12-08
Microgrid: display parameters with 2 decimals of precision

chriadam:
The default units display logic previously clipped the precision for displaying values using a particular algorithm, even if the user provided a specific precision.

This commit adds a new argument "clipPrecision" which defaults to true (to maintain behaviour for existing use-cases) which is set to false for the Microgrid page parameters, to ensure that the values there are displayed with exactly 2 decimals.

Contributes to issue #2685

### DIFF
--- a/components/QuantityLabel.qml
+++ b/components/QuantityLabel.qml
@@ -20,6 +20,7 @@ Item {
 	property alias unitText: unitLabel.text
 	property int alignment: Qt.AlignHCenter
 	property alias precision: quantityInfo.precision
+	property alias precisionAdjustmentAllowed: quantityInfo.precisionAdjustmentAllowed
 	property alias formatHints: quantityInfo.formatHints
 	property alias leftPadding: digitRow.leftPadding
 	property alias rightPadding: digitRow.rightPadding
--- a/components/listitems/core/ListQuantity.qml
+++ b/components/listitems/core/ListQuantity.qml
@@ -15,6 +15,7 @@ ListItem {
 	property alias unit: quantityLabel.unit
 	property alias unitColor: quantityLabel.unitColor
 	property alias precision: quantityLabel.precision
+	property alias precisionAdjustmentAllowed: quantityLabel.precisionAdjustmentAllowed
 
 	content.children: [
 		QuantityLabel {
--- a/pages/vebusdevice/PageMicrogrid.qml
+++ b/pages/vebusdevice/PageMicrogrid.qml
@@ -23,6 +23,7 @@ Page {
 	component MicrogridListQuantity: ListQuantity {
 		textFormat: Text.RichText
 		precision: 2
+		precisionAdjustmentAllowed: false
 	}
 
 	component ListValueRange: ListText {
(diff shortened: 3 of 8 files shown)


blammit:
@chriadam would be worthwhile to check whether we can just change the default behaviour to always show the requested number of decimals. If so then we could avoid adding this extra `clipPrecision` configuration.

chriadam:
> @chriadam would be worthwhile to check whether we can just change the default behaviour to always show the requested number of decimals. If so then we could avoid adding this extra `clipPrecision` configuration.

Good point.

@philipptrenz maybe you could help with this: currently in main branch (presumably for historical reasons) we reduce the precision (i.e. number of decimal points) with which we display numbers in two circumstances:

1) if the value has been automatically scaled from one unit to another (e.g. from kW to W), the algorithm may reduce the precision (decimals) displayed, depending on the number of non-decimal digits in the scaled value and the original precision specified for the non-scaled value.  (I assume it is because if the code specifies a label for a quantity with initial precision=2 and units = kW, expecting to show values like 12.65 kW, if the value drops below 1.0 kW (e.g. to 0.32 kW) we might display it Watts instead, and in that case, we would want to trim the number of decimals to zero (displaying e.g. 326 W instead of 326.13 W) - as in this case, keeping two decimals of precision doesn't make sense because a meter which measures in kW with 2 decimals of precision simply does not have sub-watt precision, so those "extra" decimals which might be shown in the Watts-scaled string are likely to be misleading.)  It also tries to "maintain" a maximum of 4 total digits for a post-scaled number, so even if you have a meter which measures Volts with 2-decimals of precision, if the value gets large enough that we automatically scale it to kV we try to maintain just 4 total digits (e.g. when the value steps from 999.99 V to 1000.15 V triggering the scaling logic, we do the conversion to kV (1.00015 kV) and then trim to 4 total digits (1.000 kV)).

2) if the value's natural units is kilowatt-hours (kWH) (that is, if it has NOT been scaled up from WH to kWH due to being a large number), we trim the precision to zero decimals places. (Just a guess, but possibly this is because it was assumed that meters which measure in increments of kWH aren't precise enough to offer any decimals of precision, even if the user asks for them?  Not sure.)

For (1) perhaps it is the case that this should only apply when scaling DOWN whereas it's possible (I haven't confirmed) that this might also be the case when scaling UP (in which case, keeping (or even adding more!) precision would probably be useful).

--

@blammit @DanielMcInnes  When looking into this, one thing jumped out at me: when I ported it from JS to C++ I think I might have made a mistake - as I believe this scaling / precision logic should ONLY apply when we are sure that scaling has actually occurred, but in the code I think we trigger the precision logic even if quantity.scale is the same as previousScale!!  That might be the source of this bug.

There are a few possible paths forward:

a) I just fix the bug above, to prevent precision trimming in non-scaled cases.
b) I fix the bug above, but also provide the "allowClipPrecision" parameter (which prevents the units code from messing with the number of decimals according to the algorithm I described above) so that developers can specify scaling-agnostic precision requirements for specific fields, going forward, as needed.
c) don't fix the bug above, but instead remove all precision-adjustment altogether (this is what Bea suggested above, but I suspect it's because she wasn't aware of the bug I mentioned just above).

Obviously, I favour either (a) or (b), and I personally do kind of prefer option (b) but I don't have a strong argument for it, assuming that fixing (a) resolves the issue here.

chriadam:
Done, separated into two PRs.  See https://github.com/victronenergy/gui-v2/pull/2717 for the other one.

Issue #2617, opened 2025-11-06
Hiding AC loads on DC-only systems


philipptrenz:
This is a subtask of #2570 for doing the first step: Hiding any AC Loads if they are not available. 

From Venus OS v3.70~53, there is a dedicated `/Ac/HasAcLoads` path, which indicates if AC Loads should be shown (1) or not (0). This applies to all relevant pages: Boat, Brief and Overview.

The mechanism behind this setting is that AC Loads and Essential Loads are to be shown if there is at least one of the following:
- Inverter
- Inverter/Charger
- EV Charger
- AC Load
- Heatpump

chriadam:
I recall some discussion in the meeting that the API might not suffice (or might have to be changed in future, to accommodate the future requirements in #2570 ) and so it might not be desirable to implement this subtask currently.  Did I understand correctly, or did I remember wrongly?

chriadam:
I've now updated the PR - I think it should resolve everything properly, now.

PR #2707, opened 2025-12-08
Overview: hide AC Loads widget if HasAcLoads is false

chriadam:
Also hide AC Loads related gauges on the Boat and Brief pages.

Contributes to issue #2617

### DIFF
--- a/data/mock/SystemAcImpl.qml
+++ b/data/mock/SystemAcImpl.qml
@@ -10,17 +10,18 @@ Item {
 	id: root
 
 	function setSystemValue(path, value) {
-		MockManager.setValue("com.victronenergy.system" + path, value)
+		MockManager.setValue(Global.system.serviceUid + path, value)
 	}
+
 	function systemValue(path) {
-		return MockManager.value("com.victronenergy.system" + path)
+		return MockManager.value(Global.system.serviceUid + path)
 	}
 
 	function setGaugesValue(path, value) {
 		MockManager.setValue(Global.systemSettings.serviceUid + "/Settings/Gui/Gauges" + path, value)
 	}
 	function gaugesValue(path) {
-		return MockManager.value("com.victronenergy.settings/Settings/Gui/Gauges" + path)
+		return MockManager.value(Global.systemSettings.serviceUid + "/Settings/Gui/Gauges" + path)
 	}
 
 	VeQuickItem {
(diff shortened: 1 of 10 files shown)


pages/OverviewPage.qml
```diff
@@ -344,7 +344,10 @@ SwipeViewPage {
 	}
 
 	function _resetRightWidgets() {
-		let widgets = [acLoadsWidget]
+		let widgets = []
+		if (Global.system.hasAcLoads || Global.evChargers.model.count > 0) {
+			widgets.push(acLoadsWidget)
```
blammit: I think you'll need to update the `layoutConditions` object to consider the `hasAcLoads` condition. For reference, see the `showDcLoads` property in that object and how it's used in OverviewPage to control whether the DC loads box is shown.

data/mock/SystemAcImpl.qml
```diff
@@ -306,4 +309,12 @@ Item {
 			}
 		}
 	}
+
+	FilteredServiceModel {
+		id: hasAcLoadsModel
+		serviceTypes: ["inverter", "charger", "evcharger", "acload", "heatpump"]
```
blammit: Also "vebus" and "acsystem"?

Or maybe you could just refer to the count for `Global.inverterChargers.model` instead, which has `serviceTypes: ["vebus", "acsystem", "inverter", "charger"]` since those are the services that are accessible via the Inverter/Charger drilldown.

pages/OverviewPage.qml
chriadam: Thanks, done

data/mock/SystemAcImpl.qml
chriadam: These were the specific ones mentioned by Philipp in the task, but I can update if you are sure.

blammit: @philipptrenz can you confirm please? When you said "inverter/charger" I assumed you meant that included `vebus` and `acsystem` services.

pages/OverviewPage.qml
```diff
@@ -528,6 +531,10 @@ SwipeViewPage {
 		readonly property int pvInverterCount: Global.solarInputs?.pvInverterDevices.count ?? 0
 		onPvInverterCountChanged: Qt.callLater(root._resetWidgets)
 
+		// Affects whether AcLoadsWidget is shown.
+		readonly property bool showAcLoads: Global.system?.hasAcLoads || Global.evChargers?.model.count > 0
```
blammit: I thought the system `/Ac/HasAcLoads` value would be automatically set to 1 by system calc if `evcharger` services were present, so that gui-v2 wouldn't have to check that. @philipptrenz could you confirm please?

PR #2710, opened 2025-12-09
VRM: always prompt user to reboot after VRM device instance changes

### LINKED ISSUE #2712: Always show "Reboot now?" popup after changing a VRM device instance
This popup does get shown when leaving the page via the "back" chevron and clicking "VRM" in the breadcrumbs, but not when clicking on "Settings". This leads to inconsistent behaviour, as the popup should always be shown after doing a modification to a VRM device instance. 

Please also make sure that it is shown when navigating via the keyboard.

Image

blammit:
The VRM Device Instances page should show a reboot prompt dialog if the page is popped after a VRM device instance is changed. There were two bugs here:

1. The dialog was not shown if the user changed an instance then immediately clicked a breadcrumb to go to the previous page, without first pressing Enter in the text field to "accept". This was because the new instance had not yet been written to the backend, and thus hasVrmInstanceChanges() unexpectedly returned false. Fix this by changing hasVrmInstanceChanges() to return true if a pending
change has not yet been written.

2. It was also not shown if the user clicked the first breadcrumb ("Settings") instead of the second "VRM" breadcrumb. This was because PageStack.qml only called the tryPop() function in the single-page-pop case, but not when popping all pages. Fix this by calling tryPop() from popAllPages(), and pass the destination page to tryPop() so that the VRM Device Instances page can pop to the correct page when the reboot prompt dialog is cancelled.

Fixes #2712

### DIFF
--- a/components/Page.qml
+++ b/components/Page.qml
@@ -21,7 +21,10 @@ FocusScope {
 	property int topLeftButton: VenusOS.StatusBar_LeftButton_None
 	property int topRightButton: VenusOS.StatusBar_RightButton_None
 
-	property var tryPop // optional function: returns whether the page can be poppped
+	// Optional function that is called when the stack is about to pop this page. Return true if
+	// the page can be popped, or false to deny it and remain on the page.
+	// Takes one argument: the page to which the stack will be popped (null if popping all pages)
+	property var tryPop
 
 	readonly property bool __is_venus_gui_page__: true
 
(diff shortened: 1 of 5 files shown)


citolen:
@chriadam Unfortunately this change breaks the pop action of most `ListRadioButtonGroup`, it now pops all the pages upon selecting an option.

There is always the option the add `popDestination` to all those instances, but it feels like the default behavior shouldn't be changed and when `popDestination` isn't defined it should pop by one, not all.

See:

https://github.com/user-attachments/assets/f3102b90-330e-4d50-8f86-b6a8265c0f8f

PR #2714, opened 2025-12-09
MotorDrive: consolidate device list page

citolen:
- Reorder MotorDrive page's items.
- Add Torque and Direction.

The order was consolidated by importance and in relevance with the boat page order of importance.
The DC values were consolidated into one list item.

Screenshot 2025-12-09 at 17 06 28

- Add Power to device list motordrive delegate.

Screenshot 2025-12-09 at 17 10 19


blammit (approved):

citolen:
> @citolen Looks good to me - okay if we merge this PR?

Yes, let's merge it. Thanks

PR #2715, opened 2025-12-10
VE.Bus: show phase tables on split-phase systems

### LINKED ISSUE #2677: Vebus: Quantity table missing on split phase systems
Venus OS V.3.66

On single phase and three phase systems the ACin and ACout quantities are available in the VE.Bus device overview.  These tables are missing on split phase setups.

Three phase:
Image

Split phase: 
Image

blammit:
Show the multi-phase table when there are two phases, and not only when there are three phases.

Also add a mock configuration with a MultiPlus II that has the split phase setup, for more convenient testing.

Fixes #2677


blammit:
The `split-phase` mock configuration looks like this: (it's pretty nice that we only need the `vebus` service to reproduce all of this)

image
image

and the Inverter/Charger Overview drilldown now contains the phase table, with the two phases:

image

Issue #2025, opened 2025-04-03
increase digits for display of solar yield in watts on overview screen


YetAnotherOneGuru:
The v1 GUI displays at least 5 digits, e.g. 11523 W in the "PV charger" section.
The v2 GUI displays 4 digits alright, e.g. 8543W in the "Solar yield" section of the Overview screen. Once the Power reaches 10000W, it displays "10kW" instead.  I use the v2 GUI from a browser connected via LAN.
It looks like the space on the screen can also handle 5 digit Watts, can this be adjust?
It would by also fine, if it is possible to add a comma and 2 digits after the comma, e.g. "10.34 kW" - even 1 digit after the comma would already help.

It is rather annoying for me, because I want to see the details, when testing new settings on the devices.
I assume, I am not the only one..

chriadam:
Pushed back to v1.3.0 milestone.  @philipptrenz I guess this needs discussion with you and Serj.

chriadam:
Yeah, this will be fixed as part of #3002.  Closing as duplicate.

PR #2717, opened 2025-12-10
Units: modify displayed precision adjustment logic

chriadam:
Previously, when the units of a value was scaled up, we did not increase the displayed precision.  Thus, a value with precision 0 such as 12345 W would be scaled up to 12.345 kW and then trimmed to 12 kW due to the specified precision.

This commit adjusts the logic to allow increasing the precision so long as the total number of digits is less than 4. That is, 12345 W will now be displayed as 12.35 kW.

Previously, the decimal point character seemed to have been considered to take up a "digit" when calculating whether to trim the precision in some cases (e.g. for precision 2 values with 2 non-decimal digits, precision was trimmed to just 1).

This commit removes that consideration, which allows removal of some unnecessary special-casing for Volts unit values.

Contributes to issue #2705

### DIFF
--- a/components/QuantityLabel.qml
+++ b/components/QuantityLabel.qml
@@ -20,6 +20,7 @@ Item {
 	property alias unitText: unitLabel.text
 	property int alignment: Qt.AlignHCenter
 	property alias precision: quantityInfo.precision
+	property alias precisionAdjustmentAllowed: quantityInfo.precisionAdjustmentAllowed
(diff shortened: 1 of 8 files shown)


chriadam:
See also: https://github.com/victronenergy/gui-v2/issues/2025#issuecomment-3495467201

blammit:
Obsoleted by https://github.com/victronenergy/gui-v2/pull/3064.

Issue #2724, opened 2025-12-12
Prefix energy meter power factor readings with PF


martinbosma:
It is not always clear to customers what the last column of the energy meter readings are, since the power factor does not have a unit.
Perhaps we could prefix with `PF`, e.g. `PF 0.987`.

<img width="2792" height="1694" alt="Image" src="https://github.com/user-attachments/assets/a0b0e409-e50b-4d76-abe0-89986547640b" />

chriadam:
PR: https://github.com/victronenergy/gui-v2/pull/2728

chriadam:
Per discussion in meeting, Matthijs regards this as high prio, especially since it's a simple fix.  @DanielMcInnes could you test/review my PR so we can get it into the release tomorrow, if possible?  Thanks!

PR #2728, opened 2025-12-15
Units: display a "unit" prefix for PowerFactor values

chriadam:
Power factor is a unitless ratio value, but displaying it as a unitless number in tables is confusing for users who don't know what the number is supposed to represent.

This commit ensures that we display power factor values with a "PF" prefix unit within tables.

Contributes to issue #2724

### DIFF
--- a/components/QuantityRow.qml
+++ b/components/QuantityRow.qml
@@ -57,9 +57,14 @@ Row {
 			value: quantityObject.numberValue
 			unit: quantityObject.unit
 			precision: quantityObject.precision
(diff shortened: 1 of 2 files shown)


chriadam:
Simplified so that now we only do the PF prefixing specifically within the QuantityRow (i.e. table) code, and not in the base QuantityLabel, to avoid paying the cost of more complex bindings in the 99% use case.

Can be tested in mock mode by clicking the AC Loads widget in overview, then clicking an Energy Meter device.

PR #2729, opened 2025-12-15
MotorDrive: dual-drive settings

citolen:
This is the ground work towards dual-drive support.

This PR has:
- A new link on the display & appearance page that goes to a new page for the boat page settings.
- A new page. Boat page settings.
- A refactor of the motordrive model used in the boat pages to introduce the concept of single, left and right drives.

## New link

Image

## New page

Boat page settings.
This is where the user can now configure whether the boat page is enabled and if their system has more than one motordrive they will be able to configure their dual-drive setup. Assigning which drive is left and which is right.

One key requirement was the ability to still see the configuration if the drives are powered off.
The device names will have some fallback logic in this order:
- `<product name> [<serial number>]` (if connected and serial number available)
- `<product name>` (if connected)
- `E-drive with VRM instance #<vrm instance>` (if disconnected)

Image

The options are sorted by VRM instance ID.
There is some logic preventing the user from selecting the same drive for the left and right drives. Hence the `None` option, to be able to swap the drives.

Image

## MotorDrive model and components

The `MotorDrive` model was replaced with the `MotorDrives` model which adds support for up to two drives.
This is relying on an upcoming systemcalc change that will add `com.victronenergy.system/MotorDrive/0/DeviceInstance` and `com.victronenergy.system/MotorDrive/1/DeviceInstance`.

### DIFF
--- a/data/mock/page-conf/BoatPageConfig.qml
+++ b/data/mock/page-conf/BoatPageConfig.qml
@@ -13,35 +13,73 @@ Item {
 
 	readonly property var configs: [
 		{
-			name: "Motordrive, gps, km/h",
-			gps: { speedUnit: "km/h"},
-			motorDrive: {}
+			name: "GPS, no motordrives",
+			gps: {},
 		},
 		{
-			name: "Motordrive, gps, m/s",
-			gps: { speedUnit: "m/s"},
-			motorDrive: {}
+			name: "GPS, single-drive",
+			gps: {},
+			motorDrives: [
+				{ productName: "Motor drive A", power: 1000 },
+			],
 		},
 		{
-			name: "Motordrive, gps, kt",
-			gps: { speedUnit: "kt"},
-			motorDrive: {}
+			name: "GPS, single-drive, time to go",
+			gps: {},
+			motorDrives: [
+				{ productName: "Motor drive A", power: 1000 },
+			],
+			timeToGo: 123456
 		},
 		{
-			name: "Motordrive, gps, mph",
-			gps: { speedUnit: "mph"},
-			motorDrive: {}
+			name: "GPS, single-drive, time to go, regeneration",
+			gps: {},
+			motorDrives: [
+				{ productName: "Motor drive A", power: -1000 },
+			],
+			timeToGo: 123456
 		},
 		{
-			name: "Motordrive, no gps",
-			motorDrive: {}
+			name: "GPS, dual-drive",
+			gps: {},
+			motorDrives: [
+				{ productName: "Motor drive A", power: 1000 },
+				{ productName: "Motor drive B", power: 1000 },
+			],
 		},
 		{
-			name: "No motordrive, gps",
-			gps: { speedUnit: "kt"},
+			name: "GPS, dual-drive, time to go",
+			gps: {},
+			motorDrives: [
+				{ productName: "Motor drive A", power: 1000 },
+				{ productName: "Motor drive B", power: 1000 },
+			],
+			timeToGo: 123456
 		},
 		{
-			name: "No motordrive, no gps"
+			name: "GPS, dual-drive, time to go, regeneration",
+			gps: {},
+			motorDrives: [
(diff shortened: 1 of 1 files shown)


citolen:
> this is aimed for release in gui-v2 v1.3.x series, not v1.2.x series, correct?

No, as I understood it we are aiming to release dual-motor support for v1.2.x, if all the other pieces come together in time of course, but it's on the right track.

This doesn't have to merged right now though, I just wanted to break it down into small enough pieces to review.

pages/boat/LargeCenterGauge.qml
```diff
@@ -10,17 +10,24 @@ Item {
...
 
-	readonly property VeQuickItemsQuotient motorDriveDcConsumption: root.motorDrive ? root.motorDrive.dcConsumption.quotient : null
+	readonly property VeQuickItemsQuotient motorDriveDcConsumption: root.motorDrives.dcConsumption.quotient
 	readonly property VeQuickItemsQuotient activeDataSource: root.gps.valid ? root.gps
 			: root.motorDriveDcConsumption.valid ? root.motorDriveDcConsumption
 			: null
 
 	objectName: "LargeCenterGauge"
 
+	readonly property VeQuickItemsQuotient rpm: VeQuickItemsQuotient {
```
DanielMcInnes: nit - it feels to me like `rpm` belongs in `MotorDrives`, what do you think?

citolen: Yes it could, if we do that then we'll need `rpm`, `leftRpm` and `rightRpm`.
We'll probably want to do the same for all the other values read (direction, temperature, etc...) to remain consistent which makes me question whether it's a good idea, it's just going to clog up `MotorDrives`.

The only value held in `MotorDrives` right now is `power`/`dcConsumption`, but this is overall/system power.

Not sure, as long as we do it in a consistent way.
I will look into making another object to represent a single motordrive and add the values in there.
This way we could access `motorDrives.singleDrive.rpm` or `motorDrives.leftDrive.rpm`, etc...

chriadam:
> > this is aimed for release in gui-v2 v1.3.x series, not v1.2.x series, correct?
> 
> No, as I understood it we are aiming to release dual-motor support for v1.2.x, if all the other pieces come together in time of course, but it's on the right track.
> 
> This doesn't have to merged right now though, I just wanted to break it down into small enough pieces to review.

Ok.  I will be in the office on Monday+Tuesday next week, but after that I will be away until after New Year.  I think most likely I'll be back on the 5th, since the 2nd is a Friday, I believe.  I will try to review this early next week, but if I don't get a chance, it will likely be early next year.  Let's see how we go.

data/mock/MotorDrivesImpl.qml
```diff
@@ -42,6 +42,13 @@ Item {
 				}
 			}
 
+			VeQuickItem {
+				uid: root.motorDriveUid + "/DeviceInstance"
+				onValueChanged: {
+					root.setSystemValue("/MotorDrive/0/DeviceInstance", value ?? 0)
```
chriadam: should this just be `/MotorDrive/DeviceInstance` in order to be symmetric with what is done for `/MotorDrive/Power` above?

pages/boat/BoatPage.qml
```diff
@@ -139,12 +139,12 @@ SwipeViewPage {
...
 		animationEnabled: root.animationEnabled
-		motorDrive: _motorDrive
+		motorDrives: _motorDrives
 		gps: _gps
 	}
 
-	Boat.MotorDrive {
-		id: _motorDrive
+	Boat.MotorDrives {
+		id: _motorDrives
```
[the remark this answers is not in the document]
chriadam: Not in this PR, but we should make a note to do it in v1.3.x

chriadam (approved):
LGTM, although I haven't done any non-mock-mode testing.

DanielMcInnes (approved):
Code changes lgtm. Not tested on a real installation.

PR #2740, opened 2025-12-18
Switches: improve key navigation for color switches

chriadam:
Also fix a minor bug in the color swatch radius.

Contributes to issue #2739


chriadam:
Updated handling of left/right and up/down improved to ensure that we eat up/down events only when the color button box has focus, and that we clear all internal focus when navigating away from the delegate (e.g. via left/right or tab etc).

PR #2751, opened 2026-01-06
Settings: add "Function" dropdown option to allow user to set relay purpose

### LINKED ISSUE #2731: Switch configuration page: Add "Function" dropdown
Currently, the switch configuration page does not allow setting a function, because almost all switchable outputs only support the function "Manual". However, there are some which support other functions as well, and the user should be able to set it.

Add a "Function" dropdown right below the "Type" dropdown:

Image

The mechanism is the same as for the type:

`/SwitchableOutput/x/Settings/ValidFunctions` holds a bitfield of the supported functions. Only show the functions in the dropdown that are supported. The possible functions are:

```
0: Alarm
1: Generator start/stop
2: Manual, 
3: Tank pump
4: Temperature
5: Genset helper relay
6: Opportunity load
```
And use the value of `../ValidFunctions` to only show the ones that are supported. Also described on the [wiki](https://github.com/victronenergy/venus/wiki/dbus#switch)

Upon selecting a function, write the value of that function to `/SwitchableOutput/x/Settings/Function`.
Most switchable outputs currently only support the function "Manual". If this is the case (so `../ValidFunctions` == 4), hide the dropdown.

ybott-qinetic:
Add a "Function" to switchable output that defines the purpose of the relay.
Only display function dropdown when type is set to toggle switch and when the
valid functions have options other than manual. The list of valid functions
from the settings is used to allow user selection.

Use existing Relay_Function options; add opportunity load to list of
available purposes.

Add various examples to smartswitch, they are only accessible when setting
the switch type value to toggle. Examples include error states and intentionally
limiting display, in the case of manual function or non toggle type switch.

Fixes #2731


src/switchableoutput.cpp
```diff
@@ -70,6 +70,14 @@ void SwitchableOutput::initialize(VeQItem *outputItem)
 		m_validTypesItem = validTypesItem;
 		connect(validTypesItem, &VeQItem::valueChanged, this, &SwitchableOutput::setValidTypes);
 	}
+	if (VeQItem *functionItem = m_outputItem->itemGetOrCreate(QStringLiteral("Settings/Function"))) {
+		m_relayFunctionItem = functionItem;
```
blammit: Similarly, rename this from `m_relayFunctionItem` to `m_functionItem` as it is now used for all switchable outputs, not just for those that represent system relays.

pages/settings/devicelist/switchable-outputs/PageSwitchableOutput.qml
```diff
@@ -110,6 +110,31 @@ Page {
 				defaultSecondaryText: VenusOS.switchableOutput_typeToText(root.switchableOutput.type, root.switchableOutput.outputId)
 			}
 
+			ListRadioButtonGroup {
+				//% "Function"
+				text: qsTrId("page_switchable_output_function")
+				dataItem.uid: root.switchableOutput.uid + "/Settings/Function"
+				preferredVisible: dataItem.valid
+						&& (root.switchableOutput.type === VenusOS.SwitchableOutput_Type_Toggle)
```
blammit: I don't think this `root.switchableOutput.type` condition is supposed to be here - I had to remove it to get the list item to appear.

pages/settings/devicelist/switchable-outputs/PageSwitchableOutput.qml
```diff
@@ -110,6 +110,30 @@ Page {
...
+			ListRadioButtonGroup {
+				//% "Function"
+				text: qsTrId("page_switchable_output_function")
+				dataItem.uid: root.switchableOutput.uid + "/Settings/Function"
+				preferredVisible: dataItem.valid
+						&& (root.switchableOutput.validFunctions !== (1 << VenusOS.SwitchableOutput_Function_Manual))
+				secondaryLabel.color: root.switchableOutput.hasValidFunction ? Theme.color_listItem_secondaryText : Theme.color_critical
+				optionModel: {
+					let options = []
+					for (let i = 0; i <= VenusOS.SwitchableOutput_Function_MaxSupportedType; i++) {
```
chriadam: I note that the minimum enum value is -1 (i.e. disabled) - is it intentional to start from 0 instead?  Maybe better to use `let i = VenusOS.SwitchableOutput_Function_Alarm; ...` instead for clarity?

src/switchableoutput.cpp
```diff
@@ -396,6 +437,19 @@ void SwitchableOutput::updateHasValidType()
 	}
 }
 
+void SwitchableOutput::updateHasValidFunction()
+{
+	const int functionInt = function();
+	const bool hasValidFunction = functionInt >= Enums::SwitchableOutput_Function_Alarm
```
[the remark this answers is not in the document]
ybott-qinetic: Done

PR #2753, opened 2026-01-07
Units: refactor unit of measurement storage and lookup

### LINKED ISSUE #2708: units.cpp should be refactored to avoid large switch statements
Could define the appropriate values in a lookup table:

```
static constexpr UnitMetadata UnitTable[] {
    // unit                                         label    defaultPrec scalable  veUnit
    { Enums::Units_None,                           "",        0,        false,    Unit::Default },

    { Enums::Units_Watt,                           "W",       0,        true,     Unit::Watt },
    { Enums::Units_Volt_DC,                        "V",       2,        true,     Unit::Volt },
    { Enums::Units_Volt_AC,                        "V",       0,        true,     Unit::Volt },
    ...
};
```

ybott-qinetic:
Fixes #2708


src/units.cpp
```diff
@@ -4,49 +4,71 @@
...
+	Victron::VenusOS::Enums::Units_Type unit;
+	std::string label;
+	int precision;
+	bool scalable;
+	Unit::Type veUnit;
+};
+
+static UnitMetadata UnitTable[] {
+	// unit                                                                                      label    defaultPrec scalable  veUnit
+	{ Victron::VenusOS::Enums::Units_None,                                                       "",      1,          false,    Unit::Default },
```
blammit: I'd use `0` as the default precision for Units_None to avoid confusion. Even if it's a behavioural change it is never used with a precision anyway.

src/units.cpp
```diff
@@ -4,49 +4,71 @@
...
+	{ Victron::VenusOS::Enums::Units_Kilopascal,                                              "kPa",      0,          false,    Unit::Default },
+	{ Victron::VenusOS::Enums::Units_CardinalDirection,                 DegreesSymbol.toStdString(),      0,          false,    Unit::Default },
+	{ Victron::VenusOS::Enums::Units_PowerFactor,                                                "",      3,          false,    Unit::Default },
+	{ Victron::VenusOS::Enums::Units_Time_Day,                                                  "d",      0,          false,    Unit::Default },
+	{ Victron::VenusOS::Enums::Units_Time_Hour,                                                 "h",      0,          false,    Unit::Default },
+	{ Victron::VenusOS::Enums::Units_Time_Minute,                                               "m",      0,          false,    Unit::Default },
+	{ Victron::VenusOS::Enums::Units_Altitude_Metre,                                            "m",      0,          true,     Unit::Metre },
+	{ Victron::VenusOS::Enums::Units_Altitude_Foot,                                            "ft",      0,          false,    Unit::Kelvin },
```
blammit: `Unit::Kelvin` -> `Unit::Foot`

src/units.cpp
```diff
@@ -4,49 +4,71 @@
...
+	{ Victron::VenusOS::Enums::Units_VoltAmpere,                                               "VA",      1,          true,     Unit::Default },
+	{ Victron::VenusOS::Enums::Units_VoltAmpereReactive,                                      "var",      1,          true,     Unit::Default },
+	{ Victron::VenusOS::Enums::Units_Watt,                                                      "W",      0,          true,     Unit::Default },
+	{ Victron::VenusOS::Enums::Units_Amp,                                                       "A",      1,          true,     Unit::Default },
+	{ Victron::VenusOS::Enums::Units_Hertz,                                                    "Hz",      1,          true,     Unit::Default },
+	{ Victron::VenusOS::Enums::Units_Energy_KiloWattHour,                                     "kWh",      3,          true,     Unit::Default },
+	{ Victron::VenusOS::Enums::Units_AmpHour,                                                    "",      1,          true,     Unit::Default },
+	{ Victron::VenusOS::Enums::Units_WattsPerSquareMetre,                                    "W/m2",      0,          false,    Unit::Default },
```
blammit: Can you check the `scalable` values? In `main`, `isScalingSupported()` returns true for these:

- Units_WattsPerSquareMetre
- Units_RevolutionsPerMinute
- Units_Speed_MetresPerSecond
- Units_Speed_KilometresPerHour
- Units_Speed_MilesPerHour
- Units_Speed_Knots
- Units_Volume_Litre
- Units_Volume_GallonUS
- Units_Volume_GallonImperial

But in this table, they are `false`. 

I see that you added scaling support for `Units_Altitude_Metre`, which makes sense.

src/units.cpp
```diff
@@ -237,38 +147,7 @@ QString Units::scaleToString(VenusOS::Enums::Units_Scale scale) const {
...
-	case VenusOS::Enums::Units_Temperature_Fahrenheit:
-	case VenusOS::Enums::Units_Temperature_Kelvin:
-	case VenusOS::Enums::Units_Hectopascal:
-	case VenusOS::Enums::Units_Kilopascal:
-	case VenusOS::Enums::Units_CardinalDirection:
-	case VenusOS::Enums::Units_PowerFactor:
-	default:
-		return false;
-	}
+    return UnitTable[unit].scalable;
```
blammit: This is currently a 4-space indent, should be a tab.

src/units.cpp
```diff
@@ -99,124 +121,12 @@ qreal Units::formattedNumberToReal(const QString &s) const
...
-		return QStringLiteral("Hz");
-	case VenusOS::Enums::Units_Energy_KiloWattHour:
-		return QStringLiteral("kWh");
-	case VenusOS::Enums::Units_AmpHour:
-		return QStringLiteral("Ah");
-	case VenusOS::Enums::Units_WattsPerSquareMetre:
-		return QStringLiteral("W/m2");
-	case VenusOS::Enums::Units_Percentage:
-		return QStringLiteral("%");
-	case VenusOS::Enums::Units_Temperature_Celsius:
```
blammit: You still need to check the `formatHint` for these temperature cases. Otherwise with this PR the "Environment" tab shows the C/F unnecessarily, below the gauges:

image

src/units.cpp
```diff
@@ -4,49 +4,71 @@
...
 namespace {
 
+// \u33A5 is not supported by the font, so use two characters \u006D\u00B3 instead.
+static const QString CubicMetre = QStringLiteral("m³");
 static const QString DegreesSymbol = QStringLiteral("\u00b0");
+static const QString LitreSymbol = QStringLiteral("\u2113");
+
+struct UnitMetadata {
+	Victron::VenusOS::Enums::Units_Type unit;
+	std::string label;
```
chriadam: why not store as a QString OOI?  When we return it, I think we always convert to QString, and it's better to pay the cost once (when constructing the table) rather than on every read, IMO.

src/units.cpp
ybott-qinetic: Done

src/units.cpp
```diff
@@ -4,49 +4,71 @@
...
+	{ Victron::VenusOS::Enums::Units_Temperature_Kelvin,    DegreesSymbol.toStdString().append("K"),      0,          false,    Unit::Kelvin },
+	{ Victron::VenusOS::Enums::Units_Temperature_Celsius,   DegreesSymbol.toStdString().append("C"),      0,          false,    Unit::Celsius },
+	{ Victron::VenusOS::Enums::Units_Temperature_Fahrenheit,DegreesSymbol.toStdString().append("F"),      0,          false,    Unit::Fahrenheit },
+	{ Victron::VenusOS::Enums::Units_RevolutionsPerMinute,                                    "RPM",      0,          false,    Unit::RevolutionsPerMinute },
+	{ Victron::VenusOS::Enums::Units_Speed_MetresPerSecond,                                   "m/s",      0,          false,    Unit::MetresPerSecond },
+	{ Victron::VenusOS::Enums::Units_Speed_KilometresPerHour,                                 "k/h",      0,          false,    Unit::KilometresPerHour },
+	{ Victron::VenusOS::Enums::Units_Speed_MilesPerHour,                                      "m/h",      0,          false,    Unit::MilesPerHour },
+	{ Victron::VenusOS::Enums::Units_Speed_Knots,                                               "k",      0,          false,    Unit::Knots },
```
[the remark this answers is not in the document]
ybott-qinetic: Done

src/units.cpp
```diff
@@ -4,49 +4,71 @@
...
+static const QString CubicMetre = QStringLiteral("m³");
 static const QString DegreesSymbol = QStringLiteral("\u00b0");
+static const QString LitreSymbol = QStringLiteral("\u2113");
+
+struct UnitMetadata {
+	Victron::VenusOS::Enums::Units_Type unit;
+	std::string label;
+	int precision;
+	bool scalable;
+	Unit::Type veUnit;
```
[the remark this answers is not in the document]
ybott-qinetic: Done

src/units.cpp
```diff
@@ -4,49 +4,71 @@
...
+#include "enums.h"
 
 #include <veutil/qt/unit_conversion.hpp>
 
 #include <QtMath>
 
 namespace {
 
+// \u33A5 is not supported by the font, so use two characters \u006D\u00B3 instead.
+static const QString CubicMetre = QStringLiteral("m³");
```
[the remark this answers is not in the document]
ybott-qinetic: Done

src/units.cpp
blammit: Yes I agree some of the old values are wrong, and should be fixed, but would be great if you could put them all in the separate Units_Altitude_Metre so that it's easy to find the behavioural change when looking through the commit log. E.g. the commit could be "Units: fix scaling for various units" instead of just changing it for metres.

src/units.cpp
blammit: I'm seeing some compass directions here now - do you see that in your build?

image

Also the Boat page has some extra precision now:

image

ybott-qinetic: @blammit Is that a build artifact issue? Should I look at fixing how gui_v2 builds?

src/units.cpp
```diff
@@ -141,127 +164,15 @@ QString Units::formatCoordinate(qreal decimalDegrees, VenusOS::Enums::GpsData_Fo
...
-		return QStringLiteral("µg/m³");
-	case VenusOS::Enums::Units_Lux:
-		return QStringLiteral("lux");
-	case VenusOS::Enums::Units_NewtonMeter:
-		return QStringLiteral("Nm");
-	default:
-		qWarning() << "No unit label known for unit:" << unit;
-		return QString();
+	if (formatHints == Units::FormatHint::CompactUnitFormat) {
+		return QString(UnitTable[unit].label).first(1);
```
DanielMcInnes: will this convert "kWh" -> "k"?

ybott-qinetic: It was a quick shortcut I made to output formatting. I can define a compactunitformat for every unit if desired.

src/units.cpp
```diff
@@ -4,49 +4,72 @@
...
+	{                "Nm",   Victron::VenusOS::Enums::Units_NewtonMeter,              Unit::Default,                0,   false   },
+	{               "ppm",   Victron::VenusOS::Enums::Units_PartsPerMillion,          Unit::Default,                0,   false   },
+	{                 "%",   Victron::VenusOS::Enums::Units_Percentage,               Unit::Default,                0,   false   },
+	{                  "",   Victron::VenusOS::Enums::Units_PowerFactor,              Unit::Default,                3,   false   },
+	{               "RPM",   Victron::VenusOS::Enums::Units_RevolutionsPerMinute,     Unit::RevolutionsPerMinute,   0,   true    },
+	{               "k/h",   Victron::VenusOS::Enums::Units_Speed_KilometresPerHour,  Unit::KilometresPerHour,      0,   true    },
+	{                "kn",   Victron::VenusOS::Enums::Units_Speed_Knots,              Unit::Knots,                  0,   false   },
+	{               "m/s",   Victron::VenusOS::Enums::Units_Speed_MetresPerSecond,    Unit::MetresPerSecond,        0,   true    },
+	{               "m/h",   Victron::VenusOS::Enums::Units_Speed_MilesPerHour,       Unit::MilesPerHour,           0,   false   },
```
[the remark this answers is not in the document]
ybott-qinetic: Thanks

src/units.cpp
```diff
@@ -4,49 +4,72 @@
...
+	{                 "A",   Victron::VenusOS::Enums::Units_Amp,                      Unit::Default,                1,   true    },
+	{                  "",   Victron::VenusOS::Enums::Units_AmpHour,                  Unit::Default,                1,   true    },
+	{       DegreesSymbol,   Victron::VenusOS::Enums::Units_CardinalDirection,        Unit::Default,                0,   false   },
+	{               "kWh",   Victron::VenusOS::Enums::Units_Energy_KiloWattHour,      Unit::Default,                3,   true    },
+	{               "hPa",   Victron::VenusOS::Enums::Units_Hectopascal,              Unit::Default,                1,   false   },
+	{                "Hz",   Victron::VenusOS::Enums::Units_Hertz,                    Unit::Default,                1,   true    },
+	{               "kPa",   Victron::VenusOS::Enums::Units_Kilopascal,               Unit::Default,                0,   false   },
+	{               "lux",   Victron::VenusOS::Enums::Units_Lux,                      Unit::Default,                0,   false   },
+	{ "µg/m" + CubicMetre,   Victron::VenusOS::Enums::Units_MicrogramPerCubicMeter,   Unit::Default,                1,   false   },
```
DanielMcInnes: will this yield: µg/mm³ ? I suspect it should be  "µg/" + CubicMetre

src/units.cpp
```diff
@@ -4,49 +4,72 @@
...
-	case Victron::VenusOS::Enums::Units_RevolutionsPerMinute:
-		return Unit::RevolutionsPerMinute;
-	case Victron::VenusOS::Enums::Units_Altitude_Metre:
-		return Unit::Metre;
-	case Victron::VenusOS::Enums::Units_Altitude_Foot:
-		return Unit::Foot;
-	default:
-		return Unit::Default;
-	}
+	return UnitTable[unit].veUnit;
```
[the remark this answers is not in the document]
ybott-qinetic: Done. Sensible addition.

src/units.cpp
ybott-qinetic: Fixed.

src/units.cpp
```diff
@@ -4,49 +4,80 @@
...
+	{                 "V",   Victron::VenusOS::Enums::Units_Volt_DC,                  Unit::Default,                2,   true    },
+	{          CubicMetre,   Victron::VenusOS::Enums::Units_Volume_CubicMetre,        Unit::CubicMetre,             3,   true    },
+	{               "gal",   Victron::VenusOS::Enums::Units_Volume_GallonImperial,    Unit::ImperialGallon,         0,   false   },
+	{               "gal",   Victron::VenusOS::Enums::Units_Volume_GallonUS,          Unit::UsGallon,               0,   false   },
+	{         LitreSymbol,   Victron::VenusOS::Enums::Units_Volume_Litre,             Unit::Litre,                  0,   true    },
+	{                 "W",   Victron::VenusOS::Enums::Units_Watt,                     Unit::Default,                0,   true    },
+	{              "W/m2",   Victron::VenusOS::Enums::Units_WattsPerSquareMetre,      Unit::Default,                0,   true    },
+};
+
+inline const UnitMetadata &unitMeta(Victron::VenusOS::Enums::Units_Type unit)
```
blammit: Nitpick: I believe Qt naming uses `MetaData` instead of `Metadata`

src/units.cpp
```diff
@@ -141,127 +172,19 @@ QString Units::formatCoordinate(qreal decimalDegrees, VenusOS::Enums::GpsData_Fo
...
-		return QStringLiteral("mph");
-	case VenusOS::Enums::Units_Speed_Knots:
-		return QStringLiteral("kn");
-	case VenusOS::Enums::Units_Hectopascal:
-		return QStringLiteral("hPa");
-	case VenusOS::Enums::Units_Kilopascal:
-		return QStringLiteral("kPa");
-	case VenusOS::Enums::Units_CardinalDirection:
+	if ((formatHints & CompactUnitFormat)
+        && (unit == VenusOS::Enums::Units_Type::Units_Temperature_Celsius
```
blammit: Note, these three lines are indented with spaces instead of tabs

ybott-qinetic: Fixed.

src/units.cpp
```diff
@@ -4,49 +4,84 @@
...
 
 namespace {
 
+// \u33A5 is not supported by the font
+static const QString CubicMetre = QStringLiteral("\u006D\u00B3");
 static const QString DegreesSymbol = QStringLiteral("\u00b0");
+static const QString LitreSymbol = QStringLiteral("\u2113");
+
+struct UnitMetaData {
+	QString label = QString();
```
blammit: This one can be just `QString label;` since that will have a sensible default (the empty string).

src/units.cpp
```diff
@@ -4,49 +4,84 @@
...
+	Victron::VenusOS::Enums::Units_Type unit = Victron::VenusOS::Enums::Units_None;
+	Unit::Type veUnit = Unit::Default;
+	int precision = 0;
+	bool scalable = false;
+};
+
+static const std::vector<UnitMetaData> UnitTable {
+	//              label                 unit                                        veUnit               defaultPrec  scalable
+	{                  "",   Victron::VenusOS::Enums::Units_None,                     Unit::Default,                0,   false   },
```
blammit: Use `QString()` instead of an empty literal "", to use the default constructor directly. Same for PowerFactor.

PR #2754, opened 2026-01-07
Add Electric Vehicle device support to device list

dirkjanfaber:
Adds support for displaying Electric Vehicle (EV) devices in the device list.

Site to test on is `102195` (but there is also the mock data)

### DIFF
--- a/pages/ev/EvPage.qml
+++ b/pages/ev/EvPage.qml
@@ -120,12 +120,11 @@ DevicePage {
 		ListText {
 			//% "Position"
 			text: qsTrId("ev_position")
-			secondaryText: {
-				if (latitude.valid && longitude.valid) {
-					return latitude.value.toFixed(4) + ", " + longitude.value.toFixed(4)
-				}
-				return "--"
-			}
+			secondaryText: latitude.valid && longitude.valid
+					? "%1, %2"
+						.arg(Global.systemSettings.formatLatitude(latitude.value))
+						.arg(Global.systemSettings.formatLongitude(longitude.value))
+					: "--"
 			preferredVisible: latitude.valid && longitude.valid
 		}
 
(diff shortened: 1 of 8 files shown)


blammit (commented):
Thanks for adding the mock config - that makes it much easier to test. 

I've added some comments for improvements, but let me know I should update the PR directly with the changes instead - happy to do that if it's easier.

dirkjanfaber:
> Only thing I'm not sure about is the "Battery capacity" value - is it fine for it to have two decimals of precision (which is the default precision for kWh), or should there be zero decimals of precision? See below, with the old version (left) and the new version (right).

In my opinion it is fine to have 2 decimal precision. I anticipate that, after this PR gets merged, some other changes will be discussed resulting in a new PR. (Things like the order of items and if some items need to be moved to the "Device" part).

pages/ev/EvPage.qml
```diff
@@ -0,0 +1,160 @@
...
+			unit: VenusOS.Units_Energy_KiloWattHour
+		}
+
+		ListQuantity {
+			//% "Power"
+			text: qsTrId("ev_power")
+			dataItem.uid: root.bindPrefix + "/Ac/Power"
+			unit: VenusOS.Units_Watt
+			preferredVisible: dataItem.valid && chargingState.valid &&
+							 (chargingState.value === 3 || chargingState.value === 256 || chargingState.value === 259)
```
chriadam: Does the "scheduled charging" state (259) mean that it is charging due to a schedule, OR does it mean that it is not currently charging but will in future due to a schedule?  I assume the former, because in the latter case there would be no power flow and so there would be no point showing this Power quantity?

chriadam (approved):
Tested fine, LGTM.

PR #2758, opened 2026-01-08
Battery: ensure Lynx Ion diagnostics text fits on page

blammit:
Move the date/time into the ListItem caption text so that it does not spill off the page.


blammit:
Testable with the default maximal mock conf.

Before: 

image


After: 

image

PR #2759, opened 2026-01-12
Tank: allow removing custom tank shape points

chriadam:
The "minus" icon button was unclickable due to its size being reduced too much by close button margins.  This commit ensures that the button will fill the parent delegate height, making it easy to click.

Contributes to issue #2202


ybott-qinetic (approved):
Tested; looks good.

Issue #2745, opened 2025-12-29
Online updates check: Firmware check timed out


mr-manuel:
When Venus OS is on the latest available release the notification says `Firmware check timed out` and not `Latest firmware installed` or something like it was before.

Probably introduces with https://github.com/victronenergy/gui-v2/commit/5aab62c17b08c216ea0d7f49d96d6cb870c4e722 @chriadam 

![Image](https://github.com/user-attachments/assets/425b872e-8fcf-47f7-8b30-950b9828ef9e)

chriadam:
Do you have exact repro steps (e.g. do you have to first perform a check using candidate feed, and then switch to testing feed and perform a check there?  Or, do you have to be on the latest firmware available from the testing feed?  Or something else)?  Does it only happen with a specific installation?

/edit: ah, you mention that it triggers if you are already on the latest firmware version.  I will double check this case!

PR #2761, opened 2026-01-12
Firmware Updates: fix bugs in UI after performing update check

chriadam:
- don't restart the timeout timer in idle state
- stop the timeout timer if the backend transitions to an error state
- increase the check request timeout time interval
- increase the asynchronous time allowed for the update file version to be deserialised after the backend state updates
- handle the case where the user performs an update check, changes the image feed, and then performs another update check which results in the previously-valid value becoming invalid.

Contributes to issue #2745


mr-manuel:
See https://github.com/victronenergy/gui-v2/issues/2745#issuecomment-3739419248

chriadam:
Updated.

mr-manuel (approved):
Made a few tests and all looks good now

PR #2764, opened 2026-01-14
SolarSense: add incomplete configuration detection

dirkjanfaber:
Detect when SolarSense sensors have incomplete configuration by checking bits 26-27 of the `ErrorCode` field. When configuration is incomplete, show a configuration warning with QR code link to setup instructions. 

The warning is only shown on the device detail page, allowing the device list to continue displaying irradiance and other sensor data normally. Updated mock data to include both configured and incomplete sensor examples with correct `ErrorCode` values from updated firmware. If needed, noth types of sensors can also be seen at system `102195`. 

image

And no need to worry that the link still ends up as 404. That is being worked on in parallel.


ybott-qinetic:
@ReinvdZee The url quoted above doesn't seem to exist yet, I assume that is intentional. The SolarSense-750 product we have has a different url embedded into the device and also isn't linking correctly, is that correct?

Issue #2702, opened 2025-12-05
Resolve issue with VKB / InputPanel on portrait-by-default displays


chriadam:
We currently rotate the entire main view, but the VKB might be rendered into an overlay which doesn't get rotated properly.
Can look into this on e.g. Raspberry PI.

Tentatively scheduled for v3.70 but can be pushed back if necessary.

See also issue #1446

chriadam:
PR: https://github.com/victronenergy/gui-v2/pull/2775

chriadam:
After some discussion with Leon and Jeroen, it seems that EGLFS support isn't going to happen for in v1.2.x / Venus OS v3.70 so the workaround PR should be merged for now, and the release notes should mention that dialogs don't work on rpi5 with portrait touch display.

PR #2775, opened 2026-01-20
Virtual Keyboard: get VKB working on rotated displays

chriadam:
The virtual keyboard InputPanel is a child of the Overlay item.

The QtQuick.Controls Overlay does not allow proper coordinate system transformations the way the main Window contentItem does.

As such, merely rotating the overlay item does not result in the desired transform.  We instead need to first rotate the overlay item (by 90 degrees) to ensure that this layer's rendered geometry isn't clipped by the render buffer geometry, and then counter-rotate the child InputPanel (by -90 degrees, or in other words 270 degrees) so that it is rendered appropriately within it.

Finally, we need to somehow map from the (rotated and scaled) main coordinate system to the (rotated but not scaled) overlay coordinate system, and from there map to the (counter-rotated) child coordinate system, in order to position the InputPanel at the appropriate place on the screen.

I have not yet figured out to do this programmatically, but have manually found values which worked on the device I have for testing (rpi5 with touch display 2).

Since we cannot seamlessly map between the coordinate systems, there are some shortcomings:
a) cannot animate things (as I manually figured out the final but
   not initial x and y values).
b) the input panel doesn't properly fill the screen (I am not sure
   why this is the case; if I did, mapping between the coordinate
   systems would presumably be simpler).

This commit achieves something "good enough" for now, but in future:
(i) we should not use fixed geometry (1024x600 etc), but instead should use layouts which fill the available space; (ii) we should handle rotation at the platform plugin layer, not in the UI layer, in order to avoid the coordinate system mapping altogether.

Contributes to issue #2702

### DIFF
--- a/components/dialogs/ModalDialog.qml
+++ b/components/dialogs/ModalDialog.qml
@@ -295,5 +295,15 @@ T.Dialog {
 		enabled: !!stateManager.inputItem
 		onClicked: focus = true
 	}
+
+	Component.onCompleted: {
+		if (Global.main && Global.main.requiresRotation) {
+			// we cannot manually position the header or footer.
+			// just reject the dialog for now.
+			// TODO: use eglfs and rotate the entire surface.
+			// See: issue #2702.
+			Qt.callLater(reject)
+		}
+	}
 }
 
--- a/components/dialogs/SolarDailyHistoryDialog.qml
+++ b/components/dialogs/SolarDailyHistoryDialog.qml
@@ -199,4 +199,14 @@ T.Dialog {
 			visible: enabled
 		}
 	}
+
+	Component.onCompleted: {
+		if (Global.main && Global.main.requiresRotation) {
+			// we cannot manually position the header or footer.
+			// just reject the dialog for now.
+			// TODO: use eglfs and rotate the entire surface.
+			// See: issue #2702
+			Qt.callLater(reject)
+		}
+	}
 }


leon-anavi (approved):

chriadam:
Added a commit to auto-reject any Dialog which is raised in the UI, because we cannot position the content manually, unfortunately.  This means that significant parts of the UI are simply non-functional.

This makes me doubt the usefulness of merging this PR at all.
I suspect we should just get eglfs working on rpi5, and just do:
```
QT_QPA_PLATFORM=eglfs
QT_QPA_EGLFS_ROTATION=90
QT_QPA_GENERIC_PLUGINS=evdevtouch:/dev/input/event0:rotate=90
```
or similar.

ApplicationContent.qml
```diff
@@ -112,9 +112,12 @@ FocusScope {
 	//
 	// Note this Loader is the top-most layer, to allow the idleModeMouseArea beneath to call
 	// acceptMouseEvent() when clicking outside of the focused text field, to auto-close the Qt VKB.
-	Loader {
+	property Loader _vkbLoader: Loader {
```
blammit: What's the reason for making this a property?

components/InputPanel.qml
```diff
@@ -47,9 +49,23 @@ QtVirtualKeyboard.InputPanel {
 	}
 
 	visible: Qt.inputMethod.visible || yAnimator.running
-	y: Qt.inputMethod.visible ? Theme.geometry_screen_height - root.height : Theme.geometry_screen_height
+
+	y: requiresRotation ? 312
+	 : Qt.inputMethod.visible ? Theme.geometry_screen_height - root.height
```
blammit: Indentation is a bit odd here? Also for `x` below

ApplicationContent.qml
```diff
@@ -127,8 +130,18 @@ FocusScope {
...
-		z: 1
+
+		property Rotation requiredRotation: Rotation {
+			origin.x: width / 2
+			origin.y: height / 2
+			angle: 90
+		}
+
+		Component.onCompleted: {
+			if (Global.main && Global.main.requiresRotation) {
```
blammit: Could have a comment here as well to refer to the task?

components/InputPanel.qml
chriadam: lining up the :
not really odd, we use this pattern throughout.

ApplicationContent.qml
blammit: By "wrong", do you mean semantically wrong, or that it will cause problems when reparenting?

blammit (approved):

PR #2813, opened 2026-02-13
Units: ensure utility functions have default return values

blammit:


chriadam:
I would use "Units" prefix.  This isn't just a build issue, it's that (at runtime) any code which calls those functions with invalid values (from QML, where you can provide any arbitrary int and the method will be invoked with that invalid value by the engine) will have potentially undefined behaviour.

PR #2818, opened 2026-02-16
AC Load: always show total header

ybott-qinetic:
Contributes to #2748

### DIFF
--- a/pages/loads/AcLoadListPage.qml
+++ b/pages/loads/AcLoadListPage.qml
@@ -20,9 +20,8 @@ Page {
 			readonly property alias columnWidth: loadSummary.fixedColumnWidth
 			readonly property alias columnSpacing: loadSummary.columnSpacing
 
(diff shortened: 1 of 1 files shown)


pages/loads/AcLoadListPage.qml
```diff
@@ -20,9 +20,8 @@ Page {
 			readonly property alias columnWidth: loadSummary.fixedColumnWidth
 			readonly property alias columnSpacing: loadSummary.columnSpacing
 
-			visible: root.model.count > 1 || root.measurements.phaseCount > 1
 			width: parent?.width ?? 0
-			height: visible ? phaseTable.y + phaseTable.height + bottomInset : 0
```
chriadam: does something else set visible to false?  If so, it won't appear, still, even though the space is available for it, I think.  Or am I misunderstanding?

blammit (approved):

PR #2826, opened 2026-02-18
Switches: show correct status if 'on' status is combined with others

### LINKED ISSUE #2798: Switch pane multiple status indicators: Handle combinations with 'ON'
Related to this issue: https://github.com/victronenergy/gui-v2/issues/2736

Some devices also send out a status value while the output is still turned on. For example, an output can be on but have the status "over temperature". The combined `/Status` would then be `0x0D` (13). The GUI then shows `13` but it should just show the overtemp indicator.

So, for determining the status indicator to show, mask the 'output on' bits (`0x09`) of the `/Status` field. Except for the Bilge pump (`/Type` 10), as for that one the 'Running' indicator must be shown when the `/Status` is `0x09`.

blammit:
When determining the status text, mask the 'output on' bits (0x09) of the /Status field. An exception is in the Bilge Pump case (Type=10), as for that one the 'Running' indicator must be shown when the /Status is 0x09.

This is because some devices also send out a status value while the output is still turned on. For example, an output can be on but have the status "Over temperature". The combined /Status would then be 0x0D (13). In this case the status text should be "Over temperature" instead of "13".

For the status text background, if the status for a Bilge Pump is an invalid running combination (e.g. running + tripped) then show the  background in red instead of orange.

Fixes #2798


chriadam (approved):
LGTM, but I'd like Rein to review also

PR #2831, opened 2026-02-19
Add support for generic inputs

blammit:


blammit:
Note this depends on https://github.com/victronenergy/gui-v2/pull/2751 (adding `/Settings/Function` for switchable outputs) being merged first.

ReinvdZee:
@blammit 

I can't really test this at the moment. Perhaps @citolen can give it a quick test?

ReinvdZee:
@blammit 

I tested this with the GX IO extender, which only has discrete inputs. 

The settings page currently looks like this:
image

I like that the actual value is shown in the summary, but please show them as integers without decimals

components/switches/delegates/GenericInputCardDelegate_3.qml
```diff
@@ -0,0 +1,105 @@
...
+				// the background is the border with an additional rectangle for fill
+				gradient: Gradient {
+					orientation: Qt.Horizontal
+					GradientStop { position: 0.0; color: Theme.color_temperatureslider_gradient_min_border }
+					GradientStop { position: 1.0; color: Theme.color_temperatureslider_gradient_max_border }
+				}
+			}
+		}
+
+		QuantityLabel {
```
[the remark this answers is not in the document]
blammit: Fixed.

blammit:
> @blammit
> 
> I tested this with the GX IO extender, which only has discrete inputs.
> 
> I like that the actual value is shown in the summary, but please show them as integers without decimals

Thanks, fixed in the updated PR.

src/iochannel.cpp
```diff
@@ -0,0 +1,351 @@
...
+	m_decimalsVariant = variant.toInt();
+	updateDecimals();
+}
+
+int IOChannel::getDecimals() const
+{
+	return m_decimalsVariant.toInt();
+}
+
+void IOChannel::updateDecimals()
```
chriadam: is this called by anything other than setDecimals()?  if not, can we just manually inline it into setDecimals()?

src/genericinput.cpp
```diff
@@ -0,0 +1,193 @@
...
+		{ "/stopped", qtTrId("generic_input_label_stopped") },
+		//% "Running"
+		{ "/running", qtTrId("generic_input_label_running") },
+	};
+
+	QStringList labels;
+	const QStringList rawLabels = variant.toStringList();
+	for (const QString &label : rawLabels) {
+		if (label.startsWith('/')) {
+			labels.append(reservedLabels.value(label, label));
```
chriadam: Probably not, but thought I'd raise it just in case: maybe you want to trim the leading '/' from the default value string?

src/genericinput.cpp
```diff
@@ -0,0 +1,193 @@
...
+}
+
+qreal GenericInput::rangeMax() const
+{
+	return m_rangeMax;
+}
+
+void GenericInput::setRangeMax(const QVariant &variant)
+{
+	m_rangeMax = variant.isValid() ? variant.value<qreal>() : 100;
```
chriadam: is it worth checking that the type of the variant is convertible to qreal also?  just in case someone decides to call it with `QVariant(QString("oops"))` or something

src/iochannelmodel.cpp
```diff
@@ -0,0 +1,238 @@
+/*
+** Copyright (C) 2025 Victron Energy B.V.
```
chriadam: nit: 2026

src/iochannelmodel.cpp
```diff
@@ -0,0 +1,238 @@
...
+		return outputUid;
+	case NameRole:
+		return m_entries.value(outputUid).name();
+	}
+	return QVariant();
+}
+
+QHash<int, QByteArray> SwitchableOutputModel::roleNames() const
+{
+	static QHash<int, QByteArray> roles = {
```
chriadam: can be const, don't need =

src/iochannelmodel.h
```diff
@@ -0,0 +1,86 @@
...
+		QPointer<VeQItem> functionItem;
+	};
+
+	void sourceModelRowsInserted(const QModelIndex &parent, int first, int last);
+	void sourceModelRowsAboutToBeRemoved(const QModelIndex &parent, int first, int last);
+	void clearEntries();
+	void addEntry(const QString &outputUid);
+	void updateCount();
+
+	QMap<QString, Entry> m_entries;
```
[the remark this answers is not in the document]
chriadam: The only place I could see where it linearly iterates over the entries is when clearing the datastructure, so I suspect it's unnecessary, and can be a QHash for performance reasons.  But maybe I'm wrong.

src/genericinput.cpp
```diff
@@ -147,8 +150,33 @@ QString GenericInput::primaryLabel() const
...
+{
+	const QString prevLabel = m_primaryLabel;
+
+	switch (unitType()) {
+	case Enums::Units_Speed_MetresPerSecond:
+		//% "Speed"
+		m_primaryLabel = qtTrId("generic_input_primaryLabel_speed");
+		break;
+	case Enums::Units_Temperature_Celsius:
+		m_primaryLabel = qtTrId("generic_input_primaryLabel_temperature");
```
chriadam: I guess the EE string is defined somewhere else?  Somewhat confusing that "Speed" wasn't similarly already defined.

Maybe we should adopt some convention, like having a `static void registerTranslations() { ... }` preamble function that registers all of the ids and ee strings, and then elsewhere in the source only unadorned `qtTrId(translation_id)` are used, to avoid double-registration or whatever.

Well, unrelated to this PR so ignore for now.

components/switches/delegates/GenericInputCardDelegate_0.qml
```diff
@@ -39,5 +39,31 @@ Item {
 			rightMargin: Theme.geometry_controlCard_button_margins
 			top: header.bottom
 		}
+
+		Label {
```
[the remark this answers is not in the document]
blammit: That one just touches list items. We could take this out of the background and just anchor to the background as a sibling rather than as a `parent`, if it aligns better with what we do elsewhere. I'm just used to anchoring to the parent.

src/genericinput.cpp
```diff
@@ -0,0 +1,223 @@
...
+}
+
+qreal GenericInput::rangeMin() const
+{
+	return m_rangeMin;
+}
+
+void GenericInput::setRangeMin(const QVariant &variant)
+{
+	m_rangeMin = variant.isValid() && variant.canConvert<qreal>() ? variant.value<qreal>() : 0;
```
chriadam: In the case where the input is invalid, I wonder if it's better to set the value to zero (as this does, currently), or to early-return and leave m_rangeMin as whatever it was previously?

chriadam: Yep sounds fair.

chriadam (approved):
LGTM.

Issue #2774, opened 2026-01-20
[feat] Improve readability for colourblind people


dirkjanfaber:
Colourblind person complained that the orange is difficult to see.

<img width="720" height="405" alt="Image" src="https://github.com/user-attachments/assets/987ed55a-2dfe-4f7d-9ba9-de5e21e88a3d" />

Could it get an icon/underline/... as well?

PR #2837, opened 2026-02-19
Support status: optimize page for color blind people

mr-manuel:
Use icons in addition to colors to indicate warning and alarm status

Fixes https://github.com/victronenergy/gui-v2/issues/2774


pages/settings/PageSettingsSupportStatus.qml
```diff
@@ -24,15 +25,9 @@ Page {
...
-			return qsTrId("pagesettingssupportstate_support_state_check_below_red_orange")
-		} else if (isModified) {
-			//% "Check below items in red"
-			return qsTrId("pagesettingssupportstate_support_state_check_below_red")
-		} else if (isIntegrationRunning) {
-			//% "Check below items in orange"
-			return qsTrId("pagesettingssupportstate_support_state_check_below_orange")
+		if (isModified || isIntegrationRunning) {
+			//% "Check items below"
+			return qsTrId("pagesettingssupportstate_support_state_check_items")
```
blammit: There's some text at the bottom of this page that says, "Items colored orange are...". Is that still needed? If so it needs a text update as well.

pages/settings/PageSettingsSupportStatus.qml
```diff
@@ -223,18 +212,19 @@ Page {
 	GradientListView {
 		model: VisibleItemModel {
 
-			ListText {
+			ListTextStatus {
```
blammit: In the Figma design, the "Support status" list item has been removed altogether.

components/listitems/ListTextStatus.qml
```diff
@@ -0,0 +1,41 @@
...
+import QtQuick.Controls.impl as CP
+import Victron.VenusOS
+
+ListText {
+	id: root
+
+	readonly property alias dataItem: dataItem
+	property alias secondaryText: secondaryLabel.text
+	property alias secondaryLabel: secondaryLabel
+    property int secondaryStatus: -1
```
blammit: Since the status is an `Alarm_Level` value, this could be `alarmStatus` instead of `secondaryStatus`

pages/settings/PageSettingsSupportStatus.qml
mr-manuel: Nice catch, thanks. I updated the text.

chriadam:
The toast and notification model changes LGTM.

Issue #2790, opened 2026-01-29
Switches: Support `/CustomName` and `/Group` being readonly


ReinvdZee:
For virtual switches in Node-red, the `/SwitchableOutput/x/CustomName` and `/SwitchableOutput/x/Group` paths are marked readonly because the custom name and group name are set (and stored) in the flow. A new value set through the UI would not be stored. 

So if one of these paths is not writable, just display the text without the edit field.

blammit:
@ReinvdZee what's the best way for gui-v2 to determine whether these paths are writable versus read-only?

blammit:
@ybott-qinetic Can this be closed now?

ReinvdZee:
Assuming the feature works, I believe the issue can be closed for now.
@dirkjanfaber will test this soon with NR virtual switches.

PR #2843, opened 2026-02-20
Switches: add access controls to SwitchableOutput settings

ybott-qinetic:
Adds "/Settings/Adjustable" to settings, this provides a readonly 
flag for settings. Currently only protects CustomName and Group.

Contributes to #2790

### DIFF
--- a/pages/settings/devicelist/iochannel/PageSwitchableOutput.qml
+++ b/pages/settings/devicelist/iochannel/PageSwitchableOutput.qml
@@ -11,14 +11,18 @@ Page {
 
 	required property SwitchableOutput switchableOutput
 
+	readonly property bool _writeable: !(settingsAdjustable.valid && settingsAdjustable.value === 0)
+
 	GradientListView {
 		model: VisibleItemModel {
 			ListIOChannelNameField {
 				dataItem.uid: root.switchableOutput.uid + "/Settings/CustomName"
+				interactive: _writeable
 			}
 
 			ListIOChannelGroupField {
 				dataItem.uid: root.switchableOutput.uid + "/Settings/Group"
+				interactive: _writeable
 			}
 
 			ListRadioButtonGroup {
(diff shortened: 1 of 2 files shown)


pages/settings/devicelist/iochannel/PageSwitchableOutput.qml
```diff
@@ -11,21 +11,26 @@ Page {
 
 	required property SwitchableOutput switchableOutput
 
+	readonly property bool _writeable: !(settingsAdjustable.valid && settingsAdjustable.value === 0)
```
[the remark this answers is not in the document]
blammit: Yes perhaps "writable" is better for consistency.

pages/settings/devicelist/iochannel/PageSwitchableOutput.qml
```diff
@@ -71,10 +78,12 @@ Page {
 				decimals: 0 // backend does not allow for decimal precision
 				suffix: Units.defaultUnitString(VenusOS.Units_Amp)
 				preferredVisible: dataItem.valid
+				interactive: _writeable
 			}
 
 			ListIOChannelTypeRadioButtonGroup {
 				ioChannel: root.switchableOutput
+				interactive: _writeable
```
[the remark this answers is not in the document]
blammit: Also, refer to `root._writable` instead of just `_writable`.

pages/settings/devicelist/iochannel/PageSwitchableOutput.qml
```diff
@@ -130,6 +141,7 @@ Page {
 				text: qsTrId("page_switchable_output_startup_dim_level")
 				secondaryText: startupDimLevel.valid ? startupDimLevel.value === -1 ? qsTrId("page_switchable_output_startup_state_restore_from_memory") : startupDimLevel.value + "%" : ""
 				preferredVisible: startupDimLevel.valid
+				interactive: _writable
```
blammit: Setting `interactive` here has no effect, as the user can still open the sub-page. (This is deliberate behaviour in ListNavigation, to allow the user to view the current value, even if it cannot be edited.) So, this line can be omitted.

blammit (approved):
One minor issue, otherwise LGTM.

pages/settings/devicelist/iochannel/PageSwitchableOutput.qml
blammit: Also, update the first line of the commit message - it still says only "CustomName" and "Group" are affected.

ybott-qinetic: Done

PR #2857, opened 2026-03-04
Switches: fix unit conversion and step size issues

### LINKED ISSUE #2854: Switch pane: Recalculate stepsize for destination unit
The `/StepSize` path is used to specify the stepsize of the slider components, temperature setpoint and basic slider. Currently, the stepsize is not recalculated for the destination unit, so the user is experiencing an unexpected stepsize when the selected unit is not the base unit.

Furthermore, there is an error in the stepsize conversion for temperature. When Fahrenheit is selected as unit, a stepsize of 1 corresponds to very large steps which is incorrect. This is probably derived from the conversion of 1 degrees celsius to 1 degrees fahrenheit (which is 33.8), but it should also take the offset into account.

See the report here: https://community.victronenergy.com/t/new-temperature-virtual-switch-slider-increment-stuck-at-crazy-high-values/54023

**Solution**
The best user experience would be that the configured stepsize is applied to the destination unit. So a virtual switch of type "Temperature setpoint" with a stepsize of 1 should have steps of 1 in both the degrees celsius, as well as the degrees fahrenheit unit.

Note that the conversion is only for the UI, the actual value on dbus is in the base unit.

**Example for the temperature setpoint**

Configuration:
Stepsize: 1

_GX unit set to °C_
No conversions needed. Apply a stepsize of 1 and directly display the dbus value on the slider.

_GX unit set to °F_
The base unit on dbus is in °C. Apply conversion to the value, min, max. The configured stepsize should be applied in the destination unit.

To convert from degrees x celsius to y degrees fahrenheit:
(x°C × 9/5) + 32 = y°F

So with a stepsize of 1°F, add 1°F to the value on every step, and convert the result to °C before writing to dbus.

DanielMcInnes:
Fixes #2854

### DIFF
--- a/components/switches/SwitchableOutputSlider.qml
+++ b/components/switches/SwitchableOutputSlider.qml
@@ -64,12 +64,14 @@ MiniSlider {
 		uid: root.hasDimmingProperties ? root.switchableOutput.uid + "/Settings/DimmingMax" : ""
 		sourceUnit: Units.unitToVeUnit(root.sourceUnit)
 		displayUnit: Units.unitToVeUnit(root.displayUnit)
+		onValueChanged: valueSync.updateSliderValue()
 	}
 	VeQuickItem {
 		id: dimmingMinItem
 		uid: root.hasDimmingProperties ? root.switchableOutput.uid + "/Settings/DimmingMin" : ""
 		sourceUnit: Units.unitToVeUnit(root.sourceUnit)
 		displayUnit: Units.unitToVeUnit(root.displayUnit)
+		onValueChanged: valueSync.updateSliderValue()
 	}
 	VeQuickItem {
 		id: stepSizeItem
--- a/components/SpinBoxDecimalConverter.qml
+++ b/components/SpinBoxDecimalConverter.qml
@@ -12,9 +12,9 @@ QtObject {
 	property real to
 	property real stepSize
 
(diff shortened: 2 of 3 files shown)


components/switches/SwitchableOutputSlider.qml
```diff
@@ -64,18 +64,19 @@ MiniSlider {
...
 		sourceUnit: Units.unitToVeUnit(root.sourceUnit)
 		displayUnit: Units.unitToVeUnit(root.displayUnit)
+		onValueChanged: valueSync.updateSliderValue()
 	}
 	VeQuickItem {
 		id: stepSizeItem
 		uid: root.hasDimmingProperties ? root.switchableOutput.uid + "/Settings/StepSize" : ""
-		sourceUnit: Units.unitToVeUnit(root.sourceUnit)
-		displayUnit: Units.unitToVeUnit(root.displayUnit)
+		// Do not set sourceUnit and displayUnit, as unit conversion is not applied to stepSize.
```
chriadam: should this be the case for ALL units, or only for temperature units?

chriadam: Rein says yes, for all.

chriadam:
Actually, can we squash these commits into one (or two, maybe)?

PR #2861, opened 2026-03-05
Generic Inputs: fix custom unit display

### LINKED ISSUE #2869: Generic Input API: unit not showing
While implementing for node-red and testing the functionality of the  generic input API it looks like the _Unit_ is not rendering. In this case I've set the `/GenericInput/0/Settings/Unit` path to `W`.

Image

Image

As this is a new feature I won't rule out that the problem exists on my side instead of on the gui-v2 side. But also not sure what I might be overlooking. I do recall that the unit was showing on the demo during last weeks meeting.

Testing against gui-v2 v1.3.0

blammit:
Fixes #2869 
---
Fixes to show the custom unit, as per here:

image

### DIFF
--- a/components/IOChannelQuantityLabel.qml
+++ b/components/IOChannelQuantityLabel.qml
@@ -11,7 +11,7 @@ QuantityLabel {
 
 	valueText: quantityInfo.number
 	unit: Global.systemSettings.toPreferredUnit(ioChannel.unitType)
-	unitText: quantityInfo.unit === VenusOS.Units_None ? ioChannel.unitText : quantityInfo.unit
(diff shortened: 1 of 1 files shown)


components/IOChannelQuantityLabel.qml
```diff
@@ -11,7 +11,7 @@ QuantityLabel {
 
 	valueText: quantityInfo.number
 	unit: Global.systemSettings.toPreferredUnit(ioChannel.unitType)
-	unitText: quantityInfo.unit === VenusOS.Units_None ? ioChannel.unitText : quantityInfo.unit
+	unitText: quantityInfo.unitType === VenusOS.Units_None ? ioChannel.unitText : quantityInfo.unit
```
chriadam: the last one is also `quantityInfo.unit` should it also be updated?

Issue #2205, opened 2025-06-05
Look into behaviour when localsettings or venus-platforms restarts


mpvader:
Both the local gui-v2, as well as on WASM, should preferably be able to cope with that. Or crash, exit and restart - since localsettings & venus-platform restarting themselves is not something common.

But what happens now (Jeroen saw that it gets stuck) is not good.

First thing here is to see what it would entail to improve this; probably best discussed during a next dev meeting

PR #2868, opened 2026-03-10
UI: gracefully handle localsettings / venus-platform crash

chriadam:
Contributes to issue #2205

### DIFF
--- a/Main.qml
+++ b/Main.qml
@@ -38,19 +38,22 @@ Window {
 		if (Global.mainView) {
 			Global.mainView.clearUi()
 		}
-		const demoModeChange = dataManagerLoader.active && dataManagerLoader.connectionReady
-		if (demoModeChange) {
+		const requiresReloadData = dataManagerLoader.active && dataManagerLoader.connectionReady
+		if (requiresReloadData) {
 			// we haven't lost backend connection.
-			// we must be rebuilding UI due to demo mode change.
+			// we must be rebuilding UI due to demo mode change,
+			// or detected crash in localsettings/venus-platform.
 			// manually cycle the data manager loader.
-			console.info("Main: resetting data manager due to demo mode change")
+			console.info("Main: resetting data manager due to change requiring data reload")
 			dataManagerLoader.active = false
+		} else {
+			console.info("Main: data reload not required")
 		}
 		Global.reset()
-		if (demoModeChange) {
+		gc()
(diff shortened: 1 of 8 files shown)


src/systemservicelistener.cpp
```diff
@@ -0,0 +1,86 @@
...
+	connect(model, &AllServicesModel::serviceAboutToBeRemoved,
+		this, &SystemServiceListener::handleServiceAboutToBeRemoved);
+	for (int i = 0; i < model->count(); ++i) {
+		handleServiceAdded(model->itemAt(i));
+	}
+}
+
+void SystemServiceListener::handleServiceAdded(VeQItem *serviceItem)
+{
+	if (serviceItem->uniqueId().contains(QStringLiteral("settings"))) {
```
blammit: Would be safer to check the service type (as per the `ServiceType` role), so this doesn't accidentally match against other services that have "settings" or "platform" in the uid.

Main.qml
```diff
@@ -185,5 +190,57 @@ Window {
...
+		target: SystemServiceListener
+		property bool needReload: false
+		property var toastId: null
+		function onSettingsOnlineChanged() {
+			if (Global.backendReady && !SystemServiceListener.settingsOnline) {
+				console.info("Main: settings service is now unavailable!")
+				if (!systemServiceConnections.needReload) {
+					systemServiceConnections.needReload = true
+					//% "Warning: detected localsettings service offline; reloading UI when it becomes available again..."
+					systemServiceConnections.toastId = Global.showToastNotification(VenusOS.Notification_Warning, qsTrId("main_system_service_settings_offline_warning"))
```
blammit: Does this need separate toast ids for settings and platform in case they've both crashed?

src/systemservicelistener.cpp
```diff
@@ -0,0 +1,86 @@
...
+	if (serviceType == QStringLiteral("settings")) {
+		listenToSettingsItem(serviceItem);
+	} else if (serviceType == QStringLiteral("platform")) {
+		listenToPlatformItem(serviceItem);
+	}
+}
+
+void SystemServiceListener::handleServiceAboutToBeRemoved(VeQItem *serviceItem)
+{
+	if (serviceItem->uniqueId().contains(QStringLiteral("settings"))) {
```
chriadam: Oops, I need to fix this one too

Issue #2856, opened 2026-03-03
Use runtime GPU support detection for MSAA vs Supersampling decision


chriadam:
Currently, we use some CPU arch ifdefs to determine whether to use MSAA or not.  These ifdefs basically hardcode some assumptions about known platforms (e.g. ARM CPU means CerboGX, and CerboGX has a Mali400 and Qt doesn't support MSAA on Mali400 - see QTBUG-120474), but of course CPU arch doesn't actually have anything to do with MSAA support and if in future we add a new platform with an ARM CPU but a newer GPU, Qt might support MSAA there.

So, instead of using fragile ifdefs, we should try to use run-time detection of support for MSAA in main.cpp.
e.g. request supported surface formats and sample counts, and based on that, make the appropriate decision about requesting 4 samples or not.

chriadam:
I have created this PR: https://github.com/victronenergy/gui-v2/pull/2873 but it simply cannot be done purely at runtime on the CerboGX hardware (at least, using the Qt APIs), because it reports that it supports 4 samples even though it does not.  So, the PR still maintains an ifdef (but hopefully it is more precise, as it no longer catches 64-bit arm builds).

PR #2873, opened 2026-03-12
UI: prefer run-time multi-sample render buffer detection

chriadam:
Unfortunately, we cannot entirely rely on runtime detection due to QTBUG-120474 but for most other cases, it should work.

Contributes to issue #2856


chriadam:
Do you want me to investigate further the Qt platform side of things?  It seems work correctly with the TI blobs (see https://github.com/victronenergy/gui-v2/issues/2856#issuecomment-4115258789) but if it's broken with the open mesa drivers, I can investigate and create an upstream Qt fix, potentially.

PR #2908, opened 2026-04-07
Shelly: show icons for shelly devices

blammit:
Fixes #2901


pages/settings/PageSettingsShelly.qml
```diff
@@ -104,9 +105,40 @@ Page {
...
+					case "em": return "qrc:/images/icon_energymeter_3f_32.svg"
+					default: return ""
+					}
+				}
+			}
+
+			VeQuickItem {
+				id: typeItem
+
+				readonly property string deviceUid: buddy.itemParent()?.itemParent()?.uid ?? ""
```
chriadam: what does buddy resolve to?

chriadam (approved):
One question, aside from that LGTM

ReinvdZee:
I tried to test this, but I think I need to copy the new icons somewhere because it displays nothing in front of the devices in the list:

image

I get this behavior on WASM, local and running on desktop connected over MQTT.

Where should the icons be installed?

blammit:
> I tried to test this, but I think I need to copy the new icons somewhere because it displays nothing in front of the devices in the list:
> 
> Where should the icons be installed?

If all other icons are appearing normally, then there must be something wrong with the code - would you have a VRM installation where I could check this?

pages/settings/PageSettingsShelly.qml
blammit: It's the VeQItem that comes from the VeQItemChildModel role. This is a bit icky but it's the most straightforward way to get to the parent item.

chriadam:
> > I tried to test this, but I think I need to copy the new icons somewhere because it displays nothing in front of the devices in the list:
> > Where should the icons be installed?
> 
> If all other icons are appearing normally, then there must be something wrong with the code - would you have a VRM installation where I could check this?

We've got a shelly in the office

ReinvdZee:
@blammit,

Does it work with the shelly in your office? 

Otherwise you can test on this system: https://vrm.victronenergy.com/installation/388348/overview

chriadam:
> @blammit,
> 
> Does it work with the shelly in your office?
> 
> Otherwise you can test on this system: https://vrm.victronenergy.com/installation/388348/overview

I've just set up the new cerbogx and added the shelly via the Integrations menu.  I didn't get a chance to test Bea's PR on device yet, though - will do so tomorrow.

blammit:
> @blammit,
> 
> Does it work with the shelly in your office?
> 
> Otherwise you can test on this system: https://vrm.victronenergy.com/installation/388348/overview

Thanks! I thought the path was `/Devices/<device>/Type` but it's actually `/Devices/<device>/<index>/Type`. Works on the installation now:

image

ReinvdZee:
@blammit,

Thanks, looks good now:
image


One further request though, can we omit the '1' for devices that have just one channel? Note that in this case that means there are 2 paths behind `/Devices/<device>/<index>/`.

In above screenshot, I'd like to see the channel of the plus plug s omitted.

blammit:
> One further request though, can we omit the '1' for devices that have just one channel? Note that in this case that means there are 2 paths behind `/Devices/<device>/<index>/`.
> 
> In above screenshot, I'd like to see the channel of the plus plug s omitted.

Yep, have updated the PR to do this:

image

pages/settings/PageSettingsShelly.qml
```diff
@@ -67,6 +68,20 @@ Page {
 					id: nameItem
 					uid: channelSortDelegate.deviceUid + "/Name"
 				}
+
+				// Provides a model of the channels for this device.
+				// TODO rework this whole page to use a custom C++ model, to avoid having to build
```
chriadam, line 73: is there a github issue to track this work?

chriadam (approved):

pages/settings/PageSettingsShelly.qml
blammit: Created https://github.com/victronenergy/gui-v2/issues/2924 to track this, and updated the comment to refer to it.

ReinvdZee:
Thanks @blammit 

Can confirm it works now, also with singlephase and three-phase EMs:
image

Issue #2918, opened 2026-04-13
Add "EEBus devices" menu


citolen:
For the upcoming EEBus service, we need to add a way to manage it and its devices from the UI.

Requirements are:
- new menu on the integrations page. ("EEBus Devices")
- a way to enable/disable the service.
- show the service's SKI.
- show the service's pairing QR code.
- show the discovered devices.
- a sub page per discovered device with all of its information.
- a way to set a device as trusted.

chriadam:
Can this be closed as fixed, or is there a second PR coming for this task?

citolen:
For GUI-v2 that's it for now, closing.

PR #2919, opened 2026-04-13
Integrations: add EEBus Devices page

citolen:
Contributes to: https://github.com/victronenergy/gui-v2/issues/2918

For the upcoming EEBus service, we are adding a way to manage it and its devices from the UI.

Requirements are:
- new menu on the integrations page. ("EEBus Devices")
- a way to enable/disable the service.
- show the service's SKI.
- show the service's pairing QR code.
- show the discovered devices.
- a sub page per discovered device with all of its information.
- a way to set a device as trusted.

Screenshot 2026-04-13 at 09 35 44

Screenshot 2026-04-13 at 09 35 53

Screenshot 2026-04-13 at 09 36 21

### DIFF
--- a/cmake/ModuleMock_Sources.cmake
+++ b/cmake/ModuleMock_Sources.cmake
@@ -34,6 +34,7 @@ SET(VictronMock_QML_MODULE_RESOURCES
     data/mock/conf/services/dcsystem2.json
     data/mock/conf/services/dse-genset.json
     data/mock/conf/services/digitalinput-alarm.json
+    data/mock/conf/services/eebus.json
     data/mock/conf/services/em-acload.json
     data/mock/conf/services/em-evcharger-input.json
     data/mock/conf/services/em-evcharger-output.json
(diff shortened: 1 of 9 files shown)


components/listitems/ListQrCode.qml
```diff
@@ -0,0 +1,124 @@
+/*
+** Copyright (C) 2026 Victron Energy B.V.
+** See LICENSE.txt for license information.
+*/
+
+import QtQuick
+import QtQuick.Layouts
+import Victron.VenusOS
```
blammit: It looks like this is used instead of `ListLink` because of the need to always show "Show QR code", and never show the "Open link" option (even in Wasm). Is that correct? If so, we could just make `ListLink` more configurable, to avoid the need for this duplication. In c6e8cbcf433ac9bc5df92143b2569523fe1bdae9 I removed the `ListLink` `mode` property because it was unused, but that would be really useful here. 

So in `ListLink` we could have:

```
property int mode: Qt.platform.os == "wasm" ? VenusOS.ListLink_Mode_LinkButton : VenusOS.ListLink_Mode_QRCode
```

and then in PageSettingsEeBus.qml:

```
ListLink {
     url: pairingQRCode.valid ? pairingQRCode.value : ""
     mode: VenusOS.ListLink_Mode_QRCode
     caption: ""
} 
```

citolen: Good to know, I've restored the `mode` property and removed the `ListQrCode` component. Thanks!

blammit: Looks great, thanks for making the changes.

blammit (approved):

PR #2921, opened 2026-04-14
Settings: Disable Opportunity Loads when DESS is running

### LINKED ISSUE #2913: OpportunityLoads: Disable the enable switch when DESS is running
OpportunityLoads and DESS can't be enabled together. Venus-platform will make sure that opportunity loads isn't enabled whenever DESS is. However, it must be made clear in the UI that opportunity loads can't be enabled and that DESS needs to be disabled first. 

So, when this path: `com.victronenergy.settings/Settings/DynamicEss/Mode` is greater than 0, the opportunity loads enable switch should be disabled (made uncontrollable):

Image

And show a secondary text below the enable switch saying:

"Opportunity loads cannot be enabled while Dynamic ESS is running. Disable Dynamic ESS first."

DanielMcInnes:
Fixes #2913

### DIFF
--- a/data/mock/conf/setup-common.json
+++ b/data/mock/conf/setup-common.json
@@ -9,7 +9,7 @@
     },
     "com.victronenergy.settings": {
         "/Settings/Alarm/System/GridLost": 1,
-        "/Settings/DynamicEss/Mode": 1,
+        "/Settings/DynamicEss/Mode": 0,
         "/Settings/Gps/Format": 0,
         "/Settings/Gps/SpeedUnit": "km/h",
         "/Settings/Gui/ColorScheme": 1,
--- a/pages/settings/PageControllableLoads.qml
+++ b/pages/settings/PageControllableLoads.qml
@@ -41,6 +41,14 @@ Page {
 			ListSwitch {
 				dataItem.uid: BackendConnection.serviceUidForType("platform") + "/Services/OpportunityLoads/Mode"
 				text: CommonWords.enabled
+				interactive: dEssModeItem.value === 0
+				//% "Opportunity loads cannot be enabled while Dynamic ESS is running. Disable Dynamic ESS first."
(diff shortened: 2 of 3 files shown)


data/mock/conf/services/opportunityloads.json
```diff
@@ -6,6 +6,6 @@
         "/BatteryLifeSupport": 1
     },
     "com.victronenergy.platform": {
-        "/Services/OpportunityLoads/Mode": 1
+        "/Services/OpportunityLoads/Mode": 0
```
blammit: Could this be 1 by default? It would make it easier to verify the OL feature in mock mode. It looks like there are not many UI features for DESS at the moment. What do you think?

blammit (approved):
One comment, otherwise LGTM

data/mock/conf/services/opportunityloads.json
DanielMcInnes: The problem with defaulting OL mode to 1 in mock mode, is that DESS mode (Global.systemSettings.serviceUid + "/Settings/DynamicEss/Mode") also defaults to enabled in mock mode, and you get a situation where both DynamicLoads and DESS are enabled, and the switches to control both are disabled, so you can't turn either of them off. This is only a problem in mock mode. On a real cerbo, Rein tells me that "venus-platform handles starting OL, and whenever DESS is started while OL is already running, it'll currently just disable OL."

blammit: What if we just disable DESS in mock mode by default? By changing `data/mock/conf/setup-common.json` to have `"/Settings/DynamicEss/Mode": 0`. I'm not sure if DESS is that valuable for mock mode at the moment, whereas the Opportunity Loads feature will need a fair bit of verification in the near future.

DanielMcInnes: sure, updated as suggested

PR #2926, opened 2026-04-15
Units: add rotation conversion to veutils

ybott-qinetic:
Improve VeQuickItemQuotient api to be consistent with VeQuickItem


blammit:
@ybott-qinetic aside from the `veutil` submodule update, are these changes required to fix the runtime warning, or is it more of a clean-up to align the values for `sourceUnit` and `displayUnit` throughout the codebase? If it's the latter, `SwitchableOutputSlider` also has a similar API that uses `sourceUnit` and `displayUnit` values that are gui-v2 units (those from `VenusOS.Units_*`) rather than `veutil` units (from `Unit::*`). So if the boat pages are being changed, the sliders should be changed too. In that case, it would help to improve the property names to something like `veSourceUnit` and `veDisplayUnit`, to make it clear they hold `veutil` units rather than `VenusOS.Units_*` units.

blammit (approved):

Issue #2750, opened 2026-01-05
Feature Idea: Group VRM Device Instances by service type


philipptrenz:
Recently, there were questions on Community about conflicting VRM Device Instances (see [here](https://community.victronenergy.com/t/same-instance-number-for-two-devices/50931/1)). If the VRM Device Instances overview in gui-v2 would group entries by service type / role, it would become obvious that the same device instances on different device types are not conflicting. As a bonus, it would also become more easy to find certain entries. Sorting the roles also alphabetically would support that even more.

Here a quick sketch:
<img width="1001" height="636" alt="Image" src="https://github.com/user-attachments/assets/70153884-8538-4f57-b10e-4af73325d7d8" />

(FYI @ThiemoVanEngelen)

chriadam:
Merged to main, closing as fixed.

PR #2930, opened 2026-04-18
VRM: add vrm device instance list sections

ybott-qinetic:
Contributes to #2750

### DIFF
--- a/src/classandvrminstancemodel.cpp
+++ b/src/classandvrminstancemodel.cpp
@@ -398,8 +398,8 @@ bool SortedClassAndVrmInstanceModel::lessThan(const QModelIndex &sourceLeft, con
 	const QModelIndex leftIndex = model->index(sourceLeft.row(), sourceLeft.column());
 	const QModelIndex rightIndex = model->index(sourceRight.row(), sourceRight.column());
 
-	const QString leftName = model->data(leftIndex, ClassAndVrmInstanceModel::NameRole).toString();
-	const QString rightName = model->data(rightIndex, ClassAndVrmInstanceModel::NameRole).toString();
+	const QString leftName = model->data(leftIndex, ClassAndVrmInstanceModel::DeviceClassRole).toString();
+	const QString rightName = model->data(rightIndex, ClassAndVrmInstanceModel::DeviceClassRole).toString();
 	if (leftName != rightName) {
 		if (leftName.isEmpty()) {
 			return false;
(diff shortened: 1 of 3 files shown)


src/classandvrminstancemodel.cpp
```diff
@@ -398,8 +398,8 @@ bool SortedClassAndVrmInstanceModel::lessThan(const QModelIndex &sourceLeft, con
 	const QModelIndex leftIndex = model->index(sourceLeft.row(), sourceLeft.column());
 	const QModelIndex rightIndex = model->index(sourceRight.row(), sourceRight.column());
 
-	const QString leftName = model->data(leftIndex, ClassAndVrmInstanceModel::NameRole).toString();
-	const QString rightName = model->data(rightIndex, ClassAndVrmInstanceModel::NameRole).toString();
+	const QString leftName = model->data(leftIndex, ClassAndVrmInstanceModel::DeviceClassRole).toString();
```
blammit: The sort role has changed to DeviceClassRole here, but this still uses the logic for sorting by name: in line 403-411, it checks whether leftName or rightName is empty, to ensure that unconnected devices are shown later, but this logic doesn't make sense for the Device Class, and also means the unconnected device check has been lost for the name sorting. So instead of only changing the role, swap the entire Device Class code block from 413-416 to above this one.

The comment describing the sort order in line 396 also needs updating.

ybott-qinetic: Done.

src/classandvrminstancemodel.cpp
```diff
@@ -393,11 +393,18 @@ bool SortedClassAndVrmInstanceModel::lessThan(const QModelIndex &sourceLeft, con
 		return QSortFilterProxyModel::lessThan(sourceLeft, sourceRight);
 	}
 
-	// Sort by connected devices first (i.e. those with a name), then by device class, then by VRM
-	// instance.
+	// Sort by device class first, then by device name, then by VRM instance value.
```
blammit: The part about "sorting by connected devices first" still applies, due to the `isEmpty()` string checks below, so keep that in the comment.

blammit (commented):
Sorry, forgot about this one. One minor comment, otherwise looks good to merge.

blammit (approved):

src/classandvrminstancemodel.cpp
ybott-qinetic: Done

blammit: The old comment still needs to be updated? The new sort order is:

1. Device class
2. Connected devices first
3. Device name
4. VRM instance

ybott-qinetic: Sorry for my misunderstanding.

blammit: I didn't explain it well the first time. LGTM, feel free to merge.

chriadam:
Rebased and added a quick unit test

PR #2943, opened 2026-04-23
Update file format rules, four space tab

ybott-qinetic:


.gitconfig
```diff
@@ -0,0 +1,4 @@
+[filter "tabspace"]
```
blammit, line 1: Make these changes in `.git/config` instead.

blammit (dismissed):
Can you add a commit message that describes the effect of the change?

PR #2947, opened 2026-04-27
Inverter: show correct state text in solar drilldowns

blammit:
Interpret the inverter state as per
https://github.com/victronenergy/venus/wiki/dbus#inverter, which is different from the state for solar chargers and multis.


src/enums.h
```diff
@@ -773,6 +773,15 @@ class Enums : public QObject
 	};
 	Q_ENUM(NodeRed_Mode)
 
+	// For inverter services (vedirect inverters & Inverter RS)
+	enum Inverter_State {
+		Inverter_State_Off = 0,
+		Inverter_State_LowPower = 1,
+		Inverter_State_Fault = 2,
+		Inverter_State_Inverting = 9,
```
chriadam: I think the values here should refer to e.g. System_State_Inverting etc rather than duplicating constants.

src/enums.cpp
```diff
@@ -381,6 +381,26 @@ QString Enums::digitalInput_stateToText(DigitalInput_State state) const
 	}
 }
 
+QString Enums::inverter_stateToText(Inverter_State state) const
```
chriadam: Can we just use systemStateToText() instead?
If so, maybe better to to just move that one to c++ rather than defining a subset one.

src/enums.h
chriadam: i.e. `Inverter_State_Inverting = System_State_Inverting`

src/enums.cpp
blammit: Yeah maybe that's easier. For some services, the states do not align with the system state, but in the inverter case they do. 

To be safe, we can use systemStateToText() for only the known states as documented in the API (0 / 1 / 2 / 9) and return "unknown" otherwise.

blammit: I've updated the PR to do this. It required moving systemStateToText() to enums.h as well.

chriadam (approved):

PR #2954, opened 2026-04-30
GitHub Actions & build script fixes

mr-manuel:
When build files are stored on a Windows-mounted path inside WSL, the clock skew between Windows and WSL can cause the build system to always consider files out of date. The build files are now written to a WSL-native location to avoid this issue.

### DIFF
--- a/.github/workflows/build-wasm.yml
+++ b/.github/workflows/build-wasm.yml
@@ -14,7 +14,7 @@ jobs:
 
     steps:
       - name: Checkout repository and submodules
-        uses: actions/checkout@v4
+        uses: actions/checkout@v6
         with:
           submodules: recursive
 
@@ -38,7 +38,7 @@ jobs:
           echo "OUTPUTDIR: ${{ env.OUTPUTDIR }}"
 
       - name: Check for a cached version
-        uses: actions/cache@v4
+        uses: actions/cache@v5
         id: cached-versions
         with:
           path: |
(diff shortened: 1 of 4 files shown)


mr-manuel:
Fixes also warnings in GitHub Actions:

> Node.js 20 actions are deprecated. The following actions are running on Node.js 20 and may not work as expected: actions/cache@v4, actions/upload-artifact@v4. Actions will be forced to run with Node.js 24 by default starting June 2nd, 2026. Node.js 20 will be removed from the runner on September 16th, 2026. Please check if updated versions of these actions are available that support Node.js 24. To opt into Node.js 24 now, set the FORCE_JAVASCRIPT_ACTIONS_TO_NODE24=true environment variable on the runner or in your workflow file. Once Node.js 24 becomes the default, you can temporarily opt out by setting ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION=true. For more information see: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/

mr-manuel:
> Running kernel seems to be up-to-date.
> No services need to be restarted.
> No containers need to be restarted.
> No user sessions are running outdated binaries.
> No VM guests are running outdated hypervisor (qemu) binaries on this host.
> error: unable to contact snap store
> Error: Process completed with exit code 1.

Fixes now also snap dependency problem and fails over to direct download from GitHub, if the snap store cannot be contacted

chriadam (approved):
LGTM, thanks

PR #2958, opened 2026-05-01
UI Controls: add ThreeLabelLayout labels aliases and ease requirements

blammit:
Using 'required' for various properties was useful while porting list items to use a Control base type, but in general use creates an unnecessary burden on using ThreeLabelLayout.

Also add label aliases to allow the labels to be customised more easily.

### DIFF
--- a/components/ThreeLabelLayout.qml
+++ b/components/ThreeLabelLayout.qml
@@ -24,15 +24,15 @@ GridLayout {
 	id: root
 
 	required property string primaryText
-	required property string secondaryText
-	required property string captionText
-
-	required property font primaryFont
-	required property int primaryTextFormat
-	property alias secondaryTextColor: secondaryLabel.color
+	property alias secondaryText: secondaryLabel.text
+	property alias captionText: captionLabel.text
 	property bool stretchSecondaryText
 	property real topPadding
 	property real bottomPadding
+
+	property alias primaryLabel: primaryLabel
+	property alias secondaryLabel: secondaryLabel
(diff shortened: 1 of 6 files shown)


components/ThreeLabelLayout.qml
```diff
@@ -24,16 +24,22 @@ GridLayout {
...
 	required property string primaryText
-	required property string secondaryText
-	required property string captionText
+	property string secondaryText
+	property string captionText
 
-	required property font primaryFont
-	required property int primaryTextFormat
+	// TODO since the labels are available via aliases, consider removing primaryTextFormat and
+	// secondaryTextColor properties.
```
chriadam: I would vote for this.  I get that it would make the PR larger since other things which use those properties currently would need to be changed, but I think it's a good change.

chriadam: Although it does raise the question: what about the primaryText / secondaryText / captionText properties.  Hmm.

mr-manuel: I would also remove this two. For the other aliases I would leave them, since they are the main ones used.

blammit: Sounds good - I've removed the additional primaryFont, primaryTextFormat and secondaryTextColor properties. Not sure if primaryText really needs to be `required`, but seems odd to make a ThreeLabelLayout without the primary text, so maybe it's good to force that to be explicitly set.

mr-manuel (approved):
LGTM

PR #2961, opened 2026-05-01
WASM: Add URL query string to display mock data

mr-manuel:
By adding `?mock` or `?mock=<config>` to the URL, the page will display mock data instead of real data.

Available configurations are in `data/mock/conf`. Just use the filename without the `.json` extension, e.g. `?mock=maximal`


wasm/index.html
```diff
@@ -383,7 +383,7 @@
             // If mqtt param is set: use the provided MQTT address
             // If id/shard params are set: gui-v2 will use VRM
             // If none of these are set, will use a default MQTT address.
-            const needsDefaultArgs = !searchParams.has("mqtt") && !searchParams.has("id") && !searchParams.has("shard");
+            const needsDefaultArgs = !searchParams.has("mqtt") && !searchParams.has("id") && !searchParams.has("shard") && !searchParams.has("mockdata");
```
chriadam: instead of mockdata maybe just mock?

mr-manuel: mock is already used for the mockup/Ekrano GX frame
https://github.com/victronenergy/gui-v2/wiki/Command-Line-and-Query-Parameters#webassembly-query-parameters

chriadam: Is there any chance we could make that one "frame" and then rename "mockdata" to "mock", just for symmetry with the command line arguments for gui-v2?

mr-manuel: For me that would be no problem. Should be discuss that in the next meeting, if it's ok for everyone?

mr-manuel: @chriadam done

chriadam (approved):

Issue #2971, opened 2026-05-05
Motordrive: rename motor drive to E-drive


citolen:
There are a few instances left where we still call it motor drive instead of E-drive.

Documentation must also be updated to change the term.

blammit:
@citolen Changes have been merged, can this task be closed?

PR #2972, opened 2026-05-05
Motordrive: rename to E-drive

citolen:
There were still a few instances left where we used the term "motor drive" instead of "E-drive".

Contributes to issue https://github.com/victronenergy/gui-v2/issues/2971


chriadam (approved):
LGTM

PR #2980, opened 2026-05-08
Add portrait layout

blammit:
Portrait layout is targeted at Wasm on mobile devices. 

On desktop, press 'o' in mock mode to switch between portrait and landscape.

Changes for landscape:
- In the battery list, the voltage/current/power measurements are now shown on the right (rather than left) edge of each list item. This makes it easier to align the layout with other list items with quantity rows in both portrait and landscape.


components/ThreePhaseDisplay.qml
```diff
@@ -17,7 +17,8 @@ Flow {
 	id: root
 
 	required property PhaseModel model
-	property int widgetSize: VenusOS.OverviewWidget_Size_S
+	required property int widgetSize
+	required property int fontPixeSize
 
 	// These properties are used to change the text color depending on the phase value.
```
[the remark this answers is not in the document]
blammit: Fixed.

pages/NotificationLayer.qml
```diff
@@ -59,18 +59,18 @@ Item {
...
-		spacing: Theme.geometry_toastNotification_horizontalMargin
+		// Use y binding instead of anchors so that it adjusts for portrait/landscape changes.
+		y: Theme.screenSize === Theme.Portrait ? Theme.geometry_toastNotification_verticalMargin
+				: parent.height - height - Theme.geometry_toastNotification_verticalMargin
+		height: 2*Theme.geometry_toastNotification_highlightWidth
+		spacing: Theme.geometry_page_content_verticalMargin
 		orientation: Qt.Horizontal
 		interactive: false
-		cacheBuffer: Theme.geometry_screen_width - 2*Theme.geometry_toastNotification_horizontalMargin
+		cacheBuffer: Theme.geometry_screen_width - 2*Theme.geometry_page_content_verticalMargin
```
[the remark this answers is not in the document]
blammit: Restored previous theme property.

components/dialogs/ModalDialog.qml
```diff
@@ -108,50 +139,89 @@ T.Dialog {
...
+				secondaryTitle: root.headerShowsTitles ? root.secondaryTitle : ""
+
+				// Allow the dialog to be cancelled by dragging it off-screen. It would be better to
+				// put this in the background, but on mobile the dialog background does not receive
+				// the mouse drag events.
+				DialogDragger {
+					anchors.fill: parent
+					z: -1 // avoid interfering with clicks on RoundCloseButton in the header.
+					dialog: root
+					shadow: root.shadow
```
chriadam, line 223: It's probably fine, because the way it's done there's a loose coupling and nothing is exposed by id.
But, still, it seems a bit strange to have something in the contentItem touch something in the background.

components/dialogs/ModalDialog.qml
```diff
@@ -169,10 +239,10 @@ T.Dialog {
 				right: parent.right
 				rightMargin: root.background.border.width
 			}
+			visible: Theme.screenSize !== Theme.Portrait
```
chriadam, line 251: Size?  Should it be orientation?

components/dialogs/NavBarMoreDialog.qml
```diff
@@ -0,0 +1,114 @@
...
+		}
+
+		// Spacer item, needed because the Repeater and nested NavBar cannot be stretched.
+		Item {
+			Layout.fillWidth: true
+			Layout.fillHeight: true
+		}
+
+		NavBar {
+			pages: root.pages
```
chriadam, line 109: Is this a NavBar which contains just the pages which are hidden/elided from the main NavBar?
If so, maybe the root pages property should be renamed hiddenPages for clarity, with a comment to explain where they are hidden.

components/dialogs/SecurityProfilePasswordDialog.qml
```diff
@@ -63,16 +63,19 @@ ModalDialog {
 	Component.onCompleted: resetValidation()
 
 	contentItem: ModalDialog.FocusableContentItem {
-		id: passwordEntryItem
+		implicitHeight: contentColumn.height
```
chriadam: should this be bound to the implicitHeight of the contentColumn, and then have the contentColumn effectively fill its parent?  or does ColumnLayout not expose the "children's preferred geometry" as implicitHeight the way I assume?  Or is there some other reason to do it this way?

components/EnvironmentGaugePanel.qml
```diff
@@ -6,98 +6,188 @@
...
 	}
 
-	Loader {
-		id: humidityGaugeLoader
+	contentItem: Item {
+		// In landscape, stretch the height to fill the parent.
+		// In portrait, size the height to the gauge content.
+		implicitHeight: Theme.screenSize !== Theme.Portrait ? 0
+				: (nameLabel.text.length ? nameLabel.height : 0) + gaugeFlow.height
+
```
chriadam, line 82: should define an implicitWidth if possible, also

components/NavBar.qml
```diff
@@ -56,45 +81,89 @@ FocusScope {
...
+			icon.source: "qrc:/images/icon_more_dots.svg"
+			visible: root._showMoreButton
+			checked: root.moreButtonActive
+			enabled: !root.moreButtonActive
+
+			onClicked: {
+				// The NavBarMoreDialog contains a NavBar, so open it via URI instead of a nested
+				// component. Note the component URI resolution won't work for locally-installed
+				// files on the GX, but this is fine because the "More" feature won't show there.
+				const component = Qt.createComponent("dialogs/NavBarMoreDialog.qml")
```
chriadam: this might also cause an issue with the MQTT plugin reloading stuff if we end up having to call QQmlEngine::clearComponentCache() to get the new plugin content :-/  Maybe we should disallow automatic plugin reloading, and require the user to refresh the browser in that case, instead...

chriadam: Or alternatively, maybe we can remove the NavBar from within NavBarMoreDialog, as that does seem a bit strange to me.

components/QuantityTableMetrics.qml
chriadam, line 15: I wonder if we can move this to C++?  anyway, not for this PR.

data/DcInputs.qml
```diff
@@ -21,5 +21,37 @@ QtObject {
 		uid: Global.systemSettings.serviceUid + "/Settings/Gui/Gauges/Dc/Input/Power/Max"
 	}
 
+	function overviewWidgetTypeForService(serviceType, dcMeterType = -1) {
```
chriadam, line 24: this change seems somewhat unrelated to portrait mode per-se, but perhaps I'm mistaken

chriadam:
Github is struggling with this one ;-)
For the most part, looks good to me, I had a few comments and I haven't done any thorough testing yet.
Really great work!

components/dialogs/ModalDialog.qml
blammit: Orientation would make more sense, but this also fits into how we use a different set of geometries based on the `ScreenSize` value, since we need to vary the theme geometry values based on portrait vs landscape as well.

How about changing `ScreenSize` to `ScreenGeometry`? Then it would be `Theme.screenGeometry`, referencing the enum: 

```
enum ScreenGeometry {
        FiveInch = 0,
        SevenInch,
        Portrait
    };
    Q_ENUM(ScreenGeometry)
```

components/dialogs/ModalDialog.qml
blammit: Maybe an alternative is to put `DialogShadow {}` in the root dialog, and then set its `parentItem` to the `background`?

components/dialogs/NavBarMoreDialog.qml
blammit: These are the normal pages that are shown in the NavBar. This basically shows a replica of the normal NavBar, but inside a pop-up dialog. Will add a comment to explain this.

components/NavBar.qml
blammit: Yeah, this was a bit weird. I've updated it to construct NavBarMoreDialog from within MainView, so that NavBar doesn't create a nested version of itself. Thanks!

components/QuantityTableMetrics.qml
blammit: Yeah, that would be a good idea.

data/DcInputs.qml
blammit: This is to refactor out some common code between the landscape and portrait versions of the Overview page.

components/dialogs/SecurityProfilePasswordDialog.qml
blammit: Yes, makes sense to bind to implicitHeight, done as suggested.

I've updated this elsewhere as well, wherever we bind contentItem.height to a layout height. There are some other implicit geometry bindings that I'm not sure about, that we can discuss separately.

components/EnvironmentGaugePanel.qml
blammit: Done

blammit:
Also updated to fix dialog scaling issues in Wasm.

blammit:
Updated to fix an issue with the geometry/padding for Button.

blammit:
Updated various font sizes, and fixed an issue with text field focus within dialogs.

chriadam (approved):
LGTM.  A couple of little nitpicks found during testing (mentioned offline) but they shouldn't block the PR from being merged.  Fantastic work.

blammit:
There's an issue with the scaling of the dialogs in Wasm when the browser window is shrunk beyond a certain point. Will investigate and update the PR.

blammit:
> There's an issue with the scaling of the dialogs in Wasm when the browser window is shrunk beyond a certain point. Will investigate and update the PR.

This is now fixed.

chriadam (approved):

PR #2981, opened 2026-05-08
Switches: fix auto-scroll when focusing Numeric input control in Wasm

blammit:
Delay the auto-scroll until the item actually receives focus; otherwise, the native VKB does not appear.

Also push the text field further down, so that it is not obscured by the status bar.


chriadam (approved):

mr-manuel:
Fixes the issue for me

PR #2987, opened 2026-05-12
Switches: ensure momentary button is released

### LINKED ISSUE #2949: Switches: momentary type doesn't release state on redraw
It is possible to get the momentary output type stuck in the pressed state.

If you press the button and a redraw of the switch page happens, releasing your finger won't release the button state.
It remains stuck in the pressed state until pressed again.

Here is a recording of the issue.
To reproduce I pressed the momentary switch then while keeping it pressed, I changed the type of another channel from output to input.

https://github.com/user-attachments/assets/0db0e385-cc09-485e-bba6-030d60c69434

Garmin reported this issue.

blammit:
If the momentary button is pressed, then the component is destroyed before it can be released, then automatically write State=0 to prevent the button from staying in the "on" state. This happens, for example, if the view model is updated (causing the delegates to be recreated) while the button is pressed.

Also, write State=0 (as if the button was released) as soon as the touch moves outside of the button touch zone, instead of waiting for the touch release. Otherwise, if the user presses and drags outside of the button and begins to trigger a drag in the underlying Flickable, the touch event can no longer be released, and the backend State value remains "on".

Fixes #2949

### DIFF
--- a/components/switches/delegates/SwitchableOutputCardDelegate_0.qml
+++ b/components/switches/delegates/SwitchableOutputCardDelegate_0.qml
@@ -51,6 +51,10 @@ Item {
 	}
 
 	MomentaryButton {
+		readonly property bool containsPress: pressed
+				&& pressX >= 0 && pressX <= width
+				&& pressY >= 0 && pressY <= height
+
 		anchors {
 			left: parent.left
 			right: parent.right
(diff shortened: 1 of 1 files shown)


chriadam (approved):
LGTM, tested on both wasm + desktop (windows) with mock

Issue #2963, opened 2026-05-04
Further Opportunity Loads improvements


philipptrenz:
Please find all information on Figma: [Philipps Playground – OL v3](https://www.figma.com/design/GjCtz3A3jqSGvyjVvOm6RZ/Venus-OS--GX-Touch-?node-id=19576-11070&t=mkLc3rn7qICqOX0o-1)

PR #2995, opened 2026-05-13
Opportunity loads: allow loading of generic services

blammit:
Aside from battery and evcharger services, the Opportunity Loads page may need to show acload, heatpump and any other services with /S2/0/Rm* values. Update the page to load devices generically based on service type and device instance, instead of only loading them from specific device models based on service type.

Also hide the fields for each /S2/0/Rm* field if the value is not available in the backend.

Contributes to #2963


components/listitems/ListDevicePriority.qml
```diff
@@ -49,7 +48,7 @@ ListSetting {
 	}
 
 	Component {
-		id: acLoadData
+		id: otherData
 
 		PageData {
 			property VeQuickItem powerSetting: VeQuickItem {
```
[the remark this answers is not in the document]
blammit: Oops - fixed

components/listitems/ListDevicePriority.qml
```diff
@@ -14,11 +14,10 @@ ListSetting {
 	required property string serviceType
 	required property int deviceInstance
 	required property string uniqueIdentifier
-	required property FilteredDeviceModel devices
 
-	property Device _device: devices && deviceInstance >= 0 ? devices.deviceForDeviceInstance(deviceInstance) : null
 	readonly property string pageSource: pageData.item?.pageSource || ""
 	readonly property ListModel _model: ListView.view.model
+	readonly property Device _device: AllDevicesModel.findDeviceWithTypeAndInstance(serviceType, deviceInstance)
```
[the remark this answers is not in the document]
blammit: Oops - fixed

components/listitems/ListDevicePriority.qml
DanielMcInnes, line 63: maybe a comment explaining what S2 and RmSettings mean?

DanielMcInnes (approved):

DanielMcInnes (approved):

components/listitems/ListDevicePriority.qml
DanielMcInnes: I had:

		// 'S2' is a communication standard for energy flexibility in homes and buildings, see https://s2standard.org/
		// The /S2/... interface basically makes the service controllable as a flexible load through an energy management system.
		// According to that standard, the service thereby exposes a "Resource Manager", short "RM", which is an abstract, high-level description and
		// controller for that service. The "RMSettings" are custom settings the resource manager needs to understand how to control that service/device/resource.

DanielMcInnes (approved):

components/listitems/ListDevicePriority.qml
blammit: Sounds good, updated and added you as co-author.

Also renamed `PageControllableLoadsOther.qml` to `PageControllableLoadsS2Rm.qml`.

PR #2998, opened 2026-05-15
Virtual OS keyboard: track OS keyboard height via JS→C++→QML bridge

mr-manuel:
Add Theme.keyboardHeight (EMSCRIPTEN_BINDINGS) to relay the native
keyboard height from JS to QML; Screen.heightChanged does not fire
on WASM for keyboard events.

index.html: pin canvas with visualViewport tracking, switch to
100lvh, add interactive-widget=resizes-visual-viewport, and detect
keyboard open/close via navigator.virtualKeyboard or visualViewport.

WasmVirtualKeyboardHandler: scroll only when the field is outside
the visible area; position it at the bottom of the viewport rather
than the top; add a dismiss-guard timer for Qt re-focus during close.

ModalDialog: position above the keyboard using the accurate height;
add the same dismiss-guard pattern for WASM.

### DIFF
--- a/i18n/venus-gui-v2_af.ts
+++ b/i18n/venus-gui-v2_af.ts
@@ -3208,52 +3208,52 @@ Gebruik tensy dit deur die diensverskaffer verbied word.</translation>
       <translation>Laespanning ontkoppel (ignoreer altyd)</translation>
     </message>
     <message id="batterysettings_battery_bank">
-      <location filename="../../pages/settings/devicelist/battery/PageBatterySettings.qml" line="38"/>
+      <location filename="../../pages/settings/devicelist/battery/PageBatterySettings.qml" line="46"/>
       <source>Battery bank</source>
       <translation>Battery bank</translation>
     </message>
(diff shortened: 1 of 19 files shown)


blammit:
This isn't working correctly yet on iOS - in the control cards, the card does not scroll up to the correct position. Investigation shows that `Theme.keyboardHeight` is not set at the time that `onAboutToFocusTextField()` is triggered in `WasmVirtualKeyboardHandler`, and not later in the `Qt.callLater()` call either, or even when the item becomes the window `activeFocusItem`. It appears that `Theme.keyboardHeight` is not set until the keyboard is fully on-screen. So, the current code does not work because it tries to set up the geometry at the time that the text field is clicked, which is before the `Theme.keyboardHeight` is set.

blammit:
Perhaps we can drop most of the code in WasmVirtualKeyboardHandler and just resize the view according to Theme.keyboardHeight. Will investigate.

mr-manuel:
Closed by mistake. Will open a new PR when code is ready. Copilow review changes where applied.

PR #3000, opened 2026-05-16
EVCS: show per-phase current in phase table

wafgo:
When an EV charger driver publishes per-phase current values via the standard Victron D-Bus paths .../Ac/L1/Current, .../Ac/L2/Current, .../Ac/L3/Current (already part of the com.victronenergy.evcharger spec), expose them in the per-phase rows of the EvCharger page next to the per-phase power.

Chargers that do not publish per-phase current continue to render an empty cell in that column: the underlying PhaseModel role defaults to NaN, and the QuantityObject is bound to hidden=isNaN(current), so the column layout stays identical to the previous behaviour for those drivers and only chargers that already provide the data show the new information. No new D-Bus paths are introduced; the change is purely on the UI consumer side.

The energy and charging-time columns remain summary-only (padded with hidden QuantityObjects in the phase rows), matching the existing design intent.

### DIFF
--- a/pages/evcs/EvChargerPage.qml
+++ b/pages/evcs/EvChargerPage.qml
@@ -74,14 +74,22 @@ DevicePage {
 
 					required property string name
 					required property real power
+					required property real current
 
 					headerText: name
 					model: QuantityObjectModel {
-						// The current, energy and charging time columns are only relevant to
-						// the summary and not the individual devices, so just add empty values
-						// here to pad out the remaining columns.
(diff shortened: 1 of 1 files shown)


blammit:
Hi @wafgo, thank you for your contribution. We discussed this and actually we'll be making similar changes for an upcoming feature, along with some related UI changes, so we won't merge this one. But we are looking into providing more per-phase data in this page.

Issue #2840, opened 2026-02-19
Wifi: Hide gateway field when internet access is disabled


ReinvdZee:
When the toggle "Allow using wifi for internet access" is disabled, the gateway address is cleared but still visible in the Wifi page.

Instead, the gateway field should be hidden when the toggle is disabled. 

<img width="960" height="563" alt="Image" src="https://github.com/user-attachments/assets/0779e51c-a482-49c1-8f05-d13a9ee0e8b9" />

ReinvdZee:
Raising the prio, as this should be a simple fix but valuable to the user.

ybott-qinetic:
@ReinvdZee Just to clarity the exact functionality. Do you want the Gateway value to show only when "Allow using wifi for internet access" is true? Or do you want the Gateway value to show when it has valid contents? 
I checked with gui-v1 and for wifi if I turn off the "Allow using ... internet access" toggle the gateway value will continue to show.

ybott-qinetic:
Clarity provided from Rein: Hide the field if the toggle is disabled.

chriadam:
Merged to main, closing as fixed.  Might be worth discussing the final semantics in the meeting today (i.e. the wifi gateway value seems to get set to null by the system if ethernet is also connected, so we decided to hide the wifi gateway field in that case also).

PR #3010, opened 2026-05-21
Settings: hide gateway field when wifi internet access is disabled

ybott-qinetic:
Contributes to #2840


chriadam:
commit subject should be something like: "Settings: hide gateway field when internet access is disabled" for clarity

pages/settings/NetworkSettingsPageModel.qml
```diff
@@ -155,7 +155,7 @@ VisibleItemModel {
 		//% "Gateway"
 		text: qsTrId("settings_tcpip_gateway")
 		interactive: method.userHasWriteAccess && networkServices.manual
-		preferredVisible: networkServices.wifi || ethernetGatewayEnabled.checked
```
chriadam, line 158: instead of removing the networkServices.wifi check, maybe it should just use && instead of || ?  Or am I misunderstanding what the validity constraint should be?

ybott-qinetic: Is the network settings page used for all network settings? Not just wifi?

chriadam: I'm not sure, but according to the task #2840 the toggle says "Allow using wifi for internet access" so I would imagine that the wifi would have to be available for this to be visible.  But maybe I'm wrong - I don't know the network settings stuff at all.

chriadam: Maybe ask Rein tonight at the meeting.

pages/settings/NetworkSettingsPageModel.qml
```diff
@@ -155,7 +155,7 @@ VisibleItemModel {
...
+		preferredVisible: ethernetGatewayEnabled.checked
```
DanielMcInnes: If I understand correctly, this change hides the "Gateway" setting if the cerbo is connected via wifi instead of ethernet. Why would we want to do that? Does "Gateway" only make sense for a wired connection? I don't know much about networking :)

pages/settings/NetworkSettingsPageModel.qml
blammit: > Is the network settings page used for all network settings? Not just wifi?

It's used for ethernet as well.

blammit (commented):
I can still reproduce the issue on GX. If I turn off the "Allow using wifi for internet access" option, then open the page for the connected wifi AP, it shows the "Gateway" list item with the "--" text.

Also, the commit message refers to "internet sharing" but the issue is about access rather than sharing.

pages/settings/NetworkSettingsPageModel.qml
```diff
@@ -155,10 +155,15 @@ VisibleItemModel {
...
 		interactive: method.userHasWriteAccess && networkServices.manual
-		preferredVisible: networkServices.wifi || ethernetGatewayEnabled.checked
+		preferredVisible: (networkServices.wifi && wifiGatewayEnabled.value === true) || (!networkServices.wifi && ethernetGatewayEnabled.checked)
 		writeAccessLevel: VenusOS.User_AccessType_User
 		secondaryText: networkServices.gateway
 		saveInput: function() { networkServices.setServiceProperty("Gateway", secondaryText) }
+
+		VeQuickItem {
+			id: wifiGatewayEnabled
+			uid: Global.venusPlatform.serviceUid + "/Network/Wifi/GatewayEnabled"
```
chriadam: I would do
```
uid: networkServices.wifi ? Global.venusPlatform.serviceUid + "/Network/Wifi/GatewayEnabled" : ""
```

pages/settings/NetworkSettingsPageModel.qml
```diff
@@ -155,10 +155,15 @@ VisibleItemModel {
 		//% "Gateway"
 		text: qsTrId("settings_tcpip_gateway")
 		interactive: method.userHasWriteAccess && networkServices.manual
-		preferredVisible: networkServices.wifi || ethernetGatewayEnabled.checked
+		preferredVisible: (networkServices.wifi && wifiGatewayEnabled.value === true) || (!networkServices.wifi && ethernetGatewayEnabled.checked)
```
chriadam: I would change `networkServices.wifi` to `wifiGatewayEnabled.valid` after updating the uid binding in wifiGatewayEnabled to include the networkServices.wifi check.

pages/settings/NetworkSettingsPageModel.qml
```diff
@@ -155,10 +155,16 @@ VisibleItemModel {
 		//% "Gateway"
 		text: qsTrId("settings_tcpip_gateway")
 		interactive: method.userHasWriteAccess && networkServices.manual
-		preferredVisible: networkServices.wifi || ethernetGatewayEnabled.checked
+		preferredVisible: (wifiGatewayEnabled.valid && (wifiGatewayEnabled.value === true || wifiGatewayEnabled.value === 1))
```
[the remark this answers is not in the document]
chriadam: Nope.

blammit (dismissed):

chriadam:
Indeed.  It seems that if the system has a valid ethernet connection, and that ethernet connection has Internet Gateway enabled, the system will always prefer Ethernet for internet network routing, and so the wifi gateway will be unset.  I will add a check to the wifi page, to check whether there is also a valid ethernet connection, and if so, if its gateway is enabled, and if so, to hide the wifi gateway field.

pages/settings/PageSettingsWifi.qml
```diff
@@ -187,7 +187,7 @@ Page {
 			Component {
 				id: wifiPointComponent
 
-				PageSettingsTcpIp {
+				PageSettingsEthernet {
```
blammit: Naming-wise, it seems like `PageSettingsTcpIp` should be the common component between the ethernet/wifi pages, and for`PageSettingsEthernet` to be removed instead.

blammit (approved):
LGTM, minor comment about naming.

pages/settings/PageSettingsWifi.qml
chriadam: Done

PR #3015, opened 2026-05-21
Overview: show VE.Bus state if system state is not available

blammit:
If the system state is not available, the Inverter/Charger box should show the state of the system VE.Bus service instead.

This aligns with the gui-v1 behaviour.


blammit:
Actually, as per https://victrondevelopment.slack.com/archives/C020QG3JVEW/p1779363186822549?thread_ts=1779277244.018819&cid=C020QG3JVEW, just showing the system state is correct. Closing.

PR #3016, opened 2026-05-21
Overview: rename "Solar yield" to "Solar"

### LINKED ISSUE #3014: Change Overview solar tile name from "Solar yield" to "Solar"
"Solar yield" is about the energy it yields, and the tile only shows energy values in some cases, namely when only PV chargers with solar history are present. (To reduce confusion, history bars are not shown when any PV inverters are present, as those do not provide history.)

blammit:
"Solar yield" is about the energy it yields, and the tile only shows energy values in some cases (namely when only PV chargers with solar history are present) so just "Solar" is more appropriate.

Fixes #3014

### DIFF
--- a/pages/BriefSidePanel.qml
+++ b/pages/BriefSidePanel.qml
@@ -22,8 +22,7 @@ ColumnLayout {
 	spacing: Theme.geometry_sidePanel_spacing
 
 	BriefSidePanelWidget {
-		//% "Solar yield"
-		title: qsTrId("brief_solar_yield")
+		title: CommonWords.solar
 		icon.source: "qrc:/images/solaryield.svg"
 		loadersActive: Global.solarInputs.devices.count > 0 // only show graph if there are solar inputs with history (i.e. not PV inverters)
 		visible: Global.solarInputs.inputCount > 0 // show if there are any solar inputs (PV chargers, PV inverters, etc.)
--- a/components/CommonWords.qml
+++ b/components/CommonWords.qml
@@ -496,6 +496,9 @@ QtObject {
 	//% "SOC %1"
 	readonly property string soc_with_prefix: qsTrId("common_words_soc")
 
+	//% "Solar"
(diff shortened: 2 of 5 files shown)


chriadam:
The commit message confused me a bit, perhaps it could be changed to something like:

```
Solar yield is a measure of (accumulated) energy generated over time,
and this data is only available from PV inverter/chargers with solar history.
When the system does not include a PV inverter/charger with history,
we will show the instantaneous power value on the Overview tile rather
than yield, so we should rename the tile to just "Solar" instead of "Solar Yield".
```

pages/BriefSidePanel.qml
```diff
@@ -22,8 +22,7 @@ ColumnLayout {
 	spacing: Theme.geometry_sidePanel_spacing
 
 	BriefSidePanelWidget {
-		//% "Solar yield"
-		title: qsTrId("brief_solar_yield")
+		title: CommonWords.solar
```
chriadam, line 25: maybe this should be:
```
title: loadersActive ? "Solar Yield" : CommonWords.solar
```

blammit: Would be good to keep the name consistent I think, for reference/documentation purposes as well.

blammit:
> The commit message confused me a bit, perhaps it could be changed to something like:
> 
> ```
> Solar yield is a measure of (accumulated) energy generated over time,
> and this data is only available from PV inverter/chargers with solar history.
> When the system does not include a PV inverter/charger with history,
> we will show the instantaneous power value on the Overview tile rather
> than yield, so we should rename the tile to just "Solar" instead of "Solar Yield".
> ```

Sounds good, done.

pages/BriefSidePanel.qml
chriadam: Fair enough

chriadam (approved):

PR #3017, opened 2026-05-22
Solar: order solar trackers by tracker index instead of name

### LINKED ISSUE #2852: Solar trackers are ordered differently in different drilldowns
In some places they seem to be ordered alphabetically ascending, in others alphabetically descending.  See the following screenshot:

Image

Original report: https://community.victronenergy.com/t/feature-request-keep-external-control-sorted-in-the-new-gui/53307

blammit:
In the solar input list, when there are multiple solar trackers for the same PV charger, then sort those trackers by tracker index instead of by name.

Fixes #2852

### DIFF
--- a/src/solarinput.h
+++ b/src/solarinput.h
@@ -87,6 +87,7 @@ class TrackerSolarInput : public SolarInput
 	Q_OBJECT
 public:
 	explicit TrackerSolarInput(Device *device, bool isSingleTrackerDevice, int trackerIndex, QObject *parent = nullptr);
+	int trackerIndex() const;
 
 	static bool isEnabledTracker(VeQItem *serviceItem, int trackerIndex);
 
--- a/src/solarinputmodel.cpp
+++ b/src/solarinputmodel.cpp
@@ -40,6 +40,18 @@ int SolarInputModel::count() const
 	return m_enabledInputs.count();
 }
 
+int SolarInputModel::trackerIndex(int row) const
+{
+	if (row < 0 || row >= m_enabledInputs.count()) {
+		return -1;
+	}
+	if (TrackerSolarInput *input = qobject_cast<TrackerSolarInput *>(m_enabledInputs.at(row))) {
+		return input->trackerIndex();
+	} else {
+		return -1;
+	}
+}
+
 QVariant SolarInputModel::data(const QModelIndex &index, int role) const
 {
 	const int row = index.row();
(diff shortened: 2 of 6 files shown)


src/solarinput.h
```diff
@@ -87,6 +87,7 @@ class TrackerSolarInput : public SolarInput
 	Q_OBJECT
 public:
 	explicit TrackerSolarInput(Device *device, bool isSingleTrackerDevice, int trackerIndex, QObject *parent = nullptr);
+	int trackerIndex() const;
```
chriadam, line 90: any reason why we don't want to expose this as a Q_PROPERTY ?

src/solarinputmodel.cpp
```diff
@@ -296,17 +308,28 @@ SortedSolarInputModel::SortedSolarInputModel(QObject *parent)
...
-		const QString leftName = sourceModel()->data(
-				sourceModel()->index(sourceLeft.row(), sourceLeft.column()), SolarInputModel::NameRole).toString();
-		const QString rightName = sourceModel()->data(
-				sourceModel()->index(sourceRight.row(), sourceRight.column()), SolarInputModel::NameRole).toString();
-		return leftName.localeAwareCompare(rightName) < 0;
+		if (sourceModel()->data(leftIndex, SolarInputModel::ServiceUidRole) == sourceModel()->data(leftIndex, SolarInputModel::ServiceUidRole)) {
+			// If the two inputs belong to the same service, then these must be both trackers from
+			// the same device, so sort them by tracker index.
+			if (SolarInputModel *model = qobject_cast<SolarInputModel *>(sourceModel())) {
+				return model->trackerIndex(leftIndex.row()) < model->trackerIndex(rightIndex.row());
```
chriadam: I think it would be worth ensuring that the tracker index isn't -1 when doing the comparison, and perhaps emitting a warning in the case where it is (and then falling back to name-based sorting).

blammit: Done

src/solarinput.h
blammit: It wouldn't be useful as it's not needed outside of the C++ model at the moment.

chriadam (approved):

src/solarinputmodel.cpp
blammit: Discussed the warning is not needed, as PV inverters always have trackerIndex=-1. Added a comment to explain this instead.

PR #3018, opened 2026-05-22
UI: add demo mode indicator

### LINKED ISSUE #2249: Add demo mode indicator
The old ui has a feature to show the user that demo mode is enabled. The reason to do that is that it can be quite confusing if Demo mode is enabled but as a user or technician doing trouble shooting you don't know.

Image

blammit:
When the /Settings/Gui/DemoMode local setting indicates that a demo mode is active, show a "Demo mode" label in the top-left of the app view.

Also adjust the left/right padding of SettingsListHeader so that they align with the text of adjacent list items. Otherwise, the header text is too close to the demo mode indicator.

Fixes #2249

### DIFF
--- a/components/DemoModeIndicator.qml
+++ b/components/DemoModeIndicator.qml
@@ -1,0 +1,29 @@
+/*
+** Copyright (C) 2026 Victron Energy B.V.
+** See LICENSE.txt for license information.
+*/
+
+import QtQuick
+import Victron.VenusOS
+
+Rectangle {
+	implicitWidth: modeLabel.implicitWidth
+	implicitHeight: modeLabel.implicitHeight
+	topLeftRadius: Theme.geometry_button_radius
+	topRightRadius: Theme.geometry_button_radius
+	color: Theme.color_demoModeIndicator_background
+
+	Label {
+		id: modeLabel
+
+		leftPadding: Theme.geometry_demoModeIndicator_horizontalPadding
+		rightPadding: Theme.geometry_demoModeIndicator_horizontalPadding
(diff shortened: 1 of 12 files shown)


components/DemoModeIndicator.qml
```diff
@@ -0,0 +1,32 @@
+/*
+** Copyright (C) 2026 Victron Energy B.V.
+** See LICENSE.txt for license information.
+*/
+
+import QtQuick
+import Victron.VenusOS
+
+Rectangle {
+	y: -height
```
chriadam: This seems very strange.  Normally, the positioning would be done by whatever instantiates the component.

blammit: It's to offset the rotation, but it is a bit strange - I'll move the rotation and the y change to the parent.

blammit: Done.

components/DemoModeIndicator.qml
```diff
@@ -0,0 +1,29 @@
...
+** See LICENSE.txt for license information.
+*/
+
+import QtQuick
+import Victron.VenusOS
+
+Rectangle {
+	implicitWidth: modeLabel.implicitWidth
+	implicitHeight: modeLabel.implicitHeight
+	topLeftRadius: 8
```
chriadam: I wonder if the radius should be a theme constant?

blammit: Oops, fixed

chriadam (approved):
No nitpicks left, LGTM

PR #3019, opened 2026-05-22
DESS: fix the Mode menu item and warning

jhofstee:
Oportunity loads is an optional feature and is not installed at all at the moment and won't not be available for all devices. If it is not installed the /Settings/DynamicEss/Mode in venus-platform doesn't exists, hence the comparison with zero won't work. Correct this by checking if the service is not enabled.

Also correct the warning text, opportunity loads must be disabled, not DESS.

Fixes cfa0d6f104e8 ("Settings: Disable Opportunity Loads when DESS is running")


DanielMcInnes (approved):

DanielMcInnes:
Sorry Jeroen!

PR #3020, opened 2026-05-26
UI Controls: ensure the switch in a disabled ListItem cannot be toggled

blammit:
When disabling a switch, set Switch::enabled, instead of trying to disable the behaviour from the onClicked handler. Otherwise, the user is still able to toggle the switch with a swipe/drag action.

Also remove the Switch::showEnabled property as it is no longer used.


DanielMcInnes (approved):
lgtm

PR #3021, opened 2026-05-26
Boat: add range, consumption, new gear indicator and optional temperatures

citolen:
New boat page features:
- redesigned gear indicator
- time to go moved from bottom center to bottom left slot
- add range and consumption
- add setting to display temperatures, off by default
- display temperatures

Requires a new setting: `Gui/ElectricPropulsionUI/ShowTemperatures 0 i 0 1`

The temperatures are intentionally behind the navigation bar, they will only show once the navigation bar has disappeared.

5" dual  
image

7" dual  
image

5" single  
image

7" single  
image

Settings:  
image


pages/boat/BoatPage_Landscape.qml
```diff
@@ -149,6 +131,24 @@ Boat.Background { // the blue shadows
...
+
+	Boat.Temperatures {
+		id: temperatures
+
+		visible: showTemperaturesItem.value ?? false
+
+		anchors {
+			horizontalCenter: parent.horizontalCenter
+		}
+		y: Theme.geometry_screen_height - Theme.geometry_statusBar_height - temperatures.height - Theme.geometry_boatPage_temperature_bottomMargin
```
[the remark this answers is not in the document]
citolen: This is intended, we want the temperatures to sit behind the nav bar.

blammit:
Updated some QML files to use tabs instead of spaces to align with other files in the code base.

pages/boat/BoatPage_Landscape.qml
```diff
@@ -149,6 +131,24 @@ Boat.Background { // the blue shadows
...
 
+	VeQuickItem {
+		id: showTemperaturesItem
+		uid: !!Global.systemSettings ? Global.systemSettings.serviceUid + "/Settings/Gui/ElectricPropulsionUI/ShowTemperatures" : ""
+	}
+
+	Boat.Temperatures {
+		id: temperatures
+
+		visible: showTemperaturesItem.value ?? false
```
blammit: Nitpick: would put the `visible` declaration below the anchors/y positioning, to align with code elsewhere.

pages/boat/Consumption.qml
```diff
@@ -0,0 +1,66 @@
...
+
+	VeQuickItem {
+		id: consumptionItem
+
+		uid: Global.system.serviceUid ? Global.system.serviceUid + "/MotorDrive/" + (root.useWatt ? "ConsumptionWhkm" : "ConsumptionAhkm") : ""
+		sourceUnit: Units.unitToVeUnit(root.sourceUnit)
+		displayUnit: Units.unitToVeUnit(root.displayUnit)
+	}
+
+	visible: consumptionItem.valid
```
blammit: Can you put this in line 39, so that it goes above the `VeQuickItem` children?

pages/boat/Gear.qml
```diff
@@ -5,66 +5,46 @@
...
-	spacing: Theme.geometry_boatPage_gearRow_spacing
-	visible: motorDrive.direction.valid
-
-	component GearIndicator : Item {
-		required property int gear
-		property alias text: gearLabel.text
-
-		width: Theme.geometry_boatPage_gearHighlighter_width
-		height: Theme.geometry_boatPage_gearHighlighter_height + Theme.geometry_boatPage_gear_verticalMargin + gearLabel.height
+	width: gearLabel.width
```
blammit: Can set `implicitWidth` and `implicitHeight` instead of `width` and `height` here, since this is setting the root item's preferred size based on its children.

pages/boat/LargeCenterGauge.qml
```diff
@@ -237,6 +238,40 @@ Item {
...
+
+		anchors {
+			horizontalCenter: rpmGauge.horizontalCenter
+			bottom: rpmGauge.bottom
+			bottomMargin: Theme.geometry_boatPage_singleRpmLabel_bottomMargin
+		}
+		width: 2 * rpmGauge.radius * Math.sin(((outerGauge.endAngle - outerGauge.startAngle) * Math.PI / 180) / 2)
+		visible: rpmGauge.visible
+
+		spacing: 5
```
blammit: Refer to a theme variable instead of a hardcoded value.

pages/boat/LargeCenterGauge.qml
```diff
@@ -257,23 +292,25 @@ Item {
...
+				bottom: leftRpmLabel.bottom
 			}
-			width: Theme.geometry_boatPage_dualRpmSeparator_width
+
+			verticalAlignment: Text.AlignVCenter
+			height: font.pixelSize
+			font.pixelSize: Theme.font_boatPage_rpm_label_pixelSize
 			color: Theme.color_font_secondary
-			visible: Theme.screenSize === Theme.FiveInch
+			text: qsTrId("boat_page_rpm")
```
blammit: Maybe use `text: singleRpmUnitLabel.text` instead, in case the translation id changes or disappears later on.

pages/boat/Range.qml
```diff
@@ -0,0 +1,51 @@
...
+	id: root
+
+	readonly property int displayUnit: {
+		switch (Global.systemSettings.speedUnit) {
+			case VenusOS.Units_Speed_MilesPerHour:
+				return VenusOS.Units_Mile;
+			case VenusOS.Units_Speed_Knots:
+				return VenusOS.Units_Nautical_Mile;
+			case VenusOS.Units_Speed_KilometresPerHour:
+			case VenusOS.Units_Speed_MetresPerSecond:
```
blammit, line 18: For EVs, when showing their `/RangeToGo` value in the Devices list, we use `VenusOS.Units_Metre` (rather than VenusOS.Units_Kilometre) if the system speed unit is `VenusOS.Units_Speed_MetresPerSecond`, as per [this code](https://github.com/victronenergy/gui-v2/blob/main/pages/settings/devicelist/delegates/DeviceListDelegate_ev.qml#L16). Should it be the same here, or are they different cases?

pages/boat/Range.qml
```diff
@@ -0,0 +1,51 @@
...
+
+	VeQuickItem {
+		id: rangeItem
+
+		uid: Global.system.serviceUid ? Global.system.serviceUid + "/MotorDrive/Range" : ""
+		sourceUnit: Units.unitToVeUnit(VenusOS.Units_Kilometre)
+		displayUnit: Units.unitToVeUnit(root.displayUnit)
+	}
+
+	visible: rangeItem.valid
```
blammit: Move this to above the child `VeQuickItem` declarations

pages/boat/Range.qml
```diff
@@ -0,0 +1,51 @@
...
+		//% "Range"
+		text: qsTrId("boat_page_range_label")
+	}
+
+	QuantityLabel {
+		anchors.horizontalCenter: parent.horizontalCenter
+		font.pixelSize: Theme.font_boatPage_range_value_pixelSize
+		value: rangeItem.valid ? rangeItem.value : 0
+		unit: root.displayUnit
+		unitColor: Theme.color_font_secondary
```
blammit: This is already the default unit colour.

pages/boat/Temperatures.qml
```diff
@@ -0,0 +1,122 @@
...
+		anchors.horizontalCenter: parent.horizontalCenter
+
+		font.pixelSize: Theme.font_boatPage_temperature_columnHeader_pixelSize
+		color: Theme.color_font_secondary
+	}
+
+	component Temperature: QuantityLabel {
+		font.pixelSize: Theme.font_boatPage_temperature_value_pixelSize
+		unit: Global.systemSettings.temperatureUnit
+		unitText: "°"
```
blammit: Can use `Units.degreesSymbol()` instead

pages/boat/Temperatures.qml
```diff
@@ -0,0 +1,122 @@
...
+		visible: !isNaN(value)
+	}
+
+	component TemperatureColumn : Column {
+		required property string label
+		required property string dataProperty
+
+		spacing: Theme.geometry_boatPage_temperature_temperatureColumn_spacing
+
+		visible: motorDrives.singleMotorDrive && motorDrives.singleMotorDrive[dataProperty].valid ||
```
blammit: Please put `||` at the start of the next line instead.

pages/boat/Temperatures.qml
```diff
@@ -0,0 +1,122 @@
...
+		}
+	}
+
+	component TemperatureSeparator: SeparatorBar {
+		anchors.verticalCenter: parent.verticalCenter
+		height: Theme.geometry_boatPage_temperature_temperatureSeparator_height
+		width: Theme.geometry_boatPage_temperature_temperatureSeparator_width
+	}
+
+	Column {
```
DanielMcInnes, line 71: The visibility of `batteryTemperatureColumn` works differently to all of the others. Why is `batteryTemperatureColumn` a `Column` when all of the others are `TemperatureColumn`?

DanielMcInnes:
is there a way to get the mock desktop gui to display 2  motor temperatures, 2 controller temperatures, and 2 coolant temperatures? If not, could you add mock config so there is a way to demonstrate this please?

pages/boat/Temperatures.qml
citolen: The battery temperature differs too much from the others to be able to re-use the `TemperatureColumn` component.
The other temperatures are coming from the motordrives and can be single or dual.

It did not make sense to try and re-use `TemperatureColumn` for the battery, it does not fit the use case.

citolen:
> is there a way to get the mock desktop gui to display 2 motor temperatures, 2 controller temperatures, and 2 coolant temperatures? If not, could you add mock config so there is a way to demonstrate this please?

@DanielMcInnes Yes there is a way. When on the boat page, use the `,` and `.` keys to go through all the different configurations.

pages/boat/Range.qml
citolen: It was intentional, I believe it should be different.
For range and consumption, scaling down to meters is too small and wouldn't be useful.
The consumption (Wh/m) would most likely always be zero with some decimals.

We lack analytics, but I would be surprised if a significant portion of users use the EV range in meters, it's also not practical.

pages/boat/LargeCenterGauge.qml
citolen: Thanks, I wasn't sure if it needed to be in the geometry files since it's the same value regardless of the theme size.

DanielMcInnes (approved):
lgtm

blammit (approved):

PR #3023, opened 2026-05-28
UI test fixes and file name improvements

blammit:


tests/ui/RecursivePageCapture.qml
```diff
@@ -79,19 +79,23 @@ QtObject {
...
 		} else {
 			// There is no ListView in this page, or there are no more items to be clicked in the
-			// ListView.
-			const isInitialPage = imageNameSequence.length === 1
+			// ListView. If we have returned to the initial page, finish the tests; otherwise,
+			// return to the previous page and continue the tests.
+			const isInitialPage = imageNameSequence.length <= 1 // zero if initial page was not captured
 			if (isInitialPage) {
 				// All child pages have been opened; end the tests without further captures.
 				testCase.runSteps(root.doneCallback)
```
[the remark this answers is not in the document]
blammit: Doesn't apply. If the first page is excluded, then there no pages to pop backwards to in the app.

src/uitestcase.h
```diff
@@ -97,6 +97,9 @@ class UiTestCase : public QQuickItem
 
 	// TODO add keyPress(QQuickItem *item);
 
+	// Replaces non-alphanumeric characters in the given string with an underscore.
```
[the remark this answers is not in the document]
blammit: Fixed.

src/uitestcase.cpp
```diff
@@ -335,6 +336,12 @@ bool UiTestCase::mouseClick(QQuickItem *item)
 	return true;
 }
 
+QString UiTestCase::sanitizedImageName(const QString &imageName) const
+{
+	static const QRegularExpression nonAlphanumeric("[^a-zA-Z0-9]");
+	return QString(imageName).replace(nonAlphanumeric, QString());
```
chriadam: We may also need to store a `QSet<QString> m_sanitizedImageNames` and ensure that this image name doesn't collide.

And, we probably also need to double check that the image name isn't empty (i.e. if the page title only has non-alpha-numeric characters, e.g. for Japanese language support or something) and if so, replace it with some sha256 data instead.

Pseudocode:
```
QString UiTestCase::sanitizedImageName(const QString &imageName) const
{
    static const QRegularExpression nonAlphanumeric("[^a-zA-Z0-9]");
    QString s = QString(imageName).replace(nonAlphanumeric, QString());
    if (s.isEmpty() || m_sanitizedImageNames.contains(s)) {
        s = QCryptographicHash::hash(imageName, QCryptographicHash::Sha256).toHex();
    }
    m_sanitizedImageNames.insert(s);
    return s;
}
```

chriadam: An alternative might be: instead of just removing the non-alpha-numeric characters (i.e. replacing them with empty string), we could instead define some "transliteration" function to convert non-alpha-numeric into alpha numeric (i.e. do `replaced = QString::fromLatin1(stringSegment.toUtf8().toHex());`)

blammit: The non-alphanumeric stripping was a bit of a blunt instrument. I've updated this avoid that; instead, it now calls `QString::normalize()`, strip punctuation and whitespace, and limit the string length. I expect we want to keep non-ascii characters so that we can match page titles in various languages too.

src/uitestcase.cpp
```diff
@@ -335,6 +336,28 @@ bool UiTestCase::mouseClick(QQuickItem *item)
 	return true;
 }
 
+QString UiTestCase::sanitizedImageName(const QString &imageName) const
```
chriadam, line 339: The robot suggests:

```
 QString UiTestCase::sanitizedImageName(const QString &imageName) const
 {
 	static const QString fallbackName = QStringLiteral("UNTITLED");
 	static const QRegularExpression separatorRegex(QStringLiteral("[^\\p{L}\\p{N}]+"));
 
 	constexpr int maxTokenLen = 15;
 	constexpr int maxSegmentLen = 44;
 	constexpr int hashLen = 8; // hex chars
 
 	const QString normalized = imageName.normalized(QString::NormalizationForm_C).trimmed();
 	if (normalized.isEmpty()) {
 		return fallbackName;
 	}
 
 	QStringList tokens = normalized.split(separatorRegex, Qt::SkipEmptyParts);
 	bool truncated = false;
 	for (QString &token : tokens) {
 		if (token.size() > maxTokenLen) {
 			token.truncate(maxTokenLen);
 			truncated = true;
 		}
 	}
 
 	// No internal separators; parent path parts are already joined with '_' elsewhere.
 	QString base = tokens.join(QString());
 	if (base.isEmpty()) {
 		base = fallbackName;
 	}
 
 	if (base.size() > maxSegmentLen) {
 		base.truncate(maxSegmentLen);
 		truncated = true;
 	}
 
 	if (!truncated) {
 		return base;
 	}
 
 	// if we had to truncate, append a hash to avoid collisions.
 	const QString hash = QString::fromLatin1(
 			QCryptographicHash::hash(normalized.toUtf8(), QCryptographicHash::Sha1).toHex().left(hashLen));
 
 	const int maxBaseLen = qMax(1, maxSegmentLen - (hashLen + 1)); // +1 for '-'
 	if (base.size() > maxBaseLen) {
 		base.truncate(maxBaseLen);
 	}
 
 	return QStringLiteral("%1-%2").arg(base).arg(hash);
 }
```

It also suggests moving the "limit to 255 chars" code into the _captureNextScreen function, otherwise the captureImageName there might exceed 255 char length.

tests/ui/RecursivePageCapture.qml
```diff
@@ -35,25 +31,32 @@ QtObject {
...
 
-	function start(baseImageName, doneCallback) {
-		root.viewImageIds = {}
+	// If set, each capture file name is prefixed with this string.
+	property string capturePrefix
+
+	/*
+		Starts the recursive capture process and calls 'doneCallback' when done.
+	*/
+	function start(doneCallback) {
```
chriadam, line 42: the robot pointed out that this isn't guarded against multiple calls, so if it's called multiple times, the state is not reset, and so there can be pollution (pageCaptureCounts etc).

src/uitestcase.cpp
blammit: I've moved the 255 char limit into the _captureNextScreen(), that sounds good. I don't think we need the other parts; they will add more complication than necessary. We don't have cases that would actually truncate at the moment, and when that happens we should work out a solution that doesn't involve a random hash.

tests/ui/RecursivePageCapture.qml
blammit: I've added a `_busy` marker to track this.

chriadam (approved):
LGTM

PR #3025, opened 2026-05-28
Tests: add utility for test image comparision

ybott-qinetic:


tools/uicompare/Main.qml
```diff
@@ -0,0 +1,97 @@
+import QtQuick
+import QtQuick.Layouts
+import QtQuick.Window
+import QtQuick.Controls
+import QtQuick.Dialogs
+import Qompare
+
+Window {
```
blammit, line 7: The `tools/uicompare/*.qml` files are using tabs instead of spaces for indentation.

tools/uicompare/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+- **`image-captures-baseline/`** - Contains the baseline (reference) images
+- **`image-captures/`** - Contains the current (test) images to compare against the baseline
+
+The tool will automatically discover all images in both directories and compare images with matching filenames. Supported image formats include PNG, JPG, and other formats supported by Qt's image loading.
+
+### Building and running:
+```bash
+cd tools/uicompare
+mkdir build && cd build
```
blammit: I don't think we should advise to build within the source directory, if it's not strictly necessary. Also, does this really need to include build instructions if there are no special flags or such?

tools/uicompare/CMakeLists.txt
```diff
@@ -0,0 +1,40 @@
...
+
+project(qompare VERSION 0.1 LANGUAGES CXX)
+
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+
+find_package(Qt6 REQUIRED COMPONENTS Quick)
+
+qt_standard_project_setup(REQUIRES 6.8)
+
+qt_add_executable(appqompare
```
blammit: The "app" prefix is unnecessary. Also, use the same name for both the directory and the binary, instead of using `uicompare` for one and `appqompare` for the other.

tools/uicompare/comparemodel.h
```diff
@@ -0,0 +1,112 @@
...
+    Q_PROPERTY(int failedCount READ failedCount NOTIFY failedCountChanged)
+    Q_PROPERTY(int missingBaselineCount READ missingBaselineCount NOTIFY missingBaselineCountChanged)
+    Q_PROPERTY(int missingCurrentCount READ missingCurrentCount NOTIFY missingCurrentCountChanged)
+    Q_PROPERTY(int filterMode READ filterMode WRITE setFilterMode NOTIFY filterModeChanged)
+
+public:
+    explicit CompareModel(QObject *parent = 0);
+    ~CompareModel();
+
+    struct ImageResult
```
blammit, line 32: `ImageResult` is only used within this class, so it can be private.

tools/uicompare/comparemodel.h
```diff
@@ -0,0 +1,112 @@
...
+        bool identical = false;
+        bool pending = false;
+        QString errorMessage;
+    };
+
+public:
+    enum RoleNames {
+        TitleRole = Qt::DisplayRole,
+        TextRole = Qt::UserRole,
+        similarityRole,
```
blammit: Role names should start with a capital letter.

tools/uicompare/comparemodel.h
```diff
@@ -0,0 +1,112 @@
...
+        identicalRole,
+        errorMessageRole,
+    };
+
+public: // QAbstractItemModel interface
+    int rowCount(const QModelIndex &parent = QModelIndex()) const override;
+    QVariant data(const QModelIndex &index, int role) const override;
+    bool setData(const QModelIndex &index, const QVariant &value, int role) override;
+
+public: // QListView interface
```
blammit: Were these functions meant to be removed? They are either not used or not implemented.

tools/uicompare/Main.qml
```diff
@@ -0,0 +1,97 @@
...
+            Layout.preferredHeight: 70
+            Layout.fillWidth: true
+
+            totalCount: testModel.count
+            passCount: testModel.passCount
+            failCount: testModel.failedCount
+            missingBaselineCount: testModel.missingBaselineCount
+            missingCurrentCount: testModel.missingCurrentCount
+
+            onFilterChanged: (filterMode) => testModel.setFilterMode(filterMode)
```
blammit: You can do this with a binding instead:

```
CompareModel {
    filterMode: toolBar.filterMode
}
```

tools/uicompare/Main.qml
```diff
@@ -0,0 +1,97 @@
...
+        SplitView {
+            Layout.fillHeight: true
+            Layout.fillWidth: true
+            orientation: Qt.Horizontal
+
+            ImageListView {
+                id: listView
+                SplitView.preferredWidth: 310
+                SplitView.minimumWidth: 100
+                SplitView.maximumWidth: 600
```
blammit: Would be easier to maintain if the widths are bound to the root geometry instead, e.g. `SplitView.preferredWidth: root.width / 4`

tools/uicompare/Main.qml
```diff
@@ -0,0 +1,97 @@
...
+                model: testModel
+
+                // Update currentFilename when currentIndex changes
+                onCurrentIndexChanged: {
+                    if (currentIndex >= 0) {
+                        var filename = testModel.data(testModel.index(currentIndex, 0), 0x0101) // TextRole
```
blammit: Use CompareModel.TextRole instead of 0x0101.

tools/uicompare/Main.qml
```diff
@@ -0,0 +1,97 @@
...
+
+            ImageListView {
+                id: listView
+                SplitView.preferredWidth: 310
+                SplitView.minimumWidth: 100
+                SplitView.maximumWidth: 600
+                model: testModel
+
+                // Update currentFilename when currentIndex changes
+                onCurrentIndexChanged: {
```
blammit: Would be nicer with a binding like `currentFIlename: testModel.data(...)` instead, if that also works.

tools/uicompare/ImageListView.qml
```diff
@@ -0,0 +1,37 @@
...
+
+ListView {
+    id: root
+
+    property string currentFilename: ""
+
+    clip: true
+    delegate: ImageListViewDelegate {
+        width: ListView.view.width
+        onClicked: (index, filename) => {
```
blammit: Align with the capitalisation of "file name" in Qt APIs: use `fileName` instead of `filename`, and `currentFileName` instead of `currentFilename`.

tools/uicompare/applicationsettings.cpp
```diff
@@ -0,0 +1,18 @@
+#include "applicationsettings.h"
+
+ApplicationSettings::ApplicationSettings(QObject *parent)
+    : QSettings("Qinetic", "Qompare", parent)
```
blammit: Org name should be "Victron"

tools/uicompare/comparemodel.h
```diff
@@ -0,0 +1,112 @@
...
+    void startAsyncComparison(const QString &filename);
+    void validateImage(const QString &filename);
+
+    ImageResult getResultData(QString filename) const;
+    bool passesFilter(const QString &filename) const;
+
+    QList<QString> m_allData;
+    QList<QString> m_data;
+    mutable QHash<QString, ImageResult> m_results;
+    int m_filterMode = 0;  // 0=all, 1=pass, 2=fail, 3=missing baseline, 4=missing current
```
blammit: Use enum values intsead?

tools/uicompare/applicationsettings.h
```diff
@@ -0,0 +1,33 @@
...
+    QML_SINGLETON
+
+public:
+    enum SettingKey {
+        WindowX,
+        WindowY,
+        WindowWidth,
+        WindowHeight,
+        BaselineDirectory,
+        CurrentDirectory,
```
blammit: "Current directory" could be confused with the "current working directory" - is there a better name?

tools/uicompare/ImageListViewDelegate.qml
```diff
@@ -0,0 +1,112 @@
...
+            height: 40
+
+            Rectangle {
+                id: statusBadge
+                anchors.left: parent.left
+                anchors.verticalCenter: parent.verticalCenter
+                width: 8
+                height: 30
+                radius: 4
+                color: statusColor
```
blammit: Use `root.statusColor` instead, same for elsewhere in the file

tools/uicompare/ImageListViewDelegate.qml
```diff
@@ -0,0 +1,112 @@
...
+            }
+
+            Column {
+                id: headerColumn
+                anchors.left: statusBadge.right
```
blammit: Use grouped properties like `anchors { ... }` instead.

tools/uicompare/applicationsettings.h
blammit: Seems like `BaselineDirectory` and `CurrentDirectory` are unused at the moment - can just leave them out until that feature is implemented.

tools/uicompare/Main.qml
```diff
@@ -0,0 +1,97 @@
...
+    CompareModel {
+        id: testModel
+    }
+
+    Timer {
+        id: timer
+        interval: 1
+        running: true
+        repeat: false
+        onTriggered: testModel.refresh()
```
blammit: What's the reason for this delayed refresh() call? Can the test model just load its contents when it has the required parameters?

tools/uicompare/ImageListViewDelegate.qml
```diff
@@ -0,0 +1,112 @@
...
+Item {
+    id: root
+    signal clicked(var index, var filename)
+
+    height: 200
+    readonly property var splitName: model.text.split("-")
+    readonly property bool isIdentical: model.identical
+    readonly property bool isPassing: model.similarity >= 0.999
+    readonly property bool hasError: model.errorMessage && model.errorMessage.length > 0
+    readonly property color statusColor: hasError ? "#F44336" : (isIdentical ? "#2196F3" : (isPassing ? "#4CAF50" : "#FFC107"))
```
blammit: The edge colour is yellow/orange when this is a fail, but I would expect it to be red, to match the colour of the "Fail" number in the toolbar.

Also when it passes, the colour is blue - can it be green instead?

tools/uicompare/ImageListViewDelegate.qml
```diff
@@ -0,0 +1,112 @@
+import QtQuick
+import QtQuick.Layouts
+
+Item {
+    id: root
+    signal clicked(var index, var filename)
+
+    height: 200
```
blammit: Could size to the contents as `implicitHeight: column.height`.

tools/uicompare/ImageListView.qml
```diff
@@ -0,0 +1,37 @@
...
+    highlightMoveDuration: 100
+
+    // Auto-select first item when model is populated
+    onCountChanged: {
+        if (count > 0 && currentIndex < 0) {
+            currentIndex = 0
+        }
+    }
+
+    ScrollBar.vertical: ScrollBar {
```
blammit: The scroll bar currently overlaps with the delegate highlight, so you could add some margin/padding to it.

tools/uicompare/Main.qml
ybott-qinetic: Done

tools/uicompare/README.md
ybott-qinetic: Removed

tools/uicompare/CMakeLists.txt
ybott-qinetic: Done

tools/uicompare/comparemodel.h
ybott-qinetic: The ComparisonWorker needs this to be public for the ThreadPool. Or should all this logic be folding back into CompareModel?

tools/uicompare/ImageListViewDelegate.qml
ybott-qinetic: Done

tools/uicompare/Main.qml
ybott-qinetic: I was trying to make the processing not block the ui thread. I will try to make this better.

tools/uicompare/ImageListViewDelegate.qml
ybott-qinetic: I was trying, unsuccessfully, to differenciate between a within threshold pass and an exact match. I have made it so both exact matches and passes are green.

tools/uicompare/ImageListViewDelegate.qml
ybott-qinetic: Done

tools/uicompare/ImageListView.qml
ybott-qinetic: I think this is better now.

tools/uicompare/comparemodel.h
ybott-qinetic: Done

blammit:
Rebased against `main`.

chriadam (approved):

PR #3028, opened 2026-05-28
AC Load: fix AcIn/AcOut three phase table column alignment

ybott-qinetic:


themes/geometry/FiveInch.json
```diff
@@ -498,7 +498,8 @@
     "geometry_quantityTable_maximumWidth_small": 0,
     "geometry_quantityTable_maximumWidth_large": 0,
     "geometry_quantityTable_bottomMargin": 14,
-    "geometry_quantityTable_horizontalSpacing_small": 12,
+    "geometry_quantityTable_horizontalSpacing_tiny": 4,
+    "geometry_quantityTable_horizontalSpacing_small": 10,
```
blammit: Does the `small` value need to be changed or can it remain at 12?

blammit (approved):
Seems like the `small` change is not needed, but otherwise LGTM.

themes/geometry/FiveInch.json
ybott-qinetic: It should remain at 12.

blammit (approved):

Issue #2978, opened 2026-05-07
Update resistive tank sensor step size to 0.1


chriadam:
In https://github.com/victronenergy/gui-v2/pull/2915 the step size was changed to 1 (ohm) even though the precision is displayed with 1 decimal (i.e. 0.1 ohm).  We should ensure the step size matches the available precision.

chriadam:
Merged to main, closing as fixed.

PR #3033, opened 2026-06-01
Tanks: unify tank value decimals and step size

chriadam:
Previously, resistive tank sensors were displayed with one decimal place of precision, but a step size of 1 (step by one whole ohm) because the value was expected within the range of tens to low hundreds of ohms.

This commit changes the step size to match the display precision, to ensure fine control over the tank sensor, at the cost of slightly slower initial setup.

Contributes to issue #2978


ReinvdZee:
Looks good to me. Verified by testing:

image

ReinvdZee (approved):

PR #3052, opened 2026-06-08
CAN-bus: add NMEA2000 inbound alerts setting

nmbath:
Ref: https://github.com/victronenergy/venus/issues/1586

Added a toggle switch for a new capability under Ve.Can.

image

### DIFF
--- a/pages/settings/PageSettingsCanbus.qml
+++ b/pages/settings/PageSettingsCanbus.qml
@@ -81,6 +81,13 @@ Page {
 			}
 
 			ListSwitch {
+				//% "NMEA2000 inbound alerts"
+				text: qsTrId("settings_canbus_nmea2000in_alerts")
+				dataItem.uid: root._vecanSettingsPrefix + "/AlertsInEnable"
+				preferredVisible: root._isVecan && dataItem.valid
+			}
+
+			ListSwitch {
 				//% "Reverse current polarity"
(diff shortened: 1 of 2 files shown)


pages/settings/PageSettingsCanbus.qml
```diff
@@ -80,6 +85,14 @@ Page {
 				onClicked: n2kOutEnabled.setValue(checked ? (n2kOutEnabled.value & ~valueTrue) : (n2kOutEnabled.value | valueTrue))
 			}
 
+			ListSwitch {
+				//% "NMEA2000 inbound alerts"
+				text: qsTrId("settings_canbus_nmea2000in_alerts")
+				dataItem.value: n2kAlertsInEnabled.value
+				preferredVisible: root._isVecan && n2kOutEnabled.valid && n2kAlertsInEnabled.valid
+				onClicked: n2kAlertsInEnabled.setValue(checked ? 0 : 1)
```
blammit: If you set `dataItem.uid: root._vecanSettingsPrefix + "/AlertsInEnable"` then you won't need to set `dataItem.value` and `onClicked` here, as the `ListSwitch` will take care of writing the value to the correct path and showing the value from the path.

nmbath: @blammit Thanks. I have done this. I also added a check for the value being valid in the preferredVisible.

blammit (approved):

PR #3053, opened 2026-06-09
Switches: hide some input/output settings depending on product ID

### LINKED ISSUE #2941: Switches page: properly set acceslevel on settings fields
Services that expose `/SwitchableOutputs/x/..` paths will have a switches menu, in which some settings of the particular switch can be configured. Currently, a setting is always displayed whenever the service exposes it. Depending on the access level, some settings may not be set by the user. The preferred availability and accessibility of switchable output settings can vary among different devices and manufacturers. For example, for the Aurelia, we'd like to hide settings that should not be accessible for the currently selected access level.

However, the QML is responsible for the settings page of all switches of all services, independent of the manufacturer. So to accomplish device-specific access levels for settings field, we may need to make some exceptions based on the product ID.

For the Aurelia, we'd like the following access levels:
- Green: User
- Orange: User & Installer

Image

Adding blocked label for now because we need to discuss how to handle this for other devices/products.

blammit:
For Aurelia products, some input/output settings should be hidden. We will improve this later to configure the setting visibility in the backend, instead of hardcoding it in the UI depending on the product id, but this is sufficient for now.

Fixes #2941

### DIFF
--- a/pages/settings/devicelist/iochannel/PageGenericInput.qml
+++ b/pages/settings/devicelist/iochannel/PageGenericInput.qml
@@ -11,6 +11,17 @@ Page {
 
 	required property GenericInput genericInput
 
+	// For Aurelia products, some settings are not visible at the user-access level. For now, hard
+	// code this configuration in gui-v2, but later on we will generalise this to configure the
+	// setting visibility in the backend data values instead. See #2941.
+	VeQuickItem {
+		id: productId
+
+		readonly property bool isAurelia: valid && (value === ProductInfo.ProductId_Dcdb_Aurelia)
+
+		uid: root.genericInput.serviceUid + "/ProductId"
+	}
+
 	GradientListView {
 		model: VisibleItemModel {
 			ListIOChannelNameField {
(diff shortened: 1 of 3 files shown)


pages/settings/devicelist/iochannel/PageGenericInput.qml
```diff
@@ -11,6 +11,17 @@ Page {
 
 	required property GenericInput genericInput
 
+	// For Aurelia products, some settings are not visible at the user-access level. For now, hard
+	// code this configuration in gui-v2, but later on we will generalise this to configure the
+	// setting visibility in the backend data values instead. See #2941.
+	VeQuickItem {
+		id: productId
+
+		readonly property bool isAurelia: valid && (value & ProductInfo.ProductId_Dcdb_Aurelia)
```
chriadam: & or === ?

blammit: Fixed.

chriadam (approved):
LGTM

PR #3054, opened 2026-06-09
Switches: allow type changes, hide some settings depending on product id

### LINKED ISSUE #2941: Switches page: properly set acceslevel on settings fields

blammit:
- Make the 'Type' writable at the user access level.
- Hide 'Startup switch state' and 'Startup dim level' for the user access level on Aurelia products.

Fixes #2941

### DIFF
--- a/components/listitems/ListIOChannelTypeRadioButtonGroup.qml
+++ b/components/listitems/ListIOChannelTypeRadioButtonGroup.qml
@@ -15,6 +15,7 @@ ListRadioButtonGroup {
 	text: qsTrId("iochannel_type_buttongroup_type")
 	dataItem.uid: root.ioChannel.uid + "/Settings/Type"
 	preferredVisible: dataItem.valid
+	writeAccessLevel: VenusOS.User_AccessType_User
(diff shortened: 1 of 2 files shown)


ReinvdZee:
Tested again by Lambertus, all good now.

blammit:
@chriadam This one is ready for review.

chriadam (approved):
LGTM

PR #3058, opened 2026-06-10
Dialogs: fix position of modal dialogs

### LINKED ISSUE #3049: Switch pane: RGB dialog not centered
Discovered on Venus OS 3.80~27, the RGB color selection dialog is not centered:

Image

Let me know if a test system is needed in order to troubleshoot this.

blammit:
Use the width/height properties rather than the implicitWidth/ implicitHeight to set the geometry of ModalDialog. Otherwise when the implicitWidth/implicitHeight of the contentItem changes (e.g. if the item is being laid out) then the dialog geometry does not update and so the dialog's x/y position is incorrect. This bug is only seen on Wasm, due to the scaling ratio that is applied to the window geometry.

Make a similar update to SolarDailyHistoryDialog for consistency. Also, the implicitWidth/implicitHeight there should not refer to Global.scalingRatio anyway, as this causes external changes to affect the implicit geometry.

Also, in the ColorWheelDialog, increase the horizontal space between the color wheel and the preset grid in 7" mode to align more closely with the design.

Fixes #3049

### DIFF
--- a/components/dialogs/ModalDialog.qml
+++ b/components/dialogs/ModalDialog.qml
@@ -106,14 +106,16 @@ T.Dialog {
 	rest of the app. Specifying 'width' and 'height' instead of only 'implicitWidth' and 'implicitHeight' for ModalDialog
 	makes it scale properly when you shrink the window. See https://bugreports.qt.io/browse/QTBUG-127068
 	*/
-	implicitWidth: Math.max(
-		implicitBackgroundWidth + leftInset + rightInset,
-		implicitContentWidth + leftPadding + rightPadding)
-	implicitHeight: Math.max(
-		implicitBackgroundHeight + topInset + bottomInset,
-		implicitContentHeight + topPadding + bottomPadding + implicitHeaderHeight + implicitFooterHeight)
-	width: Math.min(Theme.geometry_screen_width, implicitWidth)
-	height: Math.min(Theme.geometry_screen_height, implicitHeight)
+	width: Math.min(
+		Theme.geometry_screen_width,
+		Math.max(
+			implicitBackgroundWidth + leftInset + rightInset,
+			implicitContentWidth + leftPadding + rightPadding))
+	height: Math.min(
+		Theme.geometry_screen_height,
+		Math.max(
+			implicitBackgroundHeight + topInset + bottomInset,
+			implicitContentHeight + topPadding + bottomPadding + implicitHeaderHeight + implicitFooterHeight))
 	modal: true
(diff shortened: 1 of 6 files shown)


components/dialogs/ModalDialog.qml
```diff
@@ -106,14 +106,16 @@ T.Dialog {
...
 	*/
-	implicitWidth: Math.max(
-		implicitBackgroundWidth + leftInset + rightInset,
-		implicitContentWidth + leftPadding + rightPadding)
-	implicitHeight: Math.max(
-		implicitBackgroundHeight + topInset + bottomInset,
-		implicitContentHeight + topPadding + bottomPadding + implicitHeaderHeight + implicitFooterHeight)
-	width: Math.min(Theme.geometry_screen_width, implicitWidth)
-	height: Math.min(Theme.geometry_screen_height, implicitHeight)
+	width: Math.min(
```
chriadam, line 109: don't remove the implicitWidth and implicitHeight bindings; keep those.  But then, also ADD bindings for width and height, which do e.g. `width: Math.max(Theme.geometry_screen_width, implicitWidth)` or whatever

blammit: That's what we do in the existing code, but then the dialog position does not update as expected. I think it's to do with the scaling ratio, but not sure why it prevents it from updating dynamically.

chriadam: Hmm.  Potentially something in the transform/scaling stuff internals in QtQuick might overwrite the implicitWidth or implicitHeight from C++ which breaks the binding, perhaps?  One possibility might be to do something like:

```
readonly property real _implicitWidth: Math.max(implicitBackgroundWidth + leftInset + rightInset, implicitContentWidth + leftPadding + rightPadding)
implicitWidth: _implicitWidth
width: Math.min(_implicitWidth, Theme.geometry_screen_width)
```

That is: I would prefer if every component did expose "correctly calculated" implicit geometry if possible.

blammit: As soon as I specify the `implicitWidth` and `implicitHeight`, the layout is broken again. Perhaps this is also related to https://bugreports.qt.io/browse/QTBUG-127068, but not sure. There is something special about dialogs and how they are affected by implicitWidth and implicitHeight - perhaps because they are in a different layer.

chriadam: There is a comment on that bug saying that it doesn't reproduce in Qt 6.7, so I think this is something different (although perhaps related).  Hmm.

chriadam (approved):

PR #3059, opened 2026-06-10
Boat: charging state improvements

### LINKED ISSUE #3060: Boat: shore connected/charging state improvements
A couple of improvements needed for the boat page when the boat is plugged-in to shore and/or charging.

1. Indicate when the system battery is charging regardless of the source. Currently it's indicating a charge only when the E-Drive is regenerating.
2. Indicate when the boat is connected to shore. It doesn't necessarily mean that it is charging, just that a shore AC-in is detected.
3. Wake up the screen when shore power gets plugged-in.
4. Disable the screen blanker when the boat page is active and there is at least a gps or an E-Drive powered on.

citolen:
Updates to the boat page to improve the visualization of the charging state:
- Indicate when the battery is charging regardless of whether it is coming from regeneration, solar, etc...
- Display battery information in the center when no GPS or E-Drive is powered ON.
- Indicate when shore power is connected and wake up the display upon connection.
- Display warning when the boat is moving away with shore still connected.
- Stop the display from going to sleep when the boat page is active and there is at least a GPS or E-Drive powered ON.

No GPS, no E-Drive, no shore power
image

No GPS, no E-Drive, shore power connected
image

GPS, E-Drive, shore power still connected
image

GPS, E-Drive, charging from a source other than regeneration
image


pages/boat/SlotVerticalCenterLeft.qml
```diff
@@ -0,0 +1,47 @@
...
+		id: batteryPercentage
+
+		visible: (root.gps.valid || root.motorDrives.dcConsumption.quotient.valid) && !isNaN(Global.system.battery.stateOfCharge)
+		motorDrives: root.motorDrives
+	}
+
+	Boat.ShoreGauge {
+		id: shoreGauge
+
+		visible: !batteryPercentage.visible
```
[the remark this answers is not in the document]
citolen: This is intended.

pages/boat/SlotLeftArc.qml
```diff
@@ -0,0 +1,33 @@
...
+
+		visible: (root.gps.valid || root.motorDrives.dcConsumption.quotient.valid) && !isNaN(Global.system.battery.stateOfCharge)
+		animationEnabled: root.animationEnabled
+	}
+
+	Boat.ShoreArc {
+		id: shoreArc
+
+		visible: !batteryGauge.visible
+		animationEnabled: root.animationEnabled
```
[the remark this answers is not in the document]
citolen: This is intended.

pages/boat/ShoreArc.qml
```diff
@@ -0,0 +1,48 @@
...
+	id: root
+
+	required property bool animationEnabled
+	readonly property var _shoreInput: Global.acInputs.activeInSource === VenusOS.AcInputs_InputSource_Shore ? Global.acInputs.highlightedInput : null
+	readonly property int _leftGaugeCount: 1
+
+	objectName: "ShoreArc"
+	width: Theme.geometry_briefPage_edgeGauge_width
+	height: Gauges.gaugeHeight(root._leftGaugeCount)
+	sourceComponent: SideGauge {
```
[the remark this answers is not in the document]
citolen: This is intended.

pages/boat/BatteryPercentage.qml
```diff
@@ -12,17 +12,18 @@ Row {
...
 		anchors.verticalCenter: parent.verticalCenter
 		width: Theme.geometry_boatPage_batteryGauge_iconWidth
 		height: width
-		color: root.motorDrives.isRegenerating ? Theme.color_boatPage_regenProgress : Theme.color_font_primary
-		source: root.motorDrives.isRegenerating ? "qrc:/images/icon_battery_charging_24.svg" : "qrc:/images/icon_battery_24.svg"
+		color: (Global.system.battery.mode === VenusOS.Battery_Mode_Charging || root.motorDrives.isRegenerating)
+			? Theme.color_boatPage_regenProgress
+			: Theme.color_font_primary
+		source: (Global.system.battery.mode === VenusOS.Battery_Mode_Charging || root.motorDrives.isRegenerating)
+			? "qrc:/images/icon_battery_charging_24.svg"
```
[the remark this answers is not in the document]
citolen: No guard needed.

blammit:
@citolen Is this one intentionally still in draft mode, or is it ready to be reviewed?

citolen:
> @citolen Is this one intentionally still in draft mode, or is it ready to be reviewed?

Yes, I'm waiting on some designs.
I will remove the draft flag once it is ready for review.

pages/boat/LargeCenterGauge.qml
```diff
@@ -163,6 +190,86 @@ Item {
...
+				id: batteryLabel
+
+				verticalAlignment: Text.AlignVCenter
+				height: font.pixelSize
+				font.pixelSize: Theme.font_boatPage_batteryCenterGauge_title_pixelSize
+				color: Theme.color_font_primary
+				text: CommonWords.battery
+			}
+		}
+
```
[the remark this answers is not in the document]
citolen: Not true, it does have `bottomPadding` and it's working as expected.

pages/boat/LargeCenterGauge.qml
```diff
@@ -163,6 +190,86 @@ Item {
...
+			font.weight: Font.Medium
+			unit: VenusOS.Units_Percentage
+			value: Global.system.battery.stateOfCharge
+		}
+
+		Row {
+			visible: batteryOuterGauge.visible
+			anchors.horizontalCenter: parent.horizontalCenter
+			topPadding: Theme.geometry_boatPage_batteryCenterGauge_detailsRow_topPadding
+
```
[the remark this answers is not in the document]
citolen: Not true, it does have a `topPadding` property and it's working as expected.

pages/boat/BatteryPercentage.qml
```diff
@@ -10,19 +10,29 @@ import Victron.VenusOS
...
-		source: root.motorDrives.isRegenerating ? "qrc:/images/icon_battery_charging_24.svg" : "qrc:/images/icon_battery_24.svg"
+	Item {
+		width: Theme.geometry_boatPage_batteryGauge_iconWidth - Theme.geometry_boatPage_row_spacing
+		height: Theme.geometry_boatPage_batteryGauge_iconWidth
+
+		CP.ColorImage {
+			anchors.verticalCenter: parent.verticalCenter
+			width: Theme.geometry_boatPage_batteryGauge_iconWidth
+			height: width
+			color: (Global.system.battery.mode === VenusOS.Battery_Mode_Charging || root.motorDrives.isRegenerating)
```
DanielMcInnes: The old BatteryPercentage deliberately guarded Global.system.battery with:
`readonly property ActiveSystemBattery battery: Global.system && Global.system.battery ? Global.system.battery : null`

Does the new code ever throw `TypeError: Cannot read property 'mode' of null` ? If not, this is fine as-is.

pages/boat/BoatPage_Landscape.qml
```diff
@@ -156,11 +145,37 @@ Boat.Background { // the blue shadows
...
+
+				shorePowerListener.previousSource = Global.acInputs.activeInSource
+			}
+		}
+	}
+
+	Binding {
+		target: Global
+		property: "boatPageActive"
+		value: parent.visible && (_gps.valid || motorDrives.dcConsumption.quotient.valid)
```
DanielMcInnes, line 160: Global.boatPageActive can get stuck true and prevent the display from ever sleeping.

`boatPageActive` now gates `ScreenBlanker.enabled`. Because Global is a singleton, the last written value persists. If the page item is destroyed on navigation, the Binding is torn down and `boatPageActive` is frozen at whatever it last held — if that was true, the screen never blanks again. You can't rely on `parent.visible` being set to false before destruction, it isn't guaranteed by ordering. 

Explicitly reset Global.boatPageActive = false in Component.onDestruction.

pages/boat/SlotLeftArc.qml
```diff
@@ -0,0 +1,33 @@
...
+*/
+
+import QtQuick
+import Victron.Boat as Boat
+import Victron.VenusOS
+
+Item {
+	id: root
+
+	required property VeQuickItemsQuotient gps
```
DanielMcInnes: Is there a reason for `gps` to be of type `VeQuickItemsQuotient` rather than `Gps` ?

Here and elsewhere.

pages/boat/Background.qml
```diff
@@ -10,13 +10,26 @@ import Victron.VenusOS
 Item {
 	id: root
 
+	required property Gps gps
 	required property MotorDrives motorDrives
 
+	readonly property bool isBatteryCharging: Global.system.battery.mode === VenusOS.Battery_Mode_Charging
+	readonly property bool isShoreConnected: Global.acInputs.activeInSource === VenusOS.AcInputs_InputSource_Shore
+	readonly property bool isMoving: root.gps.valid && root.gps.numerator > 0.2
```
DanielMcInnes, line 19: isMoving uses a raw, undocumented threshold. `isMoving: root.gps.valid && root.gps.numerator > 0.2` appears in two files. `numerator` is the raw speed numerator rather than the converted speed value, and 0.2 is an unexplained magic number with no unit. If the source unit ever changes, this silently misbehaves. Suggest a named constant with a comment stating the unit, and confirm numerator is the intended field (vs. the computed quotient/value).

pages/boat/BatteryPercentage.qml
```diff
@@ -31,6 +41,18 @@ Row {
...
+
+	CP.ColorImage {
+		id: icon
+
+		anchors.verticalCenter: parent.verticalCenter
+		visible: isShoreConnected
+		width: Theme.geometry_boatPage_shoreGauge_icon_size
+		height: width
+		fillMode: Image.PreserveAspectFit
+		color: isMoving ? Theme.color_red : Theme.color_orange
```
DanielMcInnes, line 57: this seems like an odd scenario - do I understand correctly?

Our boat is plugged in to shore power, and somehow moving? Is that the requirement?

pages/boat/Background.qml
```diff
@@ -10,13 +10,26 @@ import Victron.VenusOS
 Item {
 	id: root
 
+	required property Gps gps
 	required property MotorDrives motorDrives
 
+	readonly property bool isBatteryCharging: Global.system.battery.mode === VenusOS.Battery_Mode_Charging
+	readonly property bool isShoreConnected: Global.acInputs.activeInSource === VenusOS.AcInputs_InputSource_Shore
```
DanielMcInnes: `Global.acInputs.activeInSource === VenusOS.AcInputs_InputSource_Shore` is repeated in 4 different files - refactor this to a shared readonly property.

pages/boat/Background.qml
```diff
@@ -10,13 +10,26 @@ import Victron.VenusOS
 Item {
 	id: root
 
+	required property Gps gps
 	required property MotorDrives motorDrives
 
+	readonly property bool isBatteryCharging: Global.system.battery.mode === VenusOS.Battery_Mode_Charging
```
DanielMcInnes: `Global.system.battery.mode === VenusOS.Battery_Mode_Charging` is repeated in 6 different files - refactor this to a shared readonly property.

pages/boat/ShoreArc.qml
```diff
@@ -0,0 +1,48 @@
...
+		valueType: VenusOS.Gauges_ValueType_NeutralPercentage
+		value: _shoreInput ? shoreRange.valueAsRatio * 100 : 0
+	}
+
+	ValueRange {
+		id: shoreRange
+		value: _shoreInput ? Math.abs(_shoreInput.current) : 0
+		maximumValue: _shoreInput && !isNaN(_shoreInput.inputInfo.maximumCurrent)
+			? _shoreInput.inputInfo.maximumCurrent
+			: 0;
```
DanielMcInnes: trailing `;`

DanielMcInnes: when `maximumValue` gets set to 0, do you end up with a divide-by-zero error anywhere?

pages/boat/BoatPage_Landscape.qml
```diff
@@ -13,34 +13,38 @@ Boat.Background { // the blue shadows
 	required property bool animationEnabled
 
 	motorDrives: motorDrives
+	gps: _gps
 
-	Boat.BatteryArc { // arc gauge on the far left
-		id: batteryGauge
```
DanielMcInnes, line 19: double blank line

DanielMcInnes:
Does the 'shore connected' requirement mean "there is an cable connected between the boat and the shore", or that "shore is the currently active AC input"?

pages/boat/BoatPage_Landscape.qml
```diff
@@ -156,11 +145,37 @@ Boat.Background { // the blue shadows
...
+		property int previousSource: Global.acInputs.activeInSource
+
+		Connections {
+			target: Global.acInputs
+
+			function onActiveInSourceChanged() {
+				if (shorePowerListener.previousSource !== VenusOS.AcInputs_InputSource_Shore &&
+					Global.acInputs.activeInSource === VenusOS.AcInputs_InputSource_Shore) {
+					// Wake up display when shore power gets connected
+					ScreenBlanker.setDisplayOn();
```
DanielMcInnes: The wake logic lives inside BoatPage_Landscape.qml:

Connections {
    target: Global.acInputs
    function onActiveInSourceChanged() {
        if (previousSource !== Shore && activeInSource === Shore) {
            ScreenBlanker.setDisplayOn()
        }
    }
}

If the boat page is not loaded/instantiated when shore is connected, this may not wake the display. If this is a product-level rule — “shore connection wakes display” — it may belong in a global/system-level object, not inside the visual page.

blammit: Yes, perhaps this logic belongs in `AcInputs.qml` instead?

pages/boat/BoatPage_Landscape.qml
blammit: Why does `boatPageActive` depend on whether the gps and DC consumption is valid? Can it just be based on whether the boat page is visible? If so, then this logic can be placed in `BoatPage.qml` instead of `BoatPage_Landscape`.

pages/boat/SlotVerticalCenterLeft.qml
```diff
@@ -0,0 +1,48 @@
...
+
+Item {
+	id: root
+
+	required property VeQuickItemsQuotient gps
+	required property MotorDrives motorDrives
+
+	implicitHeight: {
+		if (batteryPercentage.visible) {
+			return batteryPercentage.height;
```
blammit: To align with code elsewhere, could write this as:

```
implicitHeight: batteryPercentage.visible ? batteryPercentage.height
             : shoreGauge.visible ? shoreGauge.height
             : 0
```

Same with `implicitWidth`.

pages/boat/SlotVerticalCenterRight.qml
```diff
@@ -0,0 +1,25 @@
...
+Item {
+	id: root
+
+	required property VeQuickItemsQuotient gps
+	required property MotorDrives motorDrives
+
+	implicitHeight: consumptionGauge.height
+	implicitWidth: consumptionGauge.width
+
+	Boat.ConsumptionGauge {
```
blammit, line 19: What do you get from having this gauge as a child rather than the root item? Is there a problem with having the Column as the base type?

Global.qml
```diff
@@ -45,6 +45,7 @@ QtObject {
 	property var venusPlatform
 	property bool dataManagerLoaded
 	property bool allPagesLoaded
+	property bool boatPageActive: false
```
blammit: `false` is already the default value, would be fine to leave it out

pages/boat/BoatPage_Landscape.qml
citolen: No, it is intentional to require at least the gps or E-Drive to be valid.
The goal is when the boat page is visible and gps/E-Drive is valid then the page is considered active so that the display can remain always on without requiring the user to wake the screen or disable the timer.

pages/boat/SlotLeftArc.qml
citolen: There was a mix in a few files, I will align.
Gps is a VeQuickItemsQuotient

pages/boat/BatteryPercentage.qml
citolen: Yes that's correct, in the dangerous scenario where you forgot to unplug shore power before leaving the dock.
It happens unfortunately, this could avoid it.

pages/boat/BoatPage_Landscape.qml
blammit: Makes sense, could you add a comment here to explain this?

pages/boat/ShoreArc.qml
citolen: No it's not a problem, it's handled in `src/fastutils.h`, `FastUtils::scale` uses `qFuzzyIsNull` to avoid dividing by zero.

pages/boat/SlotVerticalCenterRight.qml
citolen: Purely to remain consistent with the left side slot.
The left slot has some conditional logic and therefore the component to display is not the root.
Eventually all the slots will be this way

pages/boat/BatteryPercentage.qml
citolen: It does not, there are multiple places across the code base which do not guard against that, I assumed it would have been seen already if it was an issue.

pages/boat/Background.qml
citolen: @DanielMcInnes Do we have an example pattern somewhere else? Not sure where to put it.
The threshold is indeed for the raw value which is in m/s.

pages/boat/BoatPage_Landscape.qml
citolen: Logic moved to `AcInputs.qml`

citolen:
>Does the 'shore connected' requirement mean "there is an cable connected between the boat and the shore", or that "shore is the currently active AC input"?

Currently all that we can reliably detect is "shore is the currently active AC input".
There is additional work to also just detect when the cable is connected, but this is not blocking this from getting released.
It will be an improvement that comes at a later time.

pages/boat/Background.qml
DanielMcInnes: I misunderstood this - I thought that if the units changed, then `root.gps.numerator` would change as well. Just put a comment saying that the numerator is always in m/s. and this is fine.

DanielMcInnes (approved):
Once the `m/s` comment is added to Background.qml, lgtm

blammit:
Updated to refer to issue #3060 in the commit messages.

PR #3081, opened 2026-06-23
BLE: Add support for encrypted instant readout devices

### LINKED ISSUE #3062: Add support for encrypted instant readout devices
We are going to add support for encrypted instant readout to dbus-ble-sensors. This requires dbus-ble-sensors to know the encryption key of a device. The path to enter this key will be `Devices/xxxxxx_xxxxxxx/Key` under the com.victronenergy.ble service (so it will be besides the Enabled path for that device). It will be possible to connect with VictronConnect over bleutooth and provide the encryption keys, but it should also be possible to manually enter the encryption key. This requires the following changed in the gui in the Integrations - Bluetooth Sensors menu:
- When a device has a `Devices/xxxxx_xxxxxx/Key` path. gui-v2 should show an input field for the key. As long as this field does not contain 32 hexadecimal characters, the slider to enable that device should be disabled.
- The input field should be limited to 32 hexadecimal characters. The input is case insensitive.
- Optionally, the input field can be changed to a readonly field when the device is Enabled.

DanielMcInnes:
Encrypted Instant Readout devices require dbus-ble-sensors to know the device's encryption key. The key is exposed on dbus at the path Devices/<id>/Key under com.victronenergy.ble, alongside the existing Enabled path for that device.

In the Bluetooth Sensors menu, devices that expose a Key path now show an input field for the key. The field is limited to 32 hexadecimal characters (case insensitive) via a validator, and the device's Enabled switch stays disabled until a valid key has been entered. The field becomes read-only once the device is enabled.

A validator alias was added to ListTextField so the input can be restricted to hex characters. Mock data for an encrypted device is included so the behaviour can be exercised.

Depends on dbus-ble-sensors exposing the Devices/<id>/Key path.

Fixes #3062

### DIFF
--- a/data/mock/conf/services/ble-sensors.json
+++ b/data/mock/conf/services/ble-sensors.json
@@ -7,6 +7,12 @@
         "/Devices/f00f00d00001/Name": "Ruuvi 41EA",
         "/Devices/f00f00d00001/Enabled": 1,
         "/Devices/afd683/Name": "SolarSense afd683",
-        "/Devices/afd683/Enabled": 1
+        "/Devices/afd683/Enabled": 1,
+        "/Devices/c2e4f10a8b3d_ruuvi/Name": "Ruuvi Garage (encrypted)",
+        "/Devices/c2e4f10a8b3d_ruuvi/Enabled": 0,
+        "/Devices/c2e4f10a8b3d_ruuvi/Key": "",
+        "/Devices/a14b9f23c5e6_ruuvi/Name": "Ruuvi Cellar (encrypted, keyed)",
+        "/Devices/a14b9f23c5e6_ruuvi/Enabled": 1,
+        "/Devices/a14b9f23c5e6_ruuvi/Key": "0123456789ABCDEF0123456789ABCDEF"
     }
 }
--- a/pages/settings/BleSensorDelegate.qml
+++ b/pages/settings/BleSensorDelegate.qml
@@ -1,0 +1,388 @@
+/*
+** Copyright (C) 2026 Victron Energy B.V.
+** See LICENSE.txt for license information.
+*/
+
+import QtQuick
+import QtQuick.Layouts
+import Victron.VenusOS
+
+/*
+	A single Bluetooth sensor in the Bluetooth Sensors settings list.
+
+	An ordinary sensor is shown as an enable switch. An encrypted Instant Readout device (one that
+	exposes a Devices/<id>/Key path) cannot be enabled until its encryption key is known, so while
+	such a device is disabled an encryption key field is shown instead of the switch. Entering a
+	valid key writes the key and enables the device; if the key is wrong, the backend disables the
+	device again.
+
+	The key field and the enable switch are never shown at the same time: the field is the only
+	control while an encrypted device is disabled, and accepting a complete key (Enter, or moving
+	focus away) enables the device and replaces the field with the switch. Switching the device off
+	again brings the field back, focused and holding the stored key, so that it can be switched back
+	on with a single Enter.
+
+	A device whose stored key is already complete is enabled as soon as this delegate learns its
+	state, so that it comes up switched on rather than showing a key field for a key that is already
+	known. Note that this applies every time the delegate is created, not only at application
+	startup: a device switched off here is switched on again by the next visit to this page.
+
+	A loader swaps between the two standard list items as the device's state changes, so neither
+	has to emulate the other. It must be a ListItemLoader rather than a plain Loader: a plain
+	Loader leaves focusPolicy at Qt.NoFocus, so Utils.acceptsKeyNavigation() rejects it and the
+	row would be skipped by key navigation.
+*/
+ListItemLoader {
+	id: root
+
+	// The device's dbus/mqtt path prefix, e.g. ".../Devices/xxxxxx_xxxxxxx".
+	required property string devicePrefix
+	required property string deviceName
+
+	// Encrypted Instant Readout devices expose a Devices/<id>/Key path; such a device is enabled
+	// by entering a valid key rather than via a switch.
+	readonly property bool keyRequired: keyItem.seen
+	readonly property bool deviceEnabled: enabledItem.value === 1
+
+	// The key currently stored for the device, if any. Shown in the key field so that a key which
+	// was rejected (or is simply out of date) can be corrected rather than retyped in full.
+	readonly property string storedKey: keyItem.valid ? keyItem.value : ""
+
+	// An enable attempt for an encrypted device runs in two stages, each tracked by its timer.
+	// First the /Enabled=1 write has to be acknowledged by the backend - on dbus/mqtt the write is
+	// asynchronous, so this is not immediate (enableAckTimer). Only once it has been acknowledged
+	// does Enabled returning to 0 mean that the backend rejected the key (validationTimer);
+	// before that, a 0 is simply the state we have not overwritten yet.
+	readonly property bool enableAttemptInFlight: enableAckTimer.running || validationTimer.running
+
+	// Set when the user switches an encrypted device off, so that the key field which replaces the
+	// switch takes the focus as it loads: switching the device off is a step towards changing or
+	// re-submitting its key, so the field is where the user is going next. Consumed by the field,
+	// so that the focus is only taken on the load that this disable caused.
+	property bool focusKeyFieldOnLoad: false
+
+	function takeFocusRequest() {
+		const requested = focusKeyFieldOnLoad
+		focusKeyFieldOnLoad = false
+		return requested
+	}
+
+	// Whether the initial state has been dealt with. Set as soon as the device's state is known,
+	// whether or not it led to the device being enabled, so that this remains a one-shot: without
+	// it, switching a device off would immediately re-enable it, as its stored key is still
+	// complete.
+	property bool initialStateApplied: false
+
+	// A stored key that is already complete is all that gates an encrypted device, so enable the
+	// device rather than presenting a key field for a key the user has no reason to re-enter.
+	// The stored key and the enabled state arrive asynchronously (and on dbus/mqtt not at all
+	// promptly), so this cannot simply run when the delegate is created; it is driven by the state
+	// changing instead, and runs on the first update in which the state is known.
+	function applyInitialState() {
+		if (initialStateApplied || !enabledItem.valid) {
+			// Already done, or the enabled state is not known yet.
+			return
+		}
+		if (!keyRequired || !keyItem.valid) {
+			// Either the key path has not been seen yet - the device may still turn out to be an
+			// encrypted one - or it has, but its stored key has not arrived. Note that an ordinary
+			// device therefore never settles here, which is harmless: it has no key to act on.
+			return
+		}
+		initialStateApplied = true
+		if (!deviceEnabled && keyIsComplete(storedKey)) {
+			enableWithKey(storedKey)
+		}
+	}
+
+	function keyIsComplete(key) {
+		return /^[0-9a-fA-F]{32}$/.test(key)
+	}
+
+	// Save the key, then enable the device. The backend validates the key and, if it is wrong,
+	// disables the device again (see the enabledItem Connections). Deferred via Qt.callLater so it
+	// is safe to call from the key field's own signal handlers - enabling swaps the field out from
+	// under them.
+	function enableWithKey(key) {
+		if (enableAttemptInFlight || deviceEnabled) {
+			// An attempt is already in flight. On dbus/mqtt the writes are asynchronous, so the key
+			// field is still shown while we wait and could otherwise submit a second, interleaved
+			// attempt.
+			return
+		}
+		enableAckTimer.restart()
+		keyItem.setValue(key)
+		enabledItem.setValue(1)
+	}
+
+	function disable() {
+		// Stop tracking first, so that the resulting Enabled -> 0 is not mistaken for the backend
+		// rejecting a key.
+		abandonEnableAttempt()
+		// Only an encrypted device swaps in a key field to focus; an ordinary device keeps its
+		// switch, and moving the focus off it would be wrong.
+		focusKeyFieldOnLoad = keyRequired
+		// The user has made an explicit choice, which the initial-state handling must not undo if
+		// it has not run by now (it normally has, since the device was enabled to be switched off).
+		initialStateApplied = true
+		enabledItem.setValue(0)
+	}
+
+	function abandonEnableAttempt() {
+		enableAckTimer.stop()
+		validationTimer.stop()
+	}
+
+	// Width is set, height is left to follow the loaded list item's implicit height.
+	width: parent ? parent.width : 0
+
+	// Show the key field while an encrypted device is disabled; otherwise (an ordinary device, or
+	// an enabled encrypted one) show the enable switch.
+	sourceComponent: (keyRequired && !deviceEnabled) ? keyFieldComponent : switchComponent
+
+	Component.onCompleted: applyInitialState()
+	onStoredKeyChanged: applyInitialState()
+	onKeyRequiredChanged: applyInitialState()
+	onDeviceEnabledChanged: applyInitialState()
+
+	VeQuickItem {
+		id: enabledItem
+		uid: root.devicePrefix + "/Enabled"
+	}
+	VeQuickItem {
+		id: keyItem
+		uid: root.devicePrefix + "/Key"
+	}
+
+	// Track the enable attempt, and detect the backend rejecting an entered key: Enabled dropping
+	// back to 0 after it acknowledged our Enabled=1, without the user having disabled the device.
+	Connections {
+		target: enabledItem
+		// A device that arrives already disabled leaves deviceEnabled at false, so its arrival is
+		// only visible as the item becoming valid.
+		function onValidChanged() {
+			root.applyInitialState()
+		}
+		function onValueChanged() {
+			if (enabledItem.value === 1) {
+				if (enableAckTimer.running) {
+					// The write landed. From here on, a return to 0 means the backend rejected the key.
+					enableAckTimer.stop()
+					validationTimer.restart()
+				}
+				return
+			}
+			// Enabled is now 0, or the item became invalid because the service dropped out. Only
+			// the former, and only while awaiting a validation result, means a rejected key: an
+			// invalid item cannot be tracked, and the Enabled=0 that the service republishes when
+			// it comes back must not be reported as a bad key either.
+			if (enabledItem.value === 0 && validationTimer.running) {
+				//% "The submitted encryption key is invalid."
+				Global.showToastNotification(VenusOS.Notification_Warning,
+						qsTrId("settings_ble_sensors_encryption_key_rejected"), 5000)
+			}
+			root.abandonEnableAttempt()
+		}
+	}
+
+	// The Enabled=1 write is asynchronous on dbus/mqtt. If the backend never echoes it back,
+	// abandon the attempt rather than waiting for a validation result that can no longer arrive.
+	Timer {
+		id: enableAckTimer
+		interval: 10000
+		onTriggered: root.abandonEnableAttempt()
+	}
+
+	// Started only once the backend has acknowledged Enabled=1, so that a slow round-trip cannot
+	// eat into the window. If the device is still enabled when this elapses, assume the key was
+	// accepted; a rejection is expected to disable the device before then (observed to take ~15
+	// seconds).
+	Timer {
+		id: validationTimer
+		interval: 30000
+	}
+
+	Component {
+		id: switchComponent
+
+		ListSwitch {
+			text: root.deviceName
+			dataItem.uid: root.devicePrefix + "/Enabled"
+			// An ordinary device toggles Enabled directly. An encrypted device only shows the
+			// switch while enabled, so clicking it disables the device (routed through disable()
+			// to keep the delegate's state in sync).
+			updateDataOnClick: !root.keyRequired
+			onClicked: {
+				if (root.keyRequired) {
+					root.disable()
+				}
+			}
+		}
+	}
+
+	Component {
+		id: keyFieldComponent
+
+		ListTextField {
+			id: keyField
+
+			readonly property bool portrait: Theme.screenSize === Theme.Portrait
+
+			function focusKeyField() { contentItem.forceInputFocus() }
+
+			// Take the focus when this field replaces the switch of a device the user just
+			// switched off. Deferred, because at this point the loader has not yet given the item
+			// a window to take focus in.
+			Component.onCompleted: {
+				if (root.takeFocusRequest()) {
+					Qt.callLater(focusKeyField)
+				}
+			}
+
+			text: root.deviceName
+			// Start from the key that is currently stored for the device, so that a key which the
+			// backend rejected can be corrected instead of retyped; empty if none is stored yet.
+			// The binding is broken by the first edit, as in ListTextField itself, so a
+			// late-arriving /Key value cannot overwrite what the user is typing.
+			secondaryText: root.storedKey
+			// Note the absence of a dataItem.uid here: the key is written via the delegate's
+			// keyItem. Pointing dataItem at /Key would make the default 'interactive' binding
+			// (which requires dataItem.valid) disable the field whenever the path has been seen
+			// but holds no valid value yet - i.e. exactly when a key still has to be entered.
+			// Instead, lock the field while an enable attempt is in flight, so a slow dbus/mqtt
+			// round-trip cannot queue a second attempt, and so the field cannot appear
+			// editable-but-inert if the device drops back to disabled while we are still awaiting
+			// a validation result.
+			interactive: !root.enableAttemptInFlight
+			maximumLength: 32
+			inputMethodHints: Qt.ImhNoAutoUppercase | Qt.ImhNoPredictiveText
+			//% "Enter a 32-character key"
+			placeholderText: qsTrId("settings_ble_sensors_encryption_key_placeholder")
+			validateInput: function() {
+				if (secondaryText.length === 0 || root.keyIsComplete(secondaryText)) {
+					return Utils.validationResult(VenusOS.InputValidation_Result_OK)
+				}
+				//% "The encryption key must be 32 hexadecimal characters."
+				return Utils.validationResult(VenusOS.InputValidation_Result_Error,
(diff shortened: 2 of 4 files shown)


ThiemoVanEngelen:
image
I think it doesn't look nice that there is a line between the product with the toggle switch and the line with the encryption key. It would be nice when both are in 1 block or perhaps even on the same.

ReinvdZee:
@DanielMcInnes @ThiemoVanEngelen 

We discussed this in the meeting today:

This should be changed to a double row listitem with the text field for the encryption key. Like we currently do for setting the security profile (but in the BLE case we use a toggle instead):

image

If possible, please show "Encryption key" on the second line right before the input/text field. No indentation necessary because it is all enclosed in one listitem.

Implementation rules:
- If the BLE device requires an encryption key and it is currently disabled, show the input field with the key.
- If the BLE device requires an encryption key and it is currently enabled, show the key as read-only text.
- If the BLE device does not require an encryption key, do not show it, the listitem should be a simple toggle switch.

DanielMcInnes:
> image I think it doesn't look nice that there is a line between the product with the toggle switch and the line with the encryption key. It would be nice when both are in 1 block or perhaps even on the same.

done

pages/settings/PageSettingsBleSensors.qml
```diff
@@ -97,9 +98,158 @@ Page {
...
+									focusPolicy: Qt.NoFocus
+									enabled: sensorDelegate.__switchClickable
+									onClicked: {
+										if (sensorDelegate.checkWriteAccessLevel()) {
+											enabledDataItem.setValue(checked ? 0 : 1)
+										}
+									}
+
+									Layout.alignment: Qt.AlignRight | Qt.AlignVCenter
+								}
```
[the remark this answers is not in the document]
DanielMcInnes: fixed

pages/settings/PageSettingsBleSensors.qml
```diff
@@ -97,9 +98,158 @@ Page {
...
+							}
+
+							// Since the root top/bottomPadding is 0, add the content margins back
+							// here. The inter-row spacing matches the top/bottom padding so that the
+							// gap above the switch (to the top of the delegate) equals the gap below
+							// it (to the text entry box); the switch is centered within its row, so
+							// its centering offset applies equally on both sides.
+							topPadding: Theme.geometry_listItem_content_verticalMargin
+							bottomPadding: Theme.geometry_listItem_content_verticalMargin
+							spacing: Theme.geometry_listItem_content_verticalMargin
```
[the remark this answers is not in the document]
DanielMcInnes: this is incorrect, ignoring

pages/settings/PageSettingsBleSensors.qml
```diff
@@ -97,9 +98,158 @@ Page {
...
+						// A valid key is either empty (clearing the key) or exactly 32 hex characters.
+						readonly property bool __keyValid: /^([0-9A-Fa-f]{32})?$/.test(secondaryText)
+						readonly property bool __enabled: enabledDataItem.value === 1
+						// An encrypted device can only be enabled once a valid 32-character
+						// hexadecimal key has been entered. An already-enabled device can always
+						// be disabled, so the user is never trapped if the backend presents an
+						// enabled device with an invalid key.
+						readonly property bool __switchClickable: userHasWriteAccess
+								&& enabledDataItem.valid
+								&& (!__keyRequired || (__keyValid && secondaryText.length > 0) || __enabled)
```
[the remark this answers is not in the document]
DanielMcInnes: done

components/listitems/core/ListTextField.qml
```diff
@@ -16,6 +16,9 @@ ListSetting {
 	property string placeholderText
 	property int echoMode: TextInput.Normal
 	property int maximumLength: 32767 // as per TextInput default
+	// Optional input validator (e.g. a RegularExpressionValidator) applied to the text field, to
+	// restrict the characters that can be entered. Null by default (no restriction).
+	property alias validator: textField.validator
```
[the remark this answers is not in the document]
DanielMcInnes: done

pages/settings/PageSettingsBleSensors.qml
```diff
@@ -97,9 +98,158 @@ Page {
...
+							return Utils.validationResult(VenusOS.InputValidation_Result_Error,
+									qsTrId("settings_ble_sensors_encryption_key_invalid"))
+						}
+
+						// Two rows, each laid out as a RowLayout so that the control on the right is
+						// vertically centered with its label on the left. The key caption/field row
+						// is only shown for encrypted devices that require a key.
+						// | Device name | Enable switch |
+						// | Key caption | Key field     |
+						contentItem: Column {
```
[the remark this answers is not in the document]
DanielMcInnes: done

pages/settings/PageSettingsBleSensors.qml
```diff
@@ -97,9 +98,158 @@ Page {
...
+									placeholderText: sensorDelegate.placeholderText
+									maximumLength: sensorDelegate.maximumLength
+									validator: sensorDelegate.validator
+
+									flickable: sensorDelegate.flickable
+									validateInput: sensorDelegate.validateInput
+									validateOnFocusLost: sensorDelegate.validateOnFocusLost
+
+									onInputValidated: sensorDelegate.saveInput()
+									onTextChanged: sensorDelegate.secondaryText = text
```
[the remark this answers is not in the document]
DanielMcInnes: fixed

pages/settings/PageSettingsBleSensors.qml
```diff
@@ -97,9 +97,52 @@ Page {
...
+							id: keyField
+
+							//% "Encryption key"
+							text: qsTrId("settings_ble_sensors_encryption_key")
+							dataItem.uid: sensorDelegate.__devicePrefix + "/Key"
+							preferredVisible: sensorDelegate.__keyRequired
+							// Limited to 32 hexadecimal characters; input is case insensitive.
+							maximumLength: 32
+							inputMethodHints: Qt.ImhNoAutoUppercase | Qt.ImhNoPredictiveText | Qt.ImhSensitiveData
+							validator: RegularExpressionValidator { regularExpression: /[0-9A-Fa-f]{0,32}/ }
```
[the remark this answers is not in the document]
DanielMcInnes: fixed

pages/settings/PageSettingsBleSensors.qml
```diff
@@ -97,9 +97,52 @@ Page {
...
+						ListTextField {
+							id: keyField
+
+							//% "Encryption key"
+							text: qsTrId("settings_ble_sensors_encryption_key")
+							dataItem.uid: sensorDelegate.__devicePrefix + "/Key"
+							preferredVisible: sensorDelegate.__keyRequired
+							// Limited to 32 hexadecimal characters; input is case insensitive.
+							maximumLength: 32
+							inputMethodHints: Qt.ImhNoAutoUppercase | Qt.ImhNoPredictiveText | Qt.ImhSensitiveData
```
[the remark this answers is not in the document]
DanielMcInnes: fixed

DanielMcInnes:
> @DanielMcInnes @ThiemoVanEngelen
> 
> We discussed this in the meeting today:
> 
> This should be changed to a double row listitem with the text field for the encryption key. Like we currently do for setting the security profile (but in the BLE case we use a toggle instead):
> image
> 
> If possible, please show "Encryption key" on the second line right before the input/text field. No indentation necessary because it is all enclosed in one listitem.
> 
> Implementation rules:
> 
>     * If the BLE device requires an encryption key and it is currently disabled, show the input field with the key.
> 
>     * If the BLE device requires an encryption key and it is currently enabled, show the key as read-only text.
> 
>     * If the BLE device does not require an encryption key, do not show it, the listitem should be a simple toggle switch.

Done, but not exactly as suggested.



image

data/mock/conf/services/ble-sensors.json
```diff
@@ -7,6 +10,12 @@
...
         "/Devices/afd683/Name": "SolarSense afd683",
-        "/Devices/afd683/Enabled": 1
+        "/Devices/afd683/Enabled": 1,
+        "/Devices/c2e4f10a8b3d_ruuvi/Name": "Ruuvi Garage (encrypted)",
+        "/Devices/c2e4f10a8b3d_ruuvi/Enabled": 0,
+        "/Devices/c2e4f10a8b3d_ruuvi/Key": "",
+        "/Devices/a14b9f23c5e6_ruuvi/Name": "Ruuvi Cellar (encrypted, keyed)",
+        "/Devices/a14b9f23c5e6_ruuvi/Enabled": 1,
+        "/Devices/a14b9f23c5e6_ruuvi/Key": "0123456789ABCDEF0123456789ABCDEF"
     }
```
chriadam, line 17: One minor gap in the mock data: no mock device with an invalid saved key (e.g., a key shorter than 32 chars already committed). This would exercise the  __savedKeyValid: false  +  __enabled: true  path where the field should remain editable for correction despite the device being enabled.

chriadam (approved):
One minor comment, aside from that LGTM.  But would be good if Rein could review and test.

ReinvdZee:
@DanielMcInnes @chriadam 

I like how it has improved!

I cannot test this atm because I don't have a device to test it with. And Thiemo is on holiday at the moment. When Thiemo is back, I'll ask him to test it.

ReinvdZee:
@DanielMcInnes 

@philipptrenz and I discussed this today and came up with the following suggestion:

If the device requires encryption, please show the input field instead of the toggle. When the encryption key is entered, GUI-v2 show enable the device. When it is valid, the device will remain enabled and can be used. When it is invalid, dbus-ble-sensors will indicate so (probably by disabling the device again), and GUI-v2 should show a toast notification saying that the submitted encryption key is invalid. When invalid, show the toast and clear the input field so the user can try again. Upon disabling the device, show an empty input field again.

DanielMcInnes:
> @DanielMcInnes
> 
> @philipptrenz and I discussed this today and came up with the following suggestion:
> 
> If the device requires encryption, please show the input field instead of the toggle. When the encryption key is entered, GUI-v2 show enable the device. When it is valid, the device will remain enabled and can be used. When it is invalid, dbus-ble-sensors will indicate so (probably by disabling the device again), and GUI-v2 should show a toast notification saying that the submitted encryption key is invalid. When invalid, show the toast and clear the input field so the user can try again. Upon disabling the device, show an empty input field again.

Updated as suggested. I notice there is a long delay (~15 seconds) between entering an incorrect key and `dbus-ble-sensors` clearing it. If there is not a good reason for this, I think it would be better if `dbus-ble-sensors` cleared the invalid key within ~ 1 second.

data/mock/conf/services/ble-sensors.json
DanielMcInnes: mock data updated as suggested

ReinvdZee:
Thanks @DanielMcInnes 

> I notice there is a long delay (~15 seconds) between entering an incorrect key and dbus-ble-sensors clearing it. If there is not a good reason for this, I think it would be better if dbus-ble-sensors cleared the invalid key within ~ 1 second.

Noted, that's something for @ThiemoVanEngelen to look into.

pages/settings/PageSettingsBleSensors.qml
```diff
@@ -112,9 +113,256 @@ Page {
...
-						text: model.item.value
-						dataItem.uid: model.item.itemParent().uid + "/Enabled"
+					// Each Bluetooth sensor is shown either as an enable switch or, for an
+					// encrypted Instant Readout device that is not yet enabled, as an encryption
+					// key entry field. A loader swaps between two standard list items as the
+					// device's state changes, so neither has to emulate the other. It must be a
+					// ListItemLoader rather than a plain Loader: a plain Loader leaves focusPolicy
+					// at Qt.NoFocus, so Utils.acceptsKeyNavigation() rejects it and every sensor
+					// row would be skipped by key navigation.
+					delegate: ListItemLoader {
```
blammit: This delegate is quite complex now - could you move it to a separate component?

pages/settings/PageSettingsBleSensors.qml
```diff
@@ -112,9 +113,256 @@ Page {
...
+									if (/^[0-9A-Fa-f]{32}$/.test(secondaryText)) {
+										Qt.callLater(sensorDelegate.enableWithKey, secondaryText)
+									}
+								}
+
+								// Single row: device name on the left, then the "Encryption key" label
+								// immediately before the entry box, instead of a caption on its own row.
+								contentItem: RowLayout {
+									function forceInputFocus() { keyTextField.forceInputFocus() }
+									function runValidation(mode) { return keyTextField.runValidation(mode) }
```
blammit: These two functions don't appear to be called?

components/listitems/core/ListTextField.qml
```diff
@@ -16,6 +16,11 @@ ListSetting {
 	property string placeholderText
 	property int echoMode: TextInput.Normal
 	property int maximumLength: 32767 // as per TextInput default
+	// Optional input validator (e.g. a RegularExpressionValidator) applied to the text field, to
+	// restrict the characters that can be entered. Null by default (no restriction).
+	// This is a plain property (not an alias to the internal textField) so that it remains valid
+	// when a caller overrides contentItem and the default textField does not exist.
+	property var validator: null
```
blammit: Is the validator necessary? In the past it's been discussed that it would be nicer to validate when the user confirms the text or moves away from the field, instead of applying restrictions during input, hence the reason for the "validator" feature in ListTextField.

pages/settings/PageSettingsBleSensors.qml
```diff
@@ -112,9 +113,256 @@ Page {
...
+										return Utils.validationResult(VenusOS.InputValidation_Result_OK)
+									}
+									return Utils.validationResult(VenusOS.InputValidation_Result_Error, _keyInvalid)
+								}
+								// Entering a complete, valid key enables the device. Fires for any
+								// input method (typing, paste); enabling swaps this field out for the
+								// switch, so defer the write until this handler has returned.
+								onSecondaryTextChanged: {
+									if (/^[0-9A-Fa-f]{32}$/.test(secondaryText)) {
+										Qt.callLater(sensorDelegate.enableWithKey, secondaryText)
```
blammit: I would prefer to not do this on secondaryText change, since text is normally validated when you explicitly save or move focus away from the field.

pages/settings/PageSettingsBleSensors.qml
```diff
@@ -112,9 +113,256 @@ Page {
...
+						Component {
+							id: keyFieldComponent
+
+							ListTextField {
+								id: keyField
+
+								//% "Encryption key"
+								readonly property string _keyCaption: qsTrId("settings_ble_sensors_encryption_key")
+								//% "The encryption key must be 32 hexadecimal characters."
+								readonly property string _keyInvalid: qsTrId("settings_ble_sensors_encryption_key_invalid")
```
blammit: Is it necessary to store these strings as properties? Both are only used from one location.

pages/settings/PageSettingsBleSensors.qml
```diff
@@ -112,9 +113,256 @@ Page {
...
+							id: keyItem
+							uid: sensorDelegate.devicePrefix + "/Key"
+						}
+
+						// Track the enable attempt, and detect the backend rejecting an entered key:
+						// Enabled dropping back to 0 after it acknowledged our Enabled=1, without
+						// the user having disabled the device.
+						// Only an explicit 0 or 1 is meaningful here. The item also becomes invalid
+						// when the service drops out, and that must not be reported as a bad key.
+						Connections {
```
blammit: Having the different state properties `awaitingEnable`, `awaitingValidation` and `userDisabling` seems a bit complicated. Can this be simplified? For example, I'm wondering if `awaitingEnable` and  `awaitingValidation` can be merged to a single `clickedByUser` property, and then handle the value changed here in this manner:

```
function onValueChanged() {
    if (enabled.value === 1) {
        if (sensorDelegate.clickedByUser) {
            // The write landed. From here on, a return to 0 means the backend rejected the key.
            sensorDelegate.clickedByUser = false
            enableAckTimer.stop()
            validationTimer.restart()            
        }
    } else if (enabled.value === 0) {
        if (sensorDelegate.clickedByUser) {
            sensorDelegate.clickedByUser = false
        } else {
            Global.showToastNotification(VenusOS.Notification_Warning, qsTrId("settings_ble_sensors_encryption_key_rejected"), 5000)
        }
        sensorDelegate.abandonEnableAttempt()
    }
}
```

pages/settings/PageSettingsBleSensors.qml
```diff
@@ -112,9 +113,256 @@ Page {
...
+										text: keyField.text
+										font: keyField.font
+										textFormat: keyField.textFormat
+										wrapMode: Text.Wrap
+
+										Layout.fillWidth: true
+										Layout.alignment: Qt.AlignVCenter
+									}
+
+									CaptionLabel {
```
blammit: The layout is too long in portrait mode - for that, you could drop the caption and text field to the next line. E.g. if you use GridLayout instead of RowLayout, it could be adapted dynamically for portrait/landscape: 

```
contentItem: GridLayout {

  columns: Theme.screenSize === Theme.Portrait ? 2 : 3

  Label { Layout.columnSpan: Theme.screenSize === Theme.Portrait ? 2 : 1  }

  CaptionLabel {}
  TextValidationField {}

}
```

blammit:
@DanielMcInnes My understanding from the meeting yesterday was that the preference is to always show the key field and the switch, so that you can re-enable the switch without entering a new key. But if that interferes with the process of editing the key and enabling/disabling it then we should do whatever works for the process instead.

philipptrenz:
To clarify: The idea from yesterday was to at all times show either the key input field or the switch.

1. No key set: The user is presented with an empty key field to enter the key, the switch is hidden.
2. After entering the correct key, the device is enabled, input field is hidden, switch is presented in On position. The user can now enable and disable the device to their liking without being prompted with the input field.


This assumes that:
- `dbus-ble-sensors` enables the device after receiving and validating the key input.
- `dbus-ble-sensors` handles a wrong key input by clearing the key, which will bring back the input field.

To my understanding, this is already the case. But please verify.

> Keep the swap, add an explicit Enable action on the key-field row, enabled once the text is 32 hex chars.

To my understanding, this is then also not needed, as the user is always (newly) presented with an empty key input field.

pages/settings/PageSettingsBleSensors.qml
DanielMcInnes: done, created BleSensorDelegate.qml

pages/settings/PageSettingsBleSensors.qml
DanielMcInnes: `forceInputFocus()` is called by ListTextField's Keys.onSpacePressed. Added a comment to clarify.
`runValidation()` has gone away.

components/listitems/core/ListTextField.qml
DanielMcInnes: no, removed.

pages/settings/PageSettingsBleSensors.qml
DanielMcInnes: removed

pages/settings/PageSettingsBleSensors.qml
DanielMcInnes: updated

pages/settings/PageSettingsBleSensors.qml
DanielMcInnes: Updated mostly as suggested. One deliberate difference from your snippet: the toast is gated on "we were waiting for a validation result" rather than "the user didn't click it". Enabled can go to 0 for other reasons — the ble service restarting and republishing Enabled=0, or the item going invalid and back — and a clickedByUser-only gate would show "encryption key is invalid" in those cases too.

pages/settings/PageSettingsBleSensors.qml
DanielMcInnes: thanks, updated as suggested

pages/settings/BleSensorDelegate.qml
```diff
@@ -0,0 +1,262 @@
...
+	A loader swaps between the two standard list items as the device's state changes, so neither
+	has to emulate the other. It must be a ListItemLoader rather than a plain Loader: a plain
+	Loader leaves focusPolicy at Qt.NoFocus, so Utils.acceptsKeyNavigation() rejects it and the
+	row would be skipped by key navigation.
+*/
+ListItemLoader {
+	id: root
+
+	// The device's dbus path prefix, e.g. ".../Devices/xxxxxx_xxxxxxx".
+	// Note: these are not 'required' properties. A required property on the root of a view
```
blammit: These can be made into required properties, if you also change the view to omit the `model.` references, like this:

```
delegate: BleSensorDelegate {
    required property VeQItem item

    devicePrefix: item.itemParent().uid
    deviceName: item.value
}
```

pages/settings/BleSensorDelegate.qml
```diff
@@ -0,0 +1,262 @@
...
+			if (enabledItem.value === 0 && validationTimer.running) {
+				//% "The submitted encryption key is invalid."
+				Global.showToastNotification(VenusOS.Notification_Warning,
+						qsTrId("settings_ble_sensors_encryption_key_rejected"), 5000)
+			}
+			root.abandonEnableAttempt()
+		}
+	}
+
+	// The Enabled=1 write is asynchronous on dbus. If the backend never echoes it back, abandon
```
blammit: "on dbus" -> "on dbus/mqtt"

pages/settings/BleSensorDelegate.qml
```diff
@@ -0,0 +1,262 @@
...
+
+	// An enable attempt for an encrypted device runs in two stages, each tracked by its timer.
+	// First the /Enabled=1 write has to be acknowledged by the backend - on dbus the write is
+	// asynchronous, so this is not immediate (enableAckTimer). Only once it has been acknowledged
+	// does Enabled returning to 0 mean that the backend rejected the key (validationTimer);
+	// before that, a 0 is simply the state we have not overwritten yet.
+	readonly property bool enableAttemptInFlight: enableAckTimer.running || validationTimer.running
+
+	// Width is set, height is left to follow the loaded list item's implicit height.
+	width: parent ? parent.width : 0
```
blammit: Place `width` and `sourceComponent` bindings after the functions?

pages/settings/BleSensorDelegate.qml
```diff
@@ -0,0 +1,262 @@
...
+					inputMethodHints: keyField.inputMethodHints
+					placeholderText: keyField.placeholderText
+					maximumLength: keyField.maximumLength
+
+					flickable: keyField.flickable
+					validateInput: keyField.validateInput
+					validateOnFocusLost: keyField.validateOnFocusLost
+
+					onInputValidated: keyField.saveInput()
+					onTextChanged: keyField.secondaryText = text
```
blammit: @DanielMcInnes  This should be `onTextEdited` instead of `onTextChanged`, to avoid rewriting the `secondaryText` binding. (I will fix this in `ListTextField` as part of https://github.com/victronenergy/gui-v2/pull/3139.)

ThiemoVanEngelen:
> > @DanielMcInnes
> > @philipptrenz and I discussed this today and came up with the following suggestion:
> > If the device requires encryption, please show the input field instead of the toggle. When the encryption key is entered, GUI-v2 show enable the device. When it is valid, the device will remain enabled and can be used. When it is invalid, dbus-ble-sensors will indicate so (probably by disabling the device again), and GUI-v2 should show a toast notification saying that the submitted encryption key is invalid. When invalid, show the toast and clear the input field so the user can try again. Upon disabling the device, show an empty input field again.
> 
> Updated as suggested. I notice there is a long delay (~15 seconds) between entering an incorrect key and `dbus-ble-sensors` clearing it. If there is not a good reason for this, I think it would be better if `dbus-ble-sensors` cleared the invalid key within ~ 1 second.

It first has to see the advertisement before it can check whether the key is correct is not. Apparently in your case it takes a long time to see the device again. So there is not much that dbus-ble-sensors can do about it.

I am now also thinking of problematic case. The check of the key is purely done on the first byte of the key, which is also present in the advertisement. Although our devices generate new keys with an increasing first byte to limit the chance that the first byte of a new key is the same as some previously stored key, we cannot prevent it. And we cannot know in any other way that the key is wrong, except from receiving garbage data. So when a user mistypes any character except the first two, the key will be accepted but the user will see garbage values coming in. In that case, the user should be able to remove the key and type again. This functionality is in my opinion not blocking for the 3.80 release, but I think this is needed in a next release.

ReinvdZee:
@ThiemoVanEngelen 

> but the user will see garbage values coming in.

So the user should conclude that the key is incorrect based on the garbage coming in? Or can `dbus-ble-sensors` determine that based on the incoming data?

ThiemoVanEngelen:
> @ThiemoVanEngelen
> 
> > but the user will see garbage values coming in.
> 
> So the user should conclude that the key is incorrect based on the garbage coming in? Or can `dbus-ble-sensors` determine that based on the incoming data?

dbus-ble-sensors could include some heuristics (a voltage changing 20V between measurements is not really realistic), but this is not programmed at the moment. So there is no 100% guarantee, but more can be done.

pages/settings/BleSensorDelegate.qml
DanielMcInnes: done

pages/settings/BleSensorDelegate.qml
DanielMcInnes: done

pages/settings/BleSensorDelegate.qml
DanielMcInnes: done

pages/settings/BleSensorDelegate.qml
```diff
@@ -0,0 +1,388 @@
...
+					// holding the stored key, and re-submitting that key is how it gets switched
+					// back on. Marking the field as having something to save on focus, when it
+					// already holds a complete key, arms both of its submit paths at once.
+					// A Connections rather than an onActiveFocusChanged handler, which would
+					// override the one this type relies on for the submit itself.
+					Connections {
+						target: keyTextField
+						function onActiveFocusChanged() {
+							if (keyTextField.activeFocus && root.keyIsComplete(keyTextField.text)) {
+								keyTextField._validateBeforeSaving = true
```
blammit, line 358: Seems a bit strange to require this workaround. Perhaps we need to improve the UI if this field behaviour does not align with existing use cases.

I'm happy to leave this as-is, but having extra workarounds like this in the codebase may create fragility later on.

blammit (approved):

PR #3092, opened 2026-06-26
WASM: instrument compile and instantiate phases with performance timing

mman:
Do not merge: diagnostics only. CC @chriadam.

Split the monolithic qtLoad() call into explicit WebAssembly.compile() and WebAssembly.instantiate() steps with performance.now() annotations, logging each phase to the console.

Helps diagnose browser engine lockups on Marine MFD devices with limited CPU/memory.

Further improved the analytics to determine where time is spent.

Added main thread blocker.

Result: 6.3s of the 12.5s startup (50%) is runtime QML compilation, which is result of CMakeLists.txt:33: `NO_CACHEGEN` set to `ON`.

IMG_0727

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -454,10 +454,34 @@ EM_JS(bool, hasNativeVirtualKeyboard, (), {
 		|| (/Macintosh/i.test(navigator.userAgent) && navigator.maxTouchPoints && navigator.maxTouchPoints > 1);
 });
 
+EM_JS(void, wasmPerfMark, (const char *name), {
+	const label = UTF8ToString(name);
+	try { performance.mark(label); } catch (e) {}
+	console.log("[WASM Perf] " + label + " [t=" + performance.now().toFixed(1) + "ms]");
+});
+
 #endif
+
+// Log a named startup phase to the browser console and DevTools performance
+// timeline, in the same format as the [WASM Perf] entries in wasm/index.html.
+// No-op on non-WASM builds.
+static inline void perfMark(const char *phase)
+{
+#if defined(VENUS_WEBASSEMBLY_BUILD)
+	wasmPerfMark(phase);
+#else
+	Q_UNUSED(phase);
+#endif
+}
 
 int main(int argc, char *argv[])
 {
+#if defined(VENUS_WEBASSEMBLY_BUILD)
+	// Timestamp Qt log lines so browser console output can be correlated with
+	// the [WASM Perf] entries logged by wasm/index.html.
+	qSetMessagePattern(QStringLiteral("[%{time process}] %{if-category}%{category}: %{endif}%{message}"));
+#endif
+	perfMark("main() entered");
 	qInfo().nospace() << "Victron gui version: v" << PROJECT_VERSION_MAJOR << "." << PROJECT_VERSION_MINOR << "." << PROJECT_VERSION_PATCH;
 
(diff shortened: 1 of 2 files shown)


ReinvdZee:
@mman 

> Do not merge: diagnostics only.

Why should this not be merged? If it adds diagnostic information it can be very useful for troubleshooting MFD issues.

ReinvdZee:
@chriadam Can you take a look? I'm fine with merging this if it is low risk. Then preferably even in 3.80.

mman:
@ReinvdZee @chriadam I think it is pretty low risk but wanted second eyes to validate before we merge it…

chriadam:
I have no concerns with this going into main, although I think MrManuel is better to review it than I am.  But, some of those copilot review comments look valid, so it would be good to fix those first too I guess.

mr-manuel:
A few things I'd address before this leaves "diagnostics only" status, since @ReinvdZee and @chriadam both sound open to actually merging it:

- **`observeMainThreadStalls()` runs unconditionally for the life of the page.** Right now every user pays for the PerformanceObserver (or the `requestAnimationFrame` fallback loop) on every load, forever, for a problem that's specific to underpowered MFD hardware. This should generate more CPU load and therefore more power consumption?
- **Compile failures show a generic message.** A `WebAssembly.compile()` failure falls into the outer catch and shows `'Error loading application!'`, same as an actual download failure. Minor, but worth a distinct message so the failure phase is obvious from the error screen alone, not just the console.
- **Naming heads-up, not a bug:** there's now a C++ `perfMark()` (→ `wasmPerfMark` via EM_JS) and a separate JS closure also called `perfMark()` inside `init()` in index.html. They're not the same function, just matching output format, worth a comment noting that, since grepping "perfMark" now hits two unrelated symbols. Better would be to rename it, if possible.

Should we add a debug mode to GUIv2 that surfaces more logging (things like this)? Something like:

- A toggle in the GUIv2 UI itself (if it's running), or
- Presence of an empty `debug` file in the gui-v2 web folder, which could itself be created/deleted from within GUIv2 or if it does not start via SSH?

mr-manuel:
As discussed in the meeting:

Debugging will only be enabled if `/var/www/venus/debug` file is present. This can be created via
- USB stick and reboot
- SSH and manual file creation

Filesystem is remounted as read-write for this.

mman:
@mr-manuel Although I basically do not have a problem with this, I think WASM loaded via browser can not access filesystem and can not determine whether profiling should be on/off. This can only work running locally on Cerbo where Qt has access to the filesystem - where it is less of an issue as the main goal of the performance monitoring is for wasm loaded via web browser on slow devices. 

So my proposal is to think whether this can be present in GUI-v2 itself somewhere under the Settings -> Debug & Develop. That way we can also instruct users how to enable it easily... rooting a device to access filesystem to gather perf data from a problematic MFD seems like a really involved process (even if it worked - which I think it will not)... 

thoughts?

mr-manuel:
We also discussed the option to enable it via the GUIv2 itself, but that is too much effort for that small number of cases.

As soon as you set a root password, the filesystem gets writable. With one command you can then create the file. On the WASM side then you can do a HEAD request and check the HTTP status you get. 404 if the file does not exist, 200 if the file exists. The file placed in `/var/www/venus/debug` is reachable via `http(s)://venus.local/debug`.

mr-manuel:
Superseeded by PR above.

PR #3098, opened 2026-07-02
Opportunity Loads menu improvements

philipptrenz:
Contributes to victronenergy/venus#1647

### DIFF
--- a/pages/settings/PageControllableLoads.qml
+++ b/pages/settings/PageControllableLoads.qml
@@ -92,7 +92,7 @@ Page {
 			width: parent?.width ?? 0
 
 			PrimaryListLabel {
-				//% "Arrange the controllable devices according to their priority; the control algorithm will control them based on the currently available PV excess."
+				//% "Devices will be controlled in order, based on available solar surplus."
 				text: qsTrId("pagecontrollableloads_arrange")
 				preferredVisible: mode.value
 			}
--- a/pages/settings/PageControllableLoads.qml
+++ b/pages/settings/PageControllableLoads.qml
@@ -55,7 +55,7 @@ Page {
 					// which animate up & down by clicking the up & down arrows.
 			y: -gradientListView.contentY
 			leftPadding: Theme.geometry_page_content_horizontalMargin
-			spacing: parent.spacing
+			spacing: Theme.geometry_gradientList_spacing
 
 			Repeater {
 				model: gradientListView.count
@@ -100,6 +100,7 @@ Page {
 			ListNavigation {
 				//% "Preferences"
 				text: qsTrId("pagecontrollableloads_preferences")
+				visible: mode.value
 				onClicked: {
 					Global.pageManager.pushPage("/pages/settings/PageControllableLoadsPreferences.qml", { title: text })
 				}
--- a/pages/settings/PageControllableLoads.qml
+++ b/pages/settings/PageControllableLoads.qml
@@ -93,14 +93,14 @@ Page {
 
 			PrimaryListLabel {
 				//% "Devices will be controlled in order, based on available solar surplus."
-				text: qsTrId("pagecontrollableloads_arrange")
+				text: qsTrId("pagecontrollableloads_controlled_in_order_based_on_available_solar_surplus")
 				preferredVisible: mode.value
 			}
 
 			ListNavigation {
 				//% "Preferences"
 				text: qsTrId("pagecontrollableloads_preferences")
-				visible: mode.value
+				preferredVisible: mode.value && loads.valid
 				onClicked: {
 					Global.pageManager.pushPage("/pages/settings/PageControllableLoadsPreferences.qml", { title: text })
 				}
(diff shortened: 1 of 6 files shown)


pages/settings/PageControllableLoads.qml
```diff
@@ -92,14 +92,15 @@ Page {
...
-				//% "Arrange the controllable devices according to their priority; the control algorithm will control them based on the currently available PV excess."
+				//% "Devices will be controlled in order, based on available solar surplus."
 				text: qsTrId("pagecontrollableloads_arrange")
 				preferredVisible: mode.value
 			}
 
 			ListNavigation {
 				//% "Preferences"
 				text: qsTrId("pagecontrollableloads_preferences")
+				visible: mode.value
```
DanielMcInnes: don't use `visible`, there is an explanation in abstractlistitem.h
preferredVisible: mode.value && loads.valid

DanielMcInnes (commented):
1 change needed in addition to copilot's comments, then lgtm

pages/settings/PageControllableLoadsBattery.qml
```diff
@@ -16,22 +16,25 @@ Page {
...
+
 			SettingsListHeader {
 				//% "Advanced"
 				text: qsTrId("pagecontrollableloads_battery_advanced")
 			}
 
 			ListQuantityField {
 				unit: VenusOS.Units_Watt
-				//% "At <font color=\"%1\">%2%</font> SOC, reserve for battery charging"
+				//% "At or above <font color=\"%1\">%2%</font> SOC"
```
DanielMcInnes, line 37: If you change the text like this, you should change the following qsTrId, otherwise the translators will not be notified that the text should change. Here & elsewhere.

The repo builds translations with lrelease -idbased and lupdate -no-obsolete (cmake/Translations.cmake). Under id-based lookup, changing the //% source text while keeping the id updates <source> but leaves the existing per-language translations finished (not flagged unfinished). Every language .ts already has finished old translations for these two ids, so non-English users would ship the old, now-incorrect text (the meaning changed, not just wording) with no signal to re-translate.
Fix: mint new ids — exactly what this PR already did for the analogous change (…battery_between_the_soc → …battery_from_configured_soc_to_100_percent). The two reused ids are just an inconsistency in the same PR.

pages/settings/PageControllableLoadsBattery.qml
```diff
@@ -41,11 +44,16 @@ Page {
 
 			ListQuantityField {
 				unit: VenusOS.Units_Watt
-				//% "At %1% SOC, reserve for battery charging"
+				//% "At %1% SOC"
 				text: qsTrId("pagecontrollableloads_battery_at_x_soc_reserve_for_battery_charging").arg(100)
 				dataItem.uid: BackendConnection.serviceUidForType("opportunityloads") + "/ReservationEndPower"
-				//% "Between the SOC set in “Allow additional loads from battery SOC” and 100% SOC, the reserved power is adjusted gradually between these values. This allows battery charging to decrease as the SOC rises, leaving more surplus power available for controlled devices."
-				caption: qsTrId("pagecontrollableloads_battery_between_the_soc")
+
```
DanielMcInnes: delete this blank line

pages/settings/PageControllableLoadsPreferences.qml
```diff
@@ -41,12 +39,17 @@ Page {
 			}
 
 			ListSwitch {
+				id: batteryLifeSupportSwitch
 				preferredVisible: dataItem.valid
-				//% "Pause Opportunity Loads when Active SOC limit exceeds 85%"
+				//% "Pause Opportunity Loads after several days without full charge"
```
DanielMcInnes: change the following qsTrId as well.

Should this ticket have the `require_docs` label? It looks like a behavior change. Are you intentionally changing the text while keeping the same dataItem.uid?

pages/settings/PageControllableLoadsPreferences.qml
```diff
@@ -41,12 +39,17 @@ Page {
...
 				text: qsTrId("page_controllable_loads_preferences_pause_opportunity_load_when_active_soc_limit_exceeds_85")
 				dataItem.uid: BackendConnection.serviceUidForType("opportunityloads") + "/BatteryLifeSupport"
-				//% "This helps the BatteryLife algorithm recharge the battery to 100%."
-				caption: qsTrId("page_controllable_loads_preferences_this_helps")
+			}
+
+			PrimaryListLabel {
+				//% "Only applies when using Optimized with BatteryLife. Opportunity Loads automatically resumes after a full charge."
```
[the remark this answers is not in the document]
philipptrenz: "Opportunity Loads" is a proprietary term, so the singular form is correct.

philipptrenz:
Hi @DanielMcInnes, thanks for the review. I guess I misunderstood when new translation IDs are necessary, I now updated all accordingly and checked via selecting a different language. I also fixed the visibility control issue and made two other small changes where I saw that the length of the label caused a layout issue in German translation, shortening the text accordingly.

DanielMcInnes:
lgtm

DanielMcInnes (approved):

PR #3103, opened 2026-07-03
cmake: Avoid AGL linker failure on macOS by removing WrapOpenGL from Qt6::Gui

philipptrenz:
This PR fixes a macOS desktop build/link issue where venus-gui-v2 fails with `ld: framework 'AGL' not found`.

On recent macOS/Xcode toolchains, AGL can be pulled in via Qt’s WrapOpenGL path. To prevent that, this change removes WrapOpenGL::WrapOpenGL from Qt6::Gui interface link libraries on APPLE builds in `BuildRequirements.cmake`.

With this in place, venus-gui-v2 links successfully again on macOS desktop builds (tested with Qt 6.8.3 on Apple Silicon).

### DIFF
--- a/cmake/BuildRequirements.cmake
+++ b/cmake/BuildRequirements.cmake
@@ -5,6 +5,20 @@ find_package(Qt6 ${REQUIRED_QT_VERSION}
         Core Gui Qml Quick QuickControls2 Svg Xml Mqtt LinguistTools ShaderTools
     REQUIRED)
 
+if(APPLE)
+    # Recent macOS/Xcode toolchains cannot link against AGL.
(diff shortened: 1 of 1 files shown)


chriadam:
LGTM but I don't have much experience with Apple stuff.  @blammit would be better to review than me.

blammit (approved):
@philipptrenz According to https://qt-project.atlassian.net/browse/QTBUG-137687 this is fixed in Qt 6.8.4. Can you add a comment to say this workaround can be removed once we upgrade to the next Qt version?

Otherwise, LGTM.

blammit (approved):

PR #3113, opened 2026-07-14
Battery: hide battery bank settings menu when empty

### LINKED ISSUE #3111: Battery: Hide battery bank settings menu when empty
The upcoming SuperPack NG VE.Can battery does not have a configurable capacity and voltage, so those settings items are not available.
The battery bank settings menu is empty and should be hidden. The restore defaults and Bluetooth Enabled items should be kept.

Image
Image

blammit:
Fixes #3111


pages/settings/devicelist/battery/PageBatterySettings.qml
```diff
@@ -34,11 +41,12 @@ Page {
 	GradientListView {
 		model: VisibleItemModel {
 			ListNavigation {
+				preferredVisible: batteryBankModelMonitor.hasVisibleItem
 				//% "Battery bank"
 				text: qsTrId("batterysettings_battery_bank")
 				onClicked: {
 					Global.pageManager.pushPage("/pages/settings/devicelist/battery/PageBatterySettingsBattery.qml",
```
chriadam: This page is now the same as emptySettingsComponent, isn't it?

blammit: Oh good point... updated to use that and remove the QML file.

chriadam (approved):

PR #3118, opened 2026-07-17
WASM: avoid pulling all Controls styles into the static build

mman:
Reworked after review: the original approach excluded QZXing entirely from the WASM build, but as @blammit pointed out, the [EE Bus settings page](https://github.com/victronenergy/gui-v2/blob/main/pages/settings/PageSettingsEebus.qml#L76) explicitly forces `ListLink_Mode_QRCode`, so QR rendering via QZXing **is** reachable on WASM. QZXing now stays in the build on all platforms.

The size win is preserved anyway: qzxing itself is small (~120 KB) — the real weight was its transitive QtQuickControls2 dependency, which made `qt_import_qml_plugins` pull every Controls style plugin (Material, Fusion, Universal, Imagine, FluentWinUI3 ≈ 7.7 MB) into the statically linked binary. This PR removes that dependency at the source:

- Bump the qzxing submodule to drop its QtQuickControls2 dependency and unqualified Controls imports (https://github.com/victronenergy/qzxing/pull/3 — must merge first)
- Switch `tools/uicompare` to `QtQuick.Controls.Basic` imports so the tool doesn't reintroduce the style pull-in

All builds from identical toolchain: Qt 6.8.3 wasm_singlethread, MinSizeRel.

| Configuration | .wasm (uncompressed) | .wasm (gzip -9) |
|---|---|---|
| Before (all Controls styles linked) | 35,751,542 bytes (34.1 MiB) | 15,739,886 bytes (15.0 MiB) |
| After (Basic style only, QZXing excluded — previous approach) | 31,615,474 bytes (30.2 MiB) | 13,992,236 bytes (13.3 MiB) |
| **After (Basic style only, QZXing included — this PR)** | **31,732,154 bytes (30.3 MiB)** | **14,031,798 bytes (13.4 MiB)** |
| **Saved vs before** | **4,019,388 bytes (−11.2%)** | **1,708,088 bytes (−10.9%)** |

Verified in the final binary: QZXing symbols present (QR pairing functional), and only the Basic Controls style is compiled in — Material, Fusion, Universal, Imagine and FluentWinUI3 no longer appear.

### DIFF
--- a/tools/uicompare/CompareView.qml
+++ b/tools/uicompare/CompareView.qml
@@ -1,5 +1,5 @@
 import QtQuick
-import QtQuick.Controls
+import QtQuick.Controls.Basic
 import QtQuick.Layouts
 
 Item {
--- a/tools/uicompare/ImageListView.qml
+++ b/tools/uicompare/ImageListView.qml
@@ -1,5 +1,5 @@
 import QtQuick
-import QtQuick.Controls
+import QtQuick.Controls.Basic
 
 ListView {
 	id: root
--- a/tools/uicompare/Main.qml
+++ b/tools/uicompare/Main.qml
@@ -1,8 +1,7 @@
 import QtQuick
 import QtQuick.Layouts
 import QtQuick.Window
-import QtQuick.Controls
-import QtQuick.Dialogs
+import QtQuick.Controls.Basic
 
 Window {
 	width: 1280
--- a/tools/uicompare/Main.qml
+++ b/tools/uicompare/Main.qml
@@ -1,8 +1,7 @@
 import QtQuick
 import QtQuick.Layouts
 import QtQuick.Window
-import QtQuick.Controls
-import QtQuick.Dialogs
+import QtQuick.Controls.Basic
 
 Window {
 	id: root
--- a/tools/uicompare/Toolbar.qml
+++ b/tools/uicompare/Toolbar.qml
@@ -1,6 +1,6 @@
 import QtQuick
 import QtQuick.Layouts
-import QtQuick.Controls
+import QtQuick.Controls.Basic
 
 Rectangle {
 	id: root


tools/uicompare/CompareView.qml
```diff
@@ -1,256 +1,308 @@
 import QtQuick
-import QtQuick.Controls
+import QtQuick.Controls.Basic
```
chriadam, line 2: instead of these changes, can we simply ensure that the uicompare tool is excluded from the wasm build altogether, so that the import scanner never sees this?

DanielMcInnes (approved):
lgtm

blammit (changes requested):
Unfortunately we do have one case where QR code generation is required in Wasm -- in the [EE Bus settings page](https://github.com/victronenergy/gui-v2/blob/main/pages/settings/PageSettingsEebus.qml#L76), where it explicitly prefers the QR code display. It would be great to exclude unused libraries in Wasm, but I don't think we can do it for QZXing in this case.

chriadam:
> Unfortunately we do have one case where QR code generation is required in Wasm -- in the [EE Bus settings page](https://github.com/victronenergy/gui-v2/blob/main/pages/settings/PageSettingsEebus.qml#L76), where it explicitly prefers the QR code display. It would be great to exclude unused libraries in Wasm, but I don't think we can do it for QZXing in this case.

Why does it show a QR code in WASM mode, though?  The user cannot take their phone out to scan the QR code in that case, because they ARE using their phone already.  @ReinvdZee is it intentional?

ReinvdZee:
> Why does it show a QR code in WASM mode, though? 

I'm not sure, but I guess it has to do with the EEBUS pairing mechanism. I'm sure @citolen can explain why we show the QR code in WASM and if it is necessary.

> because they ARE using their phone already.

Not necessarily. A user can also view it on a desktop etc.

citolen:
@chriadam 
>Why does it show a QR code in WASM mode, though?

The QR code content isn't a link hence why on WASM it's also shown.

The QR code is meant to be used to pair with an other EEBUS device.
An example would be pairing with a heat pump, the installer will need to scan the QR code from the heat pump's app to trust the GX device.

Whether they scan the GX device's screen or through VRM (WASM), it doesn't matter, it needs to show the QR code.

The pairing mechanism does not always require a QR code and is still being refined by the EEBUS initiative so we might be able to remove it at some point, but for now let's leave it in.

chriadam:
I see, this is for the "installer is inspecting the device via VRM (i.e. using gui-v2 WASM build)" case.  So they're not necessarily using their phone, they might be in their office using laptop/computer to manage user installs.  I see.

So, it looks like we cannot remove the QR code stuff from the wasm build after all, I see now.

tools/uicompare/CompareView.qml
blammit: @mman could you follow up on this? ^

mman: @chriadam I looked into this — it turns out uicompare is already not part of the wasm build (or any build: tools/uicompare is a standalone CMake project that nothing references). The scanner sees it anyway because qt_import_qml_plugins runs qmlimportscanner with -rootPath set to the repo root, which scans every .qml file in the tree regardless of what's built — same reason the qzxing submodule's examples were pulling styles in.

qmlimportscanner does have an -exclude option, but it's broken upstream: the directory iterator yields each directory both as dir and as its dir/. alias, and the basename-based exclusion only catches the first form — the alias still gets scanned, so exclusion has no effect on the output (verified with a minimalrepro on 6.8.3; the code is unchanged in dev). Even if it worked, it matches directory basenames only and doesn't prune subtrees, so we'd be maintaining a fragile blocklist of every QML-containing directory name, and CMake offers no hook to pass the flag anyway.

Given that, qualifying the imports as QtQuick.Controls.Basic seems to be the least invasive fix available: with it, a full-tree scan yields only QtQuick.Controls.Basic + impl (down from all seven styles). I'll file a QTBUG for the broken -exclude; if it gets fixed we could revisit, though the import qualification is arguably self-documenting either way.

mman:
For the reference, filed https://qt-project.atlassian.net/browse/QTBUG-149507 about `qmlimportscanner -exclude` not working at all...

PR #3129, opened 2026-07-26
WASM: change breakpoints when frame is displayed in default mode

### LINKED ISSUE #3037: WASM Frame disappearing too early
When viewing GUI-v2 on a webbrowser over WASM, the Frame disappears automatically when the windows is made smaller. However, it seems that the frame disappears quite early in this case:

Image

When the window is made larger the frame will reappear.

mr-manuel:
Fixes https://github.com/victronenergy/gui-v2/issues/3037

### DIFF
--- a/wasm/index.html
+++ b/wasm/index.html
@@ -50,7 +50,7 @@
           background: #000;
           color: #969591;
       }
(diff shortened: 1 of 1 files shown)


mr-manuel:
@blammit please merge

blammit (approved):

PR #3137, opened 2026-07-30
Fix/wasm ios orphaned touch recovery

### LINKED ISSUE #3087: As standalone iOS webapp, gui-v2 locks up after switching to another app and back
On iPhone, open safari and access GUI v2. Then share the link, and add to Home Screen to get a nice icon.

Then I can go to GUI v2 directly by tapping on the icon and the app runs as a web app in a browser engine without any navigation chrome. 

Works excellent.

Now switch to another app, do some other things and then switch back to the GUI v2 app, and it is locked up and does not react to touches.

I do not know precisely what locks it up, but I observed that waiting longer before switching back to GUI v2 will trigger the MQTT reconnect screen, but GUI will still remain locked up. The UI is still updating, values changing but no touch working.

mman:
## Summary

Touch input in the WASM app can die permanently — while the UI keeps rendering and updating — after the OS takes over a touch gesture that started on the page. On iOS home-screen webapps this happens on every home-indicator / App Switcher / Control Center swipe. On Android it happens with system back gesture, or bottom swipe up on the home indicator to reveal app switcher.

This PR adds a small, self-contained "touch input recovery guard" to `wasm/index.html` that works around the underlying Qt bug. No reload, no C++/QML changes, no effect on desktop or GX devices.

Fixes #3087.

## Root cause
  
This is a bug in the Qt WASM platform plugin, not in WebKit and not in gui-v2:

1. When the OS takes over a gesture (iOS bottom-edge swipes, Android back gesture), the browser correctly delivers `pointerdown` followed by **`pointercancel`** to the page. Verified with on-device event logging on iOS 27.
2. The Qt WASM platform plugin (`qwasmwindowclientarea.cpp` / `qwasmwindow.cpp`) keeps a per-pointer touch map, but removes entries **only on `pointerup`** — a touch ending in `pointercancel` leaks a permanently-registered, permanently-pressed touch point.
3. That stale point is included in every subsequent touch event Qt delivers, which wedges Qt Quick's touch→mouse synthesis (`touchMouseId`). Since gui-v2's tap surfaces are largely MouseArea-based (`PressArea`, `ListPressArea`, `idleModeMouseArea`), effectively all touch input goes dead until the page is reloaded.

## Upstream status

Reported upstream as [QTBUG-141505](https://qt-project.atlassian.net/browse/QTBUG-141505) (Android back gesture, closed/fixed) and [QTBUG-147635](https://qt-project.atlassian.net/browse/QTBUG-147635) (this iOS symptom, still open, not yet linked to the root cause).

Fixed by a one-line change in qtbase:
[`9147c25004` — "wasm: remove pointer ids on pointercancel"](https://github.com/qt/qtbase/commit/9147c25004) (Dec 2025, `Pick-to: 6.10 6.11`).

| Qt version | Contains the fix |
|---|---|
| 6.8.x (gui-v2 wasm toolchain: **6.8.3**) | ❌ no backport |
| 6.9.x (incl. 6.9.3) | ❌ no backport |
| 6.10.0 / 6.10.1 | ❌ |
| **6.10.2+** | ✅ |
| 6.11+ / dev | ✅ |

## The workaround
  
A capture-phase script block in `wasm/index.html`:

- After Qt has processed a trusted `pointercancel`, it dispatches the missing synthetic `pointerup` into Qt's client-area element (inside the `#screen` shadow DOM) so Qt cleans up its per-pointer state — the exact cleanup the upstream fix performs internally. The release is preceded by a `pointermove` away from
  the press position so it cannot complete the stuck press into a spurious click.
- As a safety net, touch pointers whose end event never arrived at all are released the same way, but only on provable triggers (page hidden, a new *primary* touch, a `touchstart` reporting a single remaining touch).
- `Element.prototype.setPointerCapture`/`releasePointerCapture` are wrapped to be non-throwing, because Qt calls them unguarded and a `NotFoundError` for an inactive pointer aborts Qt's event handler mid-dispatch.

## Validation

- **On device (iPhone, iOS 27 → Cerbo GX):** swiping to the App Switcher previously left touch dead until reload; with the guard, the log shows `pointercancel` followed by the guard's repair, and touch works immediately after returning.
- **On device (Samsung, Android 16 → Cerbo GX):** swiping to the App Switcher previously left touch dead until reload; with the guard, the log shows `pointercancel` followed by the guard's repair, and touch works immediately after returning.
- **Headless (Chromium + CDP trusted touch injection against the wasm build):** baseline taps/flicks unaffected; cancel-ending and never-ending gestures both recover; the repair itself causes no page change.

## Notes

- VRM ships its own `index.html`, so it needs the same guard (coordination required).
- This entire script block can be dropped once the wasm toolchain moves to Qt ≥ 6.10.2 (or the upstream fix is backported to 6.8/6.9 LTS).
- Suggested upstream follow-up: comment on QTBUG-147635 linking QTBUG-141505 and `9147c25004`, noting the mechanism reproduces on iOS standalone webapps, and request an LTS backport.

### DIFF
--- a/wasm/index.html
+++ b/wasm/index.html
@@ -610,11 +610,18 @@
         // reacting to touch until the page is reloaded. See
         // https://github.com/victronenergy/gui-v2/issues/3087.
         //
-        // This guard tracks active touch pointers and, when a pointer is
-        // provably orphaned, synthesizes the missing pointerup into the Qt
-        // window element so Qt cleans up its state. The synthetic release is
-        // preceded by a pointermove away from the press position so that the
-        // release cannot complete the orphaned press into a spurious click.
+        // This guard repairs two related WebKit/Qt failure modes, both of
+        // which corrupt Qt's touch state the same way:
+        // 1. a touch pointer whose pointerup/pointercancel is never delivered
+        //    (orphaned pointer) - flushed once it is provably orphaned;
+        // 2. a properly delivered pointercancel: Qt (up to at least 6.9)
+        //    removes touch points from its bookkeeping only on pointerup, so
+        //    a cancelled touch leaks a permanently-pressed point - repaired
+        //    by delivering the missing pointerup after the cancel. This is
+        //    the mode observed on iOS 27.
+        // Synthetic releases are preceded by a pointermove away from the
+        // press position so that the release cannot complete a stuck press
+        // into a spurious click.
         (function() {
             'use strict';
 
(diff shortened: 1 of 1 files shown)


mr-manuel (approved):
LGTM

PR #3139, opened 2026-07-31
Generator: ensure run time values are not shown as NaN

### LINKED ISSUE #3130: Generator total run time: NAN value
The generator total runtime value is shown as NAN:

Image

The related settings values are present and contain valid data, for example: 

Image

Since GUI-v2 directly pulls from `com.victronenergy.settings`, I expected this to work properly on local GUI but not on WASM, but I also see it on local GUI.

Can be reproduced by configuring relay 1 as "Genset Start/Stop".

blammit:
If a runtime VeQuickItem value is undefined (because the value is not yet loaded from the backend) then passing this value to a  JavaScript mathematical operation or function results in a NaN value. So, use zero instead of the value if the backend value is undefined (i.e. not yet available).

Fixes #3130

### DIFF
--- a/components/listitems/core/ListTextField.qml
+++ b/components/listitems/core/ListTextField.qml
@@ -100,7 +100,7 @@ ListSetting {
 			validateOnFocusLost: root.validateOnFocusLost
 
 			onInputValidated: root.saveInput()
-			onTextChanged: root.secondaryText = text
+			onTextEdited: root.secondaryText = text
 
 			Layout.minimumWidth: Theme.geometry_listItem_textField_minimumWidth
 			Layout.maximumWidth: Theme.geometry_listItem_textField_maximumWidth
--- a/pages/settings/PageGeneratorRuntimeService.qml
+++ b/pages/settings/PageGeneratorRuntimeService.qml
@@ -35,7 +35,7 @@ Page {
 	VeQuickItem {
 		id: gensetOperatingHours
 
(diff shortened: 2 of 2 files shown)


chriadam:
Does the NaN cause some exception to occur, which prevents the binding from re-evaluating when the real / correct value is available?  I ask because in the issue, the dbus path had the correct value, but the UI was displaying NaN still.

blammit:
> Does the NaN cause some exception to occur, which prevents the binding from re-evaluating when the real / correct value is available? I ask because in the issue, the dbus path had the correct value, but the UI was displaying NaN still.

Not an exception, but the mathematical operation returns NaN if you pass it an `undefined` value, which is the case if the data is not yet loaded.

chriadam:
Right, but my point is: according to the screenshot in the issue, the value was available.  So I don't understand why the UI got "stuck" showing NaN, if the actual value became available in the backend.

blammit:
Ah - I see... no, this fix is incorrect. This doesn't properly fix the issue, and also these changes shouldn't be applied across the different text fields. I think there is something going on in ListTextField instead, will take another look.

blammit:
@chriadam fixed now; the issue is that ListTextField was overwriting the binding before the final `secondaryText` value was provided by the runtime settings.

ReinvdZee:
I tested this, can confirm it is solved.
Thanks!

chriadam (approved):

PR #3142, opened 2026-07-31
EVCS: show auto mode source and hide charge current in auto mode

philipptrenz:
- add Auto mode source row below Charge mode
- show row only when mode is Auto and GxAutoMode path is available
- use GxAutoMode/Source for external label when present, fallback to External (GX device)
- show charge current only in Manual mode

### DIFF
--- a/pages/evcs/EvChargerPage.qml
+++ b/pages/evcs/EvChargerPage.qml
@@ -136,25 +136,17 @@ DevicePage {
 		ListText {
 			//% "Auto mode source"
 			text: qsTrId("evcs_auto_mode_source")
-			//% "Internal (EV Charging Station)"
-			readonly property string evcsInternalText: qsTrId("evcs_auto_mode_source_evcs_internal")
-			readonly property string externalSourceName: {
-				if (gxAutoModeSource.valid && gxAutoModeSource.value !== undefined && gxAutoModeSource.value !== null) {
-					const sourceText = String(gxAutoModeSource.value).trim()
-					if (sourceText.length > 0) {
-						return sourceText
-					}
-				}
-				//% "GX device"
-				return qsTrId("gx_device")
-			}
-			//: %1 = source string from /GxAutoMode/Source, or "GX" when not available
+			readonly property string externalSourceName: (gxAutoModeSource.value ?? "")
+					//% "GX device"
+					|| qsTrId("gx_device")
+			//: %1 = source string from /GxAutoMode/Source, or "GX device" when not available
 			//% "External (%1)"
-			secondaryText: Number(dataItem.value) === 1
+			secondaryText: dataItem.value === 1
 						   ? qsTrId("evcs_auto_mode_source_external_with_source").arg(externalSourceName)
-						   : evcsInternalText
+						   //% "Internal (EV Charging Station)"
+						   : qsTrId("evcs_auto_mode_source_evcs_internal")
 			dataItem.uid: evCharger.serviceUid + "/GxAutoMode/Enabled"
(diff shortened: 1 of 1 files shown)


philipptrenz:
Some context: With Venus OS v3.80, we're adding functionality to the GX device to control the Auto mode of the EVCS. This can be "Opportunity Loads" or other future optimisation algorithms. If the GX device does not "claim" to control the Auto mode, the EVCS uses its internal algorithm for surplus charging. This new "Auto mode source" is added to make transparent to the user which algorithm is currently in charge.

evcs_internal

evcs_external_ol

pages/evcs/EvChargerPage.qml
```diff
@@ -133,10 +133,41 @@ DevicePage {
...
+				uid: evCharger.serviceUid + "/GxAutoMode/Source"
+			}
+		}
+
 		ListEvcsSetCurrentSpinBox {
 			serviceUid: evCharger.serviceUid
 			text: CommonWords.charge_current
-			interactive: dataItem.valid && chargeMode.dataItem.value === VenusOS.Evcs_Mode_Manual
+			preferredVisible: dataItem.valid && chargeMode.dataItem.value === VenusOS.Evcs_Mode_Manual
 		}
```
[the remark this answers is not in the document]
philipptrenz: This is on purpose, control card layout should not change

pages/evcs/EvChargerPage.qml
```diff
@@ -133,10 +133,39 @@ DevicePage {
...
+					}
+				}
+				//% "GX device"
+				return qsTrId("gx_device")
+			}
+			//: %1 = source string from /GxAutoMode/Source, or "GX" when not available
+			//% "External (%1)"
+			secondaryText: Number(dataItem.value) === 1
+						   ? qsTrId("evcs_auto_mode_source_external_with_source").arg(externalSourceName)
+						   : evcsInternalText
```
blammit: Is the `Number` conversion necessary? If we know the backend value is always an int, it would be nice to be explicit and just have `dataItem.value === `.

Also if `evcsInternalText` is only used here, you could just place the qsTrId() inline, like:

```
secondaryText: Number(dataItem.value) === 1
    //: %1 = source string from /GxAutoMode/Source, or "GX" when not available
    //% "External (%1)"
   ? qsTrId("evcs_auto_mode_source_external_with_source").arg(externalSourceName)
    //% "Internal (EV Charging Station)"
   : qsTrId("evcs_auto_mode_source_evcs_internal")
```

pages/evcs/EvChargerPage.qml
```diff
@@ -133,10 +133,39 @@ DevicePage {
...
+			readonly property string externalSourceName: {
+				if (gxAutoModeSource.valid && gxAutoModeSource.value !== undefined && gxAutoModeSource.value !== null) {
+					const sourceText = String(gxAutoModeSource.value).trim()
+					if (sourceText.length > 0) {
+						return sourceText
+					}
+				}
+				//% "GX device"
+				return qsTrId("gx_device")
+			}
```
blammit: Looks like this binding could be simplified to:

```
readonly property string externalSourceName: (gxAutoModeSource.value ?? "").trim() 
       //% "GX device"
    || qsTrId("gx_device")
```

Also, is the `trim()`` needed? The codebase doesn't tend to use trim() when showing name values.

philipptrenz: Thanks, indeed not needed and done.

pages/evcs/EvChargerPage.qml
philipptrenz: Yep, both done.

blammit (approved):

PR #3150, opened 2026-08-07
Claude optimized

DanielMcInnes:


DanielMcInnes:
*Overview electron animation — further optimisations on top of PR #3121*

Benchmarked on a Cerbo GX. Baseline is _this PR's head_, not main, so these compound with Chris's numbers.

```
Metric       PR head   Branch    Delta
FPS avg        77.8      83.3    +7.1%
Total avg    12.86ms   12.01ms   -6.6%
Render avg    3.47ms    2.02ms  -41.8%
Render p95       7ms       5ms  -28.6%
Total p95       18ms      15ms  -16.7%
Total p99       26ms      24ms   -7.7%
Total max       53ms      39ms  -26.4%
```
3 x 20s captures per build (4102/4365 frames), 22s warmup, MinSizeRel, 800x480, via scripts/benchmark-overview.py.

*Where the time was going*
Sync is already ~0.03ms, so the LUT has done its job. What is left is alpha-batch invalidation and vertex re-upload. From QSG_RENDERER_DEBUG=render on device:
```
Renderer rebuild      PR head   Branch
none needed             40%      67%
batches only            42%      33%
FULL (whole scene)      18%       0%
```

*1. The electrons were faded between exactly 0.0 and exactly 1.0.* The renderer reclassifies an item at 0.001 (subtree blocked, geometry leaves the render lists) and again at 0.999 (eligible to be treated as opaque), and each reclassification forces a full rebuild of every batch in the window — re-uploading 3519 vertices across 61 batches, not just the electrons'. Every fade crossed both boundaries, twice. Fading between 0.002 and 0.996 removes them entirely, and both endpoints are within half of one 8-bit alpha step of the value they replace.

*2. Behavior on opacity { OpacityAnimator } kept the scene dirty between ticks.* An OpacityAnimator advances on the render thread on every frame. On GX the connector ticks are deliberately throttled to 20fps, but the fade animator went on dirtying the scene in between — a quarter of all frames for that reason alone. Computing the fade from the electron's progress along the path, at tick time, takes dirty frames from 60% down to 33%. The fade duration is unchanged, and it deletes the Behavior, the Animator and one QML property.

Plus loop hygiene: hoisted the invariants out of the per-electron loop, and one setPosition() instead of setX()+setY().

*Measured and rejected*
Raising the electron overlay's z so its render order sits above the rest of the page's translucent content. Right in theory, but the trace shows the only batches ever invalidated on this page are the electrons' own, so there was nothing to gain — measured no better (3.27ms vs 2.93ms). Reverted.

*Verification*
- All 24 ctest unit tests pass
- A/B screenshots of the animated Overview page vs the PR head are visually identical

*Remaining headroom, not done*
- 2 of every 3 rendered frames are pixel-identical. The 20fps throttle suppresses the property update, but FrameAnimation still forces a repaint every vsync. Fixing that needs a plain (non animation-driver) timer — bigger than belongs in this PR, and it would not move the table above, since the render loop is free-running so saved time becomes extra frames rather than idle. It would free roughly 2/3 of the page's rendering work.
- ~2.5ms of a dirty frame is fixed cost regardless of how much changed: 1.6 batches / 77 vertices costs 2.58ms, 61 batches / 3519 vertices costs 3.89ms. Looks like a Vivante driver stall on glBufferSubData rather than anything QML can fix.

components/PageStack.qml
```diff
@@ -14,13 +14,29 @@ StackView {
 	readonly property Page currentPage: opened ? currentItem : null
 
 	readonly property int animationDuration: Global.mainView && Global.mainView.allowPageAnimations ? Theme.animation_page_slide_duration : 0
-	readonly property bool animating: busy || fakePushTransition.running || fakePopTransition.running
+	// True while navigation is in flight and the stack has not settled: a page is
```
chriadam, line 17: a lot of these comments seem completely unnecessary

chriadam:
Overall - I like the overview animation improvements.  But, I do NOT like the asynchronous page building change - I guess it's nice that animations and timers aren't blocked while the page content is built asynchronously, but honestly, the UI should react immediately to user interactions.  The idea that the user might press something, and then the UI will look like NOTHING happened, until 700ms later when the page is finished constructing it gets pushed to the pagestack and the animation happens ... this is terrible.  We NEED to focus on reducing page construction time - maybe that means looking at all of the components one-by-one and shaving properties, signals, interior objects, whatever we can to reduce the papercuts that add up to 700ms page build time.

chriadam:
Maybe we need to rethink the VisibleItemModel thing - instead of requiring every instance to be constructed, only construct visible instances (via a proper delegate / model)?

chriadam:
I had a quick chat with Bea, and I will now have a crack at implementing something.

DanielMcInnes:
## Review of #3150

Reviewed at `9592c785` against merge-base `e25ed3ba` (169 files, +16350/−10876). Static review only — **nothing here was verified by building or running**; the on-device A/B benchmark is still in progress and will follow in a separate comment.

Method: 10 independent review passes over disjoint file sets, plus a second opinion from an external model on 16 batched payloads (full post-change file + diff for every changed file). Findings that were reached independently by two or more passes are marked **[×2]** / **[×3]**. I re-verified the blockers against the source myself; one widely-reported finding turned out to be a false positive and has been dropped.

The port itself is a good idea and the mechanism is sound. But `DelegateComponentModel` changes a semantic that ~144 files depended on, and the port did not account for it consistently. Most of what follows is one root cause with many instances.

---

### The root cause

`VisibleItemModel` filtered on the **item's** `effectiveVisible`, which for `ListSetting` is `preferredVisible && userHasReadAccess` (`ListSetting.qml:84`). `DelegateComponentModel` filters only on `DelegateComponent.preferredVisible` (`delegatecomponentmodel.cpp:144`, `:350`), evaluated **outside** the delegate scope, and the delegate's contents do not exist until the row is already visible.

So anything the old code read *off a child item* is now either unresolvable or not yet created. Three failure shapes follow, and they account for most of the findings below:

1. **Self-deadlock** — `preferredVisible` depends on a property whose only writer is a `Binding` *inside* the delegate it gates. It starts false, so the delegate is never built, so the writer never runs. The row can never appear.
2. **Lost scope** — `preferredVisible` references an id or property declared on the inner list item. The binding throws, and `preferredVisible` silently falls back to its C++ default `true`, so the row is now always in the model.
3. **Lost identity** — an `id:` was moved off the `DelegateComponentModel` onto the first inner `DelegateComponent`, so `model:` is handed a plain QObject.

`PageSettingsTankPump.qml:33-45` is the correct pattern for (3); `PageSettingsDvcc.qml:33` and `NetworkSettingsPageModel.qml:112` are the correct pattern for (2).

---

## Blocking

**1. `pages/ev/EvPage.qml:108-109` — orphaned expression fragment turns a binding into a function call. [×2]**

The base had `unit:` on line 66 followed by `preferredVisible:` on 67-68. The port moved line 67 up to the `DelegateComponent` and left line 68 behind:

```qml
unit: VenusOS.Units_Watt
                 (chargingState.value === 3 || chargingState.value === 256 || chargingState.value === 259)
```

JS does not insert a semicolon before `(`, so this parses as `VenusOS.Units_Watt(...)`. Open the EV page on a vehicle that is charging: the `unit` binding throws `TypeError: ... is not a function` on every evaluation and the Power row loses its `W` suffix. The line should just be deleted.

**2. Model `id:` moved onto the inner `DelegateComponent` — list renders empty. [×3]**

- `PageSettingsBluetooth.qml:19` → ids at `:23`, `:34`; both `DelegateComponentModel`s (`:21`, `:32`) are anonymous. Both branches of the Bluetooth page are blank.
- `PageSettingsGsm.qml:49` → `notConnected` is at `:53` on the `DelegateComponent`. The *disconnected* branch is broken (`modemConnected` at `:63` is correctly on the model, so the connected branch is fine).
- `PageSettingsRelayTempSensors.qml:96` → `tempRelayModel` at `:102` and `disabledModel` at `:136` are both on `DelegateComponent`s. Both branches blank.

**3. `preferredVisible` self-deadlock — the row can never appear.**

| Location | Gated on | Written by |
|---|---|---|
| `PageBattery.qml:423` **[×3]** | `batteryDetailsDC.hasAllowedItem` | `Binding` at `:436`, inside the delegate |
| `PageBattery.qml:466` **[×3]** | `batteryHistoryDC.hasAllowedItem` | `Binding` at `:478`, inside the delegate |
| `PageSettingsConnectivity.qml:98` **[×3]** | `canInterfaceRepeaterDC.count` | `Binding` at `:131` |
| `PageSettingsBleSensors.qml:124` **[×3]** | `sensorRepeaterDC.count` | `Binding` at `:146` |
| `PageSettingsDvcc.qml:211-213` **[×3]** | `optionModel.length` | `/AvailableBmsServices` handler at `:226-241` |
| `PageSolarStats.qml:62` **[×2]** | `errorModelDC.count` | `Binding` at `:77` |

Battery **Details** and **History** become unreachable on every battery; VE.Can/BMS-Can rows never appear on Connectivity; the discovered BLE sensor list is always empty; **Controlling BMS** never returns once "No BMS control" is selected; charger error history is gone. Fix is the same everywhere: hoist the model/monitor object to page scope.

**4. `preferredVisible` referencing an inner-item property — binding throws, row defaults to visible.**

- `BatterySettingsAlarmModel.qml:15,27,41,55,69,83,99` **[×2]** — `dataValid` belongs to `ListRangeSlider`. All 7 alarm sliders now always show; a BMV that publishes none of them renders 7 dead 0–100 sliders.
- `BatterySettingsRelayModel.qml:74,88,102,116,130,154,170` **[×2]** — same, and it also drops the `showSetting(...)` relay-mode filter, so all 7 thresholds show in "Manual control"/"Always open".
- `PageMeteo.qml:53` **[×2]** — `configStatus` is on the inner `ListLink`. A correctly configured sensor permanently shows the red "Configuration required" link.
- `PageSettingsCanOpenMotordrive.qml:29` **[×2]**, `PageSettingsModbus.qml:36`, `PageSettingsRvcDevice.qml:80` — `userHasWriteAccess` is a `ListSetting` property. Rows intended to be installer-only are now shown to User-level accounts. Not a privilege escalation (`checkWriteAccessLevel()` still fires on click), but wrong. `PageSettingsRvcDevice.qml:80` also loses its `isLocalSender` term, making a previously unreachable page reachable.
- `PageSettingsBatteryMeasurements.qml:39-42` **[×2]** — page-level `nameItem.uid` reads `batteryMenuItem`, which is a ListView delegate id. Batteries can no longer be renamed.

**5. `pages/settings/debug/PageDebug.qml:83, 91` — `DelegateComponent` with no delegate. [×2]**

Two fully commented-out TODO blocks were wrapped in `DelegateComponent`s. `delegatecomponentmodel.cpp:261-264` returns `nullptr`, and `QQuickItemViewPrivate::createItem` then breaks out of its fill loop — so `count` is 10 but only 8 rows can ever be created, with a `qmlWarning` on every refill. Don't wrap them.

**6. `PageSettingsModbusAddDevice.qml:51,80,103` and `PageSettingsFroniusAddLocation.qml:24,46` — "Add device" throws.**

`runValidation()` on the `DelegateComponent` references `ipAddress` / `port` / `unit`, which are ids inside the delegate `Component`. QML ids are component-scoped and do not resolve from the enclosing scope → `ReferenceError` in `onClicked`, and nothing happens. Separately, the fallback path returns `Utils.validationResult(...)` (an object) where the caller compares against an int enum, so it always reads as "not OK".

---

## High

- **`PageSettingsDvcc.qml:32` and `:93` — `ListSwitchForced` "forced on" state dropped.** `dvccActive` became `dataItem.value === 1`, but `ListSwitchForced.checked` is `value === 1 || value === Switch_ForcedOn` (3). With DVCC forced on by a managed battery/ESS, **nine** rows vanish from the DVCC page. Same at `:93` for Shared Temperature Sense.
- **`PageSettingsGsm.qml:287, 323, 334` — `checkable` switches no longer reveal their rows. [×3]** `checked` is recomputed from the backend value, which the click does not change. "Use default APN" off never shows the APN field; "Use authentication" on never shows the credential fields. Both settings become unreachable.
- **`PageSettingsHub4Feedin.qml:19` — inverted condition. [×3]** `checked: dataItem.value === 1` on `/Settings/CGwacs/PreventFeedback`, but the switch sets `invertSourceValue: true`. "Limit system feed-in" is hidden exactly when AC feed-in is enabled.
- **`PageSettingsHub4.qml:206, 245` — visibility term moved into `checked`.** The rows now display while the DVCC max-charge-current control is active / BatteryLife is Keep-charged, permanently unchecked, and each tap still writes `MaxChargePower`/`MaxDischargePower` with no visible feedback.
- **`PageTankSetup.qml:128, 141` — `.valid` guard re-pointed. [×3]** Tests `rawValueEmptyItem`/`rawValueFullItem` instead of `/Standard`. The correct id (`standardItem`) already exists at `:34`. Sensors without `/Standard` can no longer be calibrated.
- **`PageSettingsLogger.qml:164, 213` — `.valid` guard re-pointed** to `logIntervalItem` / `timeLastContactItem` instead of `vrmPortalMode`.
- **`PageRsSystemAlarms.qml:26,40,53,76` — "No system alarms" shown while alarms are published.** Guards only the top-level `/Alarms/<x>` path, but `VeBusAlarm` also renders `/Alarms/L1|L2|L3<x>`. An acsystem publishing only per-phase paths falls through to `noAlarmsModel`. The mock publishes top-level paths, so this will not reproduce under `--mock`.
- **`PageDeviceInfo.qml:18-32, 81` — `invalidate: false` dropped** on five hoisted `VeQuickItem`s. Name/Firmware/Hardware/Serial/Device-name rows now disappear when a device disconnects instead of holding their last values.
- **`PageAcCharger.qml:81` — `outputRepeater.count > 0` → `nrOfOutputs.valid`,** losing the deliberate `|| 1` fallback. A charger without `/NrOfOutputs` loses its whole battery voltage/current section.
- **`PageSettingsSystem.qml:29, 63`** — `_isCustomSystemName` returns false when `/SystemName` is invalid, but the radio group still falls back to the custom entry — so it reads "User defined" with no name field.

---

## Systemic consequences of lazy + destroyable delegates

These are consequences of the design, not slips, and they need a decision rather than a one-line fix.

- **Access-level filtering is gone at the model level.** ~35 `showAccessLevel` sites can't be hoisted, because the check lives on the delegate. Those rows now enter the model and are fully constructed for every user, then collapse to `visible:false, implicitHeight:0` via `ListItem.qml:44-55`. Not a visual leak, but it breaks `ListView.count`/index semantics, subscribes VeQuickItems to paths the user cannot read (e.g. `PageSettingsAccessAndSecurity.qml:298-300`), and pays exactly the construction cost this PR is removing.
- **`SettingsColumn` / `SettingsFlow` / `SettingsListHeader` never bind `visible` and never zero their height** — their comments say they rely on `VisibleItemModel` filtering. Any un-hoisted `preferredVisible` on those three is a *visible* regression, not a collapsed row. `PageVeBus.qml:117` is one such site (benign only because the Column is empty when the condition is false).
- **Delegates are now genuinely destroyed** (`delegatecomponentmodel.cpp:296-313` returns `Destroyed` + `deleteLater()`; the old model returned `{}`). With the default 320 px `cacheBuffer` that is ~4 rows past the viewport. Consequences found: `ListRadioButtonGroup.qml:69-93` declares its options sub-page `Component` **inside the delegate** — if the row's `preferredVisible` flips while that sub-page is open, Qt invalidates the child context recursively (`qqmlcontextdata.cpp:90-93`) and every binding on the open page silently stops. Same shape in `ListChargeSchedule.qml:82`, `PageGensetModel.qml:317`, `PageSettingsAccessAndSecurity.qml` dialogs, `PageBatterySettings.qml:95`, `PageSwitchableOutput.qml:264`. Also loses delegate-local state: `PageVeBusDebug.qml:124,157,168,179` **[×2]** (scrolling stops the setpoint timer and resets the sliders), and `PageSettingsModbusAddDevice.qml:69-120` (typed IP/port/unit reset on scroll).
- **`ObjectModelMonitor` defeats the optimisation entirely.** It is an `Instantiator`; with a `QQmlInstanceModel` it calls `object(i)` for **every** row (`qqmlinstantiator.cpp:360-414`). So `PageBatterySettings.qml:21-32` and `DcHistorySettingsColumn.qml:51-54` build everything eagerly anyway.

---

## `DelegateComponentModel` (C++)

Changeset bookkeeping is correct — all four mutation paths compute indices before mutating and pair exactly one changeset. `visibleInsertionIndex()` actually **fixes** a real bug in the old `VisibleItemModel`. Ownership is correct (`CppOwnership` set between `beginCreate`/`completeCreate`; no leak or double-delete found on normal paths). Issues:

- **`object()` ignores its `QQmlIncubator::IncubationMode` argument** (`:245-284`) and always does `beginCreate()`/`completeCreate()` synchronously. When a `ListView` fills its cache buffer it explicitly asks for `Asynchronous`; an expensive delegate is instead built on the GUI thread in that frame.
- **Reentrancy during construction** (`:277-291`): the bookkeeping record is written *after* `completeCreate()`, but `completeCreate()` runs arbitrary QML that can flip `preferredVisible`. `entryVisibilityChanged` then mutates `visibleEntries` while `d->objects` has no entry, and `object()` afterwards emits `initItem`/`createdItem` with the stale index captured at `:245`. There is no guard flag (`QQmlDelegateModel` uses one).
- **`entries_removeLast()` (`:217`) destroys the object without checking `refCount`,** unlike the guarded path at `:376-379`. `release()` (`:305-308`) and `entries_clear()` (`:184-191`) emit `destroyingItem` *inside* the `d->objects` iteration and then erase — a slot that re-enters and inserts invalidates the iterator.
- **`ObjectInfo::object` is a raw `QObject *`** (`:24`) where `VisibleItemModel` used `QPointer`. Nothing destroys a delegate externally today, but a `QPointer` here is free.
- **`variantValue()` returns empty for any non-instantiated row** (`:315-327`), which breaks `ListView.section.property`. Latent — none of the five `section.property` sites use this model.
- **The unit test is count-only.** `viewX.count` is `model->count()` (`qquickitemview.cpp:269`), so every view assertion re-asserts the model assertion above it. **The tests pass unchanged if `object()`, `release()`, `variantValue()`, `indexOf()` and `incubationStatus()` all return `{}`.** Nothing asserts a delegate is created, nothing exercises `release()`, `entries_clear()`, `entries_removeLast()` or `dataItem`. `test_nestedComponentUsesDcOwnedState` tests the *safe* pattern (Component on the `DelegateComponent`) while reading as coverage for the hazardous one (Component inside the delegate) that the port actually uses.

---

## Async page build

The interleaving analysis holds up: push-then-push, push-then-pop, double push, abandon-then-repush, synchronous completion and the error path are all handled correctly, and **no call site anywhere reads `pushPage()`'s return value** (I enumerated all 269 occurrences; `SettingsPage.qml:26` was correctly converted to the `readyCallback` form). But:

- **Navigation input is silently discarded for the whole build** (`PageStack.qml:124`, since `animating` now includes `!!_pendingBuild`). Previously the UI was frozen so the user couldn't observe input being dropped; now the UI keeps moving, so a second tap 200 ms later is thrown away with no busy affordance and no signal to the caller. This is the central UX consequence and deserves either a queue or a visible busy state.
- **`PageStack` has no `Component.onDestruction`** (`:166-188`). `Main.qml:41-64 rebuildUi()` destroys `PageStack` via `guiLoader.active = false` on connection loss / demo-mode change; if `finish()` then runs against a deleted `root`, it throws inside the incubator callback and the built page is never destroyed. `Component.onDestruction: if (_pendingBuild) _pendingBuild.forceCompletion()` fixes it.
- **The push *operation* is decided at completion time, not request time** (`:203-213`). `_animationDuration`/`_adjustedStackOperation` read `Global.allPagesLoaded` and `allowPageAnimations`, which change during the build — `MainView.qml:195` calls `goToStartPage()` while `allPagesLoaded` is false and sets it true at `:199`, so an `Immediate` start-page push can be silently upgraded to an animated one.
- **`topPageUrl` is stale for the whole build** (`:182-183`), which changes two `PageManager` outcomes: `PageManager.qml:134` misses its "already on the start page → just hide" branch, and `:100-103`'s idempotence guard fails.
- **Abandoned builds run to completion** (`:218-226`, acknowledged in the comment). `QQmlIncubator::clear()` isn't exposed to QML, so a full page tree is allocated purely to be destroyed at `:179`. Since `pushPage` already refuses a second build, `forceCompletion()` + destroy on abandon would bound the work.
- **`pages/PageManager.qml:113-117`** — the `config.stack` loop can now push only one page. Worth correcting the record: this is **not** a regression in the animated case (base's `fakePushTransition` already made `animating` true), and it is **currently unreachable** — `StartPageConfiguration.qml:88-131` only ever emits 0- or 1-element arrays. But the loop now reads as supporting N pages while being structurally incapable of it.
- **Two tests are vacuous.** `test_pushSupersededByAnotherPush` (`:240`) — the second push is dropped by the `animating` guard regardless, so the assertion is a tautology that passes with every line of abandon logic deleted. `test_uiKeepsRunningWhileAPageIsOpened` (`:67`) samples `_ticks` in a *later* step, after a 16 ms-polling `WaitUntil`, so it would pass against the old fully-synchronous implementation too. Also `pagestack.json` sets `"UIAnimations": 0`, so the whole `animating`-vs-`transitioning` split is never exercised.

The other three abandon tests are **not** vacuous — `test_pushSupersededByLeavingDoesNotOpen` and `test_leavingDuringAnOpenDoesNotBlockTheNextOne` each fail if their specific abandon path is removed.

---

## Connector / electron

`8b813def` is numerically equivalent — `spacing = 1.0/count` is bit-identical, positions/count/wrap unchanged, and no new cache was introduced (`m_lut` is base-branch code). No accumulated float drift: position, rotation and opacity are pure functions of `progress`. The `ListNavigationBase` extraction breaks nobody — all 5 `iconColor` sites are plain `ListNavigation` (which keeps it), and `SettingsListNavigation`'s `iconSource` already shadowed the inherited one.

The scene-graph rationale is **correct and verified** against Qt's source (`qsgnode.cpp:1301` `OPACITY_THRESHOLD`, `qsgbatchrenderer.cpp:82,470-476` `OPAQUE_LIMIT` → `FullRebuild`), and there's a bonus effect the comment doesn't claim: at opacity 1.0 Qt creates no `QSGOpacityNode` at all, so the old code added/removed a node every lap.

But **`7ca3e41a`'s "the fade is unchanged" is not accurate**, in three ways:

1. **Electrons no longer fade *in* when created.** Old: delegate had `opacity: 0.0` + `Behavior { OpacityAnimator }`, so a new electron ramped 0→1 over 100 ms. New: `widgetconnectorpathupdater.cpp:183` assigns `opacityAt(progress)`, which is `FadedInOpacity` immediately for anything past the first few percent. `electronRepeater.model` is reassigned whenever the count changes (`WidgetConnector.qml:70-71`), destroying and recreating every electron — so navigating to Overview, or resizing a widget across a 34 px multiple, now pops the electrons in at full brightness.
2. **On a GX the fade is ~2–3 steps instead of ~6.** `OverviewPage_Landscape.qml:601-602` advances `animationElapsed` at 20 Hz when `limitFps`; the old `OpacityAnimator` ran on the render thread at display rate. Timing is preserved exactly (I checked `fadeSpan` against the theme constants), smoothness is not. Defensible — position already steps at 20 Hz — but it should be stated.
3. **The endpoint arithmetic in `widgetconnectorpathupdater.h:45-48` is off by ~2×.** `0.002 × 255 = 0.51` (just *over* half a step) and `(1 − 0.996) × 255 = 1.02` (a full step, not half). Invisible in practice, but it's the stated justification for the constants.

Also: **`update()` now caches `count`** (`:152`, `:167-168`) where the old loop re-evaluated `electrons.count()` each iteration. `setPosition`/`setOpacity` can re-enter QML, so a mid-loop removal is now an out-of-range read. Nothing currently triggers it, but the old code was robust and this isn't — keeping `i < electrons.count()` as the condition costs nothing.

---

## The benchmark numbers need re-measuring before they're quoted

This matters because these figures are the PR's justification.

- **`c0fbad3a`'s "24.8ms → 18.2ms (−27%)" was measured on a harness this PR later proves invalid.** `9438979b`'s own message says those numbers "were mostly the cost of compiling a one-line string, not the cost of constructing the item" (119× overhead on a trivial `Item`), and re-measures the same type at **17.99 → 13.40**. `benchmarking.md:204` at head says `ListNavigation 12 ms`. So the PR contains 24.8 ms and 13.4 ms for the same item. The *delta* (~6.6 ms) survives because the string-compile overhead cancels, but −27% against a corrected baseline is ≈ **−33%**.
- **The "~46ms off Settings / ~33ms off System Setup" figures are arithmetic, not measurement** — 6.6 × 7 and 6.6 × 5 row counts. `pages/SettingsPage.qml` isn't in the benchmark's page list at all.
- **`benchmarking.md:201` and `PageStack.qml:96-97` ("median page over 300ms, worst over 700ms") are stale at head** — they were written before `b9c80eea`/`9592c785` ported 143 files to lazy construction. Head ships documentation asserting the problem this PR fixes, as a current measurement.
- **Per-page `cold` compile figures are order-dependent.** `tst_benchmark_pages.qml:70-87` runs the whole cold pass in one process with no `trimComponentCache()`, so the *first* page pays the compile of every shared type. Move `PageBattery` to the end of `pageUrls` and its "worst page, 309 ms" becomes ~30 ms with no code change.
- Each page is timed **once** per pass with ms-resolution `Date.now()`; only `COMPBENCH` amortises (N=100).
- `COMPBENCH` measures construction + binding evaluation only — instances are parented to a `UiTestCase` with no parent item (`uitest.cpp:198`), so nothing is laid out or rendered inside the timer.

Two things that are **fine** and worth recording: `PAGEBENCH-SKIP` does *not* inflate the total (`_measured` isn't incremented, and the count is printed) — and that trap was real, `e573a2b8` caught `PageSettingsDisplay.qml` being silently skipped for two commits. And pages with unsatisfied required properties are not a hole: construction runs in full and only then is the object rejected, so they are measured correctly. The measured page set is also *not* biased toward what this PR touched — `SettingsListNavigation` appears in only 1 of 19 pages.

`tests/ui/benchmark/overview/*` is base-branch, not this PR.

---

### Smaller items

`PageSettingsFirmware.qml:36`, `PageVeBus.qml:349,361`, `PageVeBusAdvanced.qml:337`, `PageVeBusError11Device.qml:66` **[×2]**, `PageAcInSetup.qml:132,169` **[×2]**, `PageAcInModel.qml:51,219,225` — `preferredVisible`/`showAccessLevel` left inside the delegate, so the row is always instantiated (hidden only by `ListItem`'s own zero-height behaviour). `GeneratorCondition.qml:48-77` hoists 8 `VeQuickItem`s out of a lazily-pushed sub-page onto an always-built row — ×6 conditions, that's ~48 subscriptions created on opening the list. `PageSettingsGeneral.qml:264-269` still overrides `ListNavigation`'s `contentItem` inline and would benefit from `ListNavigationBase`. `PageSettingsRvcDevice`/`PageDcMeter`/`PageInverter`/`DcBmsSettingsPage` carry dead duplicate `VeQuickItem`s. `PageMicrogrid.qml:219-256` has a `/* */` that straddles a `}` — it parses, but it's a trap. Missing trailing newlines in `PageSettingsBleSensors.qml`, `EvChargerPage.qml`, `GeneratorCondition.qml`, `PageDeviceInfo.qml`, `PageVeBusBackupRestore.qml`.

---

### What was not verified

No build and no run — every finding above is from reading source. Not settled statically: whether the un-hoisted `preferredVisible` sites on `SettingsColumn`/`SettingsFlow`/`SettingsListHeader` produce *visible* regressions (needs a capture A/B); whether reentrancy during `completeCreate()` actually fires under the mock/dbus backends; whether incubation progresses when the GX screen is blank (if not, `_pendingBuild` would wedge all navigation); and all on-device figures. The device A/B against `main` is running now and will follow.

One reported finding was **dropped as a false positive**: a duplicate `id:` at `PageSettingsGsm.qml:64-66` was claimed to break compilation. Line 63 is `id: modemConnected` on the `DelegateComponentModel` and line 66 is `id: statusDC` on the `DelegateComponent` — different objects, no duplicate.

DanielMcInnes:
## On-device benchmark: PR head vs `main` (Cerbo GX)

Follow-up to [the review above](https://github.com/victronenergy/gui-v2/pull/3150#issuecomment-5348851835). Every number below comes from a log collected in this session; nothing is estimated. Raw logs, per-run provenance and the analysis output are listed at the end.

**The result is good, and the control group is what makes it credible.**

| | BEFORE (`main` `97f659b5`) | AFTER (PR `9592c785`) | delta |
|---|---|---|---|
| **warm instantiate, PORTED pages (12)** | 4293 ms | 2658 ms | **−1635 ms, −38.1 %** |
| **warm instantiate, CONTROL pages (7)** | 1351 ms | 1322 ms | −29 ms, **−2.1 %** |
| warm instantiate, whole 19-page sample | 5714 ms | 3996 ms | −30.1 % |
| cold compile, whole sample | 2084 ms | 2104 ms | +1.0 % |
| **pages measured / skipped** | **19 / 0** | **19 / 0** | **no page stopped compiling** |

Median of 3 passes per revision, cold bytecode cache before every run. Every one of the 12 ported pages improved; all 7 controls sit within ±5 %, spread symmetrically around zero.

### This also settles the `c0fbad3a` number I flagged in the review

`COMPBENCH` measures `SettingsListNavigation` at **19.31 → 12.50 ms (−35.3 %)**, with non-overlapping run ranges, landing exactly on `ListNavigation`'s own cost — which is what you'd expect once the discarded contentItem is no longer built. So the optimisation is real and slightly *better* than the commit claims in relative terms, but the commit's "24.8 ms → 18.2 ms (−27 %)" baseline is wrong. Suggest amending it to the measured pair.

## Exact revisions and device

| | |
|---|---|
| **AFTER** | `9592c78501f971eacc7476f28d82be67aa611574` — *"Dev: port all instances of VisibleItemModel to DelegateComponentModel"* (PR 3150 head, branch `claude-optimized`) |
| **BEFORE** | `97f659b56184549179a77d905a52282ec5a30727` — *"Tests: only set test configuration when provided in CLI options"* (`origin/main`) |
| Repo | `/mnt/raid0/git/victron/gui-v2` |
| Device | `root@venus.local`, Cerbo GX, machine `einstein`, armv7l, 2 cores, Venus OS **v3.80~39**, kernel 6.12.90-venus-2, 79.1 MB free on `/` |
| Toolchain | Venus SDK `/opt/venus/current` (scarthgap-arm-cortexa8hf-neon), `./scripts/build-gx.sh` |
| Build times | BEFORE 161 s, AFTER 172 s |
| Binary md5 (BEFORE) | `733d3b0e11f435fa0cccf0b1d3e5a9bf`, 725 QML files deployed |
| Binary md5 (AFTER) | `d52987c4bf314a331599f25720494a58`, 727 QML files deployed |
| Submodules | `veutil` = `c69bfeb091b43ea47a100009697ddff78515f257` on **both** sides — no submodule difference confounding the A/B |

`md5sum /opt/victronenergy/gui-v2/venus-gui-v2` was read off the device and checked against the host
build **before every single pass**; it is recorded in each run's `.meta` file. Every deploy removed
`/opt/victronenergy/gui-v2` entirely before copying, so no stale file from the other revision could
survive (the two trees differ in file count).

### Note on what "BEFORE = main" includes

PR 3150 targets **`chriadam/overview-animation-improvements`** (tip `e25ed3ba`), not `main`.
`git merge-base origin/main 9592c785` is `97f659b5` = `origin/main` itself, so `main` is a clean
ancestor and the comparison is well defined, but `main..PR-head` contains three commits that belong
to the base branch rather than to this PR:

```
e25ed3ba Dev: add a script to help benchmarking perf of Overview page
eeb8ed7a Dev: ensure electrons are not children of Shape paths
36ca8cc9 Dev: use a lookup table for electron animations on Overview page
```

Those three touch only `components/widgets/WidgetConnector.qml`,
`pages/OverviewPage_Landscape.qml`, `src/widgetconnectorpathupdater.{cpp,h}`,
`.github/copilot-instructions.md` and the Overview benchmark harness — **none of the 19 benchmarked
pages, and none of the models they load**. So for this benchmark the main→head delta is attributable
to the PR's own commits.

### Provenance of the benchmark harness

`git log --oneline origin/main..pr3150-head -- tests/ui/benchmark`:

```
e573a2b8 Dev: measure page compilation with a cold bytecode cache, and explain cachegen
9438979b Dev: measure construction, not compilation, in the component benchmark
d673efc6 Dev: add a benchmark for page construction cost
e25ed3ba Dev: add a script to help benchmarking perf of Overview page
```

- `tests/ui/benchmark/overview` came from **`e25ed3ba`, i.e. the base branch**, not from this PR.
- `tests/ui/benchmark/pages` — the harness actually used here — came from **the PR** (`d673efc6`,
  later refined by `9438979b` and `e573a2b8`).

## The graft, and why it is valid

`tests/ui/benchmark/**` and `.github/benchmarking.md` do not exist on `main`; they are added by this
PR. To get a fair A/B, the harness was grafted onto a detached `main` worktree:

```bash
git worktree add -f --detach ../gxbench-main origin/main
cd ../gxbench-main
git checkout pr3150-head -- tests/ui/benchmark .github/benchmarking.md
# then, by hand, only the four benchmark entries added to cmake/ModuleUiTest_Sources.cmake
```

The whole diff of the grafted tree against `origin/main` is:

```
A  .github/benchmarking.md
 M cmake/ModuleUiTest_Sources.cmake        (+2 QML_MODULE_SOURCES, +2 QML_MODULE_RESOURCES)
A  tests/ui/benchmark/overview/overview.json
A  tests/ui/benchmark/overview/tst_benchmark_overview.qml
A  tests/ui/benchmark/pages/pages.json
A  tests/ui/benchmark/pages/tst_benchmark_pages.qml
```

**No production code was modified.** The PR's own `ModuleUiTest_Sources.cmake` was deliberately not
taken wholesale: it also registers `tests/ui/pagestack/tst_pagestack.qml` and
`tests/ui/pagestack/pagestack.json`, which do not exist on `main`. The grafted file differs from the
PR's version by exactly those two lines and nothing else.

### Validity checks performed

1. **No PR-only API is touched by the harness.** `tst_benchmark_pages.qml` was read line by line. It
   uses only `UiTestCase`, `UiTestStep.{Invoke,Wait,WaitUntil}`, `addStep`, `runSteps`, `findItem`,
   `Global.main`, `Global.mainView`, and `Qt.createComponent`/`createObject`. It **never touches
   `PageStack`**, never references `DelegateComponentModel`, and uses no new `PageStack` property or
   async signal. Notably `d673efc6` (which added it) predates `5e0ef773` ("build pages
   asynchronously"), and the test creates pages *detached* precisely so as not to involve the page
   stack.
2. **The C++ harness is byte-identical on both sides.** `git diff --stat origin/main pr3150-head --
   src/uitest.{h,cpp} src/uitestcase.{h,cpp} src/uiteststep.{h,cpp}` returns **empty**. The measuring
   instrument is literally the same code on both revisions.
3. **Every symbol the harness constructs exists on main.** All eleven `COMPBENCH` types
   (`Item`, `Label`, `VeQuickItem`, `ThreeLabelLayout`, `ListItem`, `ListSetting`, `ListNavigation`,
   `SettingsListNavigation`, `ListSwitch`, `ListQuantity`, `ListRadioButtonGroup`) and all nineteen
   page URLs resolve on `origin/main` — checked individually with `git cat-file -e`.
4. **The grafted tree builds.** `./scripts/build-gx.sh` → *Build successful*, 161 s, and the four
   harness files are present in the deployed tree at
   `build-gx_files_to_copy/Victron/UiTest/tests/ui/benchmark/…`.
5. **The grafted main produces numbers consistent with the PR author's own reference.**
   `.github/benchmarking.md` quotes, for a Cerbo, cold compile ≈ 2036 ms for the sample and warm
   instantiate ≈ 5553 ms. Grafted main measured 2084 ms and 5714 ms (medians of 3). Independent
   corroboration that the graft behaves like the harness's native environment.

**Conclusion: the graft is honest**, and no substitute baseline was needed. (For the record, the
fallback suggested to me — benchmarking against `e25ed3ba` because it "shares the harness" — would
not have worked: `e25ed3ba` carries only `tests/ui/benchmark/overview`. The *pages* harness arrives
two commits later at `d673efc6`, so `e25ed3ba` would have needed the same graft.)

## Method

- Both revisions built in their own detached worktrees (`../gxbench-main`, `../gxbench-pr`); the
  user's checkout was never touched.
- **Device backed up before the first deploy**: `ssh root@venus.local 'tar czf - -C
  /opt/victronenergy gui-v2' > /mnt/raid0/git/victron/device-gui-v2-backup.tar.gz` — 6 148 452 bytes,
  970 entries listed, 725 `.qml` files, `gui-v2/venus-gui-v2` present, verified with `tar tzf` before
  anything was deployed. Original binary md5 `891fb6989eaa98b6d07ca6baa0029103`.
- Deployed by hand (not `build-gx.sh -H`), with `svc -d /service/start-gui` first.
- **Three passes per revision** of `--ui-test benchmark/pages`.
- `/.cache/Venus/qmlcache` **and** `/data/home/root/.cache/Venus/qmlcache` deleted before **every**
  run (the second is the one this device actually uses — confirmed by watching `.qmlc` files appear
  there during a run).
- Unique log file per run (`/tmp/bench-<label>.log`), removed and confirmed absent before launch.
- Process tracked by **PID file**; `pkill -f` was never used. Liveness/cleanup used exact matching on
  `/proc/<pid>/comm`, because BusyBox `ps` has no `-eo` and BusyBox `pgrep -x venus-gui-v2` matches
  nothing.
- The app does **not** exit when the pages benchmark finishes; it was killed explicitly by PID after
  each log was collected.
- Guards: any pass finishing in under 180 s rejected; any pass not emitting 19 cold + 19 warm
  `PAGEBENCH` rows and 2 `PAGEBENCH-TOTAL` rows rejected. Actual elapsed times were 889/872/791 s
  (BEFORE) and 750/771/626 s (AFTER) — all far above the guard, and all six medians differ from each
  other, so none of the four fabrication traps fired.
- Device restored from the backup afterwards and `svc -u /service/start-gui` — see "Device restored".

### One pass was rejected, and it was re-run

`main-3` first attempt (335 s, log kept as `REJECTED-main-3-waituntil-timeout.log`) produced **zero**
`PAGEBENCH` rows. The log shows why:

```
venus.gui.test: ********* Start testing of tst_benchmark_pages *********
venus.gui.test: "initTestCase"
venus.gui.test: 	[2: WaitUntil]
venus.gui.test: FAIL!	: Step timed out!
venus.gui.test: Totals: 0 steps passed, 1 steps failed, 11234 ms
venus.gui.test: All tests finished: 0 steps passed, 1 steps failed, in 0m 11s
```

The harness's opening `WaitUntil` (waiting for a "Settings" nav-bar item) timed out after ~11 s. This
is a **real flakiness in the benchmark harness on device**, worth reporting on its own: the run still
prints `All tests finished` and exits 0, so a poller that only looks for that marker would record it
as a successful pass with an empty page list. The re-run passed cleanly (125 steps passed, 0 failed).
Both logs are kept.

## Page lists compared first

`PAGEBENCH-TOTAL` reports 19 pages on **both** revisions, in **all six** runs, and there is not a
single `PAGEBENCH-SKIP` line anywhere. The cold and warm page-URL sets are identical between BEFORE
and AFTER. **No page that compiled on main fails to compile on the PR** — no functional regression of
that kind to report, and the totals above are comparing like with like.

## Classification

A page is **PORTED** if the page itself, or the model it loads, was changed between `origin/main` and
the PR head; **CONTROL** otherwise. `PageAcIn.qml` is unchanged but loads `PageAcInModel.qml`, which
the PR does change, so it is counted as ported.

- **PORTED (12):** PageBattery, PageAcIn (via PageAcInModel), PageVeBus, SolarDevicePage,
  EvChargerPage, PageSettingsGeneral, PageSettingsConnectivity, PageSettingsSystem, PageSettingsDvcc,
  PageSettingsHub4, PageBatterySettings, PageInverter.
- **CONTROL (7):** SolarInputListPage, AcLoadListPage, DcLoadListPage, BatteryListPage,
  InverterChargerListPage, DeviceListPage, PageControllableLoads.

Control-group cleanliness was checked, not assumed: none of the seven control pages references
`VisibleItemModel` or `SettingsListNavigation` (the two things the PR changes), so the change cannot
reach them through a shared type. Three of them use plain `ListNavigation`, whose *contentItem* the
PR leaves identical — it only re-parents the type onto the new `ListNavigationBase`. `COMPBENCH`
measures `ListNavigation` directly and puts that re-parenting at −4.1 %, i.e. within noise.

## Per-page results — warm pass, instantiate (ms)

Median of 3; the three individual runs are shown so the spread is visible.

| page | grp | before (runs) | after (runs) | delta | % |
|---|---|---|---|---|---|
| `/pages/settings/devicelist/battery/PageBattery.qml` | PORTED | **604** (638, 604, 600) | **195** (198, 193, 195) | −409 | **−67.7 %** |
| `/pages/settings/PageSettingsHub4.qml` | PORTED | **331** (331, 334, 314) | **153** (150, 155, 153) | −178 | **−53.8 %** |
| `/pages/settings/devicelist/battery/PageBatterySettings.qml` | PORTED | **691** (717, 691, 670) | **356** (356, 353, 361) | −335 | **−48.5 %** |
| `/pages/settings/PageSettingsSystem.qml` | PORTED | 168 (173, 168, 160) | 95 (94, 95, 99) | −73 | −43.5 % |
| `/pages/settings/devicelist/inverter/PageInverter.qml` | PORTED | 380 (377, 380, 386) | 242 (242, 243, 234) | −138 | −36.3 % |
| `/pages/vebusdevice/PageVeBus.qml` | PORTED | 384 (400, 383, 384) | 246 (233, 264, 246) | −138 | −35.9 % |
| `/pages/settings/PageSettingsGeneral.qml` | PORTED | 179 (176, 179, 179) | 127 (126, 127, 128) | −52 | −29.1 % |
| `/pages/settings/PageSettingsDvcc.qml` | PORTED | 243 (250, 243, 235) | 176 (194, 176, 172) | −67 | −27.6 % |
| `/pages/solar/SolarDevicePage.qml` | PORTED | 150 (150, 165, 145) | 116 (113, 116, 120) | −34 | −22.7 % |
| `/pages/settings/PageSettingsConnectivity.qml` | PORTED | 128 (128, 128, 124) | 101 (102, 91, 101) | −27 | −21.1 % |
| `/pages/evcs/EvChargerPage.qml` | PORTED | 465 (465, 478, 458) | 370 (370, 374, 369) | −95 | −20.4 % |
| `/pages/settings/devicelist/ac-in/PageAcIn.qml` | PORTED | **570** (580, 570, 545) | **481** (494, 481, 478) | −89 | −15.6 % |
| `/pages/loads/AcLoadListPage.qml` | CONTROL | 9 (9, 9, 9) | 8 (8, 8, 9) | −1 | −11.1 % |
| `/pages/battery/BatteryListPage.qml` | CONTROL | 235 (235, 244, 224) | 223 (219, 228, 223) | −12 | −5.1 % |
| `/pages/settings/PageControllableLoads.qml` | CONTROL | 366 (366, 396, 363) | 349 (334, 349, 358) | −17 | −4.6 % |
| `/pages/solar/SolarInputListPage.qml` | CONTROL | 337 (343, 337, 331) | 335 (333, 335, 336) | −2 | −0.6 % |
| `/pages/loads/DcLoadListPage.qml` | CONTROL | 9 (9, 9, 8) | 9 (8, 9, 11) | 0 | 0.0 % |
| `/pages/settings/devicelist/DeviceListPage.qml` | CONTROL | 329 (352, 329, 321) | 331 (331, 330, 343) | +2 | +0.6 % |
| `/pages/invertercharger/InverterChargerListPage.qml` | CONTROL | 66 (66, 67, 65) | 67 (66, 69, 67) | +1 | +1.5 % |

**Group summary (warm instantiate):**

| group | n | before | after | delta |
|---|---|---|---|---|
| PORTED | 12 | 4293 ms | 2658 ms | **−1635 ms (−38.1 %)** |
| CONTROL | 7 | 1351 ms | 1322 ms | −29 ms (−2.1 %) |

Every one of the twelve ported pages improved; every one of the seven control pages sits within
±5.1 %, spread symmetrically around zero (three down, one flat, two up, one −11 % on a 9 ms page
where 1 ms is quantisation). That is what a genuine effect against a genuine noise floor looks like.

The worst page in the sample drops from 691 ms (PageBatterySettings) to 481 ms (PageAcIn).

## Per-page results — cold pass, compile (ms)

| group | n | before | after | delta |
|---|---|---|---|---|
| PORTED | 12 | 1589 ms | 1664 ms | +75 ms (+4.7 %) |
| CONTROL | 7 | 419 ms | 395 ms | −24 ms (−5.7 %) |
| whole sample (`PAGEBENCH-TOTAL cold`) | 19 | 2084 ms | 2104 ms | +20 ms (+1.0 %) |

**QML compilation is not improved by this PR, and should not be expected to be** — the change is
about what a document instantiates, not about the document text. The ±5 % group movements are in
opposite directions and are inside the compile column's own run-to-run noise (the BEFORE totals alone
span 1988–2150 ms, 7.8 %); `.github/benchmarking.md` warns explicitly that a single page's compile
figure is indicative only. I would not claim either group number as a real effect.

## Totals, and run-to-run spread

| metric | BEFORE (3 runs) | AFTER (3 runs) | delta |
|---|---|---|---|
| cold compile, whole sample | **2084** (1988, 2084, 2150 — spread 7.8 %) | **2104** (2086, 2104, 2106 — spread 1.0 %) | +20 ms, +1.0 % |
| cold instantiate, whole sample | **6213** (6041, 6213, 6372 — spread 5.3 %) | **4399** (4372, 4399, 4496 — spread 2.8 %) | −1814 ms, −29.2 % |
| warm instantiate, whole sample | **5714** (5521, 5714, 5765 — spread 4.3 %) | **3996** (3971, 3996, 4003 — spread 0.8 %) | −1718 ms, −30.1 % |

Spreads of 0.8–7.8 % across six independent runs, with no two runs identical on any metric — the
"identical medians" signature of a stale-log trap is absent.

## COMPBENCH — where the win actually comes from

Average ms to construct one instance (100 instances per measurement), median of 3 runs:

| type | before | after | delta | % |
|---|---|---|---|---|
| **SettingsListNavigation** | **19.31** (19.85, 19.31, 18.92) | **12.50** (12.50, 12.99, 12.46) | **−6.81** | **−35.3 %** |
| ListNavigation | 12.51 (12.51, 12.49, 12.62) | 12.00 (11.85, 12.00, 12.32) | −0.51 | −4.1 % |
| ListQuantity | 11.25 (11.96, 11.25, 11.02) | 10.82 (10.82, 10.83, 10.63) | −0.43 | −3.8 % |
| ListRadioButtonGroup | 13.56 (14.04, 13.56, 13.51) | 13.37 (13.03, 13.37, 14.09) | −0.19 | −1.4 % |
| ListSwitch | 12.73 (13.57, 12.73, 12.71) | 12.80 (12.78, 12.97, 12.80) | +0.07 | +0.5 % |
| ListSetting | 3.80 (4.10, 3.74, 3.80) | 3.75 (3.75, 3.71, 3.77) | −0.05 | −1.3 % |
| ThreeLabelLayout | 4.59 (4.64, 4.59, 4.54) | 4.50 (4.50, 4.42, 4.56) | −0.09 | −2.0 % |
| ListItem | 2.34 (2.45, 2.34, 2.28) | 2.31 (2.31, 2.42, 2.28) | −0.03 | −1.3 % |
| Label | 1.25 (1.25, 1.28, 1.23) | 1.26 (1.26, 1.37, 1.21) | +0.01 | +0.8 % |
| VeQuickItem | 0.25 (0.17, 0.25, 0.33) | 0.15 (0.14, 0.17, 0.15) | −0.10 | −40.0 % † |
| Item | 0.07 (0.07, 0.07, 0.06) | 0.07 (0.07, 0.06, 0.07) | 0.00 | 0.0 % |

† `VeQuickItem` costs 0.15–0.25 ms; a "−40 %" on a quarter of a millisecond is timer quantisation,
not a result.

**Reading:** nine of the eleven item types are flat (within ±4 %, i.e. inside the noise). One type,
`SettingsListNavigation`, is genuinely 6.8 ms cheaper per instance — that is commit `c0fbad3a`
("stop SettingsListNavigation building a contentItem it discards"), and the effect is unambiguous:
19.31 → 12.50 ms with non-overlapping run ranges, landing exactly on `ListNavigation`'s own cost, as
you would expect once the discarded contentItem is no longer built.

**So which is it — fewer items, or cheaper items? Overwhelmingly fewer.** Of the twelve ported pages,
`SettingsListNavigation` is declared on exactly **one**: `PageSettingsSystem`, five instances. At
6.81 ms each that accounts for **at most ~34 ms of that page's 73 ms** improvement. The three largest
absolute wins — PageBattery (−409 ms), PageBatterySettings (−335 ms), PageSettingsHub4 (−178 ms) —
declare **zero** `SettingsListNavigation`, so 100 % of their improvement comes from constructing
fewer items. Across the ported group, the cheaper-item effect can explain at most ~34 ms of the
−1635 ms; the remaining ~98 % is `DelegateComponentModel` building only what the view needs instead
of every declared row.

This is consistent with the mechanism in `src/delegatecomponentmodel.h`: `VisibleItemModel`
eagerly constructs every child at declaration time, `DelegateComponentModel` holds lightweight
`DelegateComponent` entries and lets the `ListView` instantiate delegates on demand.

### Why this measurement is representative rather than an artefact of detached construction

The benchmark creates each page with `component.createObject(null, {})`, i.e. detached. A fair
objection is that a detached page might have a zero-sized `ListView` that builds no delegates at all,
which would flatter the lazy model. It does not, for two reasons, both checked:

1. **The viewport is real.** `components/Page.qml` sets `implicitWidth: Theme.geometry_screen_width`
   / `implicitHeight: Theme.geometry_screen_height`, and `components/BaseListView.qml` sets
   `width: parent?.width ?? 0` / `height: parent?.height ?? 0`. A detached page therefore takes its
   implicit screen size and its list view fills it — the same viewport the page gets on screen.
2. **The measurements themselves prove delegates are being built.** Control pages that are plain
   `ListView`s over C++ models cost real time detached — SolarInputListPage 337 ms,
   PageControllableLoads 366 ms, DeviceListPage 329 ms — whereas AcLoadListPage, whose `model` is a
   `required property` that is undefined when constructed detached, costs 9 ms. Delegate construction
   is plainly inside the measured window.

## Device restored

```
md5sum /opt/victronenergy/gui-v2/venus-gui-v2
891fb6989eaa98b6d07ca6baa0029103        # identical to the pre-benchmark original
qml files: 725                          # identical to the backup listing
/service/start-gui: up (pid 14295) 317 seconds
venus-gui-v2 running, VmRSS 89044 kB
/dev/root  449.7M  342.1M  79.1M  81% /  # identical to the starting state
```

The install was removed and re-extracted from `device-gui-v2-backup.tar.gz`, the QML bytecode caches
cleared so the restored build recompiles cleanly, and the service restarted. It has been up on a
single stable PID (no crash-loop) with normal memory use, and free space is back to the exact
starting figure.

## What I could NOT verify

These are limits of what was measured. None of them is patched over with an estimate anywhere above.

1. **Nothing here measures the asynchronous `PageStack` work.** Commits `5e0ef773`, `1fe483f2`,
   `fe21e950`, `c1d5da74`, `14d495aa` change how `PageStack.pushPage()` builds pages. The benchmark
   deliberately creates pages *detached* and never goes through `PageStack`, so the "opening a page
   no longer blocks the UI" claim is **completely untested by these numbers**. My results speak only
   to how much work page construction is, not to when or on which thread it happens.
2. **Nothing here measures the Overview animation work.** `--ui-test benchmark/overview` and
   `scripts/benchmark-overview.py` were not run. Commits `7ca3e41a`, `8b813def` (and the base
   branch's `eeb8ed7a`, `36ca8cc9`) are unmeasured by me.
3. **The lazily-deferred work was not measured, only its removal from page-open.**
   `DelegateComponentModel` moves construction of off-screen rows from page-open to scroll time. This
   benchmark sees the saving at open and is structurally incapable of seeing the cost that reappears
   when the user scrolls. I did not measure scroll cost or jank at all. The improvement is real for
   time-to-first-paint of a page; it is not proof of a net reduction in total work.
4. **The "cold compile" column is not a true empty-cache endpoint.** Both bytecode caches were
   deleted before every run as `.github/benchmarking.md` requires, but the app compiles roughly 300
   QML documents during its own startup before the benchmark begins (observed: the qmlcache directory
   grew from 0 to ~296 entries during startup). Types shared between startup and the benchmarked
   pages are therefore already cached when their page is measured. This applies equally to both
   revisions, so the A/B is fair, but the absolute compile figures sit between the two endpoints
   `benchmarking.md` describes rather than on the cold one.
5. **`SettingsPage.qml` — the heaviest `SettingsListNavigation` user — is not in the benchmark's page
   list.** The 6.8 ms/instance saving will help it, but I have no measurement of that page and make
   no claim about it.
6. **No functional or visual testing was done.** The only correctness evidence I have is that all 19
   pages still compile and instantiate on the PR (zero `PAGEBENCH-SKIP`, 125 harness steps passed /
   0 failed in each accepted run). I did not run `ctest`, a capture sweep, or any interaction test,
   so I cannot speak to whether the ported pages *look* or *behave* the same.
7. **One unexplained startup stall.** During the first BEFORE pass the app spent ~7 minutes before
   the UI test started, with the main thread pegged inside `VeQItemTableModel::onValueChanged()` →
   `QCborValue::toVariant()` (confirmed by attaching `gdbserver` and taking a backtrace). It resolved
   on its own and did not recur in the other five passes. It happened entirely *before* any
   `PAGEBENCH` line, so it cannot have affected the timings, but I could not determine its cause and
   am reporting it rather than omitting it.
8. **Single device, single Venus OS version.** All six passes ran on one Cerbo GX on Venus OS
   v3.80~39. `benchmarking.md`'s reference numbers were taken on v3.80~57. Nothing here says how this
   generalises to Ekrano, GX Touch, or other hardware.
9. **`main` was benchmarked with a graft, not as it ships.** Justified in detail above and I believe
   it sound, but it remains true that the BEFORE binary is `origin/main` *plus four test-only files
   and four lines of CMake*, not a pristine `main` build.

## Raw logs

All under `/mnt/raid0/git/victron/bench-logs/`:

| file | what it is |
|---|---|
| `main-1.log`, `main-2.log`, `main-3.log` | the three accepted BEFORE passes (~717 KB each) |
| `pr-1.log`, `pr-2.log`, `pr-3.log` | the three accepted AFTER passes (~713 KB each) |
| `main-1.meta` … `pr-3.meta` | per-run provenance: label, device binary md5, elapsed seconds, start time, PID, device log path |
| `REJECTED-main-3-waituntil-timeout.log` / `.meta` | the rejected pass (harness `WaitUntil` timeout, 0 pages) — kept deliberately |
| `ANALYSIS.txt` | full machine-generated output of the analysis script, including every page in both passes |
| `build-main.log`, `build-pr.log` | the two `build-gx.sh` build logs |

Device backup: `/mnt/raid0/git/victron/device-gui-v2-backup.tar.gz` (6 148 452 bytes, 970 entries).
Worktrees: `/mnt/raid0/git/victron/gxbench-main` (grafted `origin/main`),
`/mnt/raid0/git/victron/gxbench-pr` (PR head).

Per-run provenance, verbatim:

```
main-1 md5=733d3b0e11f435fa0cccf0b1d3e5a9bf elapsed_s=889 started=2026-08-20T08:28:03+10:00 pid=24952
main-2 md5=733d3b0e11f435fa0cccf0b1d3e5a9bf elapsed_s=872 started=2026-08-20T08:43:22+10:00 pid=29044
main-3 md5=733d3b0e11f435fa0cccf0b1d3e5a9bf elapsed_s=791 started=2026-08-20T09:04:27+10:00 pid=2295
pr-1   md5=d52987c4bf314a331599f25720494a58 elapsed_s=750 started=2026-08-20T09:18:48+10:00 pid=5688
pr-2   md5=d52987c4bf314a331599f25720494a58 elapsed_s=771 started=2026-08-20T09:31:31+10:00 pid=8803
pr-3   md5=d52987c4bf314a331599f25720494a58 elapsed_s=626 started=2026-08-20T09:44:35+10:00 pid=11969
REJECTED main-3 (1st attempt) elapsed_s=335 — 0 pages, harness WaitUntil timed out
```

DanielMcInnes:
### Addendum: a guaranteed trigger for the delegate-destruction hazard

The last review pass turned up a concrete instance of the "`Component` declared inside a delegate" problem I described [above](https://github.com/victronenergy/gui-v2/pull/3150#issuecomment-5348851835) as a general hazard. This one is worth calling out separately because the trigger isn't incidental — **it is the normal way the control is used.**

`pages/vebusdevice/PageVeBusBackupRestore.qml:360-377` (and identically `:399-416` for Delete):

```qml
DelegateComponent {
    id: _restoreOptionsListDC
    preferredVisible: _backupRestoreAction.value != VenusOS.VeBusDevice_Backup_Restore_Action_Restore
            && !root.fileToRestore          // <-- goes false as soon as a file is chosen
    ListRadioButtonGroup {
        id: _restoreOptionsList
        optionModel: _availableBackupsModel
        ...
        onOptionClicked: root.fileToRestore = _availableBackupsModel.get(index).value
    }
}
```

`ListRadioButtonGroup.qml:69-93` declares its options sub-page as a `Component` **inside the delegate**, and pushes it at `:61`. So the sequence is:

1. User taps the row → options sub-page is pushed, created from a `Component` owned by this delegate.
2. User taps an option → `onOptionClicked` sets `root.fileToRestore`.
3. `preferredVisible` at `:363` immediately goes false → the row leaves the model → `DelegateComponentModel::release()` destroys the delegate (`delegatecomponentmodel.cpp:296-313`).
4. Qt invalidates the delegate's `QQmlContextData` and, recursively, every child context (`qqmlcontextdata.cpp:90-93`) — including the still-open options page's.

The sub-page is destroyed out from under itself *during its own close path*, with invalid source bindings and a dead creation context. Every backup Restore and every Delete goes through exactly this path, so unlike the other sites I listed this isn't a rare-timing bug — it's the happy path.

The general fix for this class is to declare the sub-page `Component` on the `DelegateComponent` (which is never destroyed) rather than inside the delegate — the pattern `tests/delegatecomponentmodel/tst_delegatecomponentmodel.qml:144` already demonstrates. Since `ListRadioButtonGroup` is a shared type, fixing it there would cover every user at once.

chriadam:
Replaced by the following PRs:
https://github.com/victronenergy/gui-v2/pull/3121
https://github.com/victronenergy/gui-v2/pull/3240
https://github.com/victronenergy/gui-v2/pull/3241
https://github.com/victronenergy/gui-v2/pull/3242
https://github.com/victronenergy/gui-v2/pull/3243

PR #3152, opened 2026-08-13
Settings: clean up use of SettingsListNavigation

blammit:
SettingsListNavigation is only used to provide a left-hand icon, so it doesn't need to be used in settings sub-pages. It should only be used in the main settings page, where all list items have icons. Note this means the "System setup" page will now have list items with slightly larger heights (as was defined in SettingsListNavigation).

Also rename iconSource to pageIconSource to avoid clashing with the existing property in ListNavigation.


blammit:
@chriadam @DanielMcInnes this also avoids the issue we discussed where the ListNavigation contentItem was created unnecessarily, since SettingsListNavigation no longer redefines the contentItem.

pages/settings/PageSettingsIntegrations.qml
```diff
@@ -319,8 +319,14 @@ Page {
 							? qsTrId("pagesettingsintegrations_uiplugin_integrates_with_devicelist")
 							: ""
 						indicatorColor: switchNavigationItem.color
-						pageSource: switchNavigationItem.pluginSettingsPageIntegration?.url ?? ""
 						interactive: switchNavigationItem.pluginSettingsPageIntegration !== null
+
+						onClicked: {
+							const url = switchNavigationItem.pluginSettingsPageIntegration?.url ?? ""
+							if (url) {
+								Global.pageManager.pushPage(url, { title: text })
```
chriadam, line 327: maybe: `title: Qt.binding(function() { return switchNavigationItem.text })`

pages/settings/PageSettingsSystem.qml
```diff
@@ -53,56 +53,56 @@ Page {
...
+			ListNavigation {
 				topInset: Theme.geometry_listItem_itemSeparator_height
 				//% "Opportunity Loads"
 				text: qsTrId("pagesettingssystem_opportunity_loads")
 				//% "Automate controllable devices to maximize solar self-consumption"
 				caption: qsTrId("pagesettingssystem_automate_controllable_devices")
 				secondaryText: opportunityLoadsMode.value ? CommonWords.enabled : CommonWords.disabled
-				pageSource: "/pages/settings/PageControllableLoads.qml"
 				preferredVisible: opportunityLoadsMode.valid
+				onClicked: Global.pageManager.pushPage("/pages/settings/PageControllableLoads.qml", { title: text })
```
chriadam, line 64: title isn't bound here, either

blammit: This one is okay, as the title will not change

pages/settings/PageSettingsIntegrations.qml
blammit: But does `switchNavigationItem.text` ever change? Can a plugin name change at runtime?

chriadam: Yes, due to system language change, I guess

chriadam: But maybe it's not worth worrying about?

chriadam (approved):
Nitpicks, but LGTM.

pages/settings/PageSettingsIntegrations.qml
blammit: Ah, yes I forgot... well, at the moment, we have so many instances of `{ title: text }` without bindings in the code base. Let's fix them altogether in a separate PR.

PR #3171, opened 2026-08-21
Wasm: normalize generic deleteContent input events to backspace

### LINKED ISSUE #3153: Garmin MFD Backspace is not understood by Qt 6.8.3 WASM
Complaint from a customer with extensive report and proposed fix.

----

_During physical validation of a custom application integrated with a Victron GX device through Garmin OneHelm, I have identified several reproducible input problems affecting **Victron GUI-v2 Remote Console on Garmin**.
I investigated these problems because Remote Console is accessible from our application, but the testing ultimately demonstrated that the issue is **not caused by our application or iframe integration**._

_The same behavior can be reproduced when GUI-v2 is launched directly and when Remote Console is launched from the stock Victron MFD  application._

**_Test environment_**
- Cerbo GX
- Venus OS v3.75
- Victron GUI-v2 / WebAssembly Remote Console
- GUI-v2 build based on Qt 6.8.3 WebAssembly
- Garmin OneHelm embedded browser
- Garmin browser user agent observed:
   - Mozilla/5.0 (X11; Linux aarch64)
   - AppleWebKit/605.1.15
   - (KHTML, like Gecko)
   - Version/60.5 Safari/605.1.15

**_Control tests_**
_We tested the input path at several levels._

Test | Result
-- | --
Plain HTML page in Garmin OneHelm | PASS
Plain HTML <input> + Garmin keyboard | PASS
Plain HTML Backspace | PASS
Plain HTML Tab | PASS
Plain HTML scrolling | PASS
Stock Victron /app normal tapping/navigation | PASS
Stock Victron /app continuous PWM slider drag | PASS
Direct GUI-v2 on Garmin | FAIL
Custom application iframe → GUI-v2 | FAIL
Stock Victron /app → Remote Console → GUI-v2 | FAIL

_This localizes the problem to the **Garmin WebKit ↔ Qt/WASM GUI-v2 input layer**, rather than Garmin touch support in general or our application._

_Once keyboard focus was corrected, Backspace remained non-functional.
We instrumented the actual hidden Qt <input>._

**_Garmin event sequence_**
_Normal characters are delivered through the IME/composition path and generate:_
```
keydown
key="Unidentified"
code="Unidentified"
keyCode=229
input
inputType="insertFromComposition"
```

_Backspace generates:_
beforeinput: `inputType="deleteContent"`
input: `inputType="deleteContent"`

_The hidden browser input itself correctly changes:_
_ABC → AB → A → ""
However, the QML text field remains unchanged.
The keyboard keydown cannot identify Backspace because Garmin reports:_
```
key="Unidentified"
code="Unidentified"
keyCode=229
```
_which is also used during ordinary IME input._

**_Qt behavior_**
_Qt 6.8.3's WASM QWasmInputContext explicitly handles:_
```
deleteContentBackward
deleteContentForward
```
_and synthesizes `Qt::Key_Backspace` / `Qt::Key_Delete`.
It does not handle the generic Garmin value `deleteContent`_
_Qt's implementation therefore never forwards the Garmin deletion to the QML field._

**_Confirmed workaround_**
_Normalizing the actual Garmin event before Qt receives it fixes Backspace:_
```js
document.addEventListener("input", function (event) {
const target = event.target;

if (
!target ||
target.tagName !== "INPUT" ||
typeof target["data-qinputcontext"] === "undefined" ||
event.inputType !== "deleteContent"
) {
return;
}

Object.defineProperty(event, "inputType", {
configurable: true,
value: "deleteContentBackward"
});
}, true);
```

**_Result_**
_With this normalization:
ABC → AB → A → blank
also occurs correctly in the actual GUI-v2/QML field.
This appears suitable either for a Garmin-specific compatibility normalization in GUI-v2 or, preferably, for consideration in Qt's WASM input context._

_**Relevance to Victron**_
_The defect is reproducible without our application:_
_Garmin OneHelm
↓
stock Victron /app
↓
Victron Remote Console
↓
GUI-v2 / Qt WASM
↓
same ~keyboard and~ scrolling defects_
_Therefore we believe this is a GUI-v2 / Qt WASM compatibility issue with Garmin's embedded WebKit, rather than an application integration problem.
We can provide the diagnostic HTML pages, physical test results, event traces and the two working JavaScript shims if useful._

_I have, however not yet tested what these changes and workaround produce as results on MFD's of other brands like Raymarine or Navico._

mman:
Qt's WASM input context only synthesizes `Qt::Key_Backspace`/`Qt::Key_Delete` for the directional inputTypes `deleteContentBackward`/`deleteContentForward` and silently drops the direction-agnostic `deleteContent` (Input Events Level 2). Virtual keyboards that drive the browser editor through generic editing commands rather than key events produce exactly that: WebKit maps `EditAction::Delete` and `EditAction::TypingDeleteSelection` to `"deleteContent"` (`WebCore/editing/EditCommand.cpp`), and the accompanying keydown is `key="Unidentified"`/`keyCode=229`, so Qt gets no usable signal from the keyboard event either. Result: the hidden IME `<input>` mutates but backspace never reaches the QML field. Seen in the field on Garmin OneHelm's embedded WebKit (#3153).

This adds a capture-phase document listener that rewrites `inputType` to `deleteContentBackward` on input events targeting Qt's hidden IME input (identified by the `data-qinputcontext` property Qt sets on it), before Qt's own listener runs. Semantically safe: `deleteContent` means "delete selection, direction unspecified" and backspace deletes the selection if present, else the previous character — identical in every reachable case. No mainstream browser emits `deleteContent` for a plain backspace (Blink does not implement that inputType at all), so this is inert outside the affected devices.

Based on the workaround suggested by the reporter in #3153.

### Verification

Reproduced and verified headlessly against the current wasm build (Playwright driving the app to the root password field, injecting `input` events on Qt's hidden IME input — emulating what Garmin's WebKit delivers natively):

| Injected event | without this PR | with this PR |
|---|---|---|
| `inputType: "deleteContent"` | ignored (the #3153 bug) | deletes one character |
| `inputType: "deleteContentBackward"` (control) | deletes one character | deletes one character |

Synthetic events exercise the same listener path as Garmin's trusted native events, but this has not been validated on real OneHelm hardware — confirmation by the reporter on-device is needed before merging.

Unfixed upstream as of Qt 6.8/6.9 LTS and dev; reported as [QTBUG-149393](https://bugreports.qt.io/browse/QTBUG-149393), after which this workaround can be dropped once gui-v2 builds against a fixed Qt. Note: VRM ships its own `index.html` and needs the same shim.

Fixes #3153

### DIFF
--- a/wasm/index.html
+++ b/wasm/index.html
@@ -951,5 +951,42 @@
             };
         })();
     </script>
+    <script type="text/javascript">
+        // ---- Virtual keyboard backspace normalization --------------------
+        //
+        // Qt's WASM input context only synthesizes backspace/delete for the
+        // directional inputTypes "deleteContentBackward"/"...Forward" and
+        // drops other deletion inputTypes. Virtual keyboards that drive the
+        // editor via generic editing commands (Garmin OneHelm's embedded
+        // WebKit, https://github.com/victronenergy/gui-v2/issues/3153)
+        // produce the direction-agnostic "deleteContent" instead, with
+        // keydown reported as key="Unidentified"/keyCode=229 — so backspace
+        // never reaches the QML text field.
+        //
+        // Rewrite "deleteContent" to "deleteContentBackward" on Qt's hidden
+        // IME <input> (identified by the "data-qinputcontext" JS property Qt
+        // sets on it) before Qt sees the event: a capture-phase listener on
+        // document runs first, and the own property shadows the read-only
+        // prototype getter. Semantically safe: both mean "delete selection,
+        // else the character before the cursor". Reported upstream as
+        // QTBUG-149393; needed until gui-v2 builds against a fixed Qt.
+        (function() {
+            'use strict';
+
+            document.addEventListener('input', function(event) {
+                const target = event.target;
+                if (!target
+                        || target.tagName !== 'INPUT'
+                        || typeof target['data-qinputcontext'] === 'undefined'
+                        || event.inputType !== 'deleteContent') {
+                    return;
+                }
+                Object.defineProperty(event, 'inputType', {
+                    configurable: true,
+                    value: 'deleteContentBackward'
+                });
+            }, true);
+        })();
+    </script>
   </body>
 </html>


blammit (approved):

blammit:
@mman there's a conflict due to another PR I just merged - could you update this one against main?

mman:
@blammit rebased...

PR #3173, opened 2026-08-23
Mock coverage and UI test robustness

DanielMcInnes:


tests/ui/README.md
```diff
@@ -21,6 +21,50 @@ venus-gui-v2 --mock --mock-conf barebones --ui-test smoke/generic-capture
...
+
+For unattended runs, where nobody can promise not to touch the window, use the offscreen
+platform. There is no window at all, so there is nothing to hide, and it still renders real
+screenshots via software rasterisation, so CaptureAndCompare works normally:
+
+```
+QT_QPA_PLATFORM=offscreen ./bin/venus-gui-v2 --mock --ui-test smoke/mock-maximal
+```
+
+XWayland (`QT_QPA_PLATFORM=xcb`) also avoids the frame-callback stall, and runs
```
blammit: @DanielMcInnes  I'm curious - do you normally use the `xcb` platform or `QSG_RENDER_LOOP=basic` when running the UI tests? Is that mainly for benchmarking?

tests/ui/RecursivePageCapture.qml
```diff
@@ -280,6 +289,19 @@ QtObject {
...
+		Returns a description of the current page, including its url.
+
+		The url is included so that a test run can be checked for the set of pages that it reached;
+		this shows which parts of the UI are covered by the backend data (e.g. by a specific mock
+		configuration) and which are not.
+	*/
+	function _pageDescription() {
+		const title = Global.mainView.currentPage?.title ?? ""
+		const url = Global.pageManager.pageStack.topPageUrl
+		return url ? "%1 [page url: %2]".arg(title).arg(url) : title
```
blammit: Perhaps "url" is sufficient, instead of "page url"?

tests/ui/README.md
DanielMcInnes: @blammit this is all from claude, I have only run the UI tests a few times, and have never specified which QT_QPA_PLATFORM to use

blammit: Okay, let's remove it then, if it's not explicitly useful

data/mock/TemperaturesImpl.qml
```diff
@@ -33,27 +33,62 @@ Item {
...
-			// If the auto-selected temperature service is not set, and the settings indicate the system
-			// should select one by default, then set it to this service.
-			const canAutoSelect = root.settingsValue("/Settings/SystemSetup/TemperatureService") === "default"
-			if (canAutoSelect && !autoSelectedTemperatureService.valid) {
-				console.warn("Mock: auto-set temperature service to", temperatureService.serviceIdWithPath(), temperatureService.name)
-				autoSelectedTemperatureService.setValue(temperatureService.name)
-			}
-			availableTemperatureServices.addService(temperatureService)
+			onValidChanged: updateListing()
+			onNameChanged: updateListing()
```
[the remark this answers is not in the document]
DanielMcInnes: done

data/mock/conf/split-phase.json
```diff
@@ -1,8 +1,10 @@
 {
     "files": [
+        ":/data/mock/conf/setup-common.json",
+        ":/data/mock/conf/services/generator.json",
```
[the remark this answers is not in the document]
DanielMcInnes: done

tests/ui/README.md
DanielMcInnes: mostly done, what's left is just "don't hide the window" and "use QT_QPA_PLATFORM=offscreen for unattended runs"

tests/ui/RecursivePageCapture.qml
DanielMcInnes: updated as suggested

data/mock/conf/services/meteo-imt.json
```diff
@@ -0,0 +1,34 @@
...
+        "/Mgmt/ProcessName": "dbus-imt-si-rs485tc",
+        "/Mgmt/ProcessVersion": "1.0.6",
+        "/ProductId": 45104,
+        "/ProductName": "IMT Si-RS485 Solar Irradiance Sensor",
+        "/Serial": "IMT-1234567",
+        "/Status": 0,
+        "/SwitchableOutput": null,
+        "/TodaysYield": null,
+        "/WindDirection": null,
+        "/WindSpeed": 3.2
```
blammit, line 28: I can see this file contains a few extra values compared to meteo-solarsense.json, but do you think we need the new meteo-imt.json file? It doesn't provide a separate product with special values, unlike fiamm-48tl.json, so I'm wondering if the extra values here can just be folded into meteo-solarsense.json.

data/mock/conf/services/em-heatpump-output.json
```diff
@@ -17,7 +17,14 @@
...
         "/Ac/Voltage": 151.70000000000002,
-        "/AllowedRoles": ["grid", "pvinverter", "genset", "acload", "evcharger", "heatpump"],
+        "/AllowedRoles": [
+            "grid",
+            "pvinverter",
+            "genset",
+            "acload",
+            "evcharger",
+            "heatpump"
+        ],
```
blammit: Is this change needed? Can it stay on one line to make it easier when grepping for uses of `AllowedRoles`?

data/mock/conf/services/canbus-canopen-motordrive.json
```diff
@@ -0,0 +1,5 @@
+{
+    "com.victronenergy.settings": {
+        "/Settings/Canbus/vecan0/Profile": 9
```
blammit: We already have other JSON configuration files that add a `vecan0` profile - is it useful to add this file?

data/mock/conf/services/canbus-rvc.json
```diff
@@ -0,0 +1,41 @@
+{
+    "com.victronenergy.settings": {
+        "/Settings/Canbus/vecan0/Profile": 6,
```
blammit, line 3: If we're going to commonly override these `/Settings/Canbus/vecan<X>/Profile` values in the JSON files, let's remove it from `setup-common.json` to avoid confusion.

data/mock/conf/services/evcharger1.json
```diff
@@ -22,15 +22,15 @@
         "/ProductId": 49189,
         "/ProductName": "EV Charging Station",
         "/Role": "evcharger",
+        "/S2/0/Active": 1,
```
blammit: What's the reason for changing Active=1 ? If both evcharger1.json and em-heatpump-output.json have Active=1, then there are no services remaining with Active=0 in the maximal conf, and then we won't see the "No control" text in the loads drilldown list.

data/mock/conf/services/pylontech.json
```diff
@@ -30,39 +30,39 @@
         "/Dc/1/Voltage": null,
         "/DeviceInstance": 512,
         "/Diagnostics/LastErrors/1/Error": null,
-        "/Diagnostics/Module0/Alarms/CellImbalance": null,
```
blammit, line 33: Can you move the new module-level alarm values to a different battery service? It's useful for the pylontech one to have all `null` values for the module-level alarms, because we want to verify that the "Module level alarms" list item is not shown if all values are null or if the `Id` is an empty string.

Also when you add the module-level alarms elsewhere, can you use a non-zero value for some of them so we can verify the "warning" (1) and "alarm" (2) levels as well?

data/mock/conf/multi-rs.json
```diff
@@ -1,12 +1,14 @@
 {
     "files": [
+        ":/data/mock/conf/setup-common.json",
         ":/data/mock/conf/services/dcgenset.json",
         ":/data/mock/conf/services/evcharger1.json",
+        ":/data/mock/conf/services/pump.json",
```
blammit: Nice to see `relays-tank-pump.json` added - could you delete `pump.json` and move its contents into `relays-tank-pump.json`? It doesn't really make sense to have a standalone pump service that's not attached to any relay configuration.

Also then `pump.json` can be removed from the `maximal.json` file.

data/mock/conf/split-phase.json
```diff
@@ -1,8 +1,10 @@
...
         ":/data/mock/conf/services/pylontech.json",
-        ":/data/mock/conf/services/multiplus-split-phase.json",
-        ":/data/mock/conf/setup-common.json"
+        ":/data/mock/conf/services/relays-generator.json",
```
blammit, line 5: Similarly to the comment for `pump.json` and `relays-tank-pump.json`, could you move the contents of `generator.json` into `relays-generator.json` and remove `generator.json`?

This should be safe since `generator.json` isn't used elsewhere at the moment, and it doesn't make sense to use it without having a relay set to "generator" function anyway.

data/mock/TemperaturesImpl.qml
```diff
@@ -33,27 +33,66 @@ Item {
 			// uid includes path, e.g. "mock/com.victronenergy.vebus/257/Dc/0/Temperature"
 			required property string uid
```
blammit, line 34: Could you explain what these changes are doing? Is it fixing a bug in the previous implementation?

I noticed the list of temperature sensors under `Settings > System setup > Charge control > Temperature sensor` is now a lot shorter - is that expected?

src/uitest.h
```diff
@@ -27,11 +27,13 @@ class UiTestConfiguration
 	bool isValid() const { return !dirName().isEmpty(); }
 	QString dirName() const;
 	const QVariantMap &settingsMap() const;
-	bool hasMockConfiguration() const;
+	// Returns true if this test needs the mock backend, i.e. it has a "Mock" section at all. A
+	// test may set only e.g. "TimersActive" there and leave the choice of mock configuration to
+	// the caller, so this is a weaker condition than hasMockConfiguration().
+	bool requiresMockBackend() const;
```
blammit: I think it's going to be confusing to have both `requiresMockBackend()` and `hasMockConfiguration()`. 

Thinking about it now, we can remove the `TimersActive` configuration, and mock timers can always be set to inactive during the UI tests. Otherwise, the screen capture comparisons will fail due to the effect of the mock data randomisers.

And then, just set "MockConfiguration" instead of "Mock" in the ui test JSON file. For example, in mock-maximal.json, change:

```
    "Mock": {
        "Configuration": ":/data/mock/conf/maximal.json",
        "TimersActive": false
    },
```

to:

```
    "MockConfiguration": ":/data/mock/conf/maximal.json",
```

What do you think?

src/veqitemmockproducer.cpp
```diff
@@ -32,6 +42,13 @@ VeQItemMockProducer::VeQItemMockProducer(VeQItem *root, const QString &id, QObje
 void VeQItemMockProducer::setValue(const QString &uid, const QVariant &value)
 {
 	if (VeQItemMock *item = qobject_cast<VeQItemMock*>(mProducerRoot->itemGetOrCreate(normalizedUid(uid), true, true))) {
+		// An undefined value means that the mocked device does not publish this path, so remember
+		// it as a deliberate omission rather than as a gap in the mock configuration.
```
blammit: It can also change to `undefined` if set to that value during runtime. Values can be changed by application, after the initial load by a mock configuration.

src/veqitemmockproducer.h
```diff
@@ -26,8 +27,25 @@ class VeQItemMock : public VeQItem
...
 	int setValue(QVariant const &value) override;
 
+	// True if this item is a path in its own right, i.e. something asked the backend for exactly
+	// this uid, rather than for one below it.
+	//
+	// isLeaf() cannot answer this, because adding any child clears it: if the UI asks for "/A" and
+	// then for "/A/B", "/A" stops being a leaf even though it is still a requested path with no
+	// value.
+	//
+	// This is recorded when the item is added to the tree, which is the only point we are told
```
blammit: This last paragraph appears to be implementation detail that isn't needed here

tests/mockcoverage/tst_mockcoverage.qml
```diff
@@ -0,0 +1,178 @@
...
+		VeQuickItem {}
+	}
+
+	// The uids that the UI has asked for but that are not provided, within one service.
+	function missingUnder(serviceUid) {
+		return MockManager.missingValues().filter(uid => uid.startsWith(serviceUid + "/"))
+	}
+
+	// Asks the backend for a uid, the way a binding in the UI would.
+	function request(uid) {
```
blammit: A `create` or `new` prefix for the function would be useful, to indicate it returns a new object that must be destroyed.

tests/ui/smoke/generic-capture/tst_all.qml
```diff
@@ -63,6 +63,30 @@ UiTestCase {
 		runSteps(recursivePageCapture.start)
 	}
 
+	function cleanup() {
```
blammit: Is this necessary? Shouldn't `RecursivePageCapture` have popped all pages on the stack already? If not, then that would be a bug.

tests/ui/smoke/mock-maximal/tst_overview.qml
```diff
@@ -19,11 +19,27 @@ UiTestCase {
 	}
 
 	function cleanup() {
-		Global.pageManager.popAllPages()
-
-		// No steps have been added, so call goToNextTestFunction() instead of runSteps() to
-		// continue the testing.
-		goToNextTestFunction()
+		// Pop any pages that the test function left open, and wait for them to be popped, before
```
blammit: Is the reason for this change, that the stack is not cleared immediately, without animation? If so, you can just change the `Global.pageManager.popAllPages()` line to pass the `PageStack.Immediate` parameter.

tests/ui/smoke/mock-maximal/tst_overview.qml
```diff
@@ -86,8 +102,21 @@ UiTestCase {
 	}
 
 	function test_drilldown(data) {
-		addStep(UiTestStep.Invoke, { callable: ()=> { return mouseClick(findClickableChild(findItem(Global.mainView.currentPage, data.widgetValues))) } })
-		addStep(UiTestStep.WaitUntil, { callable: ()=> { return !Global.mainView.animating } })
+		let depthBeforeClick = -1
+		addStep(UiTestStep.Invoke, { callable: ()=> {
+			depthBeforeClick = Global.pageManager.pageStack.depth
```
blammit: This assignment could be done on initialisation, on line 105

tests/ui/README.md
```diff
@@ -21,6 +21,53 @@ venus-gui-v2 --mock --mock-conf barebones --ui-test smoke/generic-capture
 ```
 
 
+## Do not hide the window while a test run is in progress
+
+On Wayland, a minimised (or otherwise unmapped) window receives no frame callbacks, and Qt waits
```
blammit: Note this is true on other platforms as well, not just for Wayland.

tests/ui/RecursivePageCapture.qml
```diff
@@ -41,11 +41,20 @@ QtObject {
 	*/
 	function start(doneCallback) {
 		if (root._busy) {
-			console.assert(false, "Cannot start while another recursive capture is in progress!")
-			return
+			// The previous capture did not finish, most likely because one of its steps failed.
+			// Warn about it, but continue, so that a single failure does not stop every following
+			// recursive capture in this test case.
+			console.warn("Previous recursive capture did not finish; restarting")
```
blammit: Should it really continue testing if a previous capture is still running? That would mean that the pages on the stack have not been popped back up to the top, which could produce errors in the next test.

tests/ui/RecursivePageCapture.qml
```diff
@@ -280,6 +297,19 @@ QtObject {
 		return searchParams
 	}
 
+	/*
+		Returns a description of the current page, including its url.
+
+		The url is included so that a test run can be checked for the set of pages that it reached;
+		this shows which parts of the UI are covered by the backend data (e.g. by a specific mock
```
blammit: Including the url seems useful in general, and not just for the coverage feature, so not sure this second paragraph is needed

data/mock/conf/README.md
```diff
@@ -11,6 +11,21 @@ To use mock data in the browser, append `?mock` or `?mock=<config>` to the `inde
 
 If --mock is set but no configuration is specified, the 'maximal' configuration is used.
 
+To find out which values the UI asked for but the loaded configuration does not provide, add the
```
blammit, line 21: Please move the coverage feature under a separate sub-heading, as it's not essential for using mock mode.

data/mock/conf/services/cgwacs-energy-meters.json
```diff
@@ -0,0 +1,25 @@
...
+        "/Settings/CGwacs/DeviceIds": "B1234567,C7654321",
+
+        "/Settings/Devices/cgwacs_B1234567/ClassAndVrmInstance": "grid:31",
+        "/Settings/Devices/cgwacs_B1234567/CustomName": "",
+        "/Settings/Devices/cgwacs_B1234567/IsMultiphase": 0,
+        "/Settings/Devices/cgwacs_B1234567/L2/ServiceType": "",
+        "/Settings/Devices/cgwacs_B1234567/Position": 0,
+        "/Settings/Devices/cgwacs_B1234567/ServiceType": "grid",
+        "/Settings/Devices/cgwacs_B1234567/SupportMultiphase": 1,
+        "/Settings/Devices/cgwacs_B1234567_S/Enabled": 0,
```
blammit: Could set Enabled=1 to show the extra "PV inverter on phase 2 Position" switch by default, in PageSettingsCGwacs.qml.

src/main.cpp
```diff
@@ -373,17 +375,27 @@ void initBackend(bool *enableFpsCounter, bool *skipSplashScreen)
 			qFatal() << "Error: --mock-conf was set but --ui-test" << parser.value(uiTest)
 					 << "already specifies a mock configuration:" << uiTestConf.dirName();
 		}
+		if (mockConfName.isEmpty() && !uiTestConf.hasMockConfiguration()) {
+			// Use "maximal" as the default mock configuration, if none is set.
+			mockConfName = "maximal";
+		}
```
blammit, line 377: Is this change here just refactoring, or was it fixing a bug?

Also looks like configurationFileName() is no longer used then, and can be removed.

src/veqitemmockproducer.h
```diff
@@ -43,6 +61,13 @@ class VeQItemMockProducer : public VeQItemProducer
 	void removeServices(const QString &serviceType);
 	void dumpValues();
 
+	// Returns the uids of all leaf items that were requested by the UI but for which the mock
+	// backend does not provide a value. Such items yield 'undefined' in QML.
+	//
+	// Items that the mock configuration deliberately set to an undefined value are not included;
```
blammit: or also, if they changed to undefined at runtime (e.g. if some action resulted in a value being cleared)

tests/ui/README.md
```diff
@@ -124,6 +171,31 @@ The `gui-v2/src` directory contains the C++ API:
 * `uiteststep.h`: defines the test steps that can be executed in a test case
 
 
+## Checking which pages a test run reached
+
+Each `RecursivePageCapture` capture step logs the url of the page it captured, e.g.
```
blammit, line 201: Note that recording the urls alone are not enough to whether all pages in the UI have been reached, because pages are also commonly pushed as Component objects rather than as standalone .qml files. The url comparisons are a great start, but we should be aware that this is not exhaustive, and it should not be documented as such.

src/main.cpp
```diff
@@ -203,6 +203,14 @@ void initBackend(bool *enableFpsCounter, bool *skipSplashScreen)
 	parser.addOption(noMockTimers);
 	optionList << noMockTimers;
 
+	QCommandLineOption mockCoverage("mock-coverage",
```
blammit, line 205: This option seems useful for checking specific services and values, though in general it's going to capture many values that are deliberately not set in the mock backend (e.g. if a value does not apply to a particular service, or does not need to be set because a similar value is already provided by another service).

blammit (commented):
Really nice, great to see more complete coverage of features in mock mode. I'm not sure  about the usefulness of the coverage feature in its current form, but let's discuss that further.

data/mock/conf/services/meteo-imt.json
DanielMcInnes: Yes, I think we do need it. It is a separate product, and without it we wouldn't be able to see PageMeteoSettings.qml in mock mode. The names are similar (meteo-imt & meteo-solarsense) but they have different manufacturers and different features.

data/mock/conf/services/em-heatpump-output.json
DanielMcInnes: moved this back to a single line

data/mock/conf/services/canbus-canopen-motordrive.json
DanielMcInnes: updated as suggested, moved into dc-only.json

data/mock/conf/services/canbus-rvc.json
DanielMcInnes: done, I think?

data/mock/conf/services/evcharger1.json
DanielMcInnes: whoops, this was a mistake. Fixed

data/mock/TemperaturesImpl.qml
DanielMcInnes: Yes, it is fixing a bug. To reproduce on main:

1. ./build/bin/venus-gui-v2 --mock
2. Settings → System setup → Charge control → Temperature sensor, and select Pylontech Force L2 (pick it, so its name is the one displayed on the collapsed row).
3. Settings → Device list → Pylontech battery → Device info → Name → type RENAMED (PageDeviceInfo.qml:36-45 writes <service>/CustomName).
4. The Device list immediately shows RENAMED. Go back to Charge control → the Temperature sensor row still reads Pylontech Force L2.

I can't reproduce the issue where you see a shorter list of temperature sensors. Can you tell me how to reproduce it?

data/mock/conf/services/pylontech.json
DanielMcInnes: done

data/mock/conf/multi-rs.json
DanielMcInnes: done

data/mock/conf/split-phase.json
DanielMcInnes: done

src/uitest.h
DanielMcInnes: done

src/veqitemmockproducer.cpp
DanielMcInnes: updated the comment

src/veqitemmockproducer.h
DanielMcInnes: removed spam

tests/mockcoverage/tst_mockcoverage.qml
DanielMcInnes: changed name to 'createVeQuickItem()'

tests/ui/smoke/generic-capture/tst_all.qml
DanielMcInnes: this has gone away

tests/ui/smoke/mock-maximal/tst_overview.qml
DanielMcInnes: yes, updated as suggested

tests/ui/smoke/mock-maximal/tst_overview.qml
DanielMcInnes: done

tests/ui/README.md
DanielMcInnes: fixed

tests/ui/RecursivePageCapture.qml
DanielMcInnes: The consequence you describe is real, but _busy is not the lever. A failed step means UiTestCase::stepGroupFinished() skips the runSteps() callback, so the recursion is abandoned with its pages still pushed — and the next test function fails even if it never touches RecursivePageCapture: forcing a failure at depth 2, test_control_cards could not click the status bar. start() cannot pop for itself either, since it has no way to tell leftovers from a page the new test legitimately pushed. So the reset now happens in UiTestCase::goToNextTestFunction(), the one place every test function is reached from, which also stops state leaking between test files. With that, continuing is safe, and the comment says why.

tests/ui/RecursivePageCapture.qml
DanielMcInnes: removed 2nd paragraph

data/mock/conf/README.md
DanielMcInnes: done

data/mock/conf/services/cgwacs-energy-meters.json
DanielMcInnes: done

src/main.cpp
DanielMcInnes: this is a bug fix. The mock backend is now set up before the UI test configuration is loaded, because the old order let a mock configuration overwrite what the test had just applied — TimersActive, and /Settings/Gui2/UIAnimations. The default-configuration guard had to change with it: it used to ask MockManager whether a configuration had been loaded yet, which only answered the real question ("does the test name one?") while the test config was loaded first. After the reorder, that proxy always says "no" and would load maximal on top of the test's configuration, silently merging the two. It now asks uiTestConf directly, which is also what made configurationFileName() dead code.

reomoved `configurationFileName()`

src/veqitemmockproducer.h
DanielMcInnes: updated comment

tests/ui/README.md
DanielMcInnes: updated

src/main.cpp
DanielMcInnes: yes, that's right. Keep it?

src/veqitemmockproducer.cpp
```diff
@@ -78,6 +102,18 @@ void VeQItemMockProducer::dumpValues()
 	});
 }
 
+QStringList VeQItemMockProducer::unsetValueUids() const
+{
+	QStringList uids;
+	mProducerRoot->forAllChildren([this, &uids](VeQItem *item) {
+		if (item->isLeaf() && !item->getValue().isValid() && !m_undefinedUids.contains(item->uniqueId())) {
```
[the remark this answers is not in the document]
DanielMcInnes: added tests/mockcoverage/tst_mockcoverage.qml

tests/ui/smoke/generic-capture/tst_all.qml
```diff
@@ -63,6 +63,20 @@ UiTestCase {
...
+		// Pop any pages that the test function left open, and wait for them to be popped, before
+		// starting the next test function. Otherwise the next test function may click while the
+		// page stack is still animating, and the click will have no effect.
+		addStep(UiTestStep.WaitUntil, { callable: ()=> {
+			if (Global.pageManager.pageStack.opened && !Global.mainView.animating) {
+				// Retry the pop on each check; a pop is ignored while the stack is animating.
+				Global.pageManager.popAllPages()
+			}
+			return !Global.mainView.animating && !Global.pageManager.pageStack.opened
```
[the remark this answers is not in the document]
DanielMcInnes: won't fix in this PR, this needs to be fixed in PageStack.qml

tests/ui/smoke/mock-maximal/tst_overview.qml
```diff
@@ -19,11 +19,17 @@ UiTestCase {
...
-		goToNextTestFunction()
+		// Pop any pages that the test function left open, and wait for them to be popped, before
+		// starting the next test function. Otherwise the next test function may click while the
+		// page stack is still animating, and the click will have no effect.
+		addStep(UiTestStep.WaitUntil, { callable: ()=> {
+			if (Global.pageManager.pageStack.opened && !Global.mainView.animating) {
+				// Retry the pop on each check; a pop is ignored while the stack is animating.
+				Global.pageManager.popAllPages()
+			}
+			return !Global.mainView.animating && !Global.pageManager.pageStack.opened
```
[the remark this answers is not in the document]
DanielMcInnes: won't fix in this PR, this needs to be fixed in PageStack.qml

src/veqitemmockproducer.h
```diff
@@ -43,6 +61,14 @@ class VeQItemMockProducer : public VeQItemProducer
 	void removeServices(const QString &serviceType);
 	void dumpValues();
 
+	// Returns the uids of all leaf items that were requested by the UI but for which the mock
+	// backend does not provide a value. Such items yield 'undefined' in QML.
+	//
+	// If an item is deliberately set to undefined in the mock configuration, or it changes to
+	// undefined at runtime, it will not be included in this list;
+	// see setValue().
```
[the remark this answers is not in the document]
DanielMcInnes: fixed

tests/ui/smoke/mock-maximal/tst_overview.qml
blammit: The `cleanup()` function has been removed altogether - was that deliberate?

data/mock/conf/services/meteo-imt.json
blammit: Okay, sounds good

data/mock/TemperaturesImpl.qml
blammit: I see, that makes sense 👍  I don't see the issue with the temperature sensor list anymore, so all good.

tests/ui/smoke/mock-maximal/tst_overview.qml
blammit: `depthBeforeClick` can be `const`.

tests/ui/smoke/mock-maximal/tst_overview.qml
blammit: Oh I see, UiTestCase::resetPageStack() takes care of this now.

src/uitestcase.cpp
```diff
@@ -242,6 +249,56 @@ void UiTestCase::goToNextTestFunction()
...
+
+	qCInfo(venusGuiTest) << qPrintable(QStringLiteral("Popping %1 page(s) left open by the previous test function").arg(depth));
+
+	// Pop without a transition: the pages must be gone before the next test function clicks
+	// anything, and a click is ignored while the page stack animates. Read the depth back in the
+	// same expression, because a pop is not guaranteed to happen: PageStack.popAllPages() returns
+	// early while the stack animates, and a page may refuse to be popped from its tryPop().
+	QQmlExpression popExpression(context, this, QStringLiteral(
+			"(Global.pageManager.popAllPages(PageStack.Immediate),"
+			" Global.pageManager.pageStack.depth)"));
```
blammit: This function is convenient, but I'm not sure it's a good idea to add it. This means the C++ side has a dependency on the QML PageManager API, which is a bit odd. I think it's better for the test to clean up after itself by calling `Global.pageManager.popAllPages(PageStack.Immediate)` from `cleanup()`.

tests/ui/RecursivePageCapture.qml
blammit: I don't think we want to reach into `Global.pageMananager.popAllPages()` from a C++ UiTestCase::goToNextTestFunction() function. 

What if we just call `Global.pageManager.popAllPages(PageStack.Immediate)` from here?

src/main.cpp
blammit: Yep okay, let's keep it 👍

data/mock/conf/setup-common.json
```diff
@@ -4,8 +4,15 @@
...
+        "/CanBus/Interfaces": [
+            { "config": 0, "interface": "vecan0", "name": "VE.Can port 1" },
+            { "config": 3, "interface": "vecan1", "name": "VE.Can port 2" }
+        ],
         "/Device/Model": "Cerbo GX",
-        "/Device/Time": 1751358535
+        "/Device/Time": 1751358535,
+        "/ModificationChecks/FsModifiedState": 0,
+        "/ModificationChecks/SystemHooksState": 0,
+        "/Network/HasBluetoothSupport": 0
```
blammit: When running with mock-maximal, I'm seeing this:

```
Warning: changing value of "com.victronenergy.platform/Network/HasBluetoothSupport" from QVariant(double, 0) to QVariant(qlonglong, 1)
Warning: changing value of "com.victronenergy.settings/Settings/Services/Modbus" from QVariant(double, 0) to QVariant(qlonglong, 1)
```

Can you remove those two values from `setup-common.json` to avoid overwriting them? I don't think these need to be set for all configurations if we're testing them in the maximal conf specifically.

tests/ui/smoke/mock-maximal/tst_overview.qml
DanielMcInnes: done

tests/ui/RecursivePageCapture.qml
DanielMcInnes: We can't call `popAllPages` because this can happen:

The sequence, from tst_overview.qml:

1. test_drilldown("Grid") → click pushes the Grid page → start() → capture aborts on a failed step → _busy stays true (it's only cleared in _captureNext()'s isInitialPage branch).
2. cleanup() runs — it does run, being a test function in its own right — and pops the Grid page. Stack empty.
3. test_drilldown("Generator") → click pushes PageGenset.qml, depth 1.
4. runSteps(recursivePageCapture.start) → start() sees _busy === true and a page it must not touch.

The reason the two can't be told apart is that by the time start() runs, cleanup() has already removed the leftovers — so a non-empty stack at start() is always the incoming test's own page, never debris. _busy says "the previous capture didn't finish"; it says nothing about what's on the stack now.

src/uitestcase.cpp
DanielMcInnes: updated as suggested

data/mock/conf/setup-common.json
DanielMcInnes: done

tests/ui/RecursivePageCapture.qml
```diff
@@ -41,11 +49,25 @@ QtObject {
...
+			console.warn("Previous recursive capture did not finish; restarting at page stack depth",
+					Global.pageManager.pageStack.depth)
 		}
 		root._busy = true
 		root.doneCallback = doneCallback
+
+		// Clear the state from any previous run, so that this run neither keeps the pages and views
+		// of the last one alive nor consults their entries.
+		root.pageCaptureCounts = new Map()
+		root.lastClickedViewItems = new Map()
```
blammit: Could assign as `{}` instead of `new Map()` to align with similar code elsewhere in the codebase, unless there's a benefit to using `new Map()` instead.

tests/ui/RecursivePageCapture.qml
blammit: Right, I see... ok, these are the options I see:

a) Add a reset() function to RecursivePageCapture, which simply clears the _busy flag. This can be called from the test cleanup() functions, so that the _busy=true case is never encountered by start(). If we do this, then I think the _busy check in start() should just assert(false) instead of printing a warning, because it means we forgot to call reset().
b) Just remove the _busy flag. It's just checking the sanity of the current state before the captures begin, and other properties are cleared/reset regardless before the captures start.

Perhaps a) is better for the sake of completeness. What do you think?

tests/ui/smoke/mock-maximal/tst_overview.qml
```diff
@@ -19,7 +19,20 @@ UiTestCase {
...
 	function cleanup() {
-		Global.pageManager.popAllPages()
+		// Pop any pages that this test function left open, so that the next one starts from the
+		// same state. A test function does not always get to finish tidily: when a step fails the
+		// rest of its steps are abandoned, and any pages it had pushed by then stay on the stack,
+		// where the next test function would click into them and fail for reasons of its own.
+		//
+		// Pop without a transition, because a click is ignored while the page stack animates.
+		Global.pageManager.popAllPages(PageStack.Immediate)
+		if (Global.pageManager.pageStack.depth > 0) {
```
blammit, line 29: Let's `console.assert()` on whether the depth is zero, instead of printing a warning, as this would mean something is buggy in PageStack.

tests/ui/README.md
```diff
@@ -21,6 +21,53 @@ venus-gui-v2 --mock --mock-conf barebones --ui-test smoke/generic-capture
 ```
 
 
+## Do not hide the window while a test run is in progress
```
blammit, line 24: I noticed all of these lines wrap at 100 chars, which doesn't align with the existing text in the file, and same with the changes to data/mock/conf/README.md.

tests/ui/smoke/generic-capture/tst_all.qml
```diff
@@ -63,6 +63,27 @@ UiTestCase {
...
 
+	function cleanup() {
+		// Pop any pages that this test function left open, so that the next one starts from the
+		// same state. A test function does not always get to finish tidily: when a step fails the
+		// rest of its steps are abandoned, and any pages it had pushed by then stay on the stack,
+		// where the next test function would click into them and fail for reasons of its own.
+		//
+		// Pop without a transition, because a click is ignored while the page stack animates.
+		Global.pageManager.popAllPages(PageStack.Immediate)
+		if (Global.pageManager.pageStack.depth > 0) {
```
blammit, line 74: Similarly as per my comment in tst_overview.qml, we should console.assert() on the stack depth instead of printing a warning.

PR #3175, opened 2026-08-24
Implement a camper view

chriadam:


src/uitest.cpp
```diff
@@ -24,6 +28,361 @@
 
 using namespace Victron::VenusOS;
 
+namespace {
```
blammit: Given this new code is a few hundred lines long, would be worthwhile to move the feature into a separate file.

chriadam: Extracted all the ui testing stuff into a separate PR: https://github.com/victronenergy/gui-v2/pull/3184

chriadam:
This is all still WIP, don't bother reviewing yet :-)

PR #3180, opened 2026-08-25
Mock: do not override the mock timer setting requested by a UI test

DanielMcInnes:
UiTest::loadConfiguration() applies the "Mock": {"TimersActive": false} of a test configuration, and the mock backend setup immediately below it then called setTimersActive() again with the command-line default, which is true unless --no-mock-timers was given. A UI test therefore always ran with mock timers on, however its configuration was written.

The visible effect is that image captures are not reproducible: MockDataRandomizer keeps shifting mock values by +/-10% for the whole sweep, so screens showing live values differ between two runs of the very same build. Measured over smoke/mock-maximal, two sweeps of one build differed on 182 of 1171 screens (worst MSE 256) before this change and on 3 of 1171 (worst MSE 17.8) after it, with 1153 screens pixel-identical.

Only apply the command-line default when the test configuration did not ask for a particular setting. An explicit --no-mock-timers still wins.

The 3 screens that still differ are not a rendering problem: the "Start page" setting resolves to a different value between runs.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -359,8 +359,12 @@ void initBackend(bool *enableFpsCounter, bool *skipSplashScreen)
 		Victron::VenusOS::MockManager *mockManager = Victron::VenusOS::MockManager::create();
 		if (parser.isSet(noMockTimers)) {
 			mockTimersEnabled = false;
+			mockManager->setTimersActive(mockTimersEnabled);
+		} else if (!uiTestConf.hasMockTimersActive()) {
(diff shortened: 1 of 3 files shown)


blammit:
The second and last paragraphs in the commit msg seem irrelevant to this patch as a standalone commit.

DanielMcInnes:
> The second and last paragraphs in the commit msg seem irrelevant to this patch as a standalone commit.

Updated commit msg, this was a cherry - pick from a bigger branch

blammit (approved):

PR #3183, opened 2026-08-27
Settings: show write permission error notification for list items

blammit:
When a button or switch list item is disabled due to the write permissions, instead of just appearing disabled, also show a notification when the user tries to activate it via a click.

To do this, provide a background with a mouse area, which will show a toast notification when clicked if the user is lacking sufficient write permissions. If the user clicks the button/switch in the contentItem but it is disabled (due to clickable=false for the write permissions) then the click falls through to the background.

### DIFF
--- a/components/listitems/ListSpinBoxRange.qml
+++ b/components/listitems/ListSpinBoxRange.qml
@@ -137,6 +137,19 @@ ListSetting {
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


blammit:
This can be triggered in mock mode, for example, if you go to Settings > Integrations and click on "Display DC Loads" while in "User" access level.

Issue #2337, opened 2025-07-14
Different current amps diplays when not producing at night


tmonxx:
Not really bad but for cosmetical reason.
My both MPPTs show current amps with -A.
My PV inverters  show current amps with 0A.
It would be nice when both MPPT and inverters show equal values e.g. 0A.

System OS is beta 3.70~9 on cerbo GX device

<img width="1267" height="581" alt="Image" src="https://github.com/user-attachments/assets/e3ef8ffe-873e-49c6-a2b0-4d93fd30e9b9" />

mpvader:
We checked, and on dbus its 0A for those top two, and gui-v2 shows the dashes.

How its wanted is that the gui shows 0 instead of the dashes.

chriadam:
Merged to main, resolving as fixed.

PR #3187, opened 2026-08-31
Solar: Verify that when the tracker power is zero, the current is also zero.

ybott-qinetic:
Contributes to #2337

### DIFF
--- a/tests/solarinputmodel/tst_solarinputmodel.qml
+++ b/tests/solarinputmodel/tst_solarinputmodel.qml
@@ -421,6 +421,71 @@ TestCase {
 					}
 				],
 			},
+			{
+				tag: "solarcharger - 0 power",
+				devices: [
+					{
+						uid: "mock/com.victronenergy.solarcharger.a",
+						children: {
+							"Yield/Power": 0,
+							"Pv/V": 20,
+							DeviceInstance: 0,
+							ProductName: "solarcharger_product",
+						},
+					}
+				],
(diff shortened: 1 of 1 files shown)


src/solarinput.cpp
```diff
@@ -239,6 +239,8 @@ void TrackerSolarInput::updateCurrent()
 	// Trackers do not have current values, so calculate it manually.
 	if (!qIsNaN(m_power) && !qIsNaN(m_voltage) && m_voltage != 0) {
 		setCurrent(m_power / m_voltage);
+	} else if (!qIsNaN(m_power) && m_power == 0) {
```
blammit: Looks like this bug is already fixed in `main`. Due to the first `if` condition above, if the power is zero and the voltage > 0, then the current will be zero.

This can also be tested with the mock files - e.g. in `data/mock/conf/services/mppt2.json`, change "/Pv/0/P" (the power for the first tracker) from `1234` to `0`. If you open the solar drilldown, you can see the current is 0:

Image

blammit: But it would still be useful to verify this in the unit tests, if nothing else. In `tst_solarinputmodel`, you could add a test to `test_trackers_data()` to verify that when a tracker's power=0 and voltage > 0, then the current is zero.

blammit (commented):
LGTM, can be merged after the v3.80 release.

blammit (approved):

Issue #2844, opened 2026-02-23
Implement WifiModel in C++ and handle venus-platform crashing gracefully


chriadam:
If venus-platform restarts, the shown wifi menu currently isn't updated properly, and requires the user to backstep and reopen the wifi menu page.

We should reimplement WifiModel in C++, and ensure that we handle the case where venus-platform crashes (for testing, can use something like `kill -9 (pidof venus-platform)`) gracefully (i.e. detect owner name change, disconnect from old and subscribe to new updates properly, etc).

Note that #2205 is also related to venus-platform and localsettings restart handling, and maybe we can look into that one first, as it affects much more than just the wifi networks.

chriadam:
When venus-platform restarts, the entire data backend and the UI will be reloaded, as per the fix for #2205, and so the main reason for refactoring this model is now gone.  Of course, it would be better if it were a C++ model rather than a QML one, so the task is still valid, but much lower priority.  Reducing prio to low.

PR #3195, opened 2026-09-02
Models: move WifiModel to C++ and add a SortedWifiModel

ybott-qinetic:
Move WifiModel to C++ to update the wifi menu correctly between venus-platform restarts. Also adds SortedWifiModel to sort on connected status and then signal strength.

Modify mock data for wifi strength to show sorting.

Contributes to #2844


src/wifimodel.cpp
```diff
@@ -0,0 +1,234 @@
...
+WifiModel::WifiModel(QObject *parent)
+	: QAbstractListModel(parent)
+{
+	const QString serviceUid = BackendConnection::create()->serviceUidForType(QStringLiteral("platform"));
+
+	m_servicesItem = VeQItems::getRoot()->itemGetOrCreate(serviceUid + QStringLiteral("/Network/Services"));
+	m_scanItem = VeQItems::getRoot()->itemGetOrCreate(serviceUid + QStringLiteral("/Network/Wifi/Scan"));
+	m_accessPointItem = VeQItems::getRoot()->itemGetOrCreate(serviceUid + QStringLiteral("/Services/AccessPoint/Enabled"));
+
+	m_servicesItem->getValueAndChanges(this, &WifiModel::update);
```
blammit: For completeness, check whether each VeQItem pointer is valid before calling the functions.

src/wifimodel.cpp
```diff
@@ -0,0 +1,234 @@
...
+	*/
+	QJsonParseError parseError;
+	const QJsonDocument doc = QJsonDocument::fromJson(m_servicesItem->getValue().toString().toUtf8(), &parseError);
+	if (parseError.error != QJsonParseError::NoError || !doc.isObject()) {
+		return;
+	}
+	const QJsonObject wifis = doc.object().value(QStringLiteral("wifi")).toObject();
+
+	// Remove networks that have been dropped.
+	QStringList services;
```
blammit: The `services` list is only used from the `m_networks` loop below, so you could just search the `wifis` list from that loop instead, and avoid building this list just for the search.

src/wifimodel.h
```diff
@@ -0,0 +1,103 @@
...
+
+protected:
+	QHash<int, QByteArray> roleNames() const override;
+
+private:
+	struct WifiNetwork {
+		QString network;
+		QString service;
+		QString state;
+		bool favorite = false;
```
blammit: Put the `bool` last, after the `int` types.

src/wifimodel.cpp
```diff
@@ -0,0 +1,234 @@
...
+
+		bool found = false;
+		for (int j = 0; j < m_networks.count(); ++j) {
+			if (!service.isEmpty() && service == m_networks.at(j).service) {
+				found = true;
+				m_networks[j].network = network;
+				m_networks[j].state = details.value(QStringLiteral("State")).toString();
+				m_networks[j].favorite = details.value(QStringLiteral("Favorite")).toString() == QStringLiteral("yes");
+				m_networks[j].strength = strength;
+				emit dataChanged(index(j), index(j), { NetworkRole, StateRole, FavoriteRole, StrengthRole });
```
blammit: Check whether each value has changed, and only include it to the list of changed roles if it has changed.

src/wifimodel.cpp
```diff
@@ -0,0 +1,234 @@
...
+	bool found = false;
+	for (const WifiNetwork &wifiNetwork : std::as_const(m_networks)) {
+		if (wifiNetwork.state == QStringLiteral("ready") || wifiNetwork.state == QStringLiteral("online")) {
+			name = wifiNetwork.network;
+			found = true;
+			break;
+		}
+	}
+
+	if (!found) {
```
blammit: Could check whether name is empty instead of needing an extra `found`?

src/wifimodel.cpp
```diff
@@ -0,0 +1,234 @@
...
+{
+	const QString serviceUid = BackendConnection::create()->serviceUidForType(QStringLiteral("platform"));
+
+	m_servicesItem = VeQItems::getRoot()->itemGetOrCreate(serviceUid + QStringLiteral("/Network/Services"));
+	m_scanItem = VeQItems::getRoot()->itemGetOrCreate(serviceUid + QStringLiteral("/Network/Wifi/Scan"));
+	m_accessPointItem = VeQItems::getRoot()->itemGetOrCreate(serviceUid + QStringLiteral("/Services/AccessPoint/Enabled"));
+
+	m_servicesItem->getValueAndChanges(this, &WifiModel::update);
+	m_scanItem->getValueAndChanges(this, &WifiModel::update);
+	m_accessPointItem->getValueAndChanges(this, &WifiModel::update);
```
blammit: Should these be queued signal connections instead, to avoid multiple updates on intitialisation?

src/wifimodel.h
```diff
@@ -0,0 +1,103 @@
...
+
+	void update();
+	void updateConnectedNetworkName();
+
+	VeQItem *m_servicesItem = nullptr;
+	VeQItem *m_scanItem = nullptr;
+	VeQItem *m_accessPointItem = nullptr;
+	QList<WifiNetwork> m_networks;
+	bool m_valid = false;
+	QString m_connectedNetworkName;
```
blammit: Order by size, i.e. QList / QString / pointers / bool

blammit (commented):
Regarding the commit message,  "Wifi" is probably a more suitable commit topic than "Models". 

Has this been tested with a venus-platform crash?

src/wifimodel.cpp
ybott-qinetic: Done

src/wifimodel.h
ybott-qinetic: Done

src/wifimodel.cpp
ybott-qinetic: Done

src/wifimodel.cpp
ybott-qinetic: Done

src/wifimodel.cpp
ybott-qinetic: Updated.

src/wifimodel.h
ybott-qinetic: Done

src/wifimodel.cpp
```diff
@@ -0,0 +1,243 @@
...
+	const QJsonObject wifis = doc.object().value(QStringLiteral("wifi")).toObject();
+
+	// Remove networks that have been dropped.
+	for (int i = m_networks.count() - 1; i >= 0; --i) {
+		if (!wifis.contains(m_networks.at(i).service)) {
+			beginRemoveRows(QModelIndex(), i, i);
+			m_networks.removeAt(i);
+			endRemoveRows();
+		}
+	}
```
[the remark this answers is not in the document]
ybott-qinetic: Fixed.

src/wifimodel.cpp
```diff
@@ -0,0 +1,243 @@
...
+	if (name.isNull()) {
+		if (m_accessPointItem->getValue().isValid()) {
+			name = m_accessPointItem->getValue().toInt() == 1
+				//% "Disconnected | AP On"
+				? qtTrId("wifimodel_disconnected_ap_on")
+				//% "Disconnected | AP Off"
+				: qtTrId("wifimodel_disconnected_ap_off");
+		} else {
+			//% "Disconnected"
+			name = qtTrId("wifimodel_disconnected");
```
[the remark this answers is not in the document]
ybott-qinetic: I am not sure this edge case is worth covering.

PR #3199, opened 2026-09-04
UI Controls: avoid using MultiEffect + mask for all bar-type gauges

### LINKED ISSUE #3186: Overview page: battery sometimes filled with a 45 degree angle
Reported on [this slack thread](https://victrondevelopment.slack.com/archives/C04RYU0UMDE/p1786969497004489).

Happens via VRM and local.

Tested on Google Chrome 151.0.7922.137 with Venus OS 3.70 and 3.80~33. Loading bar is also from QtQuick in that case.

Image

Same issue for the loading bar

Image

blammit:
UI Controls: avoid using MultiEffect + mask for all bar-type gauges

Use simple Rectangles instead of a MultiEffect with a mask to render bar
gauges, sliders and progress bars.

Remove BarGaugeBase and ClippingBarGauge, and consolidate the
implementations into BarGauge, which now renders corners using the
built-in per-corner radii support in Rectangle, instead of using a
MultiEffect with a mask. This is essentially the same as the
CheapBarGauge implementation, with some changes to the radii
calculations to progressively round the end corners when the progress
bar reaches the total. BarGauge is now used in:

- Tank gauges 
- Overview battery widget (for the background showing the SOC)

CheapBarGauge (used for the thin gauges in the Brief side panel and 
Overview) is now an extension of BarGauge that forces the corner radii 
to always remain constant, to reduce the number of binding updates. For
for the gauges in the Overview page, disable animations on the GX as
the CPU load is already high due to the electron animations; this
preserves the existing CheapBarGauge behaviour for vertical gauges.

Additionally, remove the MultiEffect and mask from:

- Slider: use BarGauge instead 
- ProgressBar: draw a moving foreground rect over a background rect
instead

This avoids browser rendering issues from using MultiEffect with a mask
to draw the gauge progress rectangle. In Chrome, those gauge rectangles
were rendered via WebGL as two triangles, and this sometimes resulted in
only one of those triangles being rendered. The MultiEffect+mask
approach was already causing issues when rendering on the GX device, so
just consolidate and avoid this approach on all platforms.

Note there is a visual change: for the thin bar gauges in the Brief side
panel and Overview, the corner radii are now always rounded, to
avoid updating the radii bindings on every value change, which may
negatively affect performance.

Also note that BarGauge now uses NumberAnimation instead of X/YAnimator
to animate the value changes. Now that the foreground rect in BarGauge
is resized, instead of being moved and clipped as the progress changes,
it can no longer use X/YAnimator. The Animator was originally introduced
in beab7c53c to avoid expensive gradient animations, but BarGauge no
longer requires gradients (as that was originally used to work around
the lack of per-corner radii in Rectangle prior to Qt 6.7) so it should
be safe to revert to using NumberAnimation again.

Other related fixes:
- Slider: do not animate value changes, as the value should jump 
immediately to the user-selected value
- BatteryWidget: only run the charge animator when the state is charging

Fixes #3186


chriadam:
Is there an advantage of using NumberAnimation specifically?  It will cause per-frame binding re-evaluation if anything depends on the property being animated, and regardless will result in per-frame property writes and signal emissions.  Of course, this may be what is required, depending on the implementation?

chriadam:
The behaviour change of the animation on the splash screen seems ... suboptimal to me.

blammit:
> Is there an advantage of using NumberAnimation specifically? It will cause per-frame binding re-evaluation if anything depends on the property being animated, and regardless will result in per-frame property writes and signal emissions. Of course, this may be what is required, depending on the implementation?

I'm not sure how we can avoid using NumberAnimation if we remove the MultiEffect+mask, since animators cannot operate on the width/height of the highlight in the bar gauge. If we use X/YAnimator, then we need to use masking as well.

We could turn off animations for the thin gauges in the Brief side panel and Overview, though in those cases, the main performance hits come respectively from the animations of the load graphs and the electron paths, rather than those of the bar gauges.


> The behaviour change of the animation on the splash screen seems ... suboptimal to me.

A couple of other options I can think of:

- The highlight area could just slide back and forth; that way it will not snap back abruptly to the beginning.
- Remove the rounded corners, and animate the highlight width to shrink/grow it, to appear as though it is moving off the edges. In this case we might need two NumberAnimations (x, width) as it may not be able to sync a width NumberAnimation with a XAnimator.

Otherwise, I'm not sure how we can keep the rounded corners without re-introducing the MultiEffect+mask... what do you think?

chriadam:
When the bar reaches its right most extent, can we animate its width rather than it's X position (and keep it anchored to the far right)?

blammit:
> When the bar reaches its right most extent, can we animate its width rather than it's X position (and keep it anchored to the far right)?

Yes, if we add some NumberAnimations around it. I've added a [commit](https://github.com/victronenergy/gui-v2/pull/3199/commits/041b0c3de5f9ad165b771ce61b295f9cc11c98b0) to do this, what do you think?

components/controls/ProgressBar.qml
```diff
@@ -6,76 +6,87 @@
...
-			layer.enabled: true
+			// For indeterminate progress bars, produce a visual effect where the bar looks as
+			// though it is entering and exiting the background area, by animating the highlight x
+			// and width at the same time.
+			SequentialAnimation {
+				running: root.indeterminate && UiConfig.applicationVisible && !ScreenBlanker.blanked
+				loops: Animation.Infinite
+				onStopped: {
+					highlightRect.x = 0
+					highlightRect.width = root.availableWidth * root.visualPosition
```
chriadam: should this be `highlightRect.width = Qt.binding(function() { return root.availableWidth * root.visualPosition })` ?

components/BarGauge.qml
```diff
@@ -4,42 +4,106 @@
...
+	property real value: 0.0
+	property int orientation: Qt.Vertical
+	property int valueType: VenusOS.Gauges_ValueType_NeutralPercentage
+	property int borderWidth
+	property color borderColor: Theme.color_overviewPage_widget_border
+	property int radius: Theme.geometry_overviewPage_widget_radius
+
+	// The status that determines the foreground/background colours. The calculation uses 'value'
+	// rather than 'animatedValue' to reduce possible colour changes.
+	readonly property int valueStatus: Theme.getValueStatus(Math.max(0, Math.min(1, value)) * 100, valueType)
```
chriadam: If value is NaN, what does this return?  Hopefully the final status is critical in this case?

components/BarGauge.qml
```diff
@@ -4,42 +4,106 @@
...
+	property color foregroundColor: Theme.color_darkOk,Theme.statusColorValue(valueStatus)
+	property color backgroundColor: Theme.color_darkOk,Theme.statusColorValue(valueStatus, true)
+
+	// Same as value, but clamped and animated for value changes.
+	readonly property real animatedValue: fgRect.animatedValue
+
+	// The radius of the two end corners (on top edge when vertical, or right edge when horizontal).
+	// These are progressively rounded as the progress reaches the total, to prevent the fgRect fill
+	// colour from spilling over the bgRect corners.
+	readonly property real endRadius: Math.max(0, root._clampedRadius - (fgRect.totalSize - fgRect.progressSize))
```
chriadam: robot says: "`r - sqrt(r^2 - (r-y)^2)` would be more exact"

components/controls/ProgressBar.qml
```diff
@@ -6,76 +6,87 @@
...
+					property: "width"
+					from: 0
+					to: root._indeterminateHighlightWidth
+					duration: Theme.animation_progressBar_duration / 2
+				}
 
-				XAnimator on x {
-					running: root.indeterminate
-					loops: Animation.Infinite
+				XAnimator {
```
chriadam, line 66: Probably just make this a NumberAnimation also, to ensure onStopped and applicationVisible order properly.

components/widgets/BatteryWidget.qml
```diff
@@ -61,79 +61,72 @@ OverviewWidget {
...
+				horizontalCenter: parent.horizontalCenter
+				bottom: parent.bottom
+				bottomMargin: batteryGauge.borderWidth
 			}
+			width: batteryGauge.width - (2 * batteryGauge.borderWidth)
+			height: (batteryGauge.height - (2 * batteryGauge.borderWidth)) * batteryGauge.animatedValue
+			visible: Global.system.battery.mode === VenusOS.Battery_Mode_Charging && root._animationReady
+			clip: true
+
+			SequentialAnimation {
```
chriadam, line 96: does this need to be disabled when the parent visible is false?

components/widgets/BatteryWidget.qml
```diff
@@ -61,79 +61,72 @@ OverviewWidget {
...
+				property bool startAnimation: root._animationReady
```
chriadam: I think this should be bound to parent.visible, too, as it should only run the animator when the state is Charging.

I know this is a previously-existing code, but worth fixing while we're here.

components/controls/ProgressBar.qml
chriadam: Ah, but then the motion is juddery during splash screen, because there's a bunch of UI thread compilation etc happening, and NumberAnimation requires UI thread for property updates and interpolation.  I see....

chriadam: Can you add a comment above this XAnimator to explain that, please?

chriadam:
Robot says: the bindings on width/height/independent-radii mean that they will re-evaluate every frame. Avoid that for overview and brief page. Restore a cheap vertical path (for overview) or use uniform radius except when animating near the caps.

blammit:
> Robot says: the bindings on width/height/independent-radii mean that they will re-evaluate every frame. Avoid that for overview and brief page. Restore a cheap vertical path (for overview) or use uniform radius except when animating near the caps.

We could also avoid animations on those pages, if that would be better, to avoid additional evaluations. Or just uses radii instead of straight edges, so that the radius never changes.

components/BarGauge.qml
blammit: No, but should it be? It could also be that the value is just not available, which may not be critical.

chriadam: At least, I think the current code assumes the valueStatus will be critical if the value is NaN.  But maybe I'm wrong.

blammit: Oh right - where does it do that? Maybe that should be fixed too

components/widgets/BatteryWidget.qml
blammit: Done

components/widgets/BatteryWidget.qml
blammit: Done

components/controls/ProgressBar.qml
blammit: Done

components/BarGauge.qml
blammit: I couldn't get that to work, so I've left it as-is

components/BarGauge.qml
blammit: Note also, the end value is different depending on whether the value type is ascending or descending

components/controls/ProgressBar.qml
blammit: Fixed

blammit:
> Robot says: the bindings on width/height/independent-radii mean that they will re-evaluate every frame. Avoid that for overview and brief page. Restore a cheap vertical path (for overview) or use uniform radius except when animating near the caps.

I've updated the PR to restore CheapBarGauge: it is now a BarGauge that uses a constant end radius, so the end-radius bindings no longer re-evaulate.

PR #3204, opened 2026-09-07
BLE: improve hint for BLE sensor encryption keys

### LINKED ISSUE #3202: Add text in Bluetooth sensors menu
When at least one sensor is shown which requires entering an Encryption key, add a hint at the very bottom of the list of found sensors:

`Use VictronConnect app over Bluetooth to add encryption keys automatically`

More background here: https://github.com/victronenergy/venus-private/issues/691

blammit:
Use "VictronConnect app" instead of "VictronConnect".

Clean up the hint-addition implementation to reuse the same VeQItemChildModel model as for the BLE sensor list.

Contributes to victronenergy/venus-private#691
Fixes #3202

### DIFF
--- a/pages/settings/PageSettingsBleSensors.qml
+++ b/pages/settings/PageSettingsBleSensors.qml
@@ -32,13 +32,24 @@ Page {
 		filterFlags: VeQItemSortTableModel.FilterOffline
 	}
 
+	VeQItemSortTableModel {
+		id: deviceChildrenNameModel
+
+		model: VeQItemChildModel {
+			model: sensors
+			childId: "Name"
+		}
+		dynamicSortFilter: true
+		filterFlags: VeQItemSortTableModel.FilterInvalid
+	}
+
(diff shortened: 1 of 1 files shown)


pages/settings/PageSettingsBleSensors.qml
```diff
@@ -115,7 +118,7 @@ Page {
 			}
 
 			ListInfoLabel {
-				//% "Use VictronConnect over Bluetooth to add encryption keys automatically."
+				//% "Use VictronConnect app over Bluetooth to add encryption keys automatically."
```
chriadam, line 121: maybe "Use the VictronConnect app..."

philipptrenz: I think we can omit “the" in this case

chriadam (approved):
LGTM, so long as Philipp has tested and it works with real devices and matches his expectations

PR #3209, opened 2026-09-08
BLE: don't re-enable a disabled sensor when showing the page

### LINKED ISSUE #3207: Disable bluetooth sensor with encryption key is re-enabled when entering or leaving the menu
When a device with a valid encryption key is disabled in the Bluetooth Sensors menu and then the user leaves the menu, the disabled sensor is automatically enabled again.

More or less the same happens when the user is not in the Bluetooth Sensors menu, disables a device via VictronConnect and the user then enters the Bluetooth Sensors menu in the gui. The device is also automatically re-enabled in that case.

DanielMcInnes:
fixes #3207

### DIFF
--- a/components/controls/TextValidationField.qml
+++ b/components/controls/TextValidationField.qml
@@ -132,6 +132,7 @@ TextField {
 		_inputCancelled = true // flag to prevent validation from running when focus is lost
 		focus = false
 		_inputCancelled = false
+		_validateBeforeSaving = false
 	}
 
 	MouseArea {
--- a/pages/settings/BleSensorDelegate.qml
+++ b/pages/settings/BleSensorDelegate.qml
@@ -22,10 +22,9 @@ import Victron.VenusOS
 	again brings the field back, focused and holding the stored key, so that it can be switched back
 	on with a single Enter.
 
-	A device whose stored key is already complete is enabled as soon as this delegate learns its
-	state, so that it comes up switched on rather than showing a key field for a key that is already
-	known. Note that this applies every time the delegate is created, not only at application
-	startup: a device switched off here is switched on again by the next visit to this page.
+	The enabled state is only ever written in response to the user acting on this page: switching a
+	device off, or submitting a key. Showing a device must not change it - a device switched off
+	here, or elsewhere (VictronConnect), stays off when the page is left and entered again.
 
 	A loader swaps between the two standard list items as the device's state changes, so neither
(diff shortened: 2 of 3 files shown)


components/controls/TextValidationField.qml
```diff
@@ -132,6 +132,7 @@ TextField {
 		_inputCancelled = true // flag to prevent validation from running when focus is lost
 		focus = false
 		_inputCancelled = false
+		_validateBeforeSaving = false
```
chriadam, line 135: What side-effects could this have?

pages/settings/BleSensorDelegate.qml
```diff
@@ -345,19 +303,22 @@ ListItemLoader {
...
+					// submittable too: after a device is switched off, its field comes back still
+					// holding the stored key, and re-submitting that key is how the device is
+					// switched back on. Marking the field as having something to save arms both of
+					// its submit paths, so it is done for the Enter keystroke and for nothing else.
+					// Arming it on focus instead - as this once did - arms the focus-loss path as
+					// well, and then merely leaving the page, which takes the focus away, submits
+					// the key and switches the device back on.
+					// The handler does not accept the event: Enter has to go on to reach the text
+					// field, whose accepted() is what actually performs the submit.
+					Keys.onReturnPressed: (event) => {
```
chriadam, line 315: should this also check keyTextField.hasActiveFocus ?

chriadam: Ah, no, I see - because the event must eventually be accepted by the text field itself.

components/controls/TextValidationField.qml
chriadam: I think it's good.

chriadam (approved):
LGTM, but would like Bea or Rein to take a look also, if possible

pages/settings/BleSensorDelegate.qml
blammit, line 315: Should we just do this in `TextValidationField` itself, and do it for all text fields, instead of just from BleSensorDelegate? It seems valid to allow validation-on-enter in general.

blammit: Or we could do that after v3.80 to reduce the scope of the bug fix. Either LGTM.

blammit (approved):

ReinvdZee:
Thiemo just confirmed the fix works.

Issue #1324, opened 2024-07-12
Add alternator temp to Overview view tile


msigman:
Add the alternator temp to the Overview tile, like the battery temp.

![image](https://github.com/user-attachments/assets/2e231118-5b0f-48a8-934b-83b8e203be96)

blammit:
@jepefe fyi. Are there any plans to expand the details shown in the overview boxes?

chriadam:
Presumably would be nice-to-have.  Should discuss early in v3.90 cycle.

msigman:
> Presumably would be nice-to-have.  Should discuss early in v3.90 cycle.

Yes, a nice-to-have, but VERY nice :)

ReinvdZee:
@ReinvdZee will check with Serj if this is still wanted

ReinvdZee:
@ybott-qinetic 

Discussed this with Serj and we decided to show the alternator temperature on the top-right corner of the widget if available. Please make sure to show the value in the GX-wide temperature unit. According to the [dbus wiki](https://github.com/victronenergy/venus/wiki/dbus#alternator-and-dcsource), the temperature is reported under the path `/Dc/0/Temperature` in degrees centigrade. Do not show anything if the path is not there.

PR #3211, opened 2026-09-09
Overview: add temperature to alternator widget

ybott-qinetic:
Refactor overview battery temperature display.
Add temperature information for alternator dcinput source.

Contributes to issue #1324


components/widgets/WidgetHeader.qml
```diff
@@ -26,4 +28,17 @@ Label {
...
+		id: secondaryQuantityLabel
+
+		anchors {
+			verticalCenter: parent.verticalCenter
+			right: parent.right
+			rightMargin: 0
+		}
+		font.pixelSize: root.font.pixelSize
+		alignment: Qt.AlignRight
+		visible: !isNaN(Global.system.battery.temperature)
```
chriadam: this is the wrong binding for something which is meant to be generic (i.e. this is WidgetHeader.qml, nothing to do with Battery etc).

components/widgets/WidgetHeader.qml
```diff
@@ -26,4 +28,17 @@ Label {
 		fillMode: Image.Pad
 		color: parent.color
 	}
+
+	QuantityLabel {
```
chriadam: Aside from Battery+Alternator, does any other widget require this secondary header label?  If not, I wonder if it's better to just copy it once, instead.

components/widgets/DcInputWidget.qml
```diff
@@ -22,6 +22,10 @@ OverviewWidget {
 			: "/pages/settings/devicelist/dc-in/PageDcMeter.qml"
 	readonly property bool _widgetOnlyPresentsDcGensets: root.serviceType === "dcgenset" && inputDeviceModel.commonMeterType !== -1
 
+	readonly property VeQuickItem _firstAlternatorTemperature: VeQuickItem {
+			uid: serviceType === "alternator" ? inputDeviceModel.firstObject.serviceUid + "/Dc/0/Temperature" : ""
+		}
```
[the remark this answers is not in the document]
chriadam: Yeah, nah, I think showing the first one is fine here, assuming that it's the same pattern we do for batteries.
But, yes, please do "guard" the dereference of the firstObject in case the count is zero etc.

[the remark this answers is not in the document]
ybott-qinetic: Done

components/widgets/DcInputWidget.qml
```diff
@@ -22,6 +22,13 @@ OverviewWidget {
 			: "/pages/settings/devicelist/dc-in/PageDcMeter.qml"
 	readonly property bool _widgetOnlyPresentsDcGensets: root.serviceType === "dcgenset" && inputDeviceModel.commonMeterType !== -1
 
+	readonly property VeQuickItem _firstAlternatorTemperature: VeQuickItem {
```
blammit, line 25: The root type is an `Item`, so this can be a plain child item rather than a property.

ybott-qinetic: Done

components/widgets/DcInputWidget.qml
```diff
@@ -57,6 +57,28 @@ OverviewWidget {
...
 	}
 
+	VeQuickItem {
+		id: firstAlternatorTemperature
+		uid: root.serviceType === "alternator" && inputDeviceModel.count === 1
+			? inputDeviceModel.firstObject.serviceUid + "/Dc/0/Temperature"
+			: ""
+	}
+
+	QuantityLabel {
```
blammit, line 67: This QuantityLabel should be inside the `contentItem`, which defines the content for the widget. Also, that way you can ensure that the "Alternator" text never overlaps with the `WidgetHeader` title text, in case the "Alternator" text is much longer in the current language.

blammit: Come to think of it, we need to do that for the battery widget as well. Currently it's possible for the "Battery" text to run over the top of the temperature display... I'll make a PR to fix that.

PR #3216, opened 2026-09-09
Switches: ensure name is correct when device is not yet available

### LINKED ISSUE #3198: Switch pane: Naming issue
Observed on Venus OS 3.80~45.

The name displaying rules are the following:

If a switch or input belongs to a group (`/Group` is set), but the switch or input does not have a custom name (`/SwitchableOutput/x/Settings/CustomName` is not set), then list the switch in the card named by the group name with <Device custom name OR device product name | Switch name>.

On this system, the switches do have a group ("Test"), the device does not have a custom name, but does have a product name. The issue is that the device productname is not displayed before the '|'.

It also was observed that on local UI this did not work for input, and on wasm it did not work for input and outputs.
After adding a customname to the device (`/CustomName`), it suddenly worked on wasm but not on local UI.
We also observed that when it reproduced on local UI, restarting the local UI resolves the issue.

There doesn't seem to be a clear pattern on whether it works or not for switches and inputs on local or WASM.

Image

blammit:
If the device details for an input/output control are not available on initialisation, ensure the displayed name is simply "<control name>", instead of " | <control name>" (i.e. starting with an empty string where the device name should be). Later when the device details become available via AllDevicesModel, update the displayed name to be "<device name> | <control name>".

Fixes #3198

### DIFF
--- a/src/iochannel.cpp
+++ b/src/iochannel.cpp
@@ -28,6 +28,15 @@ void IOChannel::initialize(VeQItem *item)
 	}
 	m_item = item;
 
+	if (m_deviceAddedConn) {
+		disconnect(m_deviceAddedConn);
+		m_deviceAddedConn = QMetaObject::Connection();
+	}
+	if (m_device) {
+		m_device->disconnect(this);
+		m_device.clear();
+	}
+
 	if (m_item) {
 		m_serviceUid = m_item->itemParent() // the /GenericInput or /SwitchableOutput parent
 				&& m_item->itemParent()->itemParent() // the root service item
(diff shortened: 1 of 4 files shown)


src/iochannel.cpp
```diff
@@ -28,6 +28,11 @@ void IOChannel::initialize(VeQItem *item)
 	}
 	m_item = item;
 
+	if (m_deviceAddedConn) {
+		disconnect(m_deviceAddedConn);
+		m_deviceAddedConn = QMetaObject::Connection();
+	}
```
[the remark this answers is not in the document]
blammit: It shouldn't happen, but it doesn't hurt to clear the m_device earlier. Done.

src/iochannel.cpp
```diff
@@ -28,6 +28,16 @@ void IOChannel::initialize(VeQItem *item)
 	}
 	m_item = item;
 
+	// Disconnect from deviceAdded() signal, and clear the device pointer.
```
chriadam: comment is superfluous

DanielMcInnes:
do we care about the case where a valid device becomes invalid?

eg. device has `ProductName="Foo"`, no `CustomName`, instance 1; grouped channel X displays `Foo (1) | X`. `/ProductName` is then cleared. `BaseDevice::maybeEmitValidChanged()` (`basedevice.cpp:123-129`) runs the property change — and its `productNameChanged` — *before* `validChanged`, so `IOChannel::updateFormattedName()` runs with the device still bound and builds `QStringLiteral("%1 (%2)").arg("").arg(1)` → prefix `" (1)"`. That is not empty, so the new `prefix.isEmpty()` guard doesn't catch it, and the row renders **`" (1) | X"`** — the same leading-empty-string shape as #3198. `validChanged` fires afterwards, `deviceAboutToBeRemoved` is ignored, `m_deviceAddedConn` is already gone, and nothing ever corrects it.

If this is too rare to worry about, I'm happy to leave it.

src/iochannel.cpp
blammit: True, removed

blammit:
> do we care about the case where a valid device becomes invalid?

The `ProductName` should never be empty, and shouldn't change at all, so I think it's safe to ignore this case.

DanielMcInnes (approved):

Issue #3154, opened 2026-08-14
Touch scrolling issue on Garmin MFD


martinbosma:
Complaint from a customer with extensive report and proposed fix.

----

_During physical validation of a custom application integrated with a Victron GX device through Garmin OneHelm, I have identified several reproducible input problems affecting **Victron GUI-v2 Remote Console on Garmin**.
I investigated these problems because Remote Console is accessible from our application, but the testing ultimately demonstrated that the issue is **not caused by our application or iframe integration**._

_The same behavior can be reproduced when GUI-v2 is launched directly and when Remote Console is launched from the stock Victron MFD  application._

**_Test environment_**
- Cerbo GX
- Venus OS v3.75
- Victron GUI-v2 / WebAssembly Remote Console
- GUI-v2 build based on Qt 6.8.3 WebAssembly
- Garmin OneHelm embedded browser
- Garmin browser user agent observed:
   - Mozilla/5.0 (X11; Linux aarch64)
   - AppleWebKit/605.1.15
   - (KHTML, like Gecko)
   - Version/60.5 Safari/605.1.15

**_Control tests_**
_We tested the input path at several levels._

Test | Result
-- | --
Plain HTML page in Garmin OneHelm | PASS
Plain HTML <input> + Garmin keyboard | PASS
Plain HTML Backspace | PASS
Plain HTML Tab | PASS
Plain HTML scrolling | PASS
Stock Victron /app normal tapping/navigation | PASS
Stock Victron /app continuous PWM slider drag | PASS
Direct GUI-v2 on Garmin | FAIL
Custom application iframe → GUI-v2 | FAIL
Stock Victron /app → Remote Console → GUI-v2 | FAIL

_This localizes the problem to the **Garmin WebKit ↔ Qt/WASM GUI-v2 input layer**, rather than Garmin touch support in general or our application._

_Stock GUI-v2 initially exhibited significant scrolling instability:_
- first drag generally worked;
- subsequent touches could jump the view toward an earlier position;
- reaching page boundaries produced particularly strong jumps.

_The same problem exists when Remote Console is opened from the stock Victron browser application, ruling out our wrapper._

**_Effect of the mousedown focus workaround_**
_The same narrow mousedown workaround used for the keyboard substantially improves scrolling.
After applying it:_
- normal mid-page drag works;
- release and subsequent re-touch works;
- scrolling up/down through the middle of a page is stable;
- fast flick/inertia works;
- inertia correctly stops at the page boundary.

[...]

PR #3224, opened 2026-09-10
VisibleItemModel: declare effectiveVisibleChanged() as a slot

mman:
`VisibleItemModelPrivate::append()` connects each source item's `effectiveVisibleChanged()` signal using the string-based `SIGNAL()`/`SLOT()` overload, but the target method was declared `Q_INVOKABLE` rather than as a slot. Qt 6.11 started warning about this on every such connect:

```
QObject::connect: the SLOT() macro is used with a non-slot function: Victron::VenusOS::VisibleItemModel::effectiveVisibleChanged(). This currently works due to backwards-compatibility reasons. In Qt7 the SLOT() macro will work only for methods marked as slots.
```

Because the connect runs once per list item, an experimental Qt 6.11.2 wasm build printed the warning hundreds of times during startup. Console output from wasm is expensive on embedded MFD browsers, and on a Raymarine MFD that build loaded noticeably slower than the same build after this fix.

The method is private and only reached through that connect, so it is now declared under `private Q_SLOTS`. No behaviour change on Qt 6.8.3; the warnings disappear on 6.11+, and the code stays valid for Qt 7, where `SLOT()` will only accept real slots.

### Verification

- Headless (Playwright, `?mock`) against a Qt 6.11.2 wasm build: the warning count went from one per list item to zero, with the UI still reaching "UI loaded and visible" in the same time.
- On-device: the fixed Qt 6.11.2 build installed on a Cerbo GX (v3.80~49) driven from a Raymarine MFD loads better and works normally.

Found while preparing Qt 6.11/6.12 test builds for #3154.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/visibleitemmodel.h
+++ b/src/visibleitemmodel.h
@@ -69,10 +69,11 @@ public:
 Q_SIGNALS:
 	void sourceModelChanged();
 
+private Q_SLOTS:
+	void effectiveVisibleChanged();
+
 private:
 	Q_DISABLE_COPY(VisibleItemModel)
-
-	Q_INVOKABLE void effectiveVisibleChanged();
 };
 
 } /* VenusOS */


src/visibleitemmodel.h
```diff
@@ -69,10 +69,11 @@ class VisibleItemModel : public QQmlInstanceModel
 Q_SIGNALS:
 	void sourceModelChanged();
 
+private Q_SLOTS:
+	void effectiveVisibleChanged();
```
chriadam, line 73: The function name is strange; it looks like a signal to me.  Should it be "handleEffectiveVisibleChanged" or "emitEffectiveVisibleChanged" or something else instead?  Also, we should avoid using string-based signal connections where possible, so I suspect that we should just fix the VisibleItemModelPrivate implementation.

But, finally: we are moving away from VisibleItemModel altogether, see e.g. the work on DelegateComponentModel in PR #3150 etc.  So eventually, the entire VisibleItemModel type might go away anyway.

Issue #3221, opened 2026-09-10
Change EVCS "Start" and "Stop" controls from switches to buttons


ReinvdZee:
The EVCS itself and VictronConnect show "Start" or "Stop" buttons to start/stop a charging session, gui-v2 and VRM instead have a toggle switch. I think the Start/Stop buttons is much more intuitive. Toggles are well suited for permanent settings, but for "actions" push buttons are more intuitive.

EVCS page:

<img width="1251" height="615" alt="Image" src="https://github.com/user-attachments/assets/5c9e5904-5e58-4b94-a0f0-70ebc1149aae" />

VictronConnect:
<img width="911" height="712" alt="Image" src="https://github.com/user-attachments/assets/a05f7419-bc30-4dff-8bff-787645a59958" />

Gui-v2:

<img width="1164" height="684" alt="Image" src="https://github.com/user-attachments/assets/e43f6716-0c75-4180-a818-0d817c83005d" />
<img width="1164" height="685" alt="Image" src="https://github.com/user-attachments/assets/b92bf5b9-90ad-4481-b242-cb88d361e1dc" />

I'll check with serj first.

ReinvdZee:
Discussed this with Serj and he agreed that it is more intuitive.
Removing design not ready label.

ReinvdZee:
@philipptrenz,

On the device page, is this a setting to enable charging as soon as the car is plugged in? Or does the user need to toggle that switch everytime the car is plugged in to start charging?

<img width="1164" height="685" alt="Image" src="https://github.com/user-attachments/assets/74c20291-d44c-468f-910f-edf6272881ff" />

Because if it is to allow the EVCS to start charging, then we could argue it should be a toggle switch.

philipptrenz:
There is a setting called "Autostart" under "Setup" that, when enabled, automatically "toggles" the "Enable charging" switch to on. If the setting is off, then the user always has to activate "Enable charging" manually.

philipptrenz:
One thing that might also be worth considering: For VictronConnect, we added the `/IsStopAllowed` path. If the path exists (it does not when using the Modbus integration) and the value is 0, then the user cannot stop the charging session. As far as I know, this is mainly to cover the case where the EVCS is between two states (starting to charge, switching phases, etc.)

ReinvdZee:
@philipptrenz 

[...]

PR #3231, opened 2026-09-14
EVCS: change enable charge switch to Start/Stop button

ybott-qinetic:
Previously the setting was enabled by a Switch this has been modified to use a button to standardise UI across platforms. This change also prevents interaction based on the IsStopAllowed path being set and enabled.

Also adds minor convenience helpers in the mocked UI.

Contributes to #3221


data/mock/EvChargersImpl.qml
```diff
@@ -19,6 +19,24 @@ Item {
...
+				uid: evcs.uid + "/StartStop"
+				onValueChanged: {
+					if (valid && value === 1) {
+						MockManager.setValue(evcs.uid + "/IsStopAllowed", 0)
+						delayIsStopAllowedTimer.start()
+					}
+				}
+				property Timer delayIsStopAllowedTimer: Timer {
+					interval: 5000
+					repeat: true
```
blammit: Looks like this shouldn't repeat, since onTriggered stops it after the first run.

Also this `Timer` object doesn't need to be a property.

data/mock/EvChargersImpl.qml
```diff
@@ -19,6 +19,24 @@ Item {
 
 			VeQuickItem { id: evcsMaxCurrent; uid: evcs.uid + "/MaxCurrent" }
 
+			VeQuickItem {
+				uid: evcs.uid + "/StartStop"
+				onValueChanged: {
+					if (valid && value === 1) {
+						MockManager.setValue(evcs.uid + "/IsStopAllowed", 0)
+						delayIsStopAllowedTimer.start()
```
blammit, line 27: The timer is triggered when the value is first initialised - e.g. if you start in mock mode and go to the control cards immediately, the "Stop" is disabled". This might cause inconsistencies in UI testing. You could avoid this by checking some flag that is set from `Component.onCompleted`.

blammit (commented):
Looks good, have comments about some minor things.

data/mock/EvChargersImpl.qml
ybott-qinetic: Removed repeat.

data/mock/EvChargersImpl.qml
ybott-qinetic: The mock data sets the StartStop to 1 so I think this is functioning correctly. I have set the path in the mock data to be 0 for testing consistency.

blammit: Okay, yep setting it to 0 avoids the issue.

blammit (approved):

Issue #3223, opened 2026-09-10
evcharger: In the control card, use /Current instead of /SetCurrent during Auto mode


philipptrenz:
Related to #3221 

During Auto mode, the "Charge current" entry in the control card for the EV Charging Station shows the value of `/SetCurrent`. However, this value is only related to the "Manual" mode. Therefore, as long as the mode is not "Manual", show the value of `/Current` as read-only, which reflects the currently used setting.

On the device page, the "Charge current" setting is already hidden and the actual current is shown above the table view. So, this is already good as is.

<img width="1237" height="1442" alt="Image" src="https://github.com/user-attachments/assets/14f44677-6b4f-4cdb-8d44-6cd4089bbc31" />

chriadam:
Merged to main, resolving as fixed.

PR #3233, opened 2026-09-14
EVCS: update set current spin box to report current

ybott-qinetic:
Contributes to issue #3223


components/listitems/ListEvcsSetCurrentSpinBox.qml
```diff
@@ -10,15 +10,17 @@ ListSpinBox {
...
 
 	text: CommonWords.charge_current
 	suffix: Units.defaultUnitString(VenusOS.Units_Amp)
-	from: 6
-	to: maxCurrent.valid ? maxCurrent.value : 32
+	from: manualMode ? 6 : 0
+	to: manualMode ? maxCurrent.valid ? maxCurrent.value : 32 : Global.int32Max
 	stepSize: 1
-	dataItem.uid: serviceUid + "/SetCurrent"
+	dataItem.uid: manualMode ? serviceUid + "/SetCurrent" : serviceUid + "/Current"
```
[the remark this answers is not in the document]
ybott-qinetic: I tested the control card in mock mode as the /Current path is updated.

components/listitems/ListEvcsSetCurrentSpinBox.qml
```diff
@@ -10,15 +10,17 @@ ListSpinBox {
...
 
 	required property string serviceUid
+	required property bool manualMode
 
 	text: CommonWords.charge_current
 	suffix: Units.defaultUnitString(VenusOS.Units_Amp)
-	from: 6
-	to: maxCurrent.valid ? maxCurrent.value : 32
+	from: manualMode ? 6 : 0
+	to: manualMode ? maxCurrent.valid ? maxCurrent.value : 32 : Global.int32Max
```
blammit: Could you rephrase this with parentheses or newlines to be more clear, e.g.

```
to: manualMode ? (maxCurrent.valid ? maxCurrent.value : 32) : Global.int32Max
```

I would bind the to/from depending on `clickable` rather than `manualMode`. In this case it's the same, but the binding would make it clear it depends on whether it's read-only.

Also, add a comment in this file to explain when/why `Current` is shown instead of `SetCurrent`.

components/listitems/ListEvcsSetCurrentSpinBox.qml
```diff
@@ -10,15 +10,17 @@ ListSpinBox {
 	id: root
 
 	required property string serviceUid
+	required property bool manualMode
```
blammit: Can you add a comment to indicate the effect of setting `manualMode` to true vs false?

blammit (commented):
Can you explain in the commit message what the change is, and why it's needed?

pages/evcs/EvChargerPage.qml
```diff
@@ -156,8 +156,8 @@ DevicePage {
 
 		ListEvcsSetCurrentSpinBox {
 			serviceUid: evCharger.serviceUid
-			text: CommonWords.charge_current
 			preferredVisible: dataItem.valid && chargeMode.dataItem.value === VenusOS.Evcs_Mode_Manual
+			interactive: chargeMode.dataItem.value === VenusOS.Evcs_Mode_Manual
```
blammit: @ybott-qinetic  This isn't needed, as the `preferredVisible` binding in the line above already ensures the spin box is only visible in "manual" mode.

ybott-qinetic: Done

blammit (approved):

PR #3234, opened 2026-09-14
Boat: do not show consumption & range on charging state layout

### LINKED ISSUE #3232: Boat: range & consumption shown incorrectly on charging state layout
When the E-drive and GPS are turned off, the screen goes to the charging state layout.
The issue is that the consumption and range metric are still shown.

Image

citolen:
Closes https://github.com/victronenergy/gui-v2/issues/3232

Screenshot 2026-09-14 at 09 33 34


chriadam (approved):
LGTM, but I'd like Daniel to take a look, too.

DanielMcInnes:
LGTM

AI review comments - FYI only, do not make changes based solely on these comments. They are often wrong.


**Verdict:** approve with comments. It fixes the reported case, but it hides Range/Consumption using a different condition from the one that selects the charging layout, so the two can disagree.

**What it does**
- Adds `MotorDrives.hasAny`, which is true when a motordrive device is present.
- Shows `Boat.Range` and `Boat.Consumption` only when their value is valid **and** `hasAny` is true. `BoatPage_Landscape` passes `motorDrives` in as a new required property.
- Adds a mock config "No GPS, no motordrives, range, consumption" that reproduces the issue.

**Against the issue**
- "E-drive and GPS off → charging layout, but range & consumption still shown" → covered **if** the motordrive disappears from D-Bus (or `/MotorDrive/0/DeviceInstance` becomes invalid) when the e-drive is switched off. See finding 1.

**Verified locally**
- Where the charging layout is chosen: `pages/boat/LargeCenterGauge.qml:21`. It shows when `activeDataSource === null`, i.e. `!gps.valid && !motorDrives.dcConsumption.quotient.valid`. The same test also appears in `SlotVerticalCenterLeft.qml` and in the `boatPageActive` Binding in `BoatPage_Landscape.qml`.
- Why the values are still there, from dbus-systemcalc-py master:
  - `delegates/motordriveconsumption.py:221` publishes ConsumptionWhkm/Ahkm whenever `consumption_window_total_distance > 0`.
  - When GPS or power goes away, `_calculate()` (~l.74–82) only resets the EMA and the last point. It never clears the window.
  - So the last consumption value stays published indefinitely. `motordriverange.py` keeps publishing a range computed from that old consumption and the live SoC, which means the range *grows* while the battery charges.
  - The GUI hiding them is a reasonable fix, but it is working around this.
- `Boat.Range` and `Boat.Consumption` are only used in `BoatPage_Landscape.qml`, because portrait shows a RotateDevicePrompt. The new required property is set everywhere it needs to be.
- No UI test depends on the index of the BoatPageConfig configs; they are only cycled through by mock shortcuts.

**Not verified**
- Not built, not run with qmllint, the mock was not run, and no screenshots were compared.
- Not checked on real hardware: whether switching off a real e-drive removes its motordrive service or its `/MotorDrive/N/DeviceInstance`.

**Findings**
1. **Hide the values using the same condition that selects the charging layout** — `pages/boat/Range.qml:26`, `pages/boat/Consumption.qml:41`, `pages/boat/MotorDrives.qml:46`
   `hasAny` only says a device exists. The layout depends on whether `dcConsumption.quotient.valid` is true.
   Failure scenario: the motordrive service is still on the bus but `/MotorDrive/Power` is invalid (drive powered down but still announced, or systemcalc stops publishing power). The page then shows the charging layout, `hasAny` is still true, and the old Range and Consumption values are still shown. That is exactly the bug in #3232.
   Suggested change: `visible: rangeItem.valid && root.motorDrives.dcConsumption.quotient.valid`, and the same in Consumption. This covers every case `hasAny` covers, because without a device `power.numeratorUid` is `""`, so the quotient is already invalid. With that change `hasAny` isn't needed.
   It also keeps the PR's good behaviour when GPS is on and there is no e-drive: the old values stay hidden, which a plain "not charging layout" test would not do.
   It would also help to add a mock config that reproduces this: motordrive present, `/MotorDrive/Power` invalid, range and consumption set.

**Nits**
- `hasAny: root.single || root.left || root.right`: `left` and `right` are always both null or both set, so this is just `!!root._firstDevice`. It is moot if finding 1 is taken.
- Separately from this PR, it may be worth raising the stale consumption window with dbus-systemcalc-py. Keeping the last consumption while moored may be intentional, but a range that rises while charging probably isn't.

**ChatGPT second opinion**
- Agreed on:
  - Finding 1, and that `quotient.valid` is the better gate. The only downside it saw: a missing gauge-max setting would also hide the values, but the center gauge already treats that state as "no e-drive data", so hiding them is consistent.
  - The fix depends on the motordrive device disappearing when the drive is switched off; checking that power is valid does not.
  - No QML correctness issues. The `||` chain converts to bool correctly. In `motorDrives: motorDrives` the right-hand side resolves to the id, not the property, so it doesn't bind to itself.
  - The new mock entry reproduces the reported case, but not the device-present, power-invalid case.
- Additionally raised: nothing new.
- I disagree with: nothing. On TimeToGo, we both think it is out of scope. It is a battery metric that only shows when TTG > 60 s, not something derived from the e-drive.

DanielMcInnes (approved):

chriadam:
I dunno.  The `visible: rangeItem.valid && root.motorDrives.dcConsumption.quotient.valid` might be technically more correct, but it's also not very obvious WHY that condition is the appropriate one.  the `hasAny` property is more self-documenting.  If systemcalc publishes an "invalid" value, that's a bug in systemcalc, so defensively covering that case doesn't sound good.  On the other hand, ensuring that the condition matches the layout condition DOES sound worthwhile to me.  Maybe it makes sense to use both `hasAny` and then add a `showMotordriveMetrics` property which is bound to `root.motordrives.dcConsumption.quotient.valid` and use both of them in the visible binding, so that it includes the layout condition also?  I don't feel strongly about it, though.

chriadam:
Well, we can follow up with this in v3.90 if necessary.  Merging.

Issue #3151, opened 2026-08-07
Opening a settings page or widget drilldown freezes the UI for 100-700ms while the page is constructed


DanielMcInnes:
Opening a settings page or an Overview widget drilldown freezes the UI while the page is built. `PageStack.pushPage()` compiles and instantiates the whole page synchronously, and only then starts the push transition, so all of that time is a hard freeze: the app does not respond to touch and animations stop.

### Measured

Added `--ui-test benchmark/pages` (see `.github/benchmarking.md`) to measure this. On a Cerbo GX, v3.80~39:

| | |
|---|---|
| Instantiate, median | **114 ms** |
| Instantiate, average | **144 ms** |
| Instantiate, worst (`PageBatterySettings`) | **736 ms** |
| First open of a page adds (QML compile) | up to **462 ms** |

Overview widget drilldowns specifically: Battery **632 ms**, AC input **482 ms**, VE.Bus **397 ms**.

Almost all of it is instantiation, not compilation: over the whole sample, warm compile is 82 ms against 24.5 s of instantiation. Qt caches the compiled unit, so caching `Component`s in `PageStack` would gain nothing.

### Where the time goes

Settings pages are built almost entirely out of list items, and `VisibleItemModel` is an instance model, so every row of a page is constructed eagerly whether or not it is on screen. Per-instance cost on device:

```
Item                    2.7 ms
Label                   4.3 ms
VeQuickItem             3.7 ms
ThreeLabelLayout        9.0 ms      (a GridLayout plus three Labels, usually two of them empty)
ListItem                6.4 ms
ListSetting             8.1 ms
ListNavigation         17.3 ms
ListSwitch             18.1 ms
ListRadioButtonGroup   18.8 ms
```

So a page with 15-40 rows costs 250-700 ms. A page with 40 rows on screen is not unusual.

### Options

1. **Build the page asynchronously** (`Component.incubateObject`) so the work is spread over frames and the UI keeps responding. Biggest win, and page-count independent. Of 257 `pushPage()` call sites only one uses the return value, so the API change is small — **but see the hazard below**.
2. **Slim the shared list items.** Making the two usually-empty labels in `ThreeLabelLayout` lazy saves ~3 ms of a 17 ms row. Low ceiling (a `Label` costs ~1.5 ms at the margin, and a `Loader` gives most of that back) and it changes an alias API used in ~56 places.
[...]

PR #3240, opened 2026-09-17
Dev: add a benchmark for page construction cost

chriadam:
Pressing a settings entry or an Overview widget freezes the UI while the page is built: PageStack.pushPage() compiles and instantiates the whole page synchronously before the push transition starts. Nothing measured that, so add a UI test that does.

It reports the compile and instantiate cost of a representative sample of pages, separately for the first open and for subsequent ones, and the construction cost of a single instance of each of the list item types that settings pages are built from.

The per-component part declares types as Components compiled once with the test file, so the timed loop only instantiates (Qt.createQmlObject from a source string was measuring compilation). Each measurement is a test step of its own with a wait after it, so deferred destruction does not contaminate later samples.

Compile figures are reported as two endpoints, because Qt's on-disk bytecode cache makes a populated cache ~4x cheaper than an empty one:

                        cache populated   cache empty
    median page               21 ms           86 ms
    worst page               124 ms          309 ms
    whole 19-page sample     526 ms         2036 ms

A real device sits between them. Enabling qmlcachegen does not change this: UrlInterceptor (VENUS_GX_BUILD) rewrites qrc URLs back to disk, so the engine never loads the C++-compiled copy.

On a Cerbo GX, median of 3 runs, a typical page still takes a few hundred milliseconds to instantiate (the UI-blocking cost this benchmark exists to track). Drop PageSettingsDisplay.qml from the page list; it has not existed for some time.

Contributes to issue #3151


.github/benchmarking.md
```diff
@@ -170,3 +170,140 @@ python scripts/benchmark-overview.py compare -a baseline.csv -b feature.csv
 > **Tip:** Close other GPU-intensive applications during benchmarking to reduce
 > variance.  Run each capture 2–3 times and compare the most representative
 > results.
+
+## Benchmarking page construction
+
+`--ui-test benchmark/pages` measures how long pages take to construct, which is
+what determines the delay between pressing a settings entry or an Overview
+widget and the page appearing.  `PageStack.pushPage()` compiles and instantiates
+the whole page synchronously before the push transition starts, so this time is
```
blammit, line 179: There's another patch to compile and instantiate pages asynchronously - depending on which change gets merged first, this documentation will need to be updated.

chriadam: Yeah, it comes later in the stack.  Will update docs then.

tests/ui/benchmark/pages/tst_benchmark_pages.qml
```diff
@@ -0,0 +1,182 @@
...
+		for (let i = 0; i < objects.length; ++i) {
+			if (objects[i]) {
+				objects[i].destroy()
+			}
+		}
+	}
+
+	property int _totalCompile
+	property int _totalCreate
+	property int _measured
```
blammit, line 162: Place this further up with the other properties?

blammit (approved):

PR #3241, opened 2026-09-17
Dev: stop Settings page icon rows constructing a discarded contentItem

chriadam:
Overriding the contentItem of a derived type does not prevent the base type's contentItem from being constructed; it is built in full and then thrown away. SettingsPage's inline SettingsListNavigation extended ListNavigation and overlaid a left-hand icon, so every icon row also built ListNavigation's ThreeLabelLayout and chevron for nothing.

Split the parts of ListNavigation that are not the contentItem out into ListNavigationBase, and extend that instead. Measured with `--ui-test benchmark/pages` on a Cerbo GX, one SettingsListNavigation goes from 24.8ms to 18.2ms (-27%).

SettingsListNavigation.qml stays deleted (it is only an inline component on the Settings root page). Sub-pages keep ListNavigation + onClicked.

Contributes to issue #3151


blammit:
I think we don't need this PR anymore, as SettingsPage is the only place where the base type would be overridden, and SettingsPage is loaded on startup anyway.

chriadam:
Unneeded

PR #3242, opened 2026-09-17
Dev: build pages asynchronously so opening one does not block the UI

chriadam:
Dev: build pages asynchronously so opening one does not block the UI

    PageStack.pushPage() compiled and instantiated the whole page before it
    pushed anything, on the UI thread. On a Cerbo GX the median page takes
    329ms to instantiate and the worst 691ms, and for that whole time the
    application was stopped.

    Build the page with an incubator instead, so it is built a piece at a
    time between frames. The page is pushed once it is complete, so the wait
    before it appears is unchanged; the rest of the application keeps running.

    If the user leaves before the page is ready it is discarded, not pushed
    on top of wherever they went. The build is abandoned as soon as the
    shown page changes, and a pending build is forced to completion (and
    destroyed) if the PageStack itself is destroyed. Compiling the page with
    Qt.createComponent() is still synchronous, so the first open of a given
    page still blocks.

    pushPage() cannot return a page it has not built yet. Pass readyCallback
    if the caller needs the page; SettingsPage.goToConnectivityPage() is the
    one call site that did.

    Contributes to issue #3151


blammit:
@chriadam do we still want to do this? It adds more complexity in page management and maintenance, and means there's less incentive to improve the cost of page creation.

chriadam:
I think we do.  Effectively it means: animations will still play, even after you press an overview widget to drill-down, and the press-effect will play (giving the user feedback that their touch was indeed registered).  So long as the instantiate time is less than the press-effect animation time, the final result will "feel" smoother and better than it does now.

blammit (commented):
As discussed, the other patch to compile page components asynchronously could be included in this PR as well.

chriadam:
> As discussed, the other patch to compile page components asynchronously could be included in this PR as well.

Done

components/PageStack.qml
```diff
@@ -72,7 +88,37 @@ StackView {
...
 	}
 
-	function pushPage(obj, properties, operation) {
+	/*
+		Pushes a page onto the stack.
+
+		'obj' is either a page url or an already-constructed page object.
+
+		A page pushed by url is built asynchronously. Building one is slow: on a
+		Cerbo GX the median page takes 329ms to instantiate and the worst 691ms,
```
blammit: I think these measurements can be left out of code comments, as they may become outdated, particularly with other optimisations we're working on at the moment.

components/PageStack.qml
```diff
@@ -82,63 +128,204 @@ StackView {
...
-			// to avoid messing up the page stack state.
-			let checkComponent = Qt.createComponent(objectOrUrl)
-			if (checkComponent.status !== Component.Ready) {
-				console.warn("Aborted attempt to push page with errors: " + obj + ": " + checkComponent.errorString())
-				return null
+			_pushFromUrl(obj, properties, operation, readyCallback)
+			return null
+		}
+		if (_isComponent(obj)) {
+			_pushFromComponent(obj, "", properties, operation, readyCallback)
```
blammit: If the `pageUrl` arg is always an empty string, I think it makes more sense for `_pushFromComponent()` to drop the parameter and pass an empty string directly when calling `_loadThenIncubate()`.

Although, I'm not sure we really need `_isComponent()`, `_pushFromUrl()` and `_pushFromComponent()` - perhaps they could be folded into this function.

pages/PageManager.qml
```diff
@@ -40,8 +40,10 @@ QtObject {
 
 	property string _hiddenStackMainPage
 
-	function pushPage(obj, properties, operation = PageStack.PushTransition) {
-		return pageStack.pushPage(obj, properties, operation)
+	// A page pushed by url does not exist yet when this returns; see PageStack::pushPage().
+	// Pass 'readyCallback' if you need the page itself.
+	function pushPage(obj, properties, operation = PageStack.PushTransition, readyCallback = undefined) {
+		return pageStack.pushPage(obj, properties, operation, readyCallback)
```
blammit: The `goToStartPage()` function in this file may call `pushPage()` repeatedly to push multiple pages, expecting that pages are created synchronously (line 114). Currently this is not a problem, because in the current start page configurations, there is no more than one page in the stack. (For example, BatteryListPage and DeviceListPage may be used as start pages, but they are pushed directly from the Overview and Settings pages, rather than from another page on the stack.) If we ever have a start page that is more than one level deep, this will fail. I can think of two options:

1. Add a comment somewhere to document that the start page cannot be more than one level deep, and remove the for loop in line 114 since it won't work. Just call `pushPage()` on the first object in the array.
2. Or, provide some parameter to pushPage() to allow the component+object to be compiled+created synchronously, and use that parameter from `goToStartPage()`.

components/PageStack.qml
```diff
@@ -82,63 +128,204 @@ StackView {
...
+		// Going back is not the only way to leave: while the stack is closed the user can
+		// also swipe to another main page, which does not touch the stack at all. So
+		// remember the page this was asked from; leaving it abandons the build, see
+		// the _shownPage handler below.
+		const origin = root._pendingOrigin
+
+		const finish = function() {
+			const stillPending = root._pendingBuild === incubator
+			if (stillPending) {
+				root._pendingBuild = null
```
blammit: Should this also clear `root._pendingOrigin`?

pages/PageManager.qml
blammit: Also, `.github/benchmarking.md` needs updating, as it says `pushPage()` operates asynchronously.

blammit (commented):
copilot says: "MainView.qml:36  now keeps  allowPageAnimations  true while a page incubates. Qt's window incubation controller only gets leftover frame budget, so on a Cerbo, Brief/Overview animations may materially stretch time-to-open." That's true, but it's still an improvement over the synchronously loading that we currently use, so I think this is fine.

components/PageStack.qml
chriadam: true, done

components/PageStack.qml
chriadam: you're right, done

pages/PageManager.qml
chriadam: I agree, did (1) for now, and also updated the docs

components/PageStack.qml
chriadam: yes, fixed

chriadam:
> copilot says: "MainView.qml:36  now keeps  allowPageAnimations  true while a page incubates. Qt's window incubation controller only gets leftover frame budget, so on a Cerbo, Brief/Overview animations may materially stretch time-to-open." That's true, but it's still an improvement over the synchronously loading that we currently use, so I think this is fine.

I've now improved the PR slightly: we now pause most animations (electron animations, gauge animations) but don't pause PressEffect animations, so the touch feedback is still given.  Technically, the incubation CAN still starve, but that seems very unlikely (and not related to animations).  Can revisit in future if someone hits it in a real system.  (e.g. a watchdog timeout followed by a synchronous force.)

components/PageStack.qml
```diff
@@ -82,63 +131,242 @@ StackView {
...
+	}
+
+	// Abandons the page currently being built, if any, so that it is discarded instead
+	// of being pushed when it is ready.
+	//
+	// An owned Component that is still compiling is destroyed, which cancels the load.
+	// An incubator cannot be aborted: the work continues in the background and its
+	// result is destroyed on completion, so a user who repeatedly starts and abandons
+	// opening pages after compilation has finished can have more than one incubation
+	// running at once.
```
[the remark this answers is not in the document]
chriadam: Fixed

components/PageStack.qml
```diff
@@ -14,13 +14,33 @@ StackView {
 	readonly property Page currentPage: opened ? currentItem : null
 
 	readonly property int animationDuration: Global.mainView && Global.mainView.allowPageAnimations ? Theme.animation_page_slide_duration : 0
```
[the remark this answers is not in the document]
chriadam: Fixed

components/PageStack.qml
```diff
@@ -72,7 +92,36 @@ StackView {
...
+		for at a time; a push made while another page is being built is ignored, as it
+		was previously ignored because the UI was blocked. An owned Component that is
+		still compiling is destroyed, which cancels the load. An incubator cannot be
+		aborted, so more than one page can still be under construction if the user
+		repeatedly starts and abandons opening pages after compilation has finished.
+
+		Because the page does not exist yet when this returns, a page object is
+		returned only when one was pushed synchronously, i.e. when 'obj' is already a
+		page object. Pass 'readyCallback' to be given the page once it is on the
+		stack; it is not called if the page was discarded or could not be built.
```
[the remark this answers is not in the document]
chriadam: Fixed

chriadam:
Unfortunately, this hits the QTBUG-123496 assert constantly.

PR #3243, opened 2026-09-17
Dev: add DelegateComponentModel to replace VisibleItemModel

chriadam:
This is a model which contains a list of DelegateComponent
    instances.  The underlying GradientListView will only see
    the entries which are not filtered out by the visibility
    constraint, so the number of actual visual items which are
    constructed is much lower (only visible items, and only items
    which are in the visible content area of the view).

    Delegates are recorded between beginCreate() and completeCreate()
    so a preferredVisible change during completeCreate() can still
    find the object. destroyObject(), entries_clear() and release()
    detach from the objects map before emitting destroyingItem, so a
    slot that inserts cannot rehash under an iterator.

    The tests assert that a hidden entry never constructs its
    delegate, and that hiding an entry destroys the delegate it built.


blammit (commented):
Ideally the DelegateComponent would be able to filter out anything with `effectiveVisible=false`, not just those with `preferredVisible=false`. For example, in `main`: if an item sets `showAccessLevel: VenusOS.User_AccessType_Service` and the current access level is "user", then the item will not be visible, as ListSetting sets `effectiveVisible: preferredVisible && userHasReadAccess`, and VisibleItemModel checks the value of `effectiveVisible` (not `preferredVisible`). 

Would that be feasible? E.g. would it help if `showAccessLevel` was in `DelegateComponent` instead:

```
DelegateComponent {
    preferredVisible: myDataItem.valid
    showAccessLevel: VenusOS.User_AccessType_Service
    
    ListItem {}
}
```

And maybe `effectiveVisible` should be in `DelegateComponent` instead, so that it can be used by `DelegateComponentModel`.

chriadam:
> Ideally the DelegateComponent would be able to filter out anything with `effectiveVisible=false`, not just those with `preferredVisible=false`. For example, in `main`: if an item sets `showAccessLevel: VenusOS.User_AccessType_Service` and the current access level is "user", then the item will not be visible, as ListSetting sets `effectiveVisible: preferredVisible && userHasReadAccess`, and VisibleItemModel checks the value of `effectiveVisible` (not `preferredVisible`).
> 
> Would that be feasible? E.g. would it help if `showAccessLevel` was in `DelegateComponent` instead:
> 
> ```
> DelegateComponent {
>     preferredVisible: myDataItem.valid
>     showAccessLevel: VenusOS.User_AccessType_Service
>     
>     ListItem {}
> }
> ```
> 
> And maybe `effectiveVisible` should be in `DelegateComponent` instead, so that it can be used by `DelegateComponentModel`.

Done this now.  We inject the enclosing DelegateComponent into the ListItem's property, and use it in the default ListItem bindings so that everything works nicely.  The model checks the DC's properties for filtering purposes.

Issue #3238, opened 2026-09-15
Ethernet IP address is shown in the Settings->Connectivity menu when ethernet cable is pulled


ThiemoVanEngelen:
While an ethernet cable is connected, the Settings->Connectivity menu shows the IP address of the ethernet interface. When the ethernet cable is disconnected, it keeps showing the IP address here. When entering the ethernet menu, the State is shown as unplugged. It would be helpful to also show "Unplugged" in the Settings->Connectivity menu.

ReinvdZee:
Thiemo and I did some troubleshooting and found that it reproduces on Venus OS 3.80~51, on Ekrano and on Cerbo. After stepping into the wifi or ethernet menu and back out of it to the connectivity menu, and then unplugging the ethernet cable, the IP address would still show. After stepping back to general and forward to connectivity again, the ethernet menu correctly shows "Unplugged".

PR #3246, opened 2026-09-17
Network: fix unplugged connection bindings

ybott-qinetic:
Use the same binding that is in PageSettingsConnectivity

Contributes to issue #3238


pages/settings/PageSettingsTcpIp.qml
```diff
@@ -15,7 +15,7 @@ Page {
 
 	GradientListView {
 		id: settingsListView
-		model: networkServices.ready ? connectedModel : disconnectedModel
+		model: networkServices.networkState !== "idle" && networkServices.networkState !== "" ? connectedModel : disconnectedModel
```
blammit: The task says the bug is that the Settings -> Connectivity page is showing an IP address instead of the "Unplugged" status text, but this patch is changing the Settings -> Connectivity -> Ethernet page. Am I missing something?

PR #3250, opened 2026-09-21
Reduce object instantiation costs

chriadam:


pages/OverviewPage_Landscape.qml
```diff
@@ -591,6 +591,7 @@ FocusScope {
 	FrameAnimation {
 		id: overviewPageRootAnimation
 
+		objectName: "overviewConnectorAnimation"
```
blammit, line 594: Is this still needed? I'd rather not add these into the non-test code, but if it's required, then this needs a comment so it's not removed accidentally.

blammit (dismissed):

Global.qml
```diff
@@ -24,6 +24,7 @@ QtObject {
 	readonly property string quantityFontFamily: _quantityFontLoader.name
 	property var dialogLayer
 	property var notificationLayer
+	property var pressEffect
```
blammit, line 27: Can this be PressEffect? Not sure

blammit:
- Tests: repair page-construction benchmark and animation opt-in -> lgtm
- Perf: disconnect FrameRateModel unless the overlay is enabled  -> lgtm
- Perf: don't switch electrons to arrows during transitions  -> lgtm
- Perf: make pushPage() both compile+instantiate asynchronously -> can squash into other PR
- Perf: load card and Brief overlays before sliding them in  -> lgtm
- Perf: omit empty secondary and caption labels until they have text -> @blammit to investigate, - perhaps we should reduce the amount of captions and secondaryText properties instead, and limit them to where they are actually used in the current designs.
- Perf: connect ListSetting to ToastModel only when showing a toast  -> lgtm
- Perf: share one PressEffect instead of creating one per PressArea  -> lgtm
- Tests: add COMPBENCH variants for row instantiate cost  -> lgtm
- Perf: Construct QuantityLabel text content natively - not sure if it's worth the complexity right now? - Let's discuss further.
- Dev: fix drill-down slide duration  -> lgtm

ApplicationContent.qml
```diff
@@ -19,6 +19,10 @@ FocusScope {
 		Component.onCompleted: Global.mainView = mainView
 	}
 
+	PagePreloader {
+		active: Global.allPagesLoaded && UiConfig.splashScreenVisible && UiConfig.showSplashAnimation
```
blammit: On Wasm, the splash animation is not shown; it will show the splash screen with the loading progress bar, but not the animation at the end. Does the `UiConfig.showSplashAnimation` condition here mean that the pages won't be preloaded for Wasm?

chriadam: Damn, I think you're right.  I will pull that out into separate PR and fix.

blammit (approved):

PR #3255, opened 2026-09-24
Perf: Construct QuantityLabel text content natively

chriadam:


chriadam:
As discussed on Tuesday, these changes might not be worth the complexity gain, separated into their own PR so they don't block the other improvements.

